PR #31, opened 2013-04-18
Enable aiMaterial::AddProperty template specializations in ASSIMP_NO_EXPORT builds

jdduke:
aiMaterial::AddProperty template specializations are used by import methods in B3DImporter and SceneProcessor; they should be defined accordingly even when ASSIMP_BUILD_NO_EXPORT is defined.


m9ee1:
> aiMaterial::AddProperty template specializations are used by import methods in B3DImporter and SceneProcessor; they should be defined accordingly even when ASSIMP_BUILD_NO_EXPORT is defined.
>

PR #435, opened 2015-01-20
Added new process - Scale-to-unit-box

orcchg:
This allows to transform input meshes, scaling its
vertices to unit box in order to let mesh become visible
for small scenes or limited frustums.

See screenshots with sample 3D models: original (unscaled) and scaled to unit box. 
original
scaled
Both cases have the same OpenGL view point and frustum (any scene parameters) conditions. Input meshes from files are different, some have compact sizes, the others are too large. The aim of ScaleToUnitBoxProcess is to scale input mesh to unit box, making mesh uploading process more universal.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -482,6 +482,8 @@ SET( PostProcessing_SRCS
 	RemoveRedundantMaterials.h
 	RemoveVCProcess.cpp
 	RemoveVCProcess.h
+	ScaleToUnitBoxProcess.cpp
+	ScaleToUnitBoxProcess.h
 	SortByPTypeProcess.cpp
 	SortByPTypeProcess.h
 	SplitLargeMeshes.cpp
--- a/code/PostStepRegistry.cpp
+++ b/code/PostStepRegistry.cpp
@@ -121,6 +121,9 @@ corresponding preprocessor flag to selectively disable steps.
 #ifndef ASSIMP_BUILD_NO_DEBONE_PROCESS
 #	include "DeboneProcess.h"
 #endif
+#ifndef ASSIMP_BUILD_NO_SCALETOUNITBOX_PROCESS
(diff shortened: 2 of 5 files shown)


acgessler:
My main concern is that this scales mesh geometry only. In a complex scene with multiple meshes that are attached to nodes with different transformations, it will simply screw up the scene. At minimum, the change to the mesh geometry (translate and scale) should be reverse-applied to its anchor node such that the scene itself is not changed.

Generally, I think unit scaling is useful but should be applied as a scaling factor to the scene root node. Such a process would traverse the scene, calculate all vertices in world coordinates, determine bounding sphere+center or min/max values for each axis and then adjusts the root transform (without changing meshes).

These are just my thoughts of course.

expipiplus1:
It's probably worth using a less naïve algorithm for calculating the average position of the points such as Kahan summation or Knuth's incremental mean.

orcchg:
Thank You for reply!

I've also checked whether this process breaks multiple meshes within the same scene - yes, it does :(

original_multimesh
scaled_multimesh

One can see original (first) and scaled (second) meshes 'dwarf.cob'. I don't actually understand, why the axe looks not properly, but in case of scaled mesh - it has been translated to the wrong position. Maybe, it is worth doing like acgessler has previously said?

orcchg:
On the other hand, it probably my own failure. Let's consider Cinema4D.dae file containing three spartially separated figures.
In case of original scene, I see wrong relative positions:
original_figures

So does after aiProcess_ScaleToUnitBox:
scaled_figures

By the way, meshes are correct in both cases. When I recognize the bug, I will report.

acgessler:
Closing this for now, feel free to open another PR with an updated version :-) My comments from above still apply: I'd prefer applying scaling to the root transform only, based on the entire scene's bounding box in bind pose.

PR #437, opened 2015-01-21
Add support for component controlled matrix transforms, i.e transform(0)...

TomMettam:
The Collada loader didn't handle skeletal animations which address a single component in a matrix per channel. This commit fixes this - tested in assimp_viewer.

Sample DAE file: https://dl.dropboxusercontent.com/u/29850039/astroBoy_walk_Max.dae

Thank you for your hard work!

Tom Mettam and the rest of the team at the Mivvo project (http://www.mivvo.com/)

### DIFF
--- a/code/ColladaLoader.cpp
+++ b/code/ColladaLoader.cpp
@@ -985,6 +985,47 @@ void ColladaLoader::CreateAnimation( aiScene* pScene, const ColladaParser& pPars
 				entry.mTransformId = srcChannel.mTarget.substr( slashPos+1);
 			}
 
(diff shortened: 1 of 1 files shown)


acgessler:
Awesome! Any chance we can add this (or another) dae file to our models database (i.e.test/models-nonbsd). It's 1.5 MiB but it compresses down to 10%.

TomMettam:
It's not my model, however I believe it was made freely available on the Collada Model Bank.

As far as I am concerned, go for it ;)

acgessler:
Merged without the file for now.

Thanks!

PR #439, opened 2015-01-22
Fix incorrect abs usage

expipiplus1:
Replace calls to abs with a float to calls to std::abs. Before int abs(int) was being called.


acgessler:
Good point, but why do we not use `fabs` instead? Makes the intent clearer.

expipiplus1:
The intent is to get the absolute value of the variable, whatever it's type. Although I think it's unlikely that the type of these variables will change in the future, using std::abs ensures that this problem doesn't crop up again.

acgessler:
Thanks!

PR #444, opened 2015-01-24
Check that zlib initialized correctly in FBX parser

Einarin:
Check the return code of inflateInit() indicates success to avoid
crashing later when zstream contains invalid data.


acgessler:
Good catch, thanks!

PR #445, opened 2015-01-25
Avoid division by zero in assimp_cmd info when scene loading succeeds bu...

turol:
Avoid division by zero in assimp_cmd info when scene loading succeeds but contains no meshes


acgessler:
Looks good, thanks!

PR #446, opened 2015-01-26
Fixed error of aiQuaterniont::Rotate()

yeonseok-yi:
The function rotates a point to opposite direction.
Conjugate() should be applied to inverse of the quaternion.


acgessler:
I do not feel confident changing anything about the quaternion code without checking every single caller within assimp - this can easily break importers. Unless of course it's not used anywhere.

yeonseok-yi:
I understand what you get worried about.
I checked that this function is not called in assimp at all.
Is it still a problem in such case?

kimkulling:
In that case we should remove it I guess.

acgessler:
This probably affects lots of other math functions as well.

Some have been added at user's request for being "math-complete" so maybe not drop them.

PR #447, opened 2015-01-26
Fix warnings under Visual Studio 12

expipiplus1:


acgessler:
We can merge here, but please also file upstream to poly2tri.

Thanks!

contrib/poly2tri/poly2tri/sweep/sweep_context.cc
```diff
@@ -96,11 +96,11 @@ void SweepContext::InitTriangulation()
...
 {
-  int num_points = polyline.size();
-  for (int i = 0; i < num_points; i++) {
-    int j = i < num_points - 1 ? i + 1 : 0;
-    edge_list.push_back(new Edge(*polyline[i], *polyline[j]));
+  std::size_t num_points = polyline.size();
+  for (std::size_t i = 0; i < num_points - 1; i++) {
+    edge_list.push_back(new Edge(*polyline[i], *polyline[i + 1]));
   }
+  edge_list.push_back(new Edge(*polyline.back(), *polyline.front()));
```
jdduke: I guess there's an implicit assumption that polyline is non-empty? Is this enforced/validated anywhere? Just asking, as the previous version handled that case gracefully (i.e., it won't crash), whereas this may not.

acgessler:
Good catch, reverted for now.

It seems to me that `sweep_context.cc` is generally written to allow empty inputs.

Since we do not directly call this code, we should be cautious. I am pretty sure we never triangulate empty triangles (since we cannot represent them in our data structures anyway ...) but it would be easy for that to slip by somewhere.

PR #450, opened 2015-01-28
Fix some bugs found with afl

turol:
Fixes assorted bugs found with afl http://lcamtuf.coredump.cx/afl/

Do not merge without review. Not regression tested since it doesn't run very well on Linux. Also there's some doubt about what we should do on corrupted input. Try to continue or just give up immediately?

Also there are many more, this is just the first batch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ test/gtest/src/gtest-stamp/Debug/gtest-build
 *.lib
 test/gtest/src/gtest-stamp/Debug/
 tools/assimp_view/assimp_viewer.vcxproj.user
+
+# Unix editor backups
+*~
--- a/code/3DSLoader.cpp
+++ b/code/3DSLoader.cpp
@@ -175,6 +175,10 @@ void Discreet3DSImporter::InternReadFile( const std::string& pFile,
 	// file.
 	for (std::vector<D3DS::Mesh>::iterator i = mScene->mMeshes.begin(),
 		 end = mScene->mMeshes.end(); i != end;++i)	{
+		if ((*i).mFaces.size() > 0 && (*i).mPositions.size() == 0)	{
+			delete mScene;
+			throw DeadlyImportError("3DS file contains faces but no vertices: " + pFile);
+		}
 		CheckIndices(*i);
 		MakeUnique  (*i);
 		ComputeNormalsWithSmoothingsGroups<D3DS::Face>(*i);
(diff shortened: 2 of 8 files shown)


acgessler:
Cool. I'll have a look tonight.

turol:
Also should I rewrite and rebase the whole series or add more commits to fix the issues you pointed out?

acgessler:
A rebase would be cleaner but if it is less effort for you, feel free to just add more commits.

turol:
V2, fix issues raised in review.

turol:
Both rebase and more commits are equally easy but it seems github loses the comments when the commits change (obviously). Hope no-one misses them :)

acgessler:
Thanks!

Running the regression test suite, we added 5 failures:

../models-nonbsd/MD2/horse.md2
../models-nonbsd/LWO/LWO2/LWSReferences/QuickDraw--GP-Gun.lwo
../models-nonbsd/LWO/LWO2/LWSReferences/QuickDraw--Chasis.lwo
../models-nonbsd/LWO/LWO2/LWSReferences/QuickDraw--Arm-Tip.lwo
../models-nonbsd/LWO/LWO2/LWSReferences/QuickDraw--Arm-Shoulder.lwo

I'll look into the lwo files later. md2 is a mystery :-)

turol:
Those LWO files are among the tests which fail on Linux. Maybe the reason they fail is the models are not quite correct and cause undefined behavior in the library?

Also no idea on md2. There's nothing in this set which should be able to cause that.

acgessler:
Apologies - I misread the output. We really need to fix the regression suite reporting.

here is the diff:

     ../models/X/test_cube_compressed.x -cfull;../results/tmp/test_cube_compressed.x_0x28c95c8a/EXPECT 
 	../models/X/test_cube_compressed.x -og -om -vds;../results/tmp/test_cube_compressed.x_0x6ce1810d/EXPECT 
 	../models/X/test_cube_compressed.x -ptv -gsn -cts -db;../results/tmp/test_cube_compressed.x_0xc2466795/EXPECT 
 	../models/X/test_cube_compressed.x -vds -jiv;../results/tmp/test_cube_compressed.x_0x1e8baf0d/EXPECT 
 	../models/X/test_cube_compressed.x ;../results/tmp/test_cube_compressed.x_0x89b6c185/EXPECT

turol:
So the .X loader fix is wrong. Can you drop that one and merge the rest?

acgessler:
I tried around and a variant of the test that seems to work is `(P + ofs > End + 2)`. Confirmed with test_cube_compressed.x and another compressed X file I have in my collection.

I have no idea why this would be the correct test, it seem we're still in danger of reading one byte off. It could be an implementation detail of zlib that I am not aware of.

Would you mind changing?

turol:
Sure, here it is.

acgessler:
Awesome. Thanks for your patience!

turol:
Don't thank me yet, I'm about to file a bug with several hundred more crashing and hanging testcases. Because there's no way I could go through all of them myself.

PR #452, opened 2015-01-29
Fix exported library name variable

ricortiz:
This should take care exported assimp library name across platforms.


kimkulling:
Looks good, thanks a lot!

Issue #454, opened 2015-01-31
Lots of crashes and hangs


turol:
https://www.dropbox.com/s/ppp297aa1idfny2/assimp-bugs.zip?dl=0

Contains 837 hanging and 908 crashing testcases discovered with afl.

The crashing testcases have been minimized and should be mostly different. Tests were run on Linux with libstdc++ in debug mode and address sanitizer enabled. If using something else be aware that most of these are probably memory corruption bugs.

turol:
OpenGEX parser wasn't in yet when I did the first fuzzing run. So I fuzzed just that one. In particular it seems very easy to make it loop indefinitely.

@kimkulling You might want to fix these while you're writing it. Also if you want to try afl yourself drop me a note and I can upload my initial test corpus.

https://www.dropbox.com/s/taf7v98tpa60xz7/ogex-crashes.zip?dl=0

kimkulling:
Thanks a lot for the test data. I will check it!

turol:
#488 and #523 should have fixed all of the crashes. Some of the hangs are still unresolved. So here's some more: https://www.dropbox.com/s/bgz5zlcaq787r2v/assimp-bugs2.zip?dl=0

turol:
This has been effectively superseded by OSS-Fuzz.

PR #455, opened 2015-01-31
Fix some infinite loops

turol:
Fixes some (but nowhere near all) of hanging inputs in #454. No new testsuite regressions on Linux which is not very useful since 81 cases currently fail even without any modifications.

Also you might want to add these testcases to the suite since at least one of them hits an error path that appears to have been completely untested until now.


acgessler:
Looking good, thanks!

I'll aim at updating the regression suite soonish. I want to cleanup some things anyway and further integrate with Travis CI.

PR #456, opened 2015-02-02
Fix division by zero

turol:
There are at least two other places where division by zero happens during processing of test/models/IFC/AC14-FZK-Haus.ifc

In PolyTools.h line 86 function PointInTriangle2D is called from TriangulateProcess.cpp line 392.

It seems that the triangle is almost but not quite degenerate. Dumping the variables I get this:
p0: 0.000000 0.680000
p1: 1.800000 0.679900
p2: 1.800000 0.680000
pp: 0.000000 0.000000
v0: 1.800000 -0.000100
v1: 1.800000 0.000000
v2: 0.000000 -0.680000
dot00: 3.240000
dot01: 3.240000
dot02: 0.000068
dot11: 3.240000
dot12: 0.000000
denom: 0.000000

Is it correct to change PointInTriangle2D to avoid divzero (by always adding FLT_MIN to denominator) or should the conditions in TriangulateProcess be changed?

In vector3.inl line 100 function Normalize() is called from GenVertexNormalsProcess.cpp line 193.
verticesFound.size() = 1
vertex 312
pMesh->mNormals[312] is a zero vector. Normalizing it ends up dividing by zero.

Should this method be changed and/or should the general case of normalizing a zero vector do something sane?

Note that changing Normalize() to either check for 0 or adding FLT_MIN to denominator causes a lot of regression failures.
Dividing by zero is undefined behavior so the compiler is theoretically free to do whatever it pleases. The most common result seems to be NaNs which are not healthy for anything.

### DIFF
--- a/code/IFCUtil.cpp
+++ b/code/IFCUtil.cpp
@@ -122,7 +122,7 @@ void TempMesh::Transform(const IfcMatrix4& mat)
 // ------------------------------------------------------------------------------
 IfcVector3 TempMesh::Center() const
 {
-	return std::accumulate(verts.begin(),verts.end(),IfcVector3()) / static_cast<IfcFloat>(verts.size());
+	return (verts.size() == 0) ? IfcVector3(0.0f, 0.0f, 0.0f) : (std::accumulate(verts.begin(),verts.end(),IfcVector3()) / static_cast<IfcFloat>(verts.size()));
 }
 
 // ------------------------------------------------------------------------------------------------


acgessler:
Thanks. As for the test failure, I think this is a flakyness on Travis part so I would take the risk and commit (I cannot find a button to re-trigger a CI build). We can rollback tomorrow :-)

Nit (feel free to correct if I am wrong!): Division by zero is well-defined by IEEE 754 and caused +INF or -INF depending on the signs involved. Any further calculation involving INFs then causes NAN.

Generally, degenerate geometry is very frequent and avoiding any numerical issues in a codebase this large (and so much not focussed on getting computational geometry right ..) is hard. There is no right way of handling a non-existent normal or a degenerate tangent space. A zero vector is as much a problem as INF or NAN: if it propagates, it kills lighting calculations somewhere in the frontend :-) We have a postprocess step that detects and removes any such stuff. Unfortunately it runs before triangulation though.

turol:
Partially true. C/C++ standard defines division by zero as undefined behavior. Clang ubsan (undefined behavior sanitizer) picks this up which is how I found it.

In practice it seems that both GCC and clang generate inf when dividing by zero. However using inf in arithmetic doesn't produce NaNs. Adding to inf produces inf and dividing finite number by inf produces 0.

You should do a test with Normalize. Fix it to not divide by zero, generate new expected outputs for all testcases on MSVC and test if this reduces failures on GCC. I don't have MSVC set up and I don't know if there's an easy way to regenerate the expected output files.

acgessler:
Indeed undefined as per the standard, but if the implementation chooses to obey IEEE 754 it is no longer undefined at all. I believe ubsan only picks it up if configured approriately.

So to speak, IEEE 754 compliance is a requirement for Assimp.

We should indeed doublecheck all of this. Not sure I'll have the time this week but I'll put it on the tab.

`test/regression/gen_db.py` re-generates test cases for the record.

PR #464, opened 2015-02-13
Clean up some warnings

turol:
Cleans up some compiler warnings. No new regression failures on Linux with GCC. Review and run testsuite on a working config before merging. Also the recent changes to the test suite seem to have made mode cases fail (101 currently vs 96 previous).

### DIFF
--- a/code/3DSExporter.cpp
+++ b/code/3DSExporter.cpp
@@ -294,7 +294,7 @@ void Discreet3DSExporter::WriteMaterials()
 			WriteColor(color);
 		}
 
-		aiShadingMode shading_mode;
+		aiShadingMode shading_mode = aiShadingMode_Flat;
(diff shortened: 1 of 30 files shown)


turol:
@acgessler ACK or NAK on this please.

acgessler:
Sorry for the delay. One comment only, otherwise looks great.

turol:
Fixed assertion.

turol:
This branch fails to build but I can't see why. I have not touched the failing file, in fact it looks like no-one has touched that file in a year. In addition the error message doesn't appear to match the code. WTF?

acgessler:
#472 should fix it, then we can proceed.

acgessler:
Looks all good now. Thanks for the patience!

PR #466, opened 2015-02-18
Fixed many doxygen errors

mbp566:
Many doxygen comments were too old (filenames have changed since, for example), so no useable documentation could be generated with doxygen.

After fixes many issues, a more or less usable doc can be generated (at least the aiMaterial help is all there, which wasn't the case before)

Other than doxygen comments I did not change anything (no code change)


acgessler:
Thanks for doing this!

PR #469, opened 2015-02-26
Support loading the MTL file for an OBJ file when using MemoryIOWrapper

smokris:
When using `MemoryIOWrapper` (e.g., by calling `aiImportFileFromMemoryWithProperties()`), the `.mtl` file associated with an `.obj` file is ignored.

This patch (by Christian Van Brussel, posted on [this thread](http://sourceforge.net/p/assimp/mailman/message/30003098/)) enables assimp to load these `.mtl` files.

### DIFF
--- a/code/Importer.cpp
+++ b/code/Importer.cpp
@@ -494,7 +494,7 @@ const aiScene* Importer::ReadFileFromMemory( const void* pBuffer,
 	IOSystem* io = pimpl->mIOHandler;
 	pimpl->mIOHandler = NULL;
 
-	SetIOHandler(new MemoryIOSystem((const uint8_t*)pBuffer,pLength));
+	SetIOHandler(new MemoryIOSystem((const uint8_t*)pBuffer,pLength,io));
 
 	// read the file and recover the previous IOSystem
 	char fbuff[128];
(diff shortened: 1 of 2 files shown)


acgessler:
Thanks - if I understand correctly though this will mean we still fetch the mtl file from disk? This may come as a surprise to API users.

kimkulling:
Seems to be broken. And it will change the default behaviour, right? So maybe we need to think about this before we can accept it.

smokris:
Thanks for your quick responses.  Regarding "we still fetch the mtl file from disk" — yes, that appears to be what is happening.  To me that seems like a slight improvement over not loading .mtl files at all, but still unexpected that the so-called memory loader would sometimes be loading stuff from disk.

Earlier today when I found that patch and submitted it here, I didn't realize assimp already had a better solution — `aiImportFileExWithProperties()`, to which I can pass my own file I/O callbacks.  And I overlooked the helpful note in the `aiImportFileFromMemory()` documentation suggesting creating a custom IOSystem.

My bad.  I'll close this pull request then.

acgessler:
Yah, `aiImportFileFromMemory` has lots of shortcomings and is basically useless for Obj, which is usually accompanied by a material file. In such a case a custom `IOSystem` is preferred. To make it easier, we could offer a utility IOSystem implementation that can be constructed from a `map<path, blob>` or similar. 

Thanks for your time :-)

PR #474, opened 2015-03-02
Fix read past end of buffer after call to TokenMatch

turol:
DO NOT MERGE IMMEDIATELY! New regression failures (116 now vs 101 previous).

IsSpaceOrNewLine returns true on end of input (NUL character). But if
TokenMatch considers a token at end of input to match it sets "in" to
one past end of buffer. This will lead to reading past the end of
buffer on any subsequent operation.

This tries to fix crashes/012101a9d2a085ba5ba382bf803cb96c (and probably many others) from #454. This case no longer crashes but this causes new failures on regression suite. Because of the large amount of baseline failures on GCC I'm unable to effectively debug this. Also this should only happen if the testcases were broken to begin with...

Can you either verify that this is correct and update the test suite or come up with a better fix?

### DIFF
--- a/code/IFCOpenings.cpp
+++ b/code/IFCOpenings.cpp
@@ -1184,16 +1184,13 @@ bool GenerateOpenings(std::vector<TempOpening>& openings,
 					profile_data = opening.profileMesh2D.get();
 					is_2d_source = true;
 				}
-				else {
-					//continue;
-				}
 			}
 			else {
 				// vertical extrusion
 				if (std::fabs(norm_extrusion_dir * nor) > 0.9) {
-					continue;
-				}
-				continue;
+					profile_data = opening.profileMesh2D.get();
+					is_2d_source = true;
+				} 
 			}
 		}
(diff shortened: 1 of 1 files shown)


abma:
without investigation: isn't

    in[len]

already an invalid access? i would assume

    in[len-1]

is correct.

turol:
No, that's still good (assuming len == strlen(token)). It only checks in[len] if strncmp returned 0 so we know there's at least len bytes (the token) and since the loader made sure the buffer is '\0'-terminated there's also one extra character there. The problem happens with "in += len+1;" which might no longer be inside the buffer.

turol:
@acgessler Ping?

acgessler:
Potentially, using TokenMatch on the last token in the input and *not* accessing `in` afterwards is fine. So I am a bit worried about regressions (yeah, regression suite still not functional, I'm looking into updating it).

We could consider setting `in` to point to the 0 character (instead of the character after it) in the EOF case. This still allow the match, but should keep any code that uses it from looking further.

PR #476, opened 2015-03-02
remove "register" keyword:

abma:
clang warns: 'register' storage class specifier is deprecated
http://stackoverflow.com/questions/10675072/is-the-register-keyword-still-used


kimkulling:
And he'S right, thanks for the pull request!

PR #480, opened 2015-03-07
Free edges map before recursive calls to reduce memory consumption

turol:
Reduces peak memory use by about 10 MB during load of SuzanneSubdiv_252.blend.

No new regression failures however this file is one of the cases which fails on Linux to begin with so take it with a grain of salt.

Also should I fix the indentation afterwards?


kimkulling:
Great, thanks!

PR #483, opened 2015-03-08
Use material names, set default direction to UP_Y, process extra tag

DenisMikhalev:
Use material names, it more informative than id
Set default direction to UP_Y, like in specification


acgessler:
Looks good, thanks for sending in the PR. It's much easier to do than patches by mail, and it acknowledges the contribution nicely.

Most collada files I have specify the direction explicitly. We still need to watch out for complaints about the direction change as people may be relying on it.

PR #484, opened 2015-03-09
Fix read past end of buffer after call to TokenMatch

turol:
This one sets "in" to point to terminating 0 character on EOF. Causes no new regression failures.


acgessler:
Thanks for updating this!

PR #485, opened 2015-03-09
Prevents NaN when normalizing a null vector

DenisMikhalev:


amc522:
Im not a contributor of assimp, just user of the library so take what I say with a grain of salt. Generally in 3D math libraries you have two normalize functions. An unsafe normalize and a safe normalize. The added logic to normalize can be a performance issue. I think the right way to do it would be to keep the existing functionality of Normalize() as is and create a new function NormalizeSafe() and use that where appropriate.

That being said, please defer to the project maintainers.

jdduke:
Yeah, I've see libraries go both ways here. Propagating NaN is nice because it's very explicit about failure, and normalizing a zero vector is, in most cases, a strong indication that something has failed.

Adding a NormalizeSafe seems quite reasonable, perhaps used in conjunction with a processing flag?

acgessler:
I am against this change. Assimp has always propagated NaN and we should not change this behaviour. We do have the `aiProcess_FindInvalidData` postprocessing step which essentially drops vertex attributes that contains NaNs et al. (however, currently it does not cover artifacts that occur later in the pipeline).

ov:
Although I agree with the safe/unsafe approach, something still needs to be done about this. The issue happened in `GenVertexNormalsProcess::GenMeshVertexNormals()` where `pcNor` was made of two vectors where the first one was the negative of the second. We probably need to fix it there instead, if you want to keep the `Normalize()` as is.

Should we make another pull request for that?

acgessler:
Hmm, technically the normal is indeed not well-defined in this case. We could substitute an arbitrary normal that is orthogonal to both vectors, but this might cause even more artifacts, i.e. lighting and tangent space discontinuities.

ov:
I believe that any vector is better than NaN-vector. What we can do is to make a local change, like:

```c++
if (pcNor.SquareLength() > 0)
    pcNor.Normalize();
```

or add `NormalizeSafe()` method and call it instead of simple `Normalize()` there.

Although we can run "fix invalid data" processor after that, I think that processors should not add invalid data, if possible.

Please let me know which approach you like more, so we can implement one.

turol:
When normalizing vectors you should use FLT_MIN trick to avoid a branch:
http://www.gamasutra.com/view/news/126691/Opinion_Practical_Floating_Point.php

ov:
That's a good point which shouldn't affect existing code too much. Need to test it, though...

turol:
I tested it already and it does affect current code (regression failures) so @acgessler NAKed the change. I think they might be cases where the code produces NaNs and then later removes them and so changing them to zeroes causes different behavior. I'll leave arguing about "correct" to someone else.

acgessler:
To clarify, I'm fine adding NormalizeSafe (and using it where appropriate) but let's not change the existing Normalize.

PR #488, opened 2015-03-11
Add an assertion and some debug prints

turol:
Do not merge. This is not a proper fix, contains debugging code and causes regression failures.

This tries to fix crashes/bee0783c79dd9a27abe2651fc16806c1 from #454. I'm not sure what the proper fix is here. Somehow numVertices is wrong and/or isn't properly checked against mNumIndices. What's the difference between numVertices and mesh->mNumVertices? @acgessler thoughts?

Also is it possible to enable logging with "assimp-cmd info" so I could use logger instead of printfs?

### DIFF
--- a/code/OFFLoader.cpp
+++ b/code/OFFLoader.cpp
@@ -189,6 +189,7 @@ void OFFImporter::InternReadFile( const std::string& pFile,
 
     // second: now parse all face indices
     buffer = old;faces = mesh->mFaces;
+    printf("numVertices: %u   mesh->mNumVertices: %u\n", numVertices, mesh->mNumVertices);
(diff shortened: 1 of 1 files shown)


acgessler:
`assimp info` does not log indeed, but you should be able to use any other command, for example `assimp export` with the `-l` and `-v` switch.

Didn't yet get to look into OffLoader.

turol:
@acgessler ping?

turol:
Rebased on top of latest master. Still not meant for merging. Still crashes when fed the bad testcase. Is there really no-one who understands this code?

kimkulling:
Hm, sounds not too bad, I will try my very best.

PR #490, opened 2015-03-11
Added 'const' to () operator on CompareVector

wilshipley:
This doesn't build on clang as-is. I don't speak C++ but the net said adding "const" is the solution, and it now appears to build fine now for iOS and OS X.

### DIFF
--- a/code/IFCUtil.cpp
+++ b/code/IFCUtil.cpp
@@ -236,7 +236,7 @@ IfcVector3 TempMesh::ComputeLastPolygonNormal(bool normalize) const
 
 struct CompareVector
 {
-	bool operator () (const IfcVector3& a, const IfcVector3& b)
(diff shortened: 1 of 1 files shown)


acgessler:
One of the travis CI builds is clang, so I am wondering which clang version rejects this?

Thanks for the patch!

wilshipley:
I’m running Xcode Version 6.3 (6D532l). Looking at the flags it looks as though the typical cmake ▶ Xcode project thing has the clang compiler flags set to be “default” for the version of C++ clang should use, instead of g11++ or g14++ (I may be getting their names wrong, this isn’t my area), and the net said this difference is a GNU vs. Clang thing, so maybe if the flags are different in CI it works just fine?

Yours,
-Wil Shipley
Chief Monster, Delicious Monster Software

PR #491, opened 2015-03-12
no one ever tried this java stuff???

arris69:
(crashes if the model has a lot of meshes...)

### DIFF
--- a/port/jassimp/jassimp-native/src/jassimp.cpp
+++ b/port/jassimp/jassimp-native/src/jassimp.cpp
@@ -703,7 +703,7 @@ static bool loadMeshes(JNIEnv *env, const aiScene* cScene, jobject& jScene)
 				wrapParams[0].l = jMatrixArr;
 				jobject jMatrix;
 				
-				if (!callStaticObject(env, "Ljassimp/Jassimp;", "wrapMatrix", "([F)Ljava/lang/Object;", wrapParams, jMatrix)) 
+				if (!callStaticObject(env, "jassimp/Jassimp", "wrapMatrix", "([F)Ljava/lang/Object;", wrapParams, jMatrix))
 				{
 					return false;
 				}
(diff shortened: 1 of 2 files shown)


port/jassimp/jassimp-native/src/jassimp.cpp
```diff
@@ -5,12 +5,30 @@
...
+#define lprintf(...) __android_log_print(ANDROID_LOG_VERBOSE, __func__, __VA_ARGS__)
+#else
 #define lprintf(...) printf (__VA_ARGS__)
+#endif /* ANDROID */
 #else
 #define lprintf
 #endif
 
 
+static void dumpRefTable(JNIEnv *env){
```
jdduke: I don't think we need this debug helper.

port/jassimp/jassimp-native/src/jassimp.cpp
```diff
@@ -5,12 +5,30 @@
 
 
 #ifdef JNI_LOG
+#ifdef ANDROID
```
jdduke: Not __ANDROID__?

jdduke:
Thanks, I wonder how difficult it would be to get Travis to test compile this against the Android NDK.

port/jassimp/jassimp-native/src/jassimp.cpp
acgessler: \_\_ANDROID\_\_?

arris69: looks like i'm not the only one...
http://stackoverflow.com/questions/15328751/android-macro-suddenly-not-defined

acgessler:
Is this complete, i.e. can be merged or work in progress?

arris69:
it can be merged, then this just fixed compile errors for android target.

acgessler:
Ok, thanks - for further fixes, just open a new PR.

Also, if you want you could add jassimp to the travis CI build.

arris69:
my question is how to do that?
http://yaml.travis-ci.org/ tells me that you can't add jdk: oracle_blabla to language: gcc
it's also not possible to use multiple languages (gcc java) at the same time.
2. how to install android ndk/sdk (normally you have to accept some braindead goole license also on commandline install), its ok for the assimp team if i provide a shellscript that install it without interaction?
3. guess it would makes sense to provide an android sample app (at least what dumps some info about the model file)?

regards

PR #501, opened 2015-03-18
Feature scale to unit box process

orcchg:
Dear Mr. Acgessler:

I've fixed Scale-To-Unit-Box process, this now traverses the whole scene and scales the whole scene to the unit sphere.
Btw, could You please explain, what does 'adjust root transform' mean? Root node has no transformation relative to the node's parent, coz there is no parent node for root.
Thus, current implementation of the process scales each vertex in each mesh of the scene.
But this works correct.

Please, answer about 'root transform'. I don't understand Assimp in such details.

Thank You!

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -482,6 +482,8 @@ SET( PostProcessing_SRCS
 	RemoveRedundantMaterials.h
 	RemoveVCProcess.cpp
 	RemoveVCProcess.h
+	ScaleToUnitBoxProcess.cpp
+	ScaleToUnitBoxProcess.h
 	SortByPTypeProcess.cpp
 	SortByPTypeProcess.h
 	SplitLargeMeshes.cpp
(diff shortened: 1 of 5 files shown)


acgessler:
Thanks. However, it is not sufficient to go through all meshes and take the bounding box as the scene extents. The actual scene as loaded by assimp is the `aiScene::mRootNode` + children. Each node has a transformation relative to its parent (typically: scaling, translation, rotation, rarely: skewing) and zero or more meshes attached. Nodes reference meshes by index into the global list of meshes which you've been looking at (this enables a single mesh to be shared between nodes).

However, the location of a mesh is solely determined by the scene node(s) it is attached to. To get a proper bounding box/sphere, you have to walk the entire tree, compute absolute transformations for each node, then apply this transformation to each point in the node's meshes and take the min/max or max radius of *all* such points.

acgessler:
Also see http://assimp.sourceforge.net/lib_html/data.html

PR #502, opened 2015-03-18
Opengex support

kimkulling:
First prototype of the OpenGEX-importer, which uses the openddl-importer.


acgessler:
Cool to get this in, keep up the good work :-)

jdduke:
I should be able to take a look at this in the next week, I expect the "todo"s should point me in the right directiON?

kimkulling:
That's the idea :-)

PR #505, opened 2015-03-20
integration of android ndk (for travis build)

arris69:
~~don't merge.~~ finished

some smart tips how to merge it into the buildmatrix?


acgessler:
Very nice. As for integrating with the build matrix, I'm not aware of any better way than to branch in the `script` section.

jdduke:
Interesting! One concern I saw voiced by developers looking at similar options was the install/build time to setup the Android/NDK environment properly with travis. Build time turnaround may not be a bottleneck now, but it's something to keep in mind.

jdduke:
Looking at the travis result, it appears that the total build time for Android (per config) is comparable to the build + test time for other configurations. Not bad at all. I imagine we'd be in pretty good shape with just the ARM configuration that builds the library + Java bindings?

workspaces/Android-NDK/jni/Android.mk
```diff
@@ -0,0 +1,134 @@
...
+LOCAL_MODULE    := assimp_static
+ASSIMP_SRC_DIR = code
+FILE_LIST := $(wildcard $(LOCAL_PATH)/$(ASSIMP_SRC_DIR)/*.cpp)
+
+IGNORE_LIST := $(wildcard $(LOCAL_PATH)/$(ASSIMP_SRC_DIR)/C4D*.cpp)
+EXCLUDES := $(IGNORE_LIST:$(LOCAL_PATH)/%=%)
+ALL_SRC_FILES := $(FILE_LIST:$(LOCAL_PATH)/%=%)
+LOCAL_SRC_FILES := $(filter-out $(EXCLUDES), $(ALL_SRC_FILES))
+
+LOCAL_SRC_FILES += contrib/clipper/clipper.cpp \
```
acgessler: By any chance, can we wildcard those as well?

workspaces/Android-NDK/jni/Android.mk
```diff
@@ -0,0 +1,134 @@
...
+	contrib/unzip/ioapi.c \
+	contrib/unzip/unzip.c \
+	contrib/irrXML/irrXML.cpp \
+	contrib/poly2tri/poly2tri/sweep/advancing_front.cc \
+	contrib/poly2tri/poly2tri/sweep/sweep_context.cc \
+	contrib/poly2tri/poly2tri/sweep/sweep.cc \
+	contrib/poly2tri/poly2tri/sweep/cdt.cc \
+	contrib/poly2tri/poly2tri/common/shapes.cc
+
+LOCAL_SRC_FILES += contrib/openddlparser/code/OpenDDLParser.cpp \
```
acgessler: Dito.

acgessler:
Thanks, looks great. Why do we end up with two ANDROID builds in Travis though? What's the difference?

arris69:
two builds came from compiler matrix? clang and cpp thats why it ends also in 8 windows and 8 linux builds...

acgessler:
I see, sure. Thanks for updating!

PR #511, opened 2015-03-26
[FBX] add support for multiple animations

VirusFree:
add support for multiple animations (by using LocalStart/LocalStop in takes) 
+
change key type from uint to int (fix bugs from negative int_t becoming junk uint_t values )


acgessler:
Looks good, some minor comments. Do you have a sample file we can add to test/ or test-nonbsd?

Thanks!

Issue #509, opened 2015-03-24
java bindings (jassimp) crashes on access to animation


arris69:
output from an example java app

```bash
converting 1 animations ...
   converting animation  ...
invalid direct buffer, expected 520, got 624
```

bug candidate port/jassimp/jassimp/src/jassimp/AiNodeAnim.java
looks like there is no padding on c struct to jni directbuffer mapping.
position key and scaling key constants should be 20 not 24.

PR #514, opened 2015-03-27
Fixes assimp/assimp#509

arris69:
looks like there is no padding on cast a c-struct into jni bytebuffer

### DIFF
--- a/port/jassimp/jassimp-native/src/jassimp.cpp
+++ b/port/jassimp/jassimp-native/src/jassimp.cpp
@@ -1340,6 +1340,14 @@ static bool loadCameras(JNIEnv *env, const aiScene* cScene, jobject& jScene)
 }
 
 
+JNIEXPORT jint JNICALL Java_jassimp_Jassimp_getVKeysize
+  (JNIEnv *env, jclass jClazz)
+{
+	const int res = sizeof(aiVectorKey);
+
+	return res;
+}
+
 JNIEXPORT jstring JNICALL Java_jassimp_Jassimp_getErrorString
   (JNIEnv *env, jclass jClazz)
 {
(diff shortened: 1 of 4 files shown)


acgessler:
I believe this is dependent on how the C compiler handles alignment for double (which is the largest member in ai***Key (see also http://stackoverflow.com/questions/11108328/double-alignment).

Can we consider a solution that is independent of underlying architecture and compiler flags?

acgessler:
I believe this is dependent on how the C compiler handles alignment for double (which is the largest member in ai***Key (see also http://stackoverflow.com/questions/11108328/double-alignment).

Can we consider a solution that is independent of underlying architecture and compiler flags?

arris69:
something like add parameters, sizeof(ai...Key)? to
```C
			const aiNodeAnim *cNodeAnim = cAnimation->mChannels[c];

			jobject jNodeAnim;
			jvalue newNodeAnimParams[6];
			newNodeAnimParams[0].l = env->NewStringUTF(cNodeAnim->mNodeName.C_Str());
			newNodeAnimParams[1].i = cNodeAnim->mNumPositionKeys;
			newNodeAnimParams[2].i = cNodeAnim->mNumRotationKeys;
			newNodeAnimParams[3].i = cNodeAnim->mNumScalingKeys;
			newNodeAnimParams[4].i = cNodeAnim->mPreState;
			newNodeAnimParams[5].i = cNodeAnim->mPostState;

			if (!createInstance(env, "jassimp/AiNodeAnim", "(Ljava/lang/String;IIIII)V", newNodeAnimParams, jNodeAnim))
			{
				return false;
			}


			/* add nodeanim to m_animations java.util.List */
```

acgessler:
Thanks for updating! I will go ahead and merge this. An alternative would be to wrap each aiVectorKey/aiQuatKey individually to Java. That's arguably nicer, but has some overhead.

PR #515, opened 2015-03-28
Regression suite fixes II

acgessler:
Mostly fixes and improvements to CompareDump code.

Achieves 100% locally for me (MSVC11).


turol:
Only (still?) 39 regression failures remain on GCC and Linux.

acgessler:
Yes. I'll look into those now that I'm confident that the regression test suite does what it is supposed to.

turol:
Interestingly enough release and debug modes cause *different* regressions. Release mode has 39, debug mode has 41. Diff between failures.csv: https://gist.github.com/turol/0cb4ac04580d2577700d

PR #517, opened 2015-03-28
Fix java final (afterparty)

arris69:
added getter for other structs too and make results as "global constants".


acgessler:
Thanks!

Issue #516, opened 2015-03-28
Android Travis build also triggers regular build


acgessler:
See for example https://travis-ci.org/assimp/assimp/jobs/56207452

arris69:
whats about cmake integration so it generates make target extra for android ndk-build and for java too?
other solution is to use shellscripts in the script: section.
i found no way to let travis select between different script: sections (but i never played around with ruby and yaml stuff)

PR #518, opened 2015-03-29
maybe Fixes assimp/assimp#516

arris69:
elif added for future use (maybe someone like to split windows and linux build or test cases too)


acgessler:
Looks good, thanks!

Issue #519, opened 2015-03-30
travis always build shared lib


arris69:
```bash
-STATIC_BUILD=$TRAVIS_STATIC_BUILD
```
makes no sense.
the cmake option that controls shared or static build of libs is: 
-DBUILD_SHARED_LIBS={ON|OFF}

acgessler:
Added in https://github.com/assimp/assimp/commit/52f3cd0afb3cba955f173ab80cec3676acf72671,

@kimkulling can you have a look?

acgessler:
Fixed by #528.

PR #522, opened 2015-04-01
Travis mingw initial

arris69:
proof of concept


kimkulling:
Please check  the travis output, android buid with mingw for windows seems to be broken. And this build config doesn't make so much sense for me as well :-).

arris69:
it makes no sense then it don't exists (android build with mingw for windows) :-)
https://travis-ci.org/assimp/assimp/builds/56724449
android build with android-ndk is ok (both compiler: gcc & llvm (clang))
linux build don't produces errors but the half of the builds are senseless then there is no difference betw. static and nonstatic builds.
i wait until #519 is fixed, after this i can take a look into fix mingw build for windows target system then it has also troubles with shared/static build (seems to be generally an error in cmakes design or in assimps cmake usage).

thnx

PR #523, opened 2015-04-01
More crash fixes

turol:
Fixes most of the remaining crashes from #454. The only one left after this is the one in #488.

No new regression failures. Please review before merge. Especially the assertion to exception and the last (too many vertices) patches. Also the last patch might require a different limit since it's still sometimes too much for Valgrind.

### DIFF
--- a/code/ACLoader.cpp
+++ b/code/ACLoader.cpp
@@ -632,6 +632,10 @@ aiNode* AC3DImporter::ConvertObjectSection(Object& object,
 								face.mIndices[1] = cur++;
 
 								// copy vertex positions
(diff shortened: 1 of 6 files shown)


kimkulling:
Please check travis output: INTPTR_MAX is not defined for all build configurations.

and thanks a lot!

turol:
Ping. I don't think this build failure is my fault since travis seems to have died on a file I didn't even change. Can someone force a rebuild?

kimkulling:
Done!

kimkulling:
Done!

PR #525, opened 2015-04-05
add strtol10_64 and fix FBX text parsing for int_64 (for animation KTime)

VirusFree:
correctly parses a negative int64 value from text


acgessler:
Looks good, thanks!

PR #531, opened 2015-04-12
double definition, i'd better open my eyes...

arris69:


acgessler:
In general, try to make the commit message as speaking as possible. It's what people read when the browse project history and file blames.

In this case, I would have suggested something along the lines of "CMake: remove duplicate BUILD_SHARED_LIBS option".

No need to fix here, but please do in future :-) Thanks!

PR #536, opened 2015-04-15
remove PCH

### LINKED ISSUE #534: "precompiled headers" slows down compilation
its not used propably. imo PCH should be removed, because:
- it pulls in a lot of unused code
- it slows down static code analyze 
- if a single header is changed, the whole code needs to be recompiled.

imo in total it doesn't safe any compilation time.

are there any thoughts about how to handle pch? imo its mostly a windows/msvc feature...

in theory there could be a cmake/macro setting added which allows to disable the usage of PCH but this would require to maintain both settings.

i personally would remove PCH as "everything depends on everything" is bad style.

abma:
this fixes #534

after this, there is still a lot todo, many existing header includes way to much / a lot of code should be moved into the .cpp files but thats out of the scope of this change as it would change everything in assimp. (its already a huge patch)

this possible breaks compile in some build environments. if so, it should be easy to fix.


kimkulling:
Thanks a lot!

Issue #295, opened 2014-06-12
Assimp 3.0 can't open PLY files it creates


kimkulling:
When I create a PLY file and then try to open it, assimp gives the following error.
"Validation failed: The mesh contains no vertices."
This seems to be the case also when trying to open it with Rhino3D. Rhino gives no error, but displays no model.

kimkulling:
Fixed

PR #538, opened 2015-04-16
Added NaN check when outputting normals in PLY exporter, to avoid '-1.#IND'

cubeman99:
This should address the issue in #295, because previously the PLY importer was throwing an exception when trying to parse '-1.#IND'.


acgessler:
Looking good, see one note on commit above.

acgessler:
Thanks!

PR #541, opened 2015-04-19
Fix compilation with libc++

Ang3lus:
Compilation errors when using libc++:

```
 assimp/code/ObjFileParser.cpp:360:30: error: use of
      undeclared identifier 'atoi'
            const int iVal = atoi( pPtr );
```

```
assimp/code/Q3BSPZipArchive.cpp:143:13: error: use of undeclared identifier 'malloc'
        m_Buffer = malloc(m_Size);
                   ^
assimp/code/Q3BSPZipArchive.cpp:147:2: error: use of undeclared identifier 'free'
        free(m_Buffer);
```

Steps to reproduce:
* Compile with libc++ library


acgessler:
I think this has already been fixed here: https://github.com/assimp/assimp/commit/f435712273a7bfec295e341e1d19910d837b4303

Ang3lus:
Nope. Both errors are still there.

acgessler:
You're right, different files. Thanks!

PR #544, opened 2015-04-20
Modifying files to let assimp build under cygwin.

aru:
Changing add_definitions() to set(CMAKE_CXX_FLAGS) where I know it is appropiate. Also adding a top-level line so that cygwin cmake stops complaining when trying to build assimp.

Renamed ByteSwap.h -> ByteSwapper.h
** Also changed file in:
workspaces/xcode3/assimp.xcodeproj/project.pbxproj
although I'm not sure how that file works, is it created by xcode?

Just noticed my emacs erased all whitespace on some lines, I apologize, I can correct this if it is a problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,4 @@
+set(CMAKE_LEGACY_CYGWIN_WIN32 0) # Remove when CMake >= 2.8.4 is required
 cmake_minimum_required( VERSION 2.8 )
 PROJECT( Assimp )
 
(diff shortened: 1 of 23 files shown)


acgessler:
Thanks for doing this! While removing the whitespace would have been nicer in a separate commit, it's definitely a good thing to do :-)

aru:
np, do tell me if something else is needed, I'll monitor the repo for changes

CMakeLists.txt
```diff
@@ -61,9 +62,9 @@ if( CMAKE_COMPILER_IS_MINGW )
 endif()
 
 if((CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX) AND NOT CMAKE_COMPILER_IS_MINGW)
-    add_definitions(-fPIC) # this is a very important switch and some libraries seem now to have it....
+    set(CMAKE_CXX_FLAGS "-fPIC") # this is a very important switch and some libraries seem now to have it....
```
jdduke: For future reference, setting flags like this *overwrites* the previous values, it does not append...

aru: I had no idea, sorry about that, do you need a fix or has it been corrected?

aru: I see that the issue has been corrected. Thanks.

jdduke:
No worries, I've surely done the same thing before.

PR #548, opened 2015-05-01
Removed FORCE attribute from CMake variable LIBRARY_SUFFIX

Highlife1911:
I don't see a reason for forcing the value of LIBRARY_SUFFIX. 
Removing the force attribute would allow experienced users to specify their own suffix format. The default behavior doesn't change with this patch.


acgessler:
Agree, thanks!

PR #549, opened 2015-05-01
Step (stp) Exporter

Madrich:


code/StepExporter.cpp
```diff
@@ -0,0 +1,380 @@
...
+// ------------------------------------------------------------------------------------------------
+// Worker function for exporting a scene to Collada. Prototyped and registered in Exporter.cpp
+void ExportSceneStep(const char* pFile,IOSystem* pIOSystem, const aiScene* pScene, const ExportProperties* pProperties)
+{
+	std::string path = "";
+	std::string file = pFile;
+
+	// We need to test both types of folder separators because pIOSystem->getOsSeparator() is not reliable.
+	// Moreover, the path given by some applications is not even consistent with the OS specific type of separator.
+	const char* end_path = std::max(strrchr(pFile, '\\'), strrchr(pFile, '/'));
```
acgessler: Please use `if` here for readability.

Also, I believe `std::max(a, b)` where one of `a`, `b` can be `nullptr` is technically undefined behaviour (obviously it works though). Pointer comparisons are well-defined only within the same object (array), which is not true if one side is `nullptr` see http://stackoverflow.com/questions/1098966/universal-less-for-pointers-in-c-standard.

code/StepExporter.cpp
```diff
@@ -0,0 +1,380 @@
...
+{
+	// see http://shodhganga.inflibnet.ac.in:8080/jspui/bitstream/10603/14116/11/11_chapter%203.pdf
+	// note, that all realnumber values must be comma separated in x files
+	mOutput.setf(std::ios::fixed);
+	mOutput.precision(16); // precission for double
+	
+	int ind = 100; // the start index to be used
+	int faceEntryLen = 30; // number of entries for a triangle/face
+	// prepare unique (count triangles and vertices)
+	std::map<aiVector3D*, int> uniqueVerts; // use a map to reduce find complexity to log(n)
```
acgessler: I think a hashmap is the preferred data structure here. Ordering doesn't matter and it'll give you amortized O(1).

Have a look at `code/FBXCompileConfig.h` for how to get a safe `unordered_map` type.

code/StepExporter.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+	mScene = pScene;
+	mSceneOwned = false;
+
+	// set up strings
+	endstr = ";\n"; 
+
+	// start writing
+	WriteFile();
+
```
acgessler: Please drop whitespace.

acgessler:
Very nice, thanks for tackling this! Some comments inline.

code/XFileExporter.cpp
```diff
@@ -514,12 +514,16 @@ void XFileExporter::WriteMesh(aiMesh* mesh)
...
-	std::replace(str.begin(), str.end(), '}', '_');
-	std::replace(str.begin(), str.end(), '$', '_');
+	std::string pref = "NN_"; // node name prefix to prevent unexpected start of string
+	std::string str = pref + std::string(name.C_Str());	
+	for (int i=0; i < (int) str.length(); ++i)
+	{
+		if ((str[i] >= 48 && str[i] <= 57) || // 0-9
+			(str[i] >= 65 && str[i] <= 90) || // A-Z
+			(str[i] >= 97 && str[i] <= 122)) // a-z
+			continue;
```
acgessler: You can directly compare against character literals such as `0`.

code/XFileExporter.cpp
```diff
@@ -514,12 +514,16 @@ void XFileExporter::WriteMesh(aiMesh* mesh)
 
 std::string XFileExporter::toXFileString(aiString &name)
 {
-	std::string str = std::string(name.C_Str());
-	std::replace(str.begin(), str.end(), '<', '_');
-	std::replace(str.begin(), str.end(), '>', '_');
-	std::replace(str.begin(), str.end(), '{', '_');
-	std::replace(str.begin(), str.end(), '}', '_');
-	std::replace(str.begin(), str.end(), '$', '_');
+	std::string pref = "NN_"; // node name prefix to prevent unexpected start of string
```
acgessler: Does this mean every name in the output file has a NN_ prefix?

code/StepExporter.h
```diff
@@ -0,0 +1,110 @@
...
+	/// The scene to be written
+	const aiScene* mScene;
+	bool mSceneOwned;
+
+	/// current line end string for simple stream insertion
+	std::string endstr;
+
+	std::map<const aiNode*, aiMatrix4x4> trafos;
+
+	typedef std::multimap<const aiNode*, unsigned int> MeshesByNodeMap;
```
acgessler: I just saw this is from 3DS, where I've used it.

It's a bad idea, we shouldn't rely on orderings of memory addresses, which is different each time.

I'll fix that one for both exporters after this PR is in.

code/XFileExporter.cpp
Madrich: Yes, to make sure the name can not start with "_". NN is for NodeName.

code/XFileExporter.cpp
Madrich: I do not understand what you mean

acgessler: `str[i] >= '0' && str[i] <= '9'`

code/XFileExporter.cpp
acgessler: I don't like such a prefix. If we re-import it, the scene would be changed. Why can it not start with _ - does the X file spec forbit it and it would not parse again?

code/StepExporter.cpp
Madrich: Done

code/StepExporter.cpp
Madrich: Done

code/StepExporter.cpp
Madrich: Done

code/XFileExporter.cpp
Madrich: I'm not sure, but i think it must start with a letter.

acgessler: Please doublecheck. If assimp's x importer is able to read it with a leading underscore and there is no other importer who can't then we should allow it.

Otherwise, please add the prefix only if the name begins with an underscore.

Madrich: Not in this Merge-Request. I will do this later. Actually, too much work for me.

acgessler: Ok, no worries. I checked: our importer handles leading underscore correctly (or rather, it does not have a check to prevent it).

Please drop the prefix.

Madrich:
All done.
In addition add Color support for Step export

acgessler:
Thanks for updating!

winkelsdorf:
Thank you! Any chance to see an importer, too?

PR #552, opened 2015-05-06
allow to compile only enabled importers

abma:
this allows to disable/enable importers via config setting. when disabled the .cpp isn't compiled at all which will reduce file size and compile time.

(again) a proof of concept:
very likely most combinations will fail atm, but the default shouldn't change anything at all.

thoughts about this?


kimkulling:
The default is using all, right? And this is a great idea, thanks for the pull request!

abma:
yes, right!

acgessler:
+1, very nice!

kimkulling:
+1 from me, too!

PR #558, opened 2015-05-16
test fail, fix file path

wise86-android:
fix the test utImporter, the file path was wrong


kimkulling:
Thanks a lot, merged!

PR #559, opened 2015-05-16
Fixed AssBin loader.

shocker-0x15:
File reads were used in ai_assert. The function has side effects, so it should not be used in an assert.
Fixed AddGTest.cmake.
There was an issue regarding Xcode project file. The path of libgtest.a was incorrectly set.


acgessler:
Looks good but I believe the change to /AddGTest.cmake broke the CI build, please fix.

shocker-0x15:
Fixed it. Now changes are only in ASSBIN loader.

acgessler:
Thanks!

PR #560, opened 2015-05-16
Fix inits and a couple of empty checks

asmaloney:
- add missing initialization of several class/struct members
- prefer check for empty() over check for size() in OptimizeGraph (empty() is guaranteed to be constant time, size() is not)

### DIFF
--- a/code/BaseImporter.cpp
+++ b/code/BaseImporter.cpp
@@ -480,7 +480,9 @@ namespace Assimp
 struct Assimp::BatchData
 {
 	BatchData()
-		:	next_id(0xffff)
+        : pIOSystem()
+        , pImporter()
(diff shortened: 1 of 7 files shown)


acgessler:
Yes, please. Also for initialization to the default value - false, NULL/nullptr and 0 the value is optional, empty parens will have the same effect, i.e. `: field()`.

asmaloney:
If this is the style you want to "encourage", maybe add it to *CodeConventions.txt*?

I was attempting to be consistent with the code around the areas I was changing.

kimkulling:
Good idea to introduce a conding convention documentation.

kimkulling:
And thanks for the work, merged!

PR #562, opened 2015-05-18
Coding Convention

asmaloney:
- use markdown for coding convention
- rewrite it


asmaloney:
@kimkulling  This is just a doc change and can be merged without CI.  The build was failing due to a merge issue which is fixed in another branch.  Doc-only changes like this shouldn't really trigger a CI build anyways...

LegalizeAdulthood:
A good way to ensure coding conventions for whitespace is to use a `.editorconfig` file and associated plugins from [EditorConfig.org](http://editorconfig.org).  This will let your editor apply the whitespace conventions consistently and automatically.  For instance, while fixing compiler warnings, I noticed that the EOL in the source files is all over the place with a mix of CRLF and LF endings :-).

asmaloney:
I'm all for adding a *.editorconfig*, but it means that checkins are going to be noisy for a while as it gets applied to the files.  Unless someone does it all in one pass...

Here's what I propose (and can add to this PR if the maintainers are on board):
```
[*.{h,hpp,c,cpp}]
# Unix-style newlines with a newline ending every file
end_of_line = lf
insert_final_newline = true

# 4 space indentation
indent_style = space
indent_size = 4

trim_trailing_whitespace = true
```

LegalizeAdulthood:
Let me split `.editorconfig` into a separate issue, which I'll be happy to apply all-at-once.

kimkulling:
I do not agree with writing foo(bar) instead of foo( bar ).

acgessler:
I think going tab->spaces etc, no trailing whitespace and unified line endings makes sense.

I'm rather against enforcing more stylistic things like braces or spaces around parameters, parens and other.

turol:
I'm against using spaces for indentation. Tabs should be used to indentation and spaces only for alignment (comments, multi-line ifs etc).

asmaloney:
@kimkulling I agree - I prefer ```foo( bar )``` - I was just looking at what the code currently looks like.  The code is not consistent.

OK - I'll let you guys fight it out.  I was just going to get the ball rolling...

acgessler:
clang format is amazing, using it daily. However not everyone has it available locally - we can run it in Travis CI and reject CLs that are not formatted correctly, but personally I find this is too much nitpicking. As a compromise, we could run it over the entire code base occasionally.

Even if, a prerequisite is to agree on one style. As this thread shows, it's a near religious question.

I also think consistent code formatting is less important for assimp going forward than more and, most of all, better test coverage.

Let's not forget that the original cause was formatting of initialization lists (iirc). There the reasoning is simple: for new code prefer a new line for each field since it gives a clearer diff.

turol:
If you're going to do a Huge Formatting Change I'd appreciate if you would also fix the line endings. Currently the files are stored in the repository with windows line endings (CR LF) and the settings are not right so they don't autoconvert on Linux.

For initializer list also remember to specify commas at the beginning of the line, not end. This way when someone adds a new member to the end of the list or reorders them there are no spurious comma-only changes.

LegalizeAdulthood:
`clang-format` is great **IF** you agree with one of their coding styles.  If you have any minor preferential deviation from their coding styles, you're basically screwed unless someone else has implemented a tweakable option to implement your preferences.  `AStyle` is much more tweakable, but doesn't support the crazy formatting that the clang/LLVM teams prefer.  I say this as a contributor to clang tools :smile_cat:.  I have my own personal preferences that already clash with some of the comments on this thread, but ultimately I am not a significant contributor to assimp and IMO the style decisions should be made by the major contributors.

kimkulling:
Do we really want to have the same coding style for everyone? Normally this causes a lot discussion which creates not any value.

Using spaces instead of tabs makes sence from my point of view. Special when people using different settings for tabs. We made really good experiences with it. 

What we could add are things like let start attributes with a m_ or a m .

acgessler:
>> What we could add are things like let start attributes with a m_ or a m .

Absolutely not ;-)

kimkulling:
Hehe, I see! :-D

asmaloney:
> Do we really want to have the same coding style for everyone?

Having worked on many different projects over many years of writing & designing software I do see a value in keeping code consistent.

Code is communication.  It's less about communicating with the computer than it is about communicating with other humans.  The more consistent a codebase is, the easier it is to read, understand, and refactor.

You will never find a format that makes everyone happy because each of us has our own style, but when you're collaborating with others, you have to let go of these little things to focus on the bigger things.  (For me the comma at the beginning of the line in the initializer or jamming things together without spaces drives me nuts - but I'm totally on board with prefixing class members with *m*!)

Like @LegalizeAdulthood I am not a significant contributor, so ultimately the stye decision should come from the main contributors, but I do think some basics (spacing, line endings, any naming conventions, preferring initializers, should build without warnings) should be defined for any project.

Or you could just rewrite in Go and use [gofmt](https://blog.golang.org/go-fmt-your-code) :-)

LegalizeAdulthood:
If we can make a little forward progress by getting agreement on whitespace conventions, then I can update pull request https://github.com/assimp/assimp/pull/566 and at least achieve consistency there.  When I fixed some warnings recently, I noticed EOL was all over the place :smile_cat:.

kimkulling:
I guess there is n agreement about whitespaces, so this could be a good starting point, right?

asmaloney:
@acgessler I disagree with using "default initializers" - I prefer it to be explicit and readable - but I got rid of the whole section on initializers when I resubmitted anyways.  I only include a section on whitespace.  (Take a look at the *Files Changed* tab above for the entire contents of this PR.)

acgessler:
@asmaloney I see, missed that.

+1 for the current change proposed from my point of view.
@kimkulling?

acgessler:
Merged :-)

asmaloney:
:+1:  Got to start somewhere!

kimkulling:
+1 from me as well, sorry for my late response!

PR #563, opened 2015-05-18
Fix overlapping merges

asmaloney:
One of the two branches should have been rebased before merging since they both touch the same code...


acgessler:
Thanks for noticing!

acgessler:
@kimkulling Let's not merge pull requests without waiting for new CI builds.

asmaloney:
If I may make another suggestion - if whoever is doing the merging has merged a PR, then before merging any others they should require they be rebased first so it's up the the requester to fix any issues.

That way you end up with clean merges and avoid the problem we had here which broke the build.

kimkulling:
@acgessler Sounds good!

PR #570, opened 2015-05-23
Collada export light and camera

wise86-android:
add the ability to export light and camera object in a collada file


acgessler:
Great. Thanks for also including tests :-)

PR #571, opened 2015-05-23
Fix issue with CMAKE_CXX_FLAGS initialization

jdduke:
Append the necessary flags to CMAKE_CXX_FLAGS, rather than replacing
the previous values. It's critical that that flags are preserved to
ensure proper and consistent compilation.


acgessler:
Good catch, thanks!

PR #580, opened 2015-06-04
Several minor improvements based on cppcheck

rob100:
I did several small code improvements based on a cppcheck result (cppcheck is a static code analyzer). The changes are minimal to prevent any breaking changes. For example:
- Usage of pre-inrement in loop instead of post-increment
- Use of correct printf format string
- Usage of .empty() instead of .size() in if statements
- Usage const reference parameter


kimkulling:
Some notes:
The ObjFileParser Default constructor and assignment operator must not be imlemented.

rob100:
You are right, I removed the implementation. 
I noticed that the pull request has to many commits, only the last two are relevant: 
fb393a8 and 518a3f3
Do you have the possibility to consider this on a merge?
I can also try to fix this but I am not so experienced in git

acgessler:
Unfortunately Github allows no cherrypicking. The best thing to do is to make a new PR I'm afraid.

rob100:
OK, I will make a new Pull Request, sorry for the trouble.

PR #581, opened 2015-06-05
Several minor improvements based on cppcheck

rob100:
Now with the correct commits :)
I did several small code improvements based on a cppcheck result (cppcheck is a static code analyzer). The changes are minimal to prevent any breaking changes. 
For example:
- Usage of pre-inrement in loop instead of post-increment
- Use of correct printf format string
- Usage of .empty() instead of .size() in if statements
- Usage const reference parameter
- Added private constructor for class that should not be copied

### DIFF
--- a/code/ASEParser.cpp
+++ b/code/ASEParser.cpp
@@ -143,9 +143,9 @@ void Parser::LogWarning(const char* szWarn)
 
 	char szTemp[1024];
 #if _MSC_VER >= 1400
-	sprintf_s(szTemp,"Line %i: %s",iLineNumber,szWarn);
+	sprintf_s(szTemp,"Line %u: %s",iLineNumber,szWarn);
 #else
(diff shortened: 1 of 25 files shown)


code/ASEParser.cpp
```diff
@@ -143,9 +143,9 @@ void Parser::LogWarning(const char* szWarn)
 
 	char szTemp[1024];
 #if _MSC_VER >= 1400
-	sprintf_s(szTemp,"Line %i: %s",iLineNumber,szWarn);
```
jdduke: Why do we use the "i" prefix for unsigned ints? There's some pretty funky hybrid Hungarian notation going on in some of these files, seems more confusing than anything else (perhaps excepting pointers?).

acgessler: It's simply very, very legacy code (for some of which I'm responsible).

PR #585, opened 2015-06-19
Fixes random crash when loading fbx files

pampersrocker:
We had an issue when loading fbx files without any debugger attached (or `_NO_DEBUG_HEAP=1` set, so memory is not filled with predefined values), because there was no trailing `\0` set when reading the fbx file.
This fixes it.

Without this fix sometimes more tokens where generated than there are actually in this file (depending on what was in the memory before), causing the parser to fail at the last tokens.


kimkulling:
Looks great, thanks a lot!

PR #587, opened 2015-06-19
Correct value for mColorAmbient for COLLADA files

rdb:
ColladaLoader currently writes out an mColorAmbient value for all lights that is equal to the diffuse and specular colors.  In engines that correctly recognize ambient colors for lights, this will result in an all-white scene.

With this patch, it now correctly assigns it to zero, except for ambient lights, in which case it writes the light color into the ambient slot only (and not to the diffuse/specular slots).


kimkulling:
Looks good, but please check the build, seems to fail for your pull request!

acgessler:
Regression suite needs updating as per the instructions at the bottom of the CI build output.

acgessler:
Friendly ping :-)

rdb:
Sorry, I hadn't forgotten about this, I'd just been too busy to get around to it.

I've updated the commit and it now passes the regression tests.

FWIW, I had a bit of trouble running the tests locally since the OutOfMemory model would lock up my machine as it is starved of resources, though this doesn't seem like a problem that is related to the collada loader.

acgessler:
Perfect, thanks for updating.

I'll look into removing the OutOfMemory file.

PR #588, opened 2015-06-20
Added mesh name importing to .X ascii format.

zippers:
assimp doesn't import mesh names from the ascii .X format.

You can verify this by viewing [this model](https://gist.github.com/zippers/c742cff8483e52cc6cc5) in the assimp viewer and observing it contains two meshes named *Mesh 0* and *Mesh 1*.

After applying the patch, you can observe the names *ninja* and *blade* are used instead.


kimkulling:
Looks great, thanks a lot!

PR #593, opened 2015-06-30
Fixed: Remove absolute path from exported file

LegalizeAdulthood:
This fixes the only test that fails for me when I run them.


kimkulling:
Thanks a lot!

PR #596, opened 2015-07-03
Initialize all members in c'tors

LegalizeAdulthood:
Corrects issues found by cppcheck for classes that don't initialize all their members


kimkulling:
looks good, thanks a lot!

PR #600, opened 2015-07-04
Prefer Format::operator<< over Format::operator,

LegalizeAdulthood:
cppcheck gives false positives for use of the comma operator, so use << operator.


kimkulling:
Looks good, thanks a lot!

PR #601, opened 2015-07-04
Windows build improvements

russell-taylor:
This lets the code build out of the box on my Windows machine, where it cannot find DirectX.  It also fixes a library issue (libm.a) on Visual Studio, where this library is not needed; this lets it build once I turn on ASSIMP_BUILD_SAMPLES and find GLUT.


acgessler:
Thanks!

PR #602, opened 2015-07-04
Builds out of the box on mac with ASSIMP_BUILD_SAMPLES turned on.

russell-taylor:
Knows where to look for glut.h on Mac with Clang compilers (command-line compilers under current XCode).

### DIFF
--- a/samples/SimpleOpenGL/Sample_SimpleOpenGL.c
+++ b/samples/SimpleOpenGL/Sample_SimpleOpenGL.c
@@ -14,7 +14,11 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


acgessler:
Thanks!

(note I'm not able to test this since I don't typically build on Mac, but it seems reasonable).

russell-taylor:
Thank you for putting this code out there.  It let me open a 3D reconstruction of flooding from disaster responders in Texas and pull it into OpenGL (at least your textured simple sample app).  Saved me a lot of time.

I did test the build on the mac and then back on Windows and Linux afterwards, so I think it is okay.

Thanks for pulling these fixes in so quickly -- I can get rid of my fork now. :-)

PR #603, opened 2015-07-04
Fix compiler warnings

russell-taylor:
Picking nits.


acgessler:
Looks good. Note that OpenDDLParser and poly2tri are external libraries so it should also be fixed upstream. Poly2tri is not the latest version, but we've had trouble porting our code to the newest one.

@kimkulling for OpenDDLParser

PR #614, opened 2015-07-21
Include correct headers, fix string length bug

bschwind:
I was integrating Assimp into an android project that uses native-activity, and ran into a few issues with the AndroidJNIIOSystem. These are the changes I had to make to get it working.

- AssimpPCH.h appears to have been removed in #536 
- There was a bug where if your model data had a null terminator, AndroidJNIIOSystem would incorrectly calculate the number of bytes it needed to write to the file (since it reads a compressed file from the Android assets, and then writes it to the internalDataPath directory for later importing)


acgessler:
Merged - thanks for patching!

bschwind:
No problem, thanks for the library!

PR #615, opened 2015-07-27
I made changes to the Collada importer so it will import SketchUp DAEs.

wilshipley:
There's a bug in the current (and previous) SketchUps in that when they output lines in DAE (Collada) format they'll output the wrong count on the number of items in the line array. The old Collada importer would throw an exception. I patched it in a couple places so that it now just emits a warning and corrects the error(s).

ALSO in this pull request I've made a project for Xcode 6 that can build a static library. This is a work in progress—it works but I'm actually going to change this to a framework (module) now that iOS supports frameworks, since they're SO much easier to deal with. (And modules can be used from Swift, which is nice.)

If you just want the Collada fixes just pull those two files—you don't need the Xcode stuff.

UNFORTUNATELY when I checked out the latest assimp/assimp to my branch I also have apparently changed a bunch of unrelated files because of a CRLF▶LF conversion. I am not sure if this was intentional or not.

So if you have instructions for me on how to revert all the irrelevant files I'm all ears!


kimkulling:
Please fix the build, there are build issues in the Collada sources.

wilshipley:
Sorry about that, I'm new to C++ AND to github and travis.

I've copied in a header import from another lie that used stdarg, so hopefully this commit will go better (for some reason it works on OS X without the header import).

I'll try to figure out how to check my travis log myself to see if this compiles on the server.

Thanks,
-W

wilshipley:
Ok if I'm reading this right the latest tests on my head worked. Is this ok now?  Do I need to submit a new pull request with my new head? Sorry I'm so new to this.

kimkulling:
Merge, thanks a lot for fixing it!

Gargaj:
There's a bunch of junk files in this commit:
```
ColladaParser upstream.cpp
ColladaParser upstream.h
ColladaParser wil.cpp
```

PR #618, opened 2015-08-02
Fix volume calculation

triplepointfive:
It includes the same part `fDelta1_y * fDelta1_z` (`fDelta1_yz` defined above as `fDelta1_yz = fDelta1_y * fDelta1_z`) on both sides of comparison, that makes no sense. The comment above shows the original intention to compare volumes. No doubt, the copy-paste issue is here.


acgessler:
Awesome - thanks for catching this!

PR #625, opened 2015-08-06
Fix out-of-bounds write on malformed input

turol:
Fixes hangs/14fa1793a3beb6e7db761f7cbfcd3f0c and ef7d28a7282aa3d29fad4fbadba7a29f from #454 . No (new) regression failures on Linux. Please test on windows before merge. @acgessler how's that regression suite fix coming along? Also feel free to change the error message to something more descriptive.


kimkulling:
Looks good, thanks.

PR #627, opened 2015-08-07
Abort early when hitting unexpected EOF in AC3D loader

turol:
Without this the code will try to loop through the specified number
of surfaces which could be very large even though none will succeed.

This helps crashes/cb821dbb50a427d2954c6dd12b9132a9 from #454. It was taking excessively long time since it declares 10 million surfaces and then EOFs. The code would loop 10 million times and not do anything useful.

Also someone who can update the test suite might want to turn that one into a real test case.


acgessler:
Looks good, thanks!

PR #629, opened 2015-08-10
Avoid reading past EOF when encountering a malformed STL file

turol:
Since IsSpaceOrNewLine returns true on '\0' we might try to read past
end of buffer on line 310. Add explicit check to avoid this.

A better fix might be to change IsSpaceOrNewLine to not accept '\0' but that might break other things.

Fixes crashes/9be812f32156c36fc89b4498b5d54ac2 from #454. No new regression failures but STL is one of the tests which always fails on Linux so check on a supported platform before merge.


kimkulling:
Merged, thanks!

PR #632, opened 2015-08-11
Random crash fixes

turol:
Fix some random crash bugs which might or might not be in #454.

Add my standard disclaimer about review, regression test etc.

Also it would probably be useful to abstract out that memory allocation check pattern in the last commit. Thoughts?

### DIFF
--- a/code/ACLoader.cpp
+++ b/code/ACLoader.cpp
@@ -587,9 +587,19 @@ aiNode* AC3DImporter::ConvertObjectSection(Object& object,
 
                 // allocate storage for vertices and normals
                 mesh->mNumFaces = (*cit).first;
+                if (mesh->mNumFaces == 0) {
+                    throw DeadlyImportError("AC3D: No faces");
+                } else if (mesh->mNumFaces > std::numeric_limits<int32_t>::max() / sizeof(aiFace)) {
+                    throw DeadlyImportError("AC3D: Too many faces, would run out of memory");
+                }
                 aiFace* faces = mesh->mFaces = new aiFace[mesh->mNumFaces];
 
                 mesh->mNumVertices = (*cit).second;
+                if (mesh->mNumVertices == 0) {
+                    throw DeadlyImportError("AC3D: No vertices");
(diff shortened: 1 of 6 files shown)


code/ACLoader.cpp
```diff
@@ -584,9 +587,19 @@ aiNode* AC3DImporter::ConvertObjectSection(Object& object,
 
                 // allocate storage for vertices and normals
                 mesh->mNumFaces = (*cit).first;
+                if (mesh->mNumFaces == 0) {
+                    throw DeadlyImportError("AC3D: No faces");
+                } else if (mesh->mNumFaces > std::numeric_limits<int32_t>::max() / sizeof(aiFace)) {
```
acgessler: This check looks like it presume the boundary of the 32 bit address space? We would run out of memory (both physical as well as virtual space) much earlier.

I'd suggest setting a constant upper bound of i.e. 100 million.

turol: Actually it's 32 bits since it uses signed int. I more or less picked it out of a hat.

The reason for both the check and and strange limit is that Valgrind and AddressSanitizer don't really handle running out of memory (or address space) and just crash the program. This makes debugging harder and gives false positives when fuzzing.

This is used in several places. By putting in one place it becomes easy to change (to whatever we think is good). Where should we put the general version and what should it be called? I was thinking something like
\#define AI_MAX_MEMORY_LIMIT(type) (bignumber / sizeof(type))

turol:
Updated the check logic so the limit can be changed in one place and lowered the default to 256M.

turol:
@heandreas @kimkulling Your latest changes conflict with this in a nontrivial way. You duplicated a block of code I'd fixed. I have fixed it locally. Should I push a new version of this or are you still going to do something in master (for example fix the new regression failures)?

heandreas:
I am looking into the regression failures now. There seems to be a problem with the node name of the imported mesh.

turol:
Rebased against latest master because of conflicts.

kimkulling:
Merged!

PR #633, opened 2015-08-13
Multiple meshes in single STL ascii file and binary bly exporter fix

heandreas:
A few month ago, I changed assimp's ascii STL loader so that it can handle STL files that contain multiple meshes. Recently, I had trouble merging your formatting changes into my fork, which reminded me that I forgot to create a pull request. I also refactored the code so that it uses std::vector for dynamic resizing of the vertex buffer, but this can be reverted if you think that the custom resize policy is crucial for performance (I do not think so).

In the meantime, I added another change that concerns the binary .ply exporter. I noticed that MeshLab could not load .ply files exported using assimp and Houdini failed in very rare cases. I opened some .ply files that MeshLab could load and noticed that it uses uchar as data type for the number of vertices per triangle and int as data type for the indices. The exporter I contributed to assimp uses uint for both types. I changed the exporter so that it uses uchar and int and now MeshLab can load all exported files, the Houdini issue seems fixed as well. In my opinion, assimp's exporter was correct, but changing assimp is the simplest and fastest way for me to fix my problem of getting thousands of fluid meshes from our simulation tool to Houdini for Rendering.
The disadvantage is that the exporter can no longer handle polygons with more than 255 vertices (for me this is completely irrelevant).


heandreas:
Hi,
thanks for merging my changes. I am surprised that you merged so quickly, what about the failed STL CI tests? I don't know how exactly your checks work so this could be normal since I changed a lot, but it is still interesting I think.

PR #635, opened 2015-08-14
IFC: Don't store temporary values by reference

turol:
This is not legal in any way and should never have worked.

This actually caused GCC to generate bad code when optimization was on. Reduces regression failures on Linux from 54 to 49. You should probably update the regression whitelist too.

### DIFF
--- a/code/IFCOpenings.cpp
+++ b/code/IFCOpenings.cpp
@@ -602,12 +602,12 @@ bool IntersectingLineSegments(const IfcVector2& n0, const IfcVector2& n1,
     const IfcVector2& m0, const IfcVector2& m1,
     IfcVector2& out0, IfcVector2& out1)
 {
-    const IfcVector2& n0_to_n1 = n1 - n0;
+    const IfcVector2 n0_to_n1 = n1 - n0;
 
-    const IfcVector2& n0_to_m0 = m0 - n0;
-    const IfcVector2& n1_to_m1 = m1 - n1;
+    const IfcVector2 n0_to_m0 = m0 - n0;
+    const IfcVector2 n1_to_m1 = m1 - n1;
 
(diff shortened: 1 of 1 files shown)


turol:
Looks like this failed on STL failures which also happen on current master. Please don't merge this yet, there are more bad references in this file. I'll fix them while I'm here.

@acgessler git blame blames you for this code.

turol:
Now you can merge it. I only checked this one file so more of these might lurk in others.

turol:
Please force a rebuild because the checks were already broken in master when I created this.

kimkulling:
Merged, thanks a lot!

turol:
Actually I was wrong about the specific cause of this bug. In C++ a const reference extends the lifetime of a temporary. However in this case (line 1230):
1. The ^ operator is evaluated, producing a temporary IfcVector3
2. Normalize() is called on this temporary. The lifetime of the temporary is extended to the lifetime of the reference. This is NOT the local variable face_nor but instead the implicit "this" parameter of Normalize().
3. Normalize() runs and returns a reference to "this". This would be safe if "this" were a normal value which in this case it isn't.
4. The reference is stored in loal reference variable face_nor. The extended lifetime of the temporary ends. The reference now points to a destroyed object. Referencing it is undefined behavior.
5. Since the lifetime of the reference ends on line 1237 and the only other variable in this scope is a float this will work fine most of the time. However on GCC with -O2 (but not -O1 or -O0) the optimizer will do Bad Things and generate bad code.

TL;DR: It was a real bug and this was the correct fix but not for the reasons I though it was. And do not EVER use this feature of C++, it's fraught with peril. Just store locals by value and avoid getting your legs blown off. All code in the project should be audited and this idiom removed.

PR #636, opened 2015-08-14
Fixed issue in STLLoader that lead to wrong node names.

heandreas:
aiNode was created twice, sorry about that.

This fixes all STL regression errors concerning the node names. There are still some STL regression errors, all of them say something like "Files are different at aiMesh.mFaces. Error is: Expected 4115780318, but actual is 2985128783". Is it possible that the tool is comparing pointers? All models load fine...


kimkulling:
Thanks for the fix and don't worry :-).

Issue #495, opened 2015-03-14
Regression: Bob.md5mesh


acgessler:

acgessler:
Also (probably same root cause) SimpleCube.md5mesh

skyjake:
Is this what you're seeing, too? Seems to affect all MD5 models.
![md5_regression](https://cloud.githubusercontent.com/assets/1225742/8607411/5175644a-269b-11e5-9934-3f271153ed2e.jpg)

kimkulling:
This issue was fixed, so I will close it. Just checked th sample model.

PR #640, opened 2015-08-18
MD5: Fix quaternion conversions

skyjake:
Some of the quaternions loaded from MD5 files were not converted to follow Assimp conventions.

https://github.com/assimp/assimp/issues/495


acgessler:
Sorry for the delay.

This looks good, but it obviously changes output for MD5. Can you run the gen_db.py script on MD5 and add the updated regression database to the patch? This will make the Travis build pass (see Travis build output).

skyjake:
Looks like the Travis build would've passed except for this:
```fatal: unable to access 'https://github.com/assimp/assimp.git/': Couldn't resolve host 'github.com'```

acgessler:
Many thanks! Yes, the other clang build passed so we should be good. I haven't found the button yet to restart a flaky CI run.

I will be brave and merge.

Issue #642, opened 2015-08-20
Broken code/cmakelist.txt on OSX


FengLinqing:
"cmake -DCMAKE_BUILD_TYPE=Release ../assimp" gives me error on OSX 10.10.5:

-- The C compiler identification is AppleClang 6.1.0.6020053
-- The CXX compiler identification is AppleClang 6.1.0.6020053
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find PkgConfig (missing:  PKG_CONFIG_EXECUTABLE) 
-- Building a non-boost version of Assimp.
-- Looking for ZLIB...
-- Could NOT find PkgConfig (missing:  PKG_CONFIG_EXECUTABLE) 
-- Found ZLIB: optimized;/Users/feng/miniconda3/lib/libz.dylib;debug;/Users/feng/miniconda3/lib/libz.dylib
CMake Error at code/CMakeLists.txt:660 (SET_TARGET_PROPERTIES):
  set_target_properties Can not find target to add properties to: assimp

-- Enabled formats: 3DS AC ASE ASSBIN ASSXML B3D BVH COLLADA DXF CSM HMP IRR LWO LWS MD2 MD3 MD5 MDC MDL NFF NDO OFF OBJ OGRE OPENGEX PLY MS3D COB BLEND IFC XGL FBX Q3D Q3BSP RAW SMD STL TERRAGEN 3D X
-- Disabled formats:
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- Found Threads: TRUE  
[...]

PR #645, opened 2015-08-26
CMake: Don't try to set property on target before it is defined.

l337r007:
This PR fixes the issue reported in #642 . Happens to me as well when trying to build HEAD via homebrew.


LegalizeAdulthood:
Shouldn't there be a corresponding addition somewhere else after the target is defined?

acgessler:
Thanks!

PR #651, opened 2015-09-02
fix #634, remove -fPIC for static builds / 32 bit

### LINKED ISSUE #634: Option -fPIC is used even when shared library is disabled
When shared library is disabled (with -DBUILD_SHARED_LIBS=0 option to CMake) it still adds -fPIC option (generate position independent code). While this is not strictly incorrect it is inefficient since among other things it wastes one register on 32-bit x86 which really doesn't have any to spare.

This is a Linux-only problem since Windows doesn't use PIC.

Probably in top-level CMakeLists.txt the condition on line 65 needs to change but I'm not entirely sure how.

abma:
-fPIC is only required on 64 bit / shared libs (afaik).

the other commit is minor, it just hides stderr


kimkulling:
Merged, thx a lot!

PR #664, opened 2015-09-28
Added auto-detection of numpy

Tcll:
if you don't have numpy installed, pyassimp will substitute local operations on raw python lists and allow you to manage data types through another interface.

note: I am not an expert at python, so be aware, there's likely better (faster) ways of doing things.


acgessler:
Looks good, thanks!

PR #666, opened 2015-09-30
remove memory.h dependancy

Lectem:
Some embedded (or exotic) systems won't provide this header, and it seems to be unused (since memcpy is in string.h).
Tested on MinGW and VS2015.


acgessler:
It looks like this won't work on all MSVCs, can you double-check?

Lectem:
It seems to be a problem with your AppVeyor script (see http://help.appveyor.com/discussions/problems/3163-builds-failing-with-unsupported-script-engine-message)
It is building fine under both MSVC2013 and MSVC2015, can't tell for previous versions.

acgessler:
True - #668 to fix.

PR #672, opened 2015-10-19
Properly build path to the mtl file when loading obj

Ang3lus:
Before patch:
```
MyIOStream(assets/raw/models/test_model/test_model.obj)
MyIOStream(test_model.mtl)
```

After patch:
```
MyIOStream(assets/raw/models/test_model/test_model.obj)
MyIOStream(assets/raw/models/test_model/test_model.mtl)
```

[Review on Reviewable](https://reviewable.io/reviews/assimp/assimp/672)

### DIFF
--- a/code/ObjFileImporter.cpp
+++ b/code/ObjFileImporter.cpp
@@ -138,7 +138,7 @@ void ObjFileImporter::InternReadFile( const std::string& pFile, aiScene* pScene,
     if ( pos != std::string::npos ) {
         modelName = pFile.substr(pos+1, pFile.size() - pos - 1);
         folderName = pFile.substr( 0, pos );
-        if ( folderName.empty() ) {
+        if ( !folderName.empty() ) {
             pIOHandler->PushDirectory( folderName );
         }
     } else {
--- a/code/ObjFileParser.cpp
+++ b/code/ObjFileParser.cpp
@@ -414,7 +414,7 @@ void ObjFileParser::getFace(aiPrimitiveType type)
 
     if ( pIndices->empty() ) {
         DefaultLogger::get()->error("Obj: Ignoring empty face");
-        // skip line and clean up 
+        // skip line and clean up
         m_DataIt = skipLine<DataArrayIt>( m_DataIt, m_DataItEnd, m_uiLine );
         delete pNormalID;
         delete pTexID;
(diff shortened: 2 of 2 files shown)


code/ObjFileImporter.cpp
```diff
@@ -138,8 +138,8 @@ void ObjFileImporter::InternReadFile( const std::string& pFile, aiScene* pScene,
     if ( pos != std::string::npos ) {
         modelName = pFile.substr(pos+1, pFile.size() - pos - 1);
         folderName = pFile.substr( 0, pos );
-        if ( folderName.empty() ) {
-            pIOHandler->PushDirectory( folderName );
+        if ( !folderName.empty() ) {
+            pIOHandler->PushDirectory( folderName + "/");
```
Ang3lus: Looked through the code and maybe this `+ "/"` should be placed in the ObjFileParser.cpp as there is the place where path is added:
```
...
    if ( m_pIO->StackSize() > 0 ) {
        const std::string &path = m_pIO->CurrentDirectory();
        absName = path + strMatName;
    } else {
        absName = strMatName;
    }
    IOStream *pFile = m_pIO->Open( absName );
...
```

What do You think ?

acgessler: @kimkulling your call!

Ang3lus:
* Used rfind and getOsSeparator to separate file from path
* Missing mtl path separator is added in the ObjFileParser.cpp when needed

Ang3lus:
* Removed hardcoded path separator from one more place

acgessler:
Looks good, two comments above.

Ang3lus:
* Reverted back to plain '/'
* Added additional check for empty path

Ang3lus:
Merge ? :)

acgessler:
I think "\\/" is what we want, right? Rest looks good, but please revert this check.

Ang3lus:
* Check for empty path removed

As for "\/" - why someone would want to escape this char ?

acgessler:
This line is changed wrt to the original. It used to be `std::string::size_type pos = pFile.find_last_of( "\\/" );` and I'd prefer not changing it since we might encounter unanticipated regressions.

Ang3lus:
Done.

acgessler:
Thanks!

PR #674, opened 2015-10-27
FBX import: If animation is not defining keys for all vector components (x, y and z), use the node transform as the default value.

suikki:
I was having problems importing FBX files created with blender after I changed from old blender ASCII FBX exporter to the newer binary exporter. Animation was not being imported correctly by assimp.

The problem is that if an animation only animates some components of a position, scale or rotation (for example only the x-axis is animated) the other axes were always reported to be 0.0 (1.0 for scale).

Here's a patch that fixed the problem for me.
I haven't messed with assimp internals before and I'm not really an expert on FBX so there were a couple of things I'm unsure of:
 - Values from different "key lists" were being combined before. Is it really possible to have an animation that defines overlapping key values like that? I had to remove this to make the default values work.
 -`InterpolateKeys` is now called a couple of times with just a hardcoded default value. It might be better to pass the node transform there too, but as far as I understand those versions of the functions are not actually ever called by the converter.

### DIFF
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -2496,8 +2496,9 @@ private:
         // need to convert from TRS order to SRT?
         if(reverse_order) {
 
-            aiVector3D def_scale, def_translate;
-            aiQuaternion def_rot;
+            aiVector3D def_scale = PropertyGet(props,"Lcl Scaling",aiVector3D(1.f,1.f,1.f));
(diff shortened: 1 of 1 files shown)


acgessler:
(subscribe myself so I don't forget reviewing this)

kimkulling:
I tried to do the review :-).

PR #675, opened 2015-10-28
Fix OFFLoader to better handle certain malformed files

turol:
A new and better version of #488. This one actually works. I think.

I'll add the testcase in a separate PR to avoid messing with the automatic checks.


kimkulling:
Looks good. Thank you very much!

PR #676, opened 2015-10-28
Add a testcase for invalid OFF file

turol:
Testcase generated by AFL which used to crash the OFF loader


acgessler:
Looks good, thanks!

(let's skip updating the DB for now. We can batch-update at some point to add new files)

turol:
Dropped the db update, left only the new file.

PR #690, opened 2015-11-15
Add new light type 'ambient' to match light.h

dankex:
Jassimp throws IllegalArgumentException without the fix.


kimkulling:
Merged, thanks a lot for your help!

PR #691, opened 2015-11-16
Fix AndroidJNI header path in CMakelists.txt

dankex:


kimkulling:
Merged, thanks a lot!

Issue #48, opened 2013-05-30
FBX embedded textures


raegnar:
The FBX loader doesn't seem to detect the presence of embedded textures in FBX files.

kimkulling:
Fixed by https://github.com/assimp/assimp/pull/694

Issue #652, opened 2015-09-02
Fbx parser not work when fbx file contains "PP" type connections


oguna:
I tried to parse StereoCamera.fbx (in Autodesk FBXSDK sample files), but parsing failed.

I found the cause. It contains "PP" type connections.
For example,

```
    C: "OO",786204432,784757920
    C: "OP",487083904,785243760, "RightCamera"
    C: "OP",487085712,785243760, "LeftCamera"
    C: "OP",786203536,487085712, "FilmOffsetX"
    C: "PP",785243760, "FilmWidth",487085712, "FilmWid
```

Now, `Assimp::Document::ReadConnections()` (FBXDocument.cpp) can parse "OP" or "OO" types connection, but "PP" type conenctions cannot be parsed.

otgerp:
Solved: now they are ignored, so the parsing does not fail anymore.

kimkulling:
Great, I will close this issue.

PR #694, opened 2015-11-23
Support for FBX embedded textures

OtgerRP:
Reads the embedded textures in FBX files which are referenced by the materials, and adds them to aiScene->mTextures (fixes issue #48).

The texture paths in the aiMaterial properties are set following the example of the COLLADA importer as `"*0"`, `"*1"`, etc. (this is documented anywhere?)

It works on the models I tested (embedded .png/.tga), but might be missing some cases, and in particular support for the LayeredTexture.

Also added a property to disable loading them (AI_CONFIG_IMPORT_FBX_READ_TEXTURES).


OtgerRP:
Seems that some tests fail because the binary FBX models had textures embedded, and now they appear at the dumps.

code/FBXDocument.h
```diff
@@ -654,6 +662,59 @@ typedef std::fbx_unordered_map<std::string, const Texture*> TextureMap;
 typedef std::fbx_unordered_map<std::string, const LayeredTexture*> LayeredTextureMap;
 
 
+/** DOM class for generic FBX videos */
+class Video : public Object
+{
+public:
+
+    Video(uint64_t id, const Element& element, const Document& doc, const std::string& name);
+    ~Video();
```
kimkulling: Not virtual.

kimkulling:
Seems so, can you update the dump in your pull request, please?

code/FBXDocument.h
OtgerRP: I just did copy the Texture class above it, which neither has it, and wanted to keep it similar. In any case they are meant to be "final" classes, so virtual destructor is not essential.

OtgerRP:
Done!

With your permission I also added a few doc lines to aiTexture explaining how the embedded textures are referenced from the materials (did not find anything about it).

OtgerRP:
I also added a quick fix to #652. FBX parser returned a parse error when it found a "PP" type connection, now it just ignores them.

kimkulling:
Merged, thanks a lot!

Kimmi

PR #696, opened 2015-11-24
aiGetExportFormatDescription() no longer uses free'd memory

yig:
aiGetExportFormatDescription() now creates an Exporter object at the beginning of the function, so that a call to `GetExportFormatDescription()` doesn't access free'd memory.


kimkulling:
Thanks, right, merged :-) ...

PR #697, opened 2015-11-27
Basic glTF support (experimental)

otgerp:
I had made a basic implementation for a glTF importer for my own use (and started working on a exporter), and as I see support for it is being added, I'd like to contribute it. It uses *RapidJSON* instead of *picojson* though (because it supports in-place parsing, avoiding making a lot of copies), which it is licensed under MIT (not sure if there are license compatibility issues).

It can read both text (gltf) and binary (glb) formats, and currently only imports a subset of the data: the node hierarchy (with transforms), materials (basic properties), textures (both file references and embedded), and the meshes (vertex positions, normals, texture coordinates and the faces).

It is still experimental, but I tried most of the glTF's repo models and they seem to work correctly in all their variants. In any case I think it could be a good starting point.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -568,6 +568,9 @@ ADD_ASSIMP_IMPORTER(GLTF
   glTFUtil.cpp
   glTFUtil.h
   glTFFileData.h
+
+  glTFExporter.h
+  glTFExporter.cpp
 )
 
 SET( Step_SRCS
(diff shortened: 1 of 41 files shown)


kimkulling:
Much better than my first baby steps. Thanks a lot!

acgessler:
Is it intentional that many test models (e.g., 2CylinderEngine) or also the model in the repository appear to have their hierarchy imported incorrectly and/or textures flipped?

Or is this a recent regression?

acgessler:
http://i.imgur.com/RIQ9uS6.png

kimkulling:
I don't think so.

PR #699, opened 2015-11-29
glTF exporter + improved the importer (and code refactor)

otgerp:
With this it can now import and export both gltf and glb formats (although there is still some stuff missing, like the animations). I tried loading most of the test models of the glTF repo, exporting them and reimporting them, and it seems to work okay.


kimkulling:
Nice work, but you need to merge master into it first.

kimkulling:
Merged, thanks a lot for your help!

Kimmi

r-lyeh:
hey guys,

i've tried to export a random model from the internet and the resulting .gltf file just lacks all "sampler" fields. i've added then manually and now it works.

i'm wondering if the exporter failed because the image URLs were just wrong (absolute paths from a random computer; even more, one of the paths was incomplete). if that's the case of the problem, shouldnt the exporter just export the data as much original as possible, even if it cant find the textures on disk?

thanks in advance :)

PR #703, opened 2015-12-08
Improvement of collada transparency handling

terziman:
All missing cases of transparency in collada have been added. In particular, the option introduced precedently to manually invert the value of the transparency setting is now useless and has been removed.

Moreover, the default value for reflectivity has also been fixed.


acgessler:
Thanks for fixing this. I'm glad there is a solution without this terrible config switch!

PR #704, opened 2015-12-09
Added missing DLL suffix for VS2015 build.

Vertexwahn:
Added missing DLL suffix for VS2015 build.


kimkulling:
Thanks for the fix, merged!

PR #705, opened 2015-12-10
Added custom target that copies generated libs and bin files to lib32/lib64 and bin32/64 folder

Vertexwahn:
Added custom target that copies generated libs and bin files to lib32/lib64 and bin32/64 folder

This should simplify deployment of VS2015 build libs.


kimkulling:
Thanks a lot! Merged!

PR #709, opened 2015-12-13
fix multiple objects generation in OBJ reader if too many 'usemtl' are used

ov:
See the test file `cube_usemtl.obj`, it has 'usemtl' before each face line and assimp makes a new object for each face. Not even mesh, but object. The reason is that the old code doesn't care what material is being set and simply make a new object.

A proper way would be to analyze the material name and ignore the line if the material is the same. If it is not, we need to make a mesh, not object instead.

The proposed patch does exactly that by comparing the new material name with the `m_pCurrentMaterial` variable and skipping all the code if the materials are the same.

It also refactors the case when the requested material is undefined, to make sure that the mesh gets a proper default material index.

Finally it fixes the typo (?) in the `needsNewMesh` method, as the `if` clause there with `or` doesn't really make sense.

As the assimp build and testing system is quite complex, all the changes are made in our local project and carefully merged into the current master code. The issues are unlikely, but possible. A good testing (especially regression testing) is required. However, I tested all the OBJ files in your test folder and they seems work fine.

Let me know if something is wrong, I will be happy to update the code.

Cheers!


acgessler:
Thanks! To clarify, all new regression test failures are due to multiple materials being merged into one?

ov:
Here are the files the regression fails on and some ideas:

- segment.obj - the old version made 3 objects, the new makes 3 meshes, this looks correct for me
- testmixed.obj - two 'usemtl' lines with the same material, our software doesn't support points and lines, so I can't test it well, but geometry should be merged too, instead of making two objects like the old version did
- cube_usemtl.obj - just a new test, needs to be added to the database


- box_mat_with_spaces.obj & space_in_material_name.obj - both having spaces in material names and (most important) around them. Not sure why the test fails now, but I've just commited one more fix that trims whitespaces both for 'usemtl' and 'newmtl' commands. Works fine for me, let's see what tests show.

ov:
OK, so the last updated fixed `space_in_material_name.obj` regression test, but `box_mat_with_spaces.obj` is still there. I believe it is worth a double-checking and probably updating the regression database.

The other three files are absolutely OK, just update the database.

kimkulling:
Shall work now!

PR #716, opened 2015-12-14
Added support for empty .MTL files.

rmitton:
Some programs such as Silo export an empty .MTL file if you export a
.OBJ that doesn't use any materials. This patch allows the loading of
such files without error.


kimkulling:
Looks fine. I guess we can accept the API-change in baseimporter.

PR #719, opened 2015-12-16
FBX Import fix

luho383:
Fixes an error when an Element has no data.

Example from (Cinema4D export):
		LayerElementNormal: 0 {
			Version: 101
			Name: ""
			MappingInformationType: "NoMappingInformation"
			ReferenceInformationType: "Direct"
		}

What's missing here is the "Normals" data block.
Cinema4D outputs this with NoMappingInformation which does not
export the "Normals" datablock. This can happen for other LayerElement* types.


acgessler:
Thanks!

PR #721, opened 2015-12-17
Fix for FBX Files that do not have a Comma on a new line in a data block

luho383:
Some FBX Exporters (Cinema4D plugin for example) do not write a Comma between
two Numbers when there is a new Line.

Example:

LayerElementColor: 0 {
			Version: 101
			Name: "colorSet1"
			MappingInformationType: "ByPolygonVertex"
			ReferenceInformationType: "IndexToDirect"
			Colors: *45640 {
				a: 0.0,0.0,0.0,1.0
				0.0,0.0,0.0,1.0
                        }
}

A better fix would be that the Tokenizer would create Comma placeholder tokens.
But this fixes it too.

### DIFF
--- a/code/FBXParser.cpp
+++ b/code/FBXParser.cpp
@@ -125,13 +125,20 @@ Element::Element(const Token& key_token, Parser& parser)
 
         if (n->Type() == TokenType_DATA) {
             tokens.push_back(n);
-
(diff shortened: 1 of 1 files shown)


acgessler:
Thanks for fixing this!

Issue #712, opened 2015-12-13
Jassimp-native: easily overflow JNI local reference table when loading big models


dankex:
Local references created while accessing Java objects, classes, strings can easily go beyond the limit of 512 entries. Though they don't cause memory leaks since they are freed automatically (when the JNI call completes and control is returned to Java), this does prevent large models from being loaded. Suggest using something like a C++ "smart pointer" to hold local references, and automatically call DeleteLocalRef(...) when it returns from a function. This might need to be done for most functions in jassimp.cpp.

PR #726, opened 2015-12-19
Use SmartLocalRef to automatically release localrefs in jassimp.cpp

dankex:
- Add SmartLocalRef class in jassimp.cpp
- Use it to automatically delete localrefs of Java objects, generated
  by FindClass, createInstance, getField, NewStringUTF, etc.
- Tested with models with more than 512 nodes.

Fixes https://github.com/assimp/assimp/issues/712


acgessler:
Thanks!

PR #732, opened 2015-12-28
Collada exporter: bind uv channel numbers with <bind_vertex_input>.

StepanHrbek:
When there are multiple uv channels used by multiple textures, <bind_vertex_input> specifies what channels to use. Without <bind_vertex_input>, at least some importers use wrong uv channels for textures.


StepanHrbek:
Hello, I'm not sure, but I think that Travis failure is not my fault. Is that ok or should I change something?

kimkulling:
You are right, this issue was caused by me. I will rebuild it!

kimkulling:
Was fine, merged! And thanks a lot for your work!

PR #735, opened 2015-12-30
Fix a crash issue in XFileExporter

JLouis-B:


kimkulling:
Merged, thanks a lot!

PR #736, opened 2015-12-31
Collada exporter: fix unicode.

StepanHrbek:


StepanHrbek:
This seems to depend on build settings, everything is good with gcc, clang, and with my own vcxproj files, only AppVeyor fails with his vcxproj file. I'm not skilled with cmake so I don't want to debug this....maybe someone knows what's wrong?

Old code calls isalnum(c) from <cctype>, my code calls isalnum(c,loc) from <locale>. And as there already is std::locale, I did not bother including <locale>, it has to be included somewhere. But then AppVeyor doesn't find isalnum(c,loc).

StepanHrbek:
Rather than debugging cmake, I would rewrite isalnum to strchr.

StepanHrbek:
Ok, it works now. 39ab716 fails for reasons I can't reproduce, so 24f0314 reimplements it.

kimkulling:
Thanks a lot, just merged it!

Issue #738, opened 2016-01-02
Compilation error on Mac OS X and Intel 64-bit processor


danke-sra:
Error log:

```
assimp/code/../contrib/rapidjson/include/rapidjson/document.h:561:14: error: constructor cannot be redeclared
    explicit GenericValue( size_t u ) RAPIDJSON_NOEXCEPT : data_(), flags_( kNumberUintFlag ) {
             ^
assimp/code/../contrib/rapidjson/include/rapidjson/document.h:550:14: note: previous declaration is here
    explicit GenericValue(unsigned long u64) RAPIDJSON_NOEXCEPT : data_(), flags_(kNumberUint64Flag) {
```  

This is because "unsigned long" and "size_t" are the same.

kimkulling:
should be fixed now.

winkelsdorf:
Still broken for iOS:

```
/Users/winkelsdorf/Projects/GitHub/assimp/code/../contrib/rapidjson/include/rapidjson/document.h:562:14: error: 
      constructor cannot be redeclared
    explicit GenericValue( size_t u ) RAPIDJSON_NOEXCEPT : data_(), flags_( kNumberUintFlag ) {
             ^
/Users/winkelsdorf/Projects/GitHub/assimp/code/../contrib/rapidjson/include/rapidjson/document.h:550:14: note: previous
      declaration is here
    explicit GenericValue(unsigned long u64) RAPIDJSON_NOEXCEPT : data_(), flags_(kNumberUint64Flag) {
             ^
```

PR #739, opened 2016-01-02
Fix compilation error for Mac OS X and 64-bit architecture

dankex:
For a 64-bit architecture, unsigned long and size_t are the same.
This causes a compilation error due to duplicate constructors in document.h.

Fixed issue https://github.com/assimp/assimp/issues/738


kimkulling:
thanks a lot for the fix, merged!

PR #748, opened 2016-01-09
Collada export bones

wise86-android:
add the ability to export joints/bone in the collada files

### DIFF
--- a/code/ColladaExporter.cpp
+++ b/code/ColladaExporter.cpp
@@ -1415,7 +1415,6 @@ void ColladaExporter::WriteNode( const aiScene* pScene, aiNode* pNode)
     for( size_t a = 0; a < pNode->mNumMeshes; ++a )
     {
         const aiMesh* mesh = mScene->mMeshes[pNode->mMeshes[a]];
-
 		// do not instanciate mesh if empty. I wonder how this could happen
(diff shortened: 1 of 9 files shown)


code/ColladaExporter.cpp
```diff
@@ -1136,21 +1136,23 @@ void ColladaExporter::WriteJointsWeightSourceNode(const size_t pIndex){
 	delete [] tempBoneWeight;
 }
 
+template <typename T>
+static void moveMatrixInArray(const aiMatrix4x4t<T> &matrix, T *outBuffer){
+	std::memcpy(outBuffer,matrix[0],16*sizeof(T));
```
acgessler: should this not be &matrix[0]?

Also wondering if perhaps a copy with two loops is better to maintain.

acgessler:
Apologies for late response. If you're still there: definitely interested in merging this. Can you rebase?

code/ColladaExporter.cpp
wise86-android: No matrix[0] return the pointer to the data buffer for be able to use matrix[i][j].

see https://github.com/assimp/assimp/blob/master/include/assimp/matrix4x4.inl#L241

kimkulling:
I merged the latest master into this one and created a new pull-request. When the new pull-request was merged I will close this on. Thanks a lot for the help.

PR #753, opened 2016-01-17
Valgrin free

wise86-android:
I run the test using valgrind and remove all the memory leaks that it found.
And fix some compile warnings.
Now there is only one problem with the [here](https://github.com/assimp/assimp/blob/master/test/unit/utSharedPPData.cpp#L106) since you will do a delete instead of delete[].
but I don't know how specialize the class Assimp::SharedPostProcessInfo::THeapData for hendle object pointer and array pointer.

### DIFF
--- a/test/unit/utSplitLargeMeshes.cpp
+++ b/test/unit/utSplitLargeMeshes.cpp
@@ -58,8 +58,7 @@ protected:
 
     SplitLargeMeshesProcess_Triangle* piProcessTriangle;
     SplitLargeMeshesProcess_Vertex* piProcessVertex;
-    aiMesh* pcMesh1;
-    aiMesh* pcMesh2;
+
 };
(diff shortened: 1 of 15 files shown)


test/unit/utSplitLargeMeshes.cpp
```diff
@@ -124,6 +85,26 @@ TEST_F(SplitLargeMeshesTest, testVertexSplit)
 {
     std::vector< std::pair<aiMesh*, unsigned int> > avOut;
 
+     aiMesh *pcMesh1 = new aiMesh();
+     pcMesh1->mNumVertices = 2100; // quersumme: 3
```
kimkulling: Do not use german comments

kimkulling:
Looks good, I will merge it!

PR #756, opened 2016-01-21
Collada exporter: improve compatibility when exported path contains ':'.

StepanHrbek:
Old code exported "c:/foo.jpg" path as "c%3a/foo.jpg".
Replacing : with %3a was probably legal, but it made paths unreadable for existing Max importers.


kimkulling:
Looks good, thanks a lot!

PR #761, opened 2016-01-22
ObjImporter: Removed unnecessary removal of '\' chars.

andrewparlane:
std::vector::erase() has complexity O(n) which for a large buffer is very
slow.

Additionally in ObjFileParser we have a copyNextLine() function which
takes line continuation with backslashes into account. This is used only
for faces ATM. I'm not sure when a '\' can occur in a .obj file, but I've
only ever seen it on face lines that have more than 3 vertices.

We could use copyNextLine() for other types of lines too if needed. This
would slow down parsing of lines a little bit, but that is probably
mitigated by not scanning through removing backslashes first (that took
1/3rd of the import time in my file that has no backslashes).

Finally I've updated the progress reporting to take into account this
change.

### DIFF
--- a/code/ObjFileImporter.cpp
+++ b/code/ObjFileImporter.cpp
@@ -145,38 +145,6 @@ void ObjFileImporter::InternReadFile( const std::string& pFile, aiScene* pScene,
         modelName = pFile;
     }
 
-    // This next stage takes ~ 1/3th of the total readFile task
-    // so should amount for 1/3th of the progress
-    // only update every 100KB or it'll be too slow
-    unsigned int progress = 0;
-    unsigned int progressCounter = 0;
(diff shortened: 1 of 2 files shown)


kimkulling:
Looks like a https://twitter.com/sharondio/status/676855973290577922/photo/1. The code is there for a reason. I will optimize the algorithm a bit at this point. Just remove this is not a solution ( only a workaround for the obj-issue takes forever ).

andrewparlane:
Ha, nice.

Yeah, I guess it depends on where '\'s can occur. I suggest we remove the
pre-parse and instead use the copyNextLine() function in the actual parsing
code for everything.

I can look into this if you wish?

Andy

kimkulling:
Of course! I am happy about all the help you offering us :-).

PR #763, opened 2016-01-23
ObjFileMtlImporter: Fixed a segfault due to NULL ptr access.

andrewparlane:
We don't handle reflection textures during MTL import. This allowed a NULL
ptr to be used. Simply return in the case of a reflection texture.


kimkulling:
Looks fine, thanks a lot for the fix!

andrewparlane:
no problem.

Issue #759, opened 2016-01-22
OBJ import takes forever


fredrik955:
When I try to import this file: https://drive.google.com/file/d/0B3PusJzfuPY0Wkc0cGwtb053ejQ/view?usp=sharing
with Assimp 3.2 it will never finish the import script.
I use these flags: aiProcess_Triangulate | aiProcess_GenUVCoords | aiProcess_GenSmoothNormals | aiProcess_OptimizeMeshes | aiProcess_OptimizeGraph | aiProcess_FindInvalidData | aiProcess_FixInfacingNormals

kimkulling:
The link is dead, sorry.

andrewparlane:
Are you building from the trunk or a particular release?

What platform?

I added changes for better progress reporting in OBJ imports, but that's
only on the trunk as of yet. If you are using the trunk it could be that,
can you try going back to the last stable release and testing there?

fredrik955:
Hi,
The link is not dead. It says "No preview", but you can see the download button in top center of the screen.
I am using this version: Version 3.2 -debug -noboost -shared -st (GIT commit c2a046)
I am using Ubuntu 14.04.3 LTS.

andrewparlane:
[image: Inline images 1]

Looks pretty dead. here.

Fine if you're using a release rather than the trunk, my changes aren't the
issue. If you can get the link working, I / we'll have a look.

fredrik955:
Here is a new link:
https://storage.googleapis.com/assimp_model_error/forever.obj

andrewparlane:
Well, that file seems to be 102MB, which is probably clue enough as to why
it's taking so long. I'm importing a 1.5MB .obj and it takes about 5s,
assuming we can linearly scale up, a 102MB one should take about 5m40s.

Try sticking a progress bar in there, or at least something that outputs
progress percent, and see how it's going.

I'm not entirely sure how you generated a .obj that big, either you have a
very complicated model, or it got exported incorrectly.

fredrik955:
Thank you for the reply. Did the import process ever finish on your
computer? I can open the OBJ file in Meshlab in less than 20 sec. Is it
possible to speed up the OBJ importer?

andrewparlane:
it's still downloading ATM. I'm currently living in Bolivia, and the
internet is a little on the slow side. Currently looking at your other seg
fault issue.

andrewparlane:
I can't seem to locate commit: c2a046 It's possibly ambiguous.

[...]

PR #765, opened 2016-01-24
Fix issue: OBJ import takes forever (#759)

andrewparlane:
Commit "ObjFileParser: Moved the parsing of line continuations (backslashes) … " is the main change, the other is just a minor change that was needed so that I could use ConstDataArrayIt instead of DataArrayIt.

I've tested this as best I can, but the .obj files I have are by no means extensive.

### DIFF
--- a/code/ObjFileImporter.cpp
+++ b/code/ObjFileImporter.cpp
@@ -145,38 +145,6 @@ void ObjFileImporter::InternReadFile( const std::string& pFile, aiScene* pScene,
         modelName = pFile;
     }
 
-    // This next stage takes ~ 1/3th of the total readFile task
-    // so should amount for 1/3th of the progress
-    // only update every 100KB or it'll be too slow
-    unsigned int progress = 0;
-    unsigned int progressCounter = 0;
-    const unsigned int updateProgressEveryBytes = 100 * 1024;
-    const unsigned int progressTotal = (3*m_Buffer.size()/updateProgressEveryBytes);
-    // process all '\'
-    std::vector<char> ::iterator iter = m_Buffer.begin();
-    while (iter != m_Buffer.end())
-    {
-        if (*iter == '\\')
-        {
-            // remove '\'
-            iter = m_Buffer.erase(iter);
-            // remove next character
-            while (*iter == '\r' || *iter == '\n')
-                iter = m_Buffer.erase(iter);
-        }
-        else
-            ++iter;
-
-        if (++progressCounter >= updateProgressEveryBytes)
-        {
-            m_progress->UpdateFileRead(++progress, progressTotal);
-            progressCounter = 0;
-        }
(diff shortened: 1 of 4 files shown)


kimkulling:
Sorry, but I cannot detect the better algorithm for filtering wrong lline endings. Could you please explain where this code is placed. 
And please try to separate such changed in 2 different branches. It is really hard to disdiguish between them.

andrewparlane:
OK, so previously we have two stages.

1) \ removal
for every char in buffer
{
  if char is '\'
  {
    buffer.delete(thisChar);
    buffer.delete(newLines);
  }
}

This code was in ObjFileImporter.cpp and has been deleted. The problem with
this is std::vector::delete() has complexity O(n), and with a file with
\r\n line endings, we do 3 deletes every time we find a \.

2) parsing
while (itStart != itEnd)
{
  parseStuff();
}


Instead of doing this, I use the copyNextLine() function in
ObjFileParser.cpp. See how in the parse loop (line 137) I use
copyNextLine() to copy the next line from the data buffer into a helper
buffer. Have a look at copyNextLine(). It copies the line, but will ignore
\r or \n characters if they are proceeded by a \.

So if your .obj looked like:

v 0.000124 1.234567\
9.87654
v 123\
456\
789

The first call to copyNextLine would give:
v 0.000124 1.234567 9.87654

and the second:
v 123 456 789

It does this by simply not copying certain characters, rather than deleting
them, and since using an iterator access is O(1) and a push_back() is O(1),
then total complexity for copy of each line is O(m) where m is the line
length. Since we copy every line eventually, we have complexity O(n) where
n is the buffer length. This can be compared to the original removal of
whitespace which was O(n) to check every single character, with additional
O(n). Which gives O(n^2). You still have to parse the data in both cases,
which is character by character parsing so also O(n).

Therefore if I'm correct, the old method was either: O(n^2) or O(n log n),
and the new method is O(n).

Hope that helps to clarify why this is better, if you have more questions
just ask.

Finally, I'm not sure I understand your comment on branches? You would
prefer me to create a new branch based off master, and push my 2 commits
there, and use that as a pull request? Surely that would just be as hard to
read, since my new branch would be the same as master. I'm happy to make
modifications as needed, just let me know what you would like.

kimkulling:
Thanks a lot! Now I got it!

kimkulling:
Merged! Great work!

andrewparlane:
excellent. Glad I can help.

I quit my job a few months ago and moved to Bolivia. Now that I don't spend
my days coding for work, I can actually do stuff for fun (I teach kids
circus skills).

I was hoping to do a bunch of open source stuff while out here, to bolster
my CV and what not, so I'm glad I can help out.

I don't know much about assimp ATM, but I'll probably be helping out from
time to time when something comes up.

Andy

PR #773, opened 2016-01-27
Import mClipPlane{Near,Far} and mHorizontalFOV from Blender.

boxdot:
This will populate aiCamera with real values. I am still not sure, how to get mAspect from Blender.


kimkulling:
Merged, thanks a lot!

PR #779, opened 2016-02-02
SIBImporter: Properly fix C++11 issues for Clang

turol:
C-style cast has a higher precedence than & -operator so this was getting
parsed differently than Kim assumed. Thou shalt not use C-style casts.


kimkulling:
Merged, thanks a lot!

PR #789, opened 2016-02-07
Revert "Fix issue: OBJ import takes forever (#759) (attempt 2)"

kimkulling:
Reverts assimp/assimp#771


kimkulling:
Will not work with obj-material loading.

andrewparlane:
How so? If you let me know what is wrong, I'll see if I can fix it today or
tomorrow.

Andy

PR #791, opened 2016-02-10
Collada animation clip support

trond:
Attempt at adding support for `<library_animation_clips>`, which is used by **Better Collada Exporter** for **Blender**, by the **Godot** engine team. The exporter is useful to pack multiple animations into one file, and it also supports baking of constraints.

The exporter is available here:
http://www.godotengine.org/projects/godot-engine/documents

It seems to work for me so far, and also the regression tests should still pass like they did before this patch. I added a file for regression testing `(library_animation_clips.dae),` but I did not create a new db.zip, as it seems like there are more files that need to be regenerated.


kimkulling:
Merged, thank you very much for your contribution.

trond:
Great, thanks!

PR #792, opened 2016-02-11
Fix loading of pk3 q3bsp maps

rdb:
This fixes loading of pk3 maps.  The pk3 files contain directory entries with a size of 0, which triggered the following assertion:

    Q3BSPZipArchive.cpp:142: Assimp::Q3BSP::ZipFile::ZipFile(size_t): Assertion `m_Size != 0' failed.

The fix ignores "files" with size of 0, which seems to work perfectly for me.


kimkulling:
Looks great, thank you very much for the contribution!

PR #800, opened 2016-02-18
Can't add integer to string pointer

pkholland:
The code probably meant to create a new string of the form "animation_52", right?  The way it is now it attempts to add 52 to the base address of the string "animations_" and then make an std::string from that pointer - or something like that.


code/ColladaParser.cpp
```diff
@@ -297,7 +297,7 @@ void ColladaParser::ReadAnimationClipLibrary()
 				else if (indexID >= 0)
 					animName = mReader->getAttributeValue(indexID);
 				else
-					animName = "animation_" + mAnimationClipLibrary.size();
+					animName = std::string("animation_") + std::to_string(mAnimationClipLibrary.size());
```
kimkulling: std::to_string( ... ) is a C++11 call ( see http://de.cppreference.com/w/cpp/string/basic_string/to_string ).

kimkulling:
C++11 is enabled, so thanks a lot for the contribution.

PR #812, opened 2016-03-02
Fix warning when compiling using -Wswitch-enum

tgoulart:
No changes to functionality, only removing a warning due to unhandled value in a switch statement. The missing value is not of any interest there, so let the default block handle it. Found while building an application with assimp static library + headers on Xcode 7.2.1.


kimkulling:

PR #814, opened 2016-03-04
Add missing files  so that xcode6 project builds successfully

tgoulart:
Seems like the xcode6 project got slightly outdated. Three of the loaders wouldn't compile, but it was just a matter of adding the required files to the project, along with adding a required dependency to the headers search path.

It now builds successfully on Xcode 7.2.1 with a single warning (I'll make another pull request later for that).


kimkulling:

PR #820, opened 2016-03-13
Fix some crashes on malformed input

turol:
Fix various crashes from #454.


kimkulling:
Merged, thanks a lot.

PR #821, opened 2016-03-14
Update CMakeLists.txt

Wietse:
Changed "add_definitions" to "add_compile_options" to fix building with NMake.
Using "add_definitions" adds "/MP" to the resource compiler options which results in a fatal error when run with NMake.


kimkulling:
Look good, thanks a lot for your contribution.

Issue #824, opened 2016-03-17
Replace Boost static_assert with C++11 version


sherief:
Self-explanatory task tracking issue.

kimkulling:
Fixed by 49d60dc8567b55dd1263de22edffe6e48641d040

PR #832, opened 2016-03-19
Removed usage of Boost static assert.

sherief:
Fixes issue #824.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,10 +70,12 @@ if((CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX) AND NOT CMAKE_COMPILER_
      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
   endif()
   # hide all not-exported symbols
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden -Wall" )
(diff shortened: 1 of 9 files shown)


sherief:
@kimkulling / @acgessler  Can I get a review?

kimkulling:
Travis build is broken

turol:
You need to set C++11 mode. This is slightly complicated because older versions of GCC need "std=c++0x" but newer versions prefer "std=c++11". You might also have to check for some minimum version of GCC/clang.

sherief:
Can someone point me in the right direction / help with enabling C++0x / C++11 on all configs?

kimkulling:
Check the global cmake-file, look out for the cpp-flags. I can do this later this week. Unfortunately i am ill at the moment.

sherief:
Sorry to hear that, hope you feel better soon!

sherief:
@kimkulling Any help on this one?

mensinda:
Adding `set_property( GLOBAL PROPERTY CXX_STANDARD 11 )` somewhere in the root CMakeLists.txt (maybe ca. line 23) should do the trick.

This will/should enable c++11 support for all compilers.

Links:

https://cmake.org/cmake/help/v3.5/prop_tgt/CXX_STANDARD.html
https://cmake.org/cmake/help/v3.5/command/set_property.html

kimkulling:
SO last but not least we are supporting c++11.

sherief:
WOOT!

PR #835, opened 2016-03-22
Vcs 1030/obj mtl filename fallback

robertliebo:
* this fallback strategy for cases when a .mtl file is not found serves two use cases:
 * 1.): workaround for a common user mistake: we occasionally get renamed .obj/.mtl files where the creator didn't know that this breaks the reference in the .obj file to the .mtl file
 * 2.): workaround for file name encoding issue: on a debian system that we use, a "Köln.obj" refering to "Köln.mtl" is within assimp processed as "KÃ¶ln.obj" (file open works with this badly encoded file name containing "Ã¶", but not with the correct "ö"). but opening the refered .mtl file also only works when m_pIO->Open() is called with "KÃ¶ln.mtl"


kimkulling:
Looks good, thanks a lot for your contribution.

PR #837, opened 2016-03-27
Python 3 compatibility fix for xrange in PyAssimp

stevenjt:
This is a fix that sets xrange to range in Python 3.0 and above. In Python 3 xrange was renamed range and the original range function was removed.


kimkulling:
Makes sense for me, merged! And thank you verymuch for your contribution!

PR #839, opened 2016-03-29
Import reflectance factor from Blender.

boxdot:


kimkulling:
Looks good, thanks a lot for your contribution!

PR #854, opened 2016-04-18
fix SceneCombiner copy of aiNode not setting mParent field

tomacd:
After performing a deep copy of a scene, the mParent fields of the nodes of the new scene were pointing to the nodes of the original scene.


acgessler:
Thanks for noticing!!

PR #861, opened 2016-04-22
only check a library folder if it actually exists

v4hn:
otherwise this breaks for no reason if /usr/local/lib is missing.


kimkulling:
Looks right, thanks for your cntribution.

PR #863, opened 2016-04-22
Fix texture coords exporting on binary PLY files

mpersano:


kimkulling:
Thanks for the contribution, merged!

Issue #865, opened 2016-04-25
GLTF exporter doesn't work (but Binary GLTF does)


KeepItOneHundred:
I'm trying to convert the [Chinese Dragon](http://graphics.cs.williams.edu/data/meshes.xml) model using the assimp command line tool, but exporting to GLTF just writes a header and nothing else. Other output formats (including Binary GLTF) seem to work fine.

```
$ assimp export dragon.obj dragon.gltf
assimp export: select file format: 'gltf' (GL Transmission Format)
Launching asset import ...           OK
Validating postprocessing flags ...  OK
Importing file ...                   OK
   import took approx. 7.74268 seconds

Launching asset export ...           OK
Exporting file ...                   OK
   export took approx. 6.28994 seconds

assimp export: wrote output file: dragon.gltf

$ assimp export dragon.obj dragon.glb
assimp export: select file format: 'glb' (GL Transmission Format (binary))
Launching asset import ...           OK
Validating postprocessing flags ...  OK
Importing file ...                   OK
   import took approx. 7.99809 seconds

Launching asset export ...           OK
Exporting file ...                   OK
   export took approx. 6.30317 seconds

assimp export: wrote output file: dragon.glb

$ ls -l
total 87760
-rw-rw-r-- 1 kioh kioh 15763392 Apr 25 13:55 dragon.glb
-rw-rw-r-- 1 kioh kioh      148 Apr 25 13:55 dragon.gltf
-rw-rw-r-- 1 kioh kioh 74094075 Jul 23  2011 dragon.obj

$ cat dragon.gltf
{
    "asset": {
        "version": 1,
        "generator": "Open Asset Import Library (assimp v3.2.122064037)"
    },
    "scene": "defaultScene"
}
```

kimkulling:
Closed by 11e52dc

Issue #833, opened 2016-03-19
gltf import (with separate .bin) using ReadFileFromMemory


pixelnerve:
Hello,

I was wondering how will assimp deal with the case of loading gltf with external .bin file when using ReadFileFromMemory. Did a quick test and it crashes when accessing the data

kimkulling:
Closed by 11e52dc

Issue #818, opened 2016-03-09
gltf export problems


ftrastour:
Assimp application failed to create gltf files:
- with some models, the application crash,

[crash.zip](https://github.com/assimp/assimp/files/165413/crash.zip)

- with some others the application generates invalid gltf files with only : 
{
    "asset": {
        "version": 1,
        "generator": "Open Asset Import Library (assimp v3.2.160146130)"
    },
    "scene": "defaultScene"
}

[empty_gltf.zip](https://github.com/assimp/assimp/files/165403/empty_gltf.zip)

The assimp application used for test is from latest team builds (from artifacts) - x86.
Same problems appear with cygwin build of master.

Note that both models load well in assimp_viewer.

kimkulling:
Thank you very much for the report, we will take a look.

kimkulling:
Closed by 11e52dc

Issue #785, opened 2016-02-05
glTF importer crash


turol:
On linux, 64-bit release with debug info

assimp info test/models/glTF/CesiumMilkTruck/CesiumMilkTruck.gltf
Launching asset import ...           OK
Validating postprocessing flags ...  OK
Segmentation fault

backtrace from gdb:
0  0x00007ffff71ea59c in std::string::assign(char const*, unsigned long) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
1  0x00007ffff7b9c8ea in assign (
    __s=0x42925f "collada2gltf@ceec062e3d5793f2f249f53cbd843aee382ad40b", 
    this=0x7fffffffc680) at /usr/include/c++/4.7/bits/basic_string.h:1128
2  operator= (
    __s=0x42925f "collada2gltf@ceec062e3d5793f2f249f53cbd843aee382ad40b", 
    this=0x7fffffffc680) at /usr/include/c++/4.7/bits/basic_string.h:552
3  Read (out=
    <error reading variable: Cannot access memory at address 0xffffffffffffffe8>, val=...) at assimp/code/glTFAsset.inl:78
4  glTF::(anonymous namespace)::ReadMember<std::basic_string<char> > (
    obj=..., id=id@entry=0x7ffff7bf95d8 "generator", out=
    <error reading variable: Cannot access memory at address 0xffffffffffffffe8>) at assimp/code/glTFAsset.inl:96
5  0x00007ffff7ba0ed4 in glTF::AssetMetadata::Read (
    this=this@entry=0x7fffffffc678, doc=...)
    at assimp/code/glTFAsset.inl:826
6  0x00007ffff7baf01b in glTF::Asset::Load (this=this@entry=0x7fffffffc5f0, 
    pFile="test/models/glTF/CesiumMilkTruck/CesiumMilkTruck.gltf", 
    isBinary=<optimized out>) at assimp/code/glTFAsset.inl:944
7  0x00007ffff7b9e9b7 in Assimp::glTFImporter::InternReadFile (this=0x426390, 
    pFile="test/models/glTF/CesiumMilkTruck/CesiumMilkTruck.gltf", 
    pScene=0x4278c0, pIOHandler=<optimized out>)
    at assimp/code/glTFImporter.cpp:604
8  0x00007ffff781ca51 in Assimp::BaseImporter::ReadFile (this=0x426390, 
    pImp=<optimized out>, pFile=
[...]

PR #868, opened 2016-04-28
Fixed a few GLTF importer/exporter bugs

otgerrogla:
Should solve  #785, #818, #833, #865, and a pair of other bugs I found.

Also, two small changes to the CMakeList.txt:
- Added a compilation flag "/bigobj" for MSVC to the file "IFCReaderGen.cpp", as when I tried compiling ASSIMP on VS2015 64bit, it failed with:

        1>D:\assimp\code\IFCReaderGen.cpp : fatal error C1128: number of sections exceeded object
        file format limit: compile with /bigobj

- Added a cmake option to manually disable building `assimp_view` on windows (in my system it was finding only some of the DirectX libs, so it was enabled but failed to compile)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,9 +269,13 @@ option ( ASSIMP_BUILD_ASSIMP_TOOLS
   ON
 )
 IF ( ASSIMP_BUILD_ASSIMP_TOOLS )
-  IF ( WIN32 AND DirectX_FOUND )
-    ADD_SUBDIRECTORY( tools/assimp_view/ )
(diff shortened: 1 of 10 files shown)


kimkulling:
Thanks for your contribution, merged.

otgerrogla:
There are some linking problems on Travis. I tried to reproduce the error with the same gcc version but it works for me, so it seems a problem with the linker version.

kimkulling:
Saw this, but I did not have time to look at. Sorry.

PR #876, opened 2016-05-05
Fix travis compile: glTFImporter chokes on unavailability of WriteLaz…

acgessler:
…yDict<T>. It does not use it directly, but instantiation of LazyDict<T> creates a reference.

My understanding is that compilers were correct in rejecting this, but I may be missing some detail of C++' template instantiation rules.


code/glTFAsset.h
```diff
@@ -786,8 +790,10 @@ namespace glTF
         void AttachToDocument(Document& doc);
         void DetachFromDocument();
 
+#ifdef  GLTF_ASSET_WITH_EXPORT_SUPPORT
```
turol: This is not correct. GCC 6 with LTO gives me this:
warning: virtual table of type ‘struct LazyDict’ violates one definition rule [-Wodr]
note: the conflicting type defined in another translation unit has virtual table table with more entries

Looks like glTFAssetWriter.h defines GLTF_ASSET_WITH_EXPORT_SUPPORT and then includes this header while other translation units don't define it. This violates the one definition rule. I suspect this would break devirtualization in LTO and might even break virtual dispatch in normal compiles.

acgessler: Makes sense. The more obvious fix is then to include the exporter implementation headers in all cases. I'll do that.

Issue #883, opened 2016-05-10
glTF Support: 


kimkulling:
Many test models (e.g., 2CylinderEngine) and also the model in the repository appear to have their hierarchy imported incorrectly and/or textures flipped.

See for instance:

http://i.imgur.com/RIQ9uS6.png

smalcom:
in last version everything is fine

![gltf_motor](https://cloud.githubusercontent.com/assets/19540675/17292472/f737374e-57f2-11e6-9a07-cc27f98fcf94.png)

PR #885, opened 2016-05-10
Fixed GLTF bugs, and added a few test models

otgerrogla:
Fixes issue #883 and a few others:
- Node transform matrices being ignored (it was a typo)
- Texture coords are now flipped on Y, which seems to fix the texturing issues
- Crash on loading old GLBs, due to spec changes
- Materials under KHR_materials_common extension were not imported correctly

Also, added some test models and updated the test DB.


kimkulling:

PR #889, opened 2016-05-16
Adds support for area lights.

ciechowoj:
Adds support for importing blender area lights.


kimkulling:
Great work, thanks a lot for the contribution ( even with a unittest ).

I will merge it!

ciechowoj:
Thanks.

PR #890, opened 2016-05-17
Add fix for IRR vs IRRMESH importer registry issue

TheCodeInside:
When compiling without IRR file support, there was an error with the importer registry not differentiating between the IRR and IRRMESH support macros, the latter of which was not defined in the CMake script. This change just adds in the IRRMESH macro and supports building both (IRR and IRRMESH), either, or neither.


kimkulling:

PR #891, opened 2016-05-18
Blender materials

ciechowoj:
Adds blender material parameters (part of) under custom keys ("$mat.blend.parameter_name") in aiMaterial.


acgessler:
Thanks!

PR #850, opened 2016-04-06
Replaced boost with c++11 features

mensinda:
Resolves Issues:
  - #409
  - #825
  - #826 (used `std::to_string()`)
  - #827
  - #828 (used `std::unique_ptr<T[]>`)
  - #829

I did not remove the BoostWorkaround yet.

PR #898, opened 2016-05-22
Use C++11 range-based for loops

turol:
Used clang-modernizer to refactor some for loops and then cleaned up by hand. These are only those which the modernizer considers "safe".

### DIFF
--- a/code/LWOMaterial.cpp
+++ b/code/LWOMaterial.cpp
@@ -90,8 +90,8 @@ bool LWOImporter::HandleTextures(aiMaterial* pcMat, const TextureList& in, aiTex
     aiString s;
     bool ret = false;
 
-    for (TextureList::const_iterator it = in.begin(), end = in.end();it != end;++it)    {
-        if (!(*it).enabled || !(*it).bCanUse)
+    for (const auto &texture : in)    {
+        if (!texture.enabled || !texture.bCanUse)
             continue;
         ret = true;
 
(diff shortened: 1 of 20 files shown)


acgessler:
@kimkulling 

I may lack some recent status here, but I thought we explicitly don't do C++11 yet.

I occasionally use MSVC11 to compile, which does not support all of it (it would support range based for loops however).

turol:
C++11 has been required since #850 was merged. Range-based for loops are already used. I'm not sure what MSVC minimum requirement has officially been set, if any, and what C++11 features are missing. Is "auto" supported? It would allow removal of lots of iterator boilerplate.

code/LWOMaterial.cpp
```diff
@@ -343,24 +343,24 @@ void LWOImporter::ConvertMaterial(const LWO::Surface& surf,aiMaterial* pcMat)
 
     // Now we need to know which shader to use .. iterate through the shader list of
     // the surface and  search for a name which we know ...
-    for (ShaderList::const_iterator it = surf.mShaders.begin(), end = surf.mShaders.end();it != end;++it)   {
-        //if (!(*it).enabled)continue;
```
kimkulling: Remove dead code.

kimkulling:
This quetsion comes up some months ago: use c++11 or not. Most of the votes were for using it. So ...

code/LWOMaterial.cpp
turol: It was there already. Do you know who commented it out and why?

kimkulling:
Merged, thanks a lot!

PR #903, opened 2016-05-26
fix strncpy not null terminated string

dontbmh:
The project was built under vs2015. A char array with uninitialized values will cause not null terminated string error when using strncopy. Thus Marshal.PtrToStringAnsi will return garbled text in AssimpNet.

### DIFF
--- a/code/AssimpCExport.cpp
+++ b/code/AssimpCExport.cpp
@@ -70,11 +70,11 @@ ASSIMP_API const aiExportFormatDesc* aiGetExportFormatDescription( size_t index)
     }
 
     aiExportFormatDesc *desc = new aiExportFormatDesc;
-    desc->description = new char[ strlen( orig->description ) + 1 ];
+    desc->description = new char[ strlen( orig->description ) + 1 ]();
     ::strncpy( (char*) desc->description, orig->description, strlen( orig->description ) );
-    desc->fileExtension = new char[ strlen( orig->fileExtension ) + 1 ];
+    desc->fileExtension = new char[ strlen( orig->fileExtension ) + 1 ]();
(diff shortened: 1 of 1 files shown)


code/AssimpCExport.cpp
```diff
@@ -70,11 +70,11 @@ ASSIMP_API const aiExportFormatDesc* aiGetExportFormatDescription( size_t index)
...
 
     aiExportFormatDesc *desc = new aiExportFormatDesc;
-    desc->description = new char[ strlen( orig->description ) + 1 ];
+    desc->description = new char[ strlen( orig->description ) + 1 ]();
     ::strncpy( (char*) desc->description, orig->description, strlen( orig->description ) );
-    desc->fileExtension = new char[ strlen( orig->fileExtension ) + 1 ];
+    desc->fileExtension = new char[ strlen( orig->fileExtension ) + 1 ]();
     ::strncpy( ( char* ) desc->fileExtension, orig->fileExtension, strlen( orig->fileExtension ) );
-    desc->id = new char[ strlen( orig->id ) + 1 ];
+    desc->id = new char[ strlen( orig->id ) + 1 ]();
```
kimkulling: Is this portable to other compilers?

kimkulling:
Hi,

I am not sure if using a memset to \0 makes more sence, because VS2015 behaviour is not implemented on all other supported compilers.

code/AssimpCExport.cpp
dontbmh: I think zero initialized array is compatible to other compilers. But I'm not sure because currently I only have vs environment at hand.

kimkulling:
Merged, thanks for the fix!

PR #908, opened 2016-06-06
Added 'joint' node type to Collada export

Adjective-Object:
Allows skeleton information to be exported to Collada by adding the node 'type' field with options 'NODE" and 'JOINT'.


kimkulling:
I will update the regression test suite as well. Thanks a lot for the work!

PR #916, opened 2016-06-10
Fixed some gltf errors

likangning93:
We were playing with assimp's gltf exporter and noticed a few small errors in the files we were getting.
* the byte offsets were being placed in the accessors instead of the bufferViews and the bufferViews all overlapped as a result
* there was a logic error preventing bufferViews looking at index values from being written as targeting index arrays.

I've attached the sample files we used. We exported `cube.obj` as `cube_unedited.gltf` using a build based on d64006b6b341a224cff23917d81b511f37578eb4 and exported `cube_assimp_rebuilt.gltf` using d64006b6b341a224cff23917d81b511f37578eb4 with our changes.

[cube.zip](https://github.com/assimp/assimp/files/309100/cube.zip)


likangning93:
That last commit: correction, tabs to spaces. My bad.

kimkulling:
Thanks a lot, merged!

pjcozzi:
Thanks for the prompt support @likangning93 @kimkulling!

@kimkulling FYI - we are integrating assimp into http://cesiumjs.org/convertmodel.html to support more input formats that COLLADA.  This will be a huge benefit to the glTF community thanks to assimp!

turol:
@pjcozzi: It looks like you're uploading the files to your server and doing the conversion there. This is very dangerous since assimp is not robust in face of malformed inputs. It's pretty trivial to find memory and stack corruption bugs in it.

pjcozzi:
@turol we'll see about adding a validation layer at the start of the pipeline.

kimkulling:
I guess being not stable in face of malformat inputs is bad for assimp. So we shall work on this, too.

pjcozzi:
Awesome, @kimkulling keep me posted on your progress!

turol:
@pjcozzi I'm not sure how validation would help since the crash would happen while assimp is parsing the input file before any validation layer has a chance to run. If you tried to put validation before assimp loading you'd end up doing essentially the same job assimp is already doing.

You could sidestep the issue entirely by using emscripten to compile your software to js and running entirely in the user's browser. This way any bad effects should be sandboxed by the browser. Model size might be limited by memory issues though.

@kimkulling If you want to make assimp more robust start with #454. I've been slowly working through those issues but my time (and understanding of assimp codebase) is limited.

PR #923, opened 2016-06-24
Fix compilation of iOS static libraries

tgoulart:
This fixes two different issues when generating iOS static libs.

1. Building for any targets using clang-703.0.31 failed with multiple errors. Using c++11 or c++14 as -std builds successfully for me. Added option to change the -std via command line, even though I still don't understand why c++03 is supposed to work.
2. Building for arm64 failed compilation due to duplicate constructors in rapidjson's document.h.

This is the command I used to build (from ports/iOS): $ ./build.sh --archs="x86_64 arm64"
Compiler version: Apple LLVM version 7.3.0 (clang-703.0.31)

### DIFF
--- a/contrib/rapidjson/include/rapidjson/document.h
+++ b/contrib/rapidjson/include/rapidjson/document.h
@@ -557,7 +557,7 @@ public:
             flags_ |= kIntFlag;
     }
 
-#if !defined(__x86_64__)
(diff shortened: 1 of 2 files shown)


kimkulling:
Thank you so much for that fix!

Issue #809, opened 2016-02-29
Vertex colours support for Obj


sahilagnihotri:
It seems that in the outer hirearchy we have vertex colours but in objparser we use getVector3 if the line starts with v. So even if there are vertex colours we discard them

kimkulling:
I see, the obj format supports some more surprises for me: http://gamedev.stackexchange.com/questions/21303/how-can-i-include-vertex-color-information-in-obj-files

Thanks for the hint, I will fix that issue.

sahilagnihotri:
You are welcome :) Another suggestion in formats where vertex colors doesnt exists, we give zero since that is default value which can sometimes give illusion that might not exist in file. Perhaps -1 or smthng else is a better defaullt maybe ? then you can have a compare againt that and know that ok its not that the library didnt read it from file, format doesnt support it.

kimkulling:
Fixed by  886cea2.

aybe:
Could you upgrade the OBJ exporter to also write vertex colors when there are any ?

Thank you !

PR #927, opened 2016-06-28
adding support for per-vertex colors in OBJ files.

royshil:
This add the option to parse OBJ files for per-vertex colors, specified in the `v <xx> <yy> <zz> <rr> <gg> <bb>` format.
This format is used by Meshlab and MeshMixer, which are very popular.
Also will then close issue #809


royshil:
Awesome thanks!

PR #931, opened 2016-07-01
C4D Support: Update to Newest Melange SDK

cmdrf:
This adds support for the V16 Melange SDK. It removes support for the outdated V6 SDK, as this would probably result in some ugly #ifdeffing. All versions of MSVC supported by the SDK are handled by CMake. I only tested VS2015 though.

The Melange SDK can be downloaded here: https://developers.maxon.net/?page_id=1112


kimkulling:
Thanks a lot the that contribution. But what about older versions. If I understand your changes right the older version will not build/work anymore, right?

cmdrf:
Yes, support for pre-V15 SDKs is dropped. Since the recent SDKs can be downloaded as easily as the old one, I didn't see much use for supporting the old one. However, the new SDKs drop support for:
* VS2008 (which is ancient, probably doesn't matter)
* 32-bit builds (don't know if this still matters today)
So you had to trade off 32-bit support against a more complex configuration and ugly preprocessor code.

kimkulling:
Ok, lets do it! Thanks for the contribution.

PR #932, opened 2016-07-04
Allowing to enable building of zlib manually through CMake option.

r-chris:
Currently, `assimp`'s `cmake` automatically tries to find `zlib` and reverts to building an internal version if the library was not found. There is no easy option to force the internal build to go ahead. Preventing the `find_package(zlib)` call can also be useful to use another, externally build zlib version when embedding `assimp` into a larger project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Open Asset Import Library (assimp)
 # ----------------------------------------------------------------------
-# 
+#
 # Copyright (c) 2006-2016, assimp team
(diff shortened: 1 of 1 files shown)


kimkulling:
Good idea, thanks a lot for the contribution.

r-chris:
No problem - btw - do you have any release schedule in mind for assimp? It looks like you have quite a few changes since 3.2 :+1:

r-chris:
Great to see a new release :rocket:

Issue #638, opened 2015-08-17
Double precision support


kimkulling:
Hi,
I am trying to import an .obj file that contains some geo-referenced model. All vertices are defined with an absolute offset (no mesh transformation) and because of that, the vertices coordinates are very big, for instance:
v 1024199.614999999 6280752.805999999 0.9109999999890448
Now, the .obj parser, but also the resulting imported mesh, store the data in arrays of aiVector3D, which is defined as follow:
typedef aiVector3t<float> aiVector3D;
I would like all the import pipeline to read my values (at least the vertices position) using double precision floating points.
I guess I will have to hack the source at some locations, but before I start this, I would like to check with you guys if there was a better existing method ?
Thanks a lot,
Greg

r-chris:
If I understand correctly, you made a branch to tackle this problem here:
https://github.com/assimp/assimp/tree/feature/configured_precision

Any plan to merge this to master - seems like a pretty solid contribution and opt-in for anyone who doesn't need it.

r-chris:
I rebased that branch to master, but not sure if there are many outstanding fixes to be done since float is still used in many instances:
https://github.com/r-chris/assimp/tree/configured_precision

My actual problem here is with the export, though - I currently see all exported values written like this:

```
v  557760 4.07449e+06 -49.1995
v  557760 4.07449e+06 -49.095
v  557760 4.0745e+06 -49.0082
v  557760 4.0745e+06 -49.1127
```

It looks like  you are using

```
public:
    std::ostringstream mOutput, mOutputMat;
```

and then something like this (obj example):

```
mOutput << "# " << vp.size() << " vertex positions" << endl;
for(const aiVector3D& v : vp) {
    mOutput << "v  " << v.x << " " << v.y << " " << v.z << endl;
}
mOutput << endl;
```

Apparently, you can technically set the `stringstream` precision to 16 (http://www.cplusplus.com/reference/ios/ios_base/precision/) to get more than 6 digits as output.

I understand that `mOutput` is a public variable, but I am calling `Assimp::Exporter exporter;`, which doesn't expose this to me - it would be great to hear if you have any advice how I should increase the output precision (or will this require source code changes in `assimp`)?

r-chris:
After some digging in the different exporters it appears that some of them already set a higher precision for export, while others don't. I have made this behaviour more consistent and will send a PR #935 your way.

r-chris:
[...]

PR #935, opened 2016-07-06
setting default export precision to 17 for all exporters

r-chris:
It appears like a good idea to increase the export precision to a consistent number (discussed below), so (`stl`, `obj`, `ply`, etc) exporters are not cutting values of the models. Currently, some exporters did define higher precisions, but most did not - this will fix that and ensure you don't loose too much precision upon export. _It may be a good idea to also allow double precision in general (#638)._

_from: http://stackoverflow.com/questions/554063/how-do-i-print-a-double-value-with-full-precision-using-cout_

> The precision needs to be 17 (or std::numeric_limits<double>::digits10 + 2) because 2 extra digits are needed when converting from decimal back to the binary representation to ensure the value is rounded to the same original value. Here is a paper with some details: http://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html
> 
> [C++11 introduces max_digits10](http://www2.open-std.org/JTC1/SC22/WG21/docs/papers/2005/n1822.pdf) to denote the same.

### DIFF
--- a/code/ColladaExporter.cpp
+++ b/code/ColladaExporter.cpp
@@ -94,7 +94,7 @@ ColladaExporter::ColladaExporter( const aiScene* pScene, IOSystem* pIOSystem, co
 {
     // make sure that all formatting happens using the standard, C locale and not the user's current locale
     mOutput.imbue( std::locale("C") );
-    mOutput.precision(17);
+    mOutput.precision(16);
 
     mScene = pScene;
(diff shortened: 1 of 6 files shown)


r-chris:
Ok, even though the export with precision 17 does work - it may be a problem for other readers to process this properly - I might revert back to 16 ...

kimkulling:

PR #939, opened 2016-07-07
FIX: Fixed UB for aiMatrix4x4t<TReal>::operator[]

SahinVardar:
I was trying to use the operator[] in my code but it did not work. I saw your comment and fixed the issue. See commit message.

### DIFF
--- a/include/assimp/matrix4x4.inl
+++ b/include/assimp/matrix4x4.inl
@@ -243,8 +243,7 @@ inline TReal* aiMatrix4x4t<TReal>::operator[](unsigned int p_iIndex) {
         return NULL;
     }
 
-    // XXX this is UB. Has been for years. The fact that it works now does not make it better.
-    return &this->a1 + p_iIndex * 4;
+    return &(&a1)[p_iIndex];
 }
 
 // ----------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


turol:
This is not any better. The undefinedness comes from the fact that the elements of the matrix are separate member variables and not arrays. Therefore trying to access them by either pointer arithmetic or array indexing is wrong. Either there needs to be some union trickery or operator[] must use a switch.

SahinVardar:
HI Turol,

thanks for you time. You are right, in theory it the compiler could add something in between, but most likely it wont. 

I will think  about to solve this. I imagine a union with an unnamed struct and an float array.

But in any case the pointer magic is wrong, because you multiply the index with 4. The intention here seems like you want to increase the pointer by 4 bytes. But since your pointer is a float pointer every increase or decrease occurs in a multiple of sizeof( pointer type aka float in this case )

So when you increase the float pointer by one. The address of the float pointer will increase by 4 bytes
In general every pointer arithmetic will automatically take the size of the pointer type into account.

In case you want to increase your pointer byte wise, you have to case your pointer to a byte sized type like char.

SahinVardar:
Sorry my intention was not to close it, but I can create a new commit or you just delete the 4.

sagaceilo:
The pull request intention is wrong, as that the 4 was put to offset by 4 bytes. It returns offset to row, so that You could access matrix this what :    matrixObject[a][b] (First bracket returns offset to row/column - depending of your order of thinking).
The best way of deal with this would be to add a MatrixRow that would hold 4 floats, and Matrix would hold 4 of MatrixRows. Both classes could have operator[] with some assert if index exceeds <0..3> boundry.

SahinVardar:
Ah thanks for the clarification. I was so used to work with matrices as an float array [16].
Now it makes sense for me.

PR #947, opened 2016-07-15
Optional double precision

r-chris:
I spend some time to try and get some basic `double` precision support working in `assimp`. It's an optional compile-time flag that will allow anyone to request `double` precision through `CMake`.

For us this is important when exporting largely displaced (or very large) models, where too much precision would be lost when relying only on 32-bit `float`s. **There is still work required to also make the Importers benefit from this**, but it's laying the ground work and I think one massive pull-request is enough for now :).

I understand that these are a lot of changes at once, but unfortunately, there isn't really a way to do this incrementally. I am pretty confident with most of the changes, but please check if you find some time. 

> Please look at the change in: `include/assimp/material.h : 518`, where I am adding `aiPTI_Double   = 0x2` to the `aiPropertyTypeInfo enum` and wasn't 100% sure which value I really should be using here, but `0x2` appears to make sense and was not used elsewhere, while `0x1` and `0x3` were actually assigned.

Any comments and suggestions are of course welcome.


r-chris:
**Update:** I added double precision import support for the formats that can be exported.

I most likely have to change the `config.h` file to include a `#cmakedefine AI_DOUBLE_PRECISION 1` to allow external projects to pick-up this define easily. Let me know if you have a better / different way in mind for this!?

r-chris:
Build is broken due to: #948

CMakeLists.txt
```diff
@@ -95,9 +95,17 @@ configure_file(
   ${CMAKE_CURRENT_BINARY_DIR}/revision.h
 )
 
+SET( HEADER_PATH include/assimp )
+
+configure_file(
+  ${HEADER_PATH}/config.h.in
+  ${ASSIMP_INCLUDE_INSTALL_DIR}/assimp/config.h
```
r-chris: @kimkulling looks like this is causing issues on the build server - maybe `ASSIMP_INCLUDE_INSTALL_DIR` is not defined at this point?

r-chris: @kimkulling  alright, I changed this and it hopefully works now

r-chris: ignore that - it's late over here - I am losing focus I think ... let me try something else

r-chris:
Ok, I'm off for now - this should technically work , but I have only build it on linux - so there might be some testing required for windows / android / iOS / etc.

kimkulling:
Looks good for me. Let's try it. Thanks a lot for your great work.

PR #953, opened 2016-07-24
Fix blender vertex colors being negative, and fix blender vertex colors scaling to be 0 to 1

TrianglesPCT:
Blender vertex color was importing as negative values. This was because it was being converted to a signed char. Needed to be unsigned.

Also aiColor4 specifies that the color will be 0 to 1, but blender importer was dumping out -128 to +127. Rescaled it so that it is now 0 to 1

### DIFF
--- a/code/BlenderLoader.cpp
+++ b/code/BlenderLoader.cpp
@@ -1118,7 +1118,7 @@ void BlenderImporter::ConvertMesh(const Scene& /*in*/, const Object* /*obj*/, co
             const aiFace& f = out->mFaces[out->mNumFaces++];
 
             aiColor4D* vo = &out->mColors[0][out->mNumVertices];
-			const ai_real scaleZeroToOne = 1.f / 255.f;
+			const ai_real scaleZeroToOne = 1.f/255.f;
             for (unsigned int j = 0; j < f.mNumIndices; ++j,++vo,++out->mNumVertices) {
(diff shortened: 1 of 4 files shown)


code/BlenderLoader.cpp
```diff
@@ -1118,12 +1118,13 @@ void BlenderImporter::ConvertMesh(const Scene& /*in*/, const Object* /*obj*/, co
...
                 const MLoopCol& col = mesh->mloopcol[v.loopstart + j];
-                vo->r = col.r;
-                vo->g = col.g;
-                vo->b = col.b;
-                vo->a = col.a;
+                vo->r = float(col.r) * scaleZeroToOne;
+                vo->g = float(col.g) * scaleZeroToOne;
+                vo->b = float(col.b) * scaleZeroToOne;
+                vo->a = float(col.a) * scaleZeroToOne;
             }
```
r-chris: Hi, could you please use `ai_real` instead of `float` here?
(`aiColor4D` is a `typedef` dependent on `ai_real` and should contain `double` when compiling for higher precision)

TrianglesPCT: Hi chris, it uses ai_real now

r-chris:
Looks like that failed build was just a random `gtest` fetch error.

acgessler:
Thanks!

PR #956, opened 2016-07-25
Fixes a few build issues with MSVC 2015 and gcc

adishavit:
There are still a whole bunch of `conversion from 'double' to 'ai_real', possible loss of data` warnings on MSVC2015, but the first commit fix already removes many.
The second fix a prevent a collision with the gcc "keyword" `typeof` by renaming it to `type_of`.


kimkulling:
Looks good, thanks for the contriubution.

PR #958, opened 2016-07-27
Fixed library to use less memory when loading and processing models.

### LINKED ISSUE #944: Large model issues
I'm trying to import a relatively large model, it's 8 million vertices and 380MB in OBJ format and 130MB in FBX format. The memory consumption from assimp is quite excessive, at its peak it was around 29GB before settling down to 11GB. This process took over an hour due to my laptop only having 8GB of physical RAM.

I've managed to narrow this issue down to the mini-allocations that occur all throughout the code for the `aiFace` struct. I replaced this with one big allocation (similar to what is done in the "Improve Cache Locality" processing step) and this reduced the peak memory footprint to 1.7GB and loaded in under 16 seconds.

Unfortunately, my solution is very hacky and I've only made this change to the FBX importer and the main culprits in the post processing step. Currently my solution makes the assumption that the different file importers will never do mini-allocations for the indices but this isn't very intuitive nor is it possible to prevent this from occurring.

I think the best solution would be to have another array in the `aiMesh` struct for the indices, and change the `aiFace` struct to have an index into that array instead of a pointer. Unless there's a better or preferred solution to this problem?

I'm also more than happy to make the changes to get this working.

samsinsane:
This is a pretty big change, but it removes all of the mini-allocations that occurred for indices. These mini-allocations would result in excessive memory usage, at least on Windows (potentially not an issue on Windows 10 though). Resolves #944

### DIFF
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -1286,8 +1286,6 @@ unsigned int Converter::ConvertMeshSingleMaterial( const MeshGeometry& mesh, con
         if ( uvs.empty() ) {
             break;
         }
-        out_mesh->mNumIndices = std::accumulate(faces.begin(), faces.end(), 0);
-        out_mesh->mIndices = new unsigned int[out_mesh->mNumIndices];
 
         aiVector3D* out_uv = out_mesh->mTextureCoords[ i ] = new aiVector3D[ vertices.size() ];
         for( const aiVector2D& v : uvs ) {
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -1209,12 +1209,14 @@ unsigned int Converter::ConvertMeshSingleMaterial( const MeshGeometry& mesh, con
     // generate dummy faces
     out_mesh->mNumFaces = static_cast<unsigned int>( faces.size() );
     aiFace* fac = out_mesh->mFaces = new aiFace[ faces.size() ]();
+    out_mesh->mNumIndices = std::accumulate(faces.begin(), faces.end(), 0);
+    out_mesh->mIndices = new unsigned int[out_mesh->mNumIndices];
 
     unsigned int cursor = 0;
     for( unsigned int pcount : faces ) {
         aiFace& f = *fac++;
         f.mNumIndices = pcount;
-        f.mIndices = new unsigned int[ pcount ];
+        f.mIndices = cursor;
         switch ( pcount )
         {
         case 1:
@@ -1231,7 +1233,7 @@ unsigned int Converter::ConvertMeshSingleMaterial( const MeshGeometry& mesh, con
             break;
         }
         for ( unsigned int i = 0; i < pcount; ++i ) {
-            f.mIndices[ i ] = cursor++;
+            out_mesh->mIndices[ f.mIndices + i ] = cursor++;
         }
     }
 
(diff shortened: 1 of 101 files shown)


include/assimp/mesh.h
```diff
@@ -129,27 +129,24 @@ struct aiFace
     //! The maximum value for this member is #AI_MAX_FACE_INDICES.
     unsigned int mNumIndices;
 
-    //! Pointer to the indices array. Size of the array is given in numIndices.
-    unsigned int* mIndices;
```
jdduke: This will break every single Assimp client in existence, and require non-trivial changes when upgrading. Ideally this kind of optimization would preserve the public-facing struct API in some fashion, otherwise it'll probably have to wait until the next major version bump.

I took a stab at something similar a while ago in an attempt to minimize heap usage (https://github.com/assimp/assimp/compare/master...jdduke:noheap_face). It preserves the outward-facing aiFace struct data members, but eliminates heap allocations for faces with index count <= 4 (albeit in a somewhat crude fashion, index storage on the mesh is a better long term solution, but unless we make mIndices a pointer into the mesh index array, which has its own set of issues, this is a breaking change).

samsinsane: If I change it to be a pointer that references the indices array in the mesh, will this PR be accepted? This solution works for us, and I'm happy to make the change if that's the only thing stopping it from being accepted, but I've already spent way too much time working on this.

kimkulling:
I guess we need to release this changes ( when they are tested and reviewed ) in a release 4.0. As jdduke mentioned before this will break the API and every client need to change his code.

samsinsane:
> I guess we need to release this changes ( when they are tested and reviewed ) in a release 4.0.

What's the time frame on the 4.0 release?

> As jdduke mentioned before this will break the API and every client need to change his code.

If the breaking the API is the only thing blocking this from being merged, I can change the `mIndices` back to a pointer that references the indices array in the mesh? It's up to you, if you want this to wait to 4.0 anyway, then I'll leave it as is.

kimkulling:
I think we can release 4.0 next .....

Samuel Surtees <notifications@github.com> schrieb am Fr., 29. Juli 2016
03:54:

acgessler:
I think I'd be in favor of not breaking the API and have mIndices refer to the indices array.

Also it should be backwards-compatible -- importers should still be able to allocate their own data arrays. I am not sure how it would work with deleting, perhaps we should add a boolean to aiFace so it knows whether to free the data.

samsinsane:
> Also it should be backwards-compatible -- importers should still be able to allocate their own data arrays. I am not sure how it would work with deleting, perhaps we should add a boolean to aiFace so it knows whether to free the data.

This kind of defeats the purpose of this PR. The large memory consumption I'm experiencing is due to the faces having their own allocations for indices. Additionally, adding more data per face doesn't help with the data consumption problems. This is why I favour the API changing, it prevents this issue from reoccurring by not allowing you to make mini-allocations for each face.

kimkulling:
One question: will this pull request in the current state break the ABI? If not: we should merge it!

kimkulling:
I thought quiet a long time about this patch. We should try a different approach before merging this one. 
My idea in short:
Using pool allocators for faces. So a big bundle of faces will be allocated in the beginning and we just using this to store data inside of the faces. So we do not have any API-changes ( we have a lot at the moment and from my point of view the patch will create a lot of anger ) and we can get you idea in.

I will prepare a special branch for testing this.

code/FBXConverter.cpp
```diff
@@ -1284,6 +1286,8 @@ unsigned int Converter::ConvertMeshSingleMaterial( const MeshGeometry& mesh, con
         if ( uvs.empty() ) {
             break;
         }
+        out_mesh->mNumIndices = std::accumulate(faces.begin(), faces.end(), 0);
+        out_mesh->mIndices = new unsigned int[out_mesh->mNumIndices];
```
OsiJr: This line wipes out the existing mIndices (withing freeing) and results in the index buffer being unallocated.

PR #965, opened 2016-08-08
fix long option name for --om

waldyrious:
to match the actual code in tools/assimp_cmd/Main.cpp


kimkulling:

PR #970, opened 2016-08-08
Added documentation for embedded textures

nyorain:
Added to the documentation of aiMaterial::GetTexture that the returned path string is a '*' followed by the texture id if the texture is embedded.

I searched quite a while on how to handle embedded textures and could not find anything in the docs. I think such a hint would be quite useful.
Maybe I was just to stupid to find the place this is already documented, then feel free to tell me so and close this request.
I additionally fixed some indentation issues for the documentation of the GetTexture function.


acgessler:
Thank you!

PR #983, opened 2016-08-15
adding support to store all the layered textures

roshanch:
1. Storing all the layered textures and blend mode.
2. Some of the FBX files has "SpecularFactor" with specular texture, adding support for that.


kimkulling:
Thanks for the fix ...

PR #985, opened 2016-08-18
Building own zlib on 64bit Linux shared library requires -fPIC for C compiler.

Sailsman63:
Some of the files from the internally built zlib are too big to be placed in a .so when compiled without -fPIC on 64 bit platform. Tested on up to date Arch Linux, running virtualized on x86_64.

CMake 3.6.1
GNU make 4.2.1
gcc 6.1.1 20160802

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,9 @@ ENDIF()
 IF((CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX) AND NOT CMAKE_COMPILER_IS_MINGW)
   IF (BUILD_SHARED_LIBS AND CMAKE_SIZEOF_VOID_P EQUAL 8) # -fPIC is only required for shared libs on 64 bit
      SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+     IF(ASSIMP_BUILD_ZLIB)
+	     SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
+     ENDIF()
   ENDIF()
   # hide all not-exported symbols
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden -Wall -std=c++0x" )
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -164,6 +164,9 @@ ENDIF()
 IF((CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX) AND NOT CMAKE_COMPILER_IS_MINGW)
   IF (BUILD_SHARED_LIBS AND CMAKE_SIZEOF_VOID_P EQUAL 8) # -fPIC is only required for shared libs on 64 bit
      SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+     IF(ASSIMP_BUILD_ZLIB)
+	     SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```
acgessler: Should we just set this independently of whether we build our own zlib? We set it for C++, so why not do it for C.

Sailsman63: That might well be. I'm just a newb to C/C++ development, this was the minimum circumstance that would allow compilation in the configuration I was testing. 

When doing the searching to figure things out this far, I found that -fPIC will inflate the size of the *.o files. I doubt that the few C files affected would be significant when all of the C++ are already using the option. Don't know enough to evaluate any other possible side effects.

BTW, I think this would have been caught automatically if the travis build had a configuration with `ASSIMP_BUILD_ZLIB=ON` would that be worth adding? I can run up a separate PR if you like.

acgessler: That's right in principle, but it will make Travis builds take even longer and be more vulnerable to flaky failures :-( I'd say, let's just always specify FPIC. Yes, it can blow up binary sizes but I also think that it won't matter much for the amount of C code we compile.

Sailsman63:
Sorry, I messed up the history trying to pull in the latest from master. (I'm still learning) would you prefer a clean branch?

kimkulling:
I do agree to enable the feature globally. And the change looks good for me! So I will merge it.

And thanks for your contribution.

PR #998, opened 2016-09-14
Flip UVs in glTFExporter

johnmaf:
Related to #883. Fixes texture flipping problem, but not hierarchy.


kimkulling:
Ok, merged as well!

PR #999, opened 2016-09-14
Add all glTF data to single buffer

johnmaf:
Rather than creating one buffer per mesh, which is unnecessary, add all data to a single buffer.


kimkulling:

johnmaf:
No problem! Thanks for the quick response

PR #1001, opened 2016-09-20
Export glTF buffer data in a separate .bin file

johnmaf:
Changes:
* Export a `.bin` file containing buffer data, and reference that file in the `.gltf`, rather than embedding a base64 blob
* Factor GLB file logic out of `AssetWriter::WriteFile` into its own function
* Split meshes in constructor to cover GLB files
* use file name as base for buffer IDs (instead of 'buffer')


kimkulling:

PR #1004, opened 2016-09-21
Add Triangulate post-processing step to glTF exporters

### LINKED ISSUE #1002: glTF geometry buffer offset issue
I have an OBJ/MTL model that I need to convert to glTF.

The original source files are here: (OBJ format) https://github.com/richtr/three.js/tree/gh-pages/examples/models/gltf/Minion/source/. The output files from the coversion are here: (glTF format) https://github.com/richtr/three.js/tree/gh-pages/examples/models/gltf/Minion/glTF/.

I then use [`assimp`](https://github.com/assimp/assimp) to convert these files to the glTF format on the command line as follows:

`assimp export ./source/Minion.obj ./glTF/Minion.gltf`

This command completes successfully.

I can inspect the original source file and get the following:

``` bash
$> assimp info ./source/Minion.obj

Importing file ...                   OK
   import took approx. 0.20682 seconds

Memory consumption: 621408 B
Nodes:              4
Maximum depth       2
Meshes:             10
Animations:         0
Textures (embed.):  0
Materials:          9
Cameras:            0
Lights:             0
Vertices:           9259
Faces:              13572
Bones:              0
Animation Channels: 0
Primitive Types:    linestriangles
Average faces/mesh  1357
Average verts/mesh  925
Minimum point      (-4.818348 -2.837892 -5.243814)
Maximum point      (1.430433 2.149038 -2.422743)
Center point       (-1.693957 -0.344427 -3.833279)

Named Materials:
    'Detalhe'
    'Material'
    'Minion'
    'OLHOS'
    'Roupa'
    'Sapato'
    'boca'
    'fita'
    'oCULOS'

Node hierarchy:
'Minion.obj', meshes: 0
-- 'Plane', meshes: 2
-- 'Cylinder.002', meshes: 3
-- 'Sphere', meshes: 5
<--

```

However, when I run the following I get an error:

```
$> assimp info ./glTF/Minion.gltf

Assertion failed: (i*stride < accessor.bufferView->byteLength), function GetValue, file /Users/richt/Apps/assimp/code/glTFAsset.inl, line 584.
[1]    65714 abort      ./assimp info ~/glTF/Minion.gltf
```

Additionally, when I then load this glTF file via the glTF loader available in THREE.js [here](https://github.com/mrdoob/three.js/tree/dev/examples/js/loaders/gltf) there is then an issue with the output rendering that looks to be related.

It looks like only half the triangles are being rendered from this model.

[HERE IS THE LIVE VERSION](https://richtr.github.io/three.js/examples/?q=gltf#webgl_loader_gltf)

Here's a screenshot of the current rendering issue via THREE.js:

screen shot 2016-09-21 at 12 33 13 am

It feels like the offset or vertex stride is not being taken in to account in the glTF loader.

Any other thoughts/ideas as to the source of this issue would be welcome as I'm not sure where to begin with this.

johnmaf:
Fixes #1002.
However, it doesn't address a similar issue where a single mesh contains both lines and triangles.

### DIFF
--- a/code/Exporter.cpp
+++ b/code/Exporter.cpp
@@ -139,9 +139,9 @@ Exporter::ExportFormatEntry gExporters[] =
 
 #ifndef ASSIMP_BUILD_NO_GLTF_EXPORTER
     Exporter::ExportFormatEntry( "gltf", "GL Transmission Format", "gltf", &ExportSceneGLTF,
-        aiProcess_JoinIdenticalVertices /*| aiProcess_SortByPType*/),
+        aiProcess_JoinIdenticalVertices | aiProcess_Triangulate /*| aiProcess_SortByPType*/),
     Exporter::ExportFormatEntry( "glb", "GL Transmission Format (binary)", "glb", &ExportSceneGLB,
-        aiProcess_JoinIdenticalVertices /*| aiProcess_SortByPType*/),
+        aiProcess_JoinIdenticalVertices | aiProcess_Triangulate /*| aiProcess_SortByPType*/),
 #endif
 
 #ifndef ASSIMP_BUILD_NO_ASSBIN_EXPORTER


smalcom:
What if i do not need triangles?

johnmaf:
I'm not sure I understand your question. I think this makes sense, since the [glTF spec](https://github.com/KhronosGroup/glTF/tree/master/specification#primitivemode) only supports triangular faces:

> primitive.mode

> The type of primitives to render. Allowed values are 0 (POINTS), 1 (LINES), 2 (LINE_LOOP), 3 (LINE_STRIP), 4 (TRIANGLES), 5 (TRIANGLE_STRIP), and 6 (TRIANGLE_FAN).

> Type: integer
> Required: No, default: 4
> Allowed values: 0, 1, 2, 3, 4, 5, 6

So the triangulation has to happen somewhere, right? It's just a matter of whether it's before or during the glTF export.

smalcom:
Ah, understood.

PR #1006, opened 2016-09-24
Fixed Appveyor x64 builds

odalet:
I've noticed that the build script used by appveyor was always generating x86 binaries no matter what platform was defined. I fixed this in this PR. 

I made sure the generated DLL was indeed x64 by opening it in Depends then by using it from the Assimp-net wrapper.

You can check the generated binaries here: https://ci.appveyor.com/project/odalet/assimp/build/job/ulygh9s2j99sx9d8/artifacts

Hope this helps!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,9 +28,16 @@ install:
 - call c:\projects\assimp\scripts\appveyor\compiler_setup.bat
 
 build_script:
- - cd c:\projects\assimp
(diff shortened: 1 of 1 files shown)


kimkulling:
Ah, I tried to figure that out as well. Thanks a lot!

odalet:
You're welcome!

PR #1015, opened 2016-10-03
Fix glTF validator errors

ascandal:
This PR attempts to fix the errors when reading an exported glTF file into the glTF validator (Live drag-n-drop tool).

The validator can be found here.
githuh: https://github.com/KhronosGroup/glTF-Validator
Live drag-n-drop tool: http://github.khronos.org/glTF-Validator/

However, this PR is not finished and I am hoping somebody can help me with a few issues:

## Issues

* In `struct Accessor`,
 * `max` and `min` are `std::vector<float>` however they may be another type.
* In `inline Ref<Accessor> ExportData`, 
 * the Large value number is hard-coded as `float float_MAX = 10000000000000;` however I assume there is a macro constant somewhere in the code defined for a large value.
* In `struct Sampler`,
 * `magFilter` and `minFilter` are hard-coded because I can not find where this information is stored in the internal `assimp` structures` after import.
 * see lines: 280 - 281 in file `code/glTFExporter.cpp`.

I used `robot.dae` as an import test model to debug.
[robot-dae.zip](https://github.com/assimp/assimp/files/506748/robot-dae.zip)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,22 @@
 CHANGELOG
 ----------------------------------------------------------------------
 
+3.2.1 (2016-10-01)
+
+FEATURES:
+  - Updated glTF exporter to meet 1.0 specification.
+
+FIXES/HOUSEKEEPING:
+  - Fixed glTF Validator errors for exported glTF format.
+
+ISSUES:
+  - Hard coded sampler setting for
+    - magFilter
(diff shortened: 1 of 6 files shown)


kimkulling:
Merged and thx for the contribution.

PR #1016, opened 2016-10-03
Sort by primitive type before exporting as glTF

johnmaf:
Currently the glTF exporter [assumes](https://github.com/assimp/assimp/blob/master/code/glTFExporter.cpp#L382) that every face in a given mesh has the same number of indices. This assumption fails when a mesh contains e.g. both lines and triangles, and causes the exported geometry to be corrupted.

This change fixes that problem by setting `aiProcess_SortByPType` on the glTF exporter.

A potential alternative would be to separate mesh primitives by primitive type in the exporter itself.


kimkulling:
Looks good, will be merged.

PR #1018, opened 2016-10-04
Parse homogeneous vertex coordinates in OBJs

johnmaf:
This change allows import of OBJ files that define vertex positions in homogeneous coordinates (as allowed by the [OBJ spec](http://www.martinreddy.net/gfx/3d/OBJ.spec)).


kimkulling:
Oh, nice finding. Thanks a lot for the bugfix!

johnmaf:
No problem!

PR #1022, opened 2016-10-06
[+] Few models for AMF and X3D.

smalcom:
One AMF model i compressed because uncompressed size is near 70 MB.


kimkulling:
Gerät thanks!

PR #1026, opened 2016-10-10
glTF skinning animation

ascandal:
This PR is in regard to animation export for glTF format.

## FEATURES
* `Bugfix` for exporting rotation data for node animations.
* Enabled skin animation export by uncommenting code.
* Routines to find node which contains the given mesh id.
* Routine to find the root joint node of the skeleton.

## ISSUES
* There needs to be a better, more reliable, method to find the root joint of the skeleton.
* `glTF::Node` structure does not copy over the `assimp::node->mParent` information, making it difficult to track up the node hierarchy. Either have to add the parent information to `glTF::Node` or have all methods that search in the node hierarchy operation on the aiNode instead.
 * The difficultly being that there is a potential for node names to be different on export.

## Testing

`sampleModels/BoxAnimated`  - works good
`sampleModels/RiggedSimple` - works good
`robot`  - issue finding root joint node.


code/glTFExporter.cpp
```diff
@@ -753,9 +810,12 @@ inline void ExtractAnimationData(Asset& mAsset, std::string& animId, Ref<Animati
     //-------------------------------------------------------
     // Extract rotation parameter data
     if(nodeChannel->mNumRotationKeys > 0) {
-        C_STRUCT aiQuaternion* rotationData = new aiQuaternion[nodeChannel->mNumRotationKeys];
+        vec4* rotationData = new vec4[nodeChannel->mNumRotationKeys];
```
richtr: Rotation should be represented here as a quaternion.

ascandal: `nodeChannel->mRotationKeys` is represented as a quaternion internally to assimp. That is `C_STRUCT aiQuaternion`. However when it is exported we need it as a simple vector of length 4.

`rotationData` is only used to export the `mRotationKeys` to the buffer. During this export into the buffer, it issues the [] operator which does not exist for the `aiQuaternion` struct.

This also makes sure that it is output to the buffer in the order of `x, y, z, w` which is the glTF specification.

kimkulling:
Merged, thanks a lot!

PR #1027, opened 2016-10-10
Fixed: unzip.h was still included even if the D3MF importer is disabled

corporateshark:


kimkulling:
Ok, works. Thanks for the contribution.

corporateshark:
I guess this one can be closed.

PR #1029, opened 2016-10-11
fix bug when reading strings in quote in MD5 models

samitc:
When trying to load md5 models with string that has spaces in bones or shaders assimp can not load this model.
Fix this by read the string with the spaces inside the quote/


kimkulling:
Merged, thanks for the fix!

PR #1033, opened 2016-10-15
Fix mingw build (mingw supports stat64 nowadays)

ashdnazg:


kimkulling:
Ok, merge! Tanks for the contribution.

PR #1034, opened 2016-10-16
glTF compact skins

ascandal:
This PR is in regard to skins in glTF export.

## FEATURES
* Iterating over the aiMeshes, all the bones are collected into one skin.
* `parent` node reference member is added to the glTF asset Node to help when traversing up the node hierarchy.
* A vertex can only have at most four joint weights. After the first four, all extras are ignored.

## TEST MODELS
| Model  | Export Success |
| ------------- | ------------- |
| `gltf/sampleModels/BoxAnimated`  | OK  |
| `gltf/sampleModels/RiggedSimple`  | OK  |
| `gltf/sampleModels/CesiumMilkTruck`  | OK  |
| `gltf/sampleModels/RiggedFigure`  | OK  |
| `gltf/sampleModels/CasiumMan`  | OK  |
| `gltf/sampleModels/Monster`  | OK  |
| `gltf/sampleModels/BrainStem`  | OK  |
| `robot` | OK  |

## ISSUES
* `glTFExporter.cpp` needs some code refactoring. Specifically, the `ExportSkins` procedure is intricately tied into the `ExportMeshes` function.
* The `Animations` and `Skins` need to be considered for compression lines 640 - 729.
* Seems to work well for scenes with one complex model. however I am not sure how it will perform for a scene with multiple complex models.
* I have been fixing the skins and animations on a model-by-model basis. Ideally I think there must be a more correct way to do this.

### DIFF
--- a/code/glTFAsset.h
+++ b/code/glTFAsset.h
@@ -811,6 +811,8 @@ namespace glTF
         Ref<Skin>  skin;                          //!< The ID of the skin referenced by this node.
         std::string jointName;                    //!< Name used when this node is a joint in a skin.
 
+        Ref<Node> parent;                         //!< This is not part of the glTF specification. Used as a helper.
+
         Node() {}
         void Read(Value& obj, Asset& r);
     };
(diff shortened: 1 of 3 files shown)


kimkulling:
Thanks a lot for the contribution, merged!

PR #1037, opened 2016-10-17
cfileio: fix leaks by not closing files in the destructor

ystreet:
Numerous importers rely on the default C++ IOSystem implementation
having the destructor close the file for them.  The C IOSystem wrapper
wasn't and instead assumed that the Close() method was going to be
called.

This brings the C IOSystem wrapper in line with the default C++ IOSystem
by having the destructor close the file.


kimkulling:

PR #1040, opened 2016-10-17
Fix OBJ parser mtllib statement parsing bug.

### LINKED ISSUE #1041: regression: 0359ded94621a4bbba6a021ee5d62314916f6bd2 causes .mtl files referenced by mtllib in .obj to not be loaded
The reason for this seems to be the:
190	                else if (name == "mtllib")
(gdb) p name
$1 = "mtllib jeep.mtl"

johnmaf:
A previous commit used `getName` to parse lines beginning with 'm'. But since `getName` returns the whole line including spaces, the comparisons (e.g. `name == "mtllib"`) always fail, causing the line to be skipped.

In particular, any OBJ imported will be missing its material.

I added some additional parsing to get the correct behaviour back, but it might be best to revert the [commit](https://github.com/assimp/assimp/commit/0359ded94621a4bbba6a021ee5d62314916f6bd2) that introduced the problem.

Fixes #1041

### DIFF
--- a/code/ObjFileParser.cpp
+++ b/code/ObjFileParser.cpp
@@ -185,7 +185,11 @@ void ObjFileParser::parseFile()
                 std::string name;
 
                 getName(m_DataIt, m_DataItEnd, name);
-                name = name.substr(0, name.find(" "));
+
+                size_t nextSpace = name.find(" ");
+                if (nextSpace != std::string::npos)
+                    name = name.substr(0, nextSpace);
+
                 if (name == "mg")
                     getGroupNumberAndResolution();
                 else if(name == "mtllib")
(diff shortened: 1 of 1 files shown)


code/ObjFileParser.cpp
```diff
@@ -182,9 +182,10 @@ void ObjFileParser::parseFile()
 
         case 'm': // Parse a material library or merging group ('mg')
             {
-				std::string name;
+                std::string name;
 
-				getName(m_DataIt, m_DataItEnd, name);
+                getName(m_DataIt, m_DataItEnd, name);
+                name = name.substr(0, name.find(" "));
```
johnmaf: Thanks for reviewing my code.

If I understand correctly, `name.find(" ")` will return `npos` if `name` doesn't contain a space, and `name.substr(0, npos)` [should `== name`](http://en.cppreference.com/w/cpp/string/basic_string/substr). So if `name` contains no spaces, won't the extra line have no effect?

johnmaf: gotcha. Thanks for taking a look

kimkulling:
Merged!

PR #1042, opened 2016-10-19
Added support for 64 bit version header introduced in FbxSdk2016

VeganPower:
This add support to read FBX binary file with 64bit header.

Just after the magic word there is a 32 flags mask that was ignored before.
I'm using these flag to check if the file is 32 or 64 bit.
But I'm not 100% sure what these flags do.

Tested on Maya 2017 and 2015, works in both versions.


kimkulling:
Thanks for the vegan pull request, merged.

PR #1048, opened 2016-10-27
Keep original materials names in RemoveRedundantMaterials, ColladaExporter

tomacd:
Our workflow requires materials names to stay the same after process/conversion.
RemoveRedundantMaterials generates names "JoinedMaterials_#" and the ColladaExporter renames materials to "m#XXX".
This patch keeps the original materials names when possible.


kimkulling:
Merged, thanks a lot!

PR #1053, opened 2016-11-02
Fix compile error with MSVC 2015

rdb:
This patch fixes compilation with MSVC 2015.  WIN32 is not a preprocessor symbol that is defined by MSVC, but _WIN32 is.


kimkulling:
Merged, thanks a Lot...

rdb:
Thanks!

PR #1064, opened 2016-11-10
Fix obj .mtl file loading

snowzurfer:
Fix the obj file loader by adding a new method which allows
a name to be read considering the space in the middle between two
words and use that for parsing the "mtlib" line in the .obj file
parsing method.

Before, the method used in the obj parsing function would have
returned the string "mtlib NAME_OF_MTL" instead of "mtlib" only,
which resulted in the .mtl file being never parsed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,7 @@ OPTION( ASSIMP_BUILD_ZLIB
 )
 option( ASSIMP_BUILD_ASSIMP_TOOLS
   "If the supplementary tools for Assimp are built in addition to the library."
-  ON
+  OFF 
 )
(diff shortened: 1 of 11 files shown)


snowzurfer:
Having pulled, I have noticed how the original method has been updated basically to reflect what I have done in the new one. The method comment hasn't been updated though, and probably it would be a good idea to have two methods with both a specific functionality than one.

snowzurfer:
I thought I could open another pull request from master, having made a branch and rebased on top of master to then do the pull request more cleanly, but it interferes with this one; hence the revert.

kimkulling:
Merged, thx a lot for your contribution!

PR #1066, opened 2016-11-14
MSVC14 x64 /bigobj missing causes compile error

jaredmulconry:
Fixed a compile error on MSVC14 x64 caused by the /bigobj flag failing to be set
for the 1 and 2-suffixed versions introduced in
commit 0a25b076b8968b7ea2aa96d7d1b4381be2d72ce6.


kimkulling:
Ops, good finding! Thanks a lot!

kimkulling:
Merged!

PR #1071, opened 2016-11-18
AMF Postprocessed Metadata given type bool regardless of original type

jaredmulconry:
Fixed a bug in the AMF Importer Postprocessing where metadata would be incorrectly recorded as having type bool for all entries.
The warning originally surfaced inside of metadata.h, which indicated the original const char* was being interpreted as a bool, rather than an aiString as was obviously intended.


kimkulling:
Merged thanks a lot for the fix!

PR #1073, opened 2016-11-20
Update STLLoader.cpp

elect86:
If `faceVertexCounter` does not start equal 3, then at line 291 you will get always a warning (only one, but still..)


kimkulling:
Merged, thanks a lot for the fix.

Issue #123, opened 2013-09-28
mtl shininess value


kimkulling:
Using Assimp 3.0.1270 when reading the following wavefront obj mtl file:
newmtl Material_floor.jpg
Ns 96.078431
Ka 0.000000 0.000000 0.000000
Kd 0.640000 0.640000 0.640000
Ks 0.500000 0.500000 0.500000
Ni 1.000000
d 1.000000
illum 2
map_Kd floor.jpg
For shininess I get back 384.314, and shininess strength is 1.0. 
It seems that Ns is multiplied by 4 in line 534 in ObjFileImporter.cpp:
pCurrentMaterial->shineness *= 4.f;
According to the mtl specs (http://paulbourke.net/dataformats/mtl/):
Ns exponent
"exponent" is the value for the specular exponent. A high exponent
results in a tight, concentrated highlight. Ns values normally
range from 0 to 1000.
I'm not sure what the shininess range in Assimp is, It is 0-128 in OpenGl.
But now shininess is scaled to 0-4000.
I believe shininess should be the same regardless of the import format, e.g.
if I import the model from collada for example.

see: https://sourceforge.net/p/assimp/bugs/87/

kimkulling:
Closed by 56e9bce

PR #1079, opened 2016-11-25
Remove scaling of specular exponent in OBJFileImporter.cpp

johnmaf:
Currently the specular exponent in MTL files is being multiplied by 4 on import, causing imported models to be exponentially shinier than specified.

In addition, this scaling isn't matched in the exporter, so, for example, `assimp export my.obj my-new.obj` will yield a model with a much larger specular exponent than the original. Since Assimp isn't the only tool in our pipeline, this is causing us some headaches.

This was previously mentioned in #123, but not implemented. Is there any interest in revisiting this?


kimkulling:
I guess we should use this patch to avoid more cobfusions in the future.

kimkulling:

PR #1080, opened 2016-11-25
Properly export COLLADA <transparency> value

johnmaf:
This value is handled properly on import, but is inverted in the exporter. Since the default interpretation of this value is such that 0.0 is fully transparent and 1.0 fully opaque, which matches how the importer stores the value, inverting it on export is incorrect.

For example, running `assimp export my.dae my-new.dae` where `my.dae` is in `A_ONE` mode for transparency (the default), will yield a .dae with the transparency inverted from the original. You can also see this happening when converting e.g. OBJs to COLLADA.


kimkulling:
You're right. Thanks for the contribution!

Issue #1065, opened 2016-11-13
307 warnings with MSVC14 (Visual Studio 2015) in x64 build


TamasSzep:
Mostly coming from from sizeof(size_t) != sizeof(unsigned).

kimkulling:
Hm, we need to move to size_T I guess. Thanks for the report!

TamasSzep:
I would strongly discourage the usage of size_t overall. In fact, there are virtually no possible candidate that stores a size and needs more than 4 bytes. I like to follow the practice of using unsigned everywhere, and interacting with STL by casting it's size_t-s at the very first encounter:

auto countElements = static_cast<unsigned>(vector.size());
for(unsigned i = 0; i < countElements; i++) { /* */ }

It becomes important when storing a lot of indices, which is kind of the best practice of organizing your data, since it flattens your memory usage patterns, which makes your application efficient. As far as I known. : )))

jaredmulconry:
This seems like a question of design and intended use cases. Is Assimp intended to support meshes with more than 2^32 indices? Considering the presence of 'unsigned int' in all of the commonly used interfaces, I would guess not.
I think it doesn't make sense for Assimp to be using unsigned types in many of the places it is currently, since if any one of those overflows and wraps, I don't think good things would happen.
That said, it may not be practical to translate everything into signed types.

I'm interested in addressing the many build warnings that Assimp produces, so I'd be interested in discussing a way of approaching this in general, not just for MSVC.

TamasSzep:
Meshes with more than 2^32 indices, or even with more than a million indices are fairly impractical, one would split them for sure to enable more materials, separate culling, etc. But even if this should be supported, it's clear that sizes like the number of meshes, nodes, animations, materials will never be more than 2^32. So I would keep the current design and simply kill the warnings by casting to unsigned.
About signed vs. unsigned: signed becomes practical in implementations, where you want to iterate backwards. Since that is algorithmic question and has nothing to do with the libraries interface, I would keep the unsigned values since they express explicitly that you are dealing with sizes and indices. If somebody doesn't like it, he/she can cast it to signed anyway. : )))

kimkulling:
We had a discussion regarding this topic some years ago. For point of potability size_t is the right choise because this type was introduced to make size definitions better portable.
From the point of data we will not need this kind of type I guess. And it will be an API-change.

[...]

PR #1083, opened 2016-11-27
Fixed warnings on MSVC14 x64 (Issue #1065)

jaredmulconry:
This is intended to fix warnings when building with MSVC14 the x64 target.
Most of this was caused by mishandling of size_t. This is likely to be a recurring issue, unless we have a point in the style guide that requires conversions from size_t to be done with a static_cast.


kimkulling:
That were a lot of static casts :-). Thanks a lot for the contribution!

PR #1091, opened 2016-12-02
FBX: woraround for issue 1089.

kimkulling:
Workaround: 
- remove 64-bit flag lookup and use old way to read token.
- add free FBX-model 
- add FBX-importer unittest


ongamex:
WOW thanks for the quick action! I'll give it a try as soon as I get home.

PR #1099, opened 2016-12-08
Fixes two type warnings and a missing case option

asmaloney:


kimkulling:

PR #1106, opened 2016-12-09
OBJ, added the index to handle clamp option for each cubemap texture

elect86:
As titled 😄


kimkulling:
Looks good, thanks for the contribution! Merged.

PR #1119, opened 2016-12-20
Replace use of nullptr in headers with NULL

rdb:
It's fine if assimp uses nullptr internally, but not all projects that use it compile in C++11 mode, so this patch removes the 3 occurrences of nullptr from the public headers.

This allows projects that use assimp to build in C++03 mode.


kimkulling:
OK, thanks a lot! Merged!

PR #1123, opened 2016-12-26
Avoid unneeded copy-ctor calls when calling getAiType

iamsergio:
Specially for aiString, which does a memcpy.
Found with clazy static analyzer.


kimkulling:
Thanks a lot, merged!

Issue #807, opened 2016-02-25
FBX & Camera broken


vd3d:
Hi,

It seems that importing the camera with FBX is broken. I'm not the first one to notice this problem but they don't got any feedback (https://github.com/assimp/assimp/issues/643).

Does someone has some news about this problem ?

Thanks

vd3d:
Here is my code : http://stackoverflow.com/questions/35536994/assimp-import-camera

kimkulling:
Hi,

sorry for the delay. Trying to get some knowledge about this format. 

Kimmi

kimkulling:
I just checked the model from #643. Camera position is on the wrong side of the scene. And there is another camera in the scene, which was not parsed at all. I am still investigating this issue. At the moment I trying to get the right transformation of the camera via blender.

kimkulling:
Ok, after parsing with Assimp the camera from untitled.fbx contains a data definition for transform x = 1207 instead of 12.07 checked in Blender. Z differs also about factor 10. Y seems to be wrong. Any ideas where these factor comes from? I checked the model from #643. 
and forget what I said about a second camera: it was the default camera from blender which I saw in blender, but not in the model.

Issue #643, opened 2015-08-23
broken FBX camera position/orientation


promethe42:
Hello,

using the same Blender file I export a DAE and a FBX.
The DAE loads and renders as expected.
But the FBX camera position is completely broken.

I tried to load the FBX in other programs - such as 3DSMax or Blender itself - and the camera position is fine.
Where could I look in ASSIMP to diagnose the problem further?
Is this a known issue?

https://dl.dropboxusercontent.com/u/39247193/untitled.blend
https://dl.dropboxusercontent.com/u/39247193/untitled.dae
https://dl.dropboxusercontent.com/u/39247193/untitled.fbx

Thank you very much.

promethe42:
So I'm loading both the FBX and the DAE with the exact same (ASSIMP) code, and here is what I get for the camera matrices:

```
main.fbx
transform: mat4(
-100.000000, 0.000004, -0.000039, 0.000000, 
0.000004, 100.000000, 0.000000, 0.000000, 
0.000039, -0.000000, -100.000000, 0.000000, 
1054.672363, 198.692459, 13.696361, 1.000000)

main.dae
transform: mat4(
-0.000000, 1.000000, 0.000000, 0.000000, 
0.000000, 0.000000, 1.000000, 0.000000, 
1.000000, 0.000000, -0.000000, 0.000000, 
10.546721, -0.136964, 1.986925, 1.000000)
```

The last line show there is an obvious difference : the FBX matrix translation is 100x the DAE one.
So you'd think the scale of the FBX matrix would be 0.01, but instead it's 100. So in the end, the FBX matrix has a scale 100 * 100 = 10 000 times bigger than the DAE one.

Where could this come from?

Regards,

kimkulling:
I don't know. Do you have any example?

On Mon, Aug 24, 2015 at 2:01 PM, Jean-Marc Le Roux <notifications@github.com

vd3d:
Is there a fix ? or a way to solve it ?

vd3d:
I have post something on StackOverflow... seems nobody leave here ?

http://stackoverflow.com/questions/35536994/assimp-import-camera

kimkulling:
I am investigating this issue here: https://github.com/assimp/assimp/issues/807

kimkulling:
I guess this is caused by #775 .

kimkulling:
Fixed by  a97a4fb

PR #1131, opened 2017-01-09
FBX: Fix light and camera transformations and add some missing properties.

### LINKED ISSUE #851: Missing parameters for lights in FBX importing
I am importing fbx model files with lighting using assimp. Lights are arranged into the scene hierarchy and each light has transformation matrix. But direction vector is not present for the lights.

Sibras:
Fixes #851 #643 #807


code/FBXConverter.cpp
```diff
@@ -708,10 +719,13 @@ void Converter::ConvertCamera( const Model& model, const Camera& cam )
...
 
     out_camera->mAspect = cam.AspectWidth() / cam.AspectHeight();
-    out_camera->mPosition = cam.Position();
-    out_camera->mUp = cam.UpVector();
-    out_camera->mLookAt = cam.InterestPosition() - out_camera->mPosition;
+    //cameras are defined along positive x direction
+    out_camera->mPosition = aiVector3D(0.0f);
+    out_camera->mLookAt = aiVector3D(1.0f, 0.0f, 0.0f);
+    out_camera->mUp = aiVector3D(0.0f, 1.0f, 0.0f);
     out_camera->mHorizontalFOV = AI_DEG_TO_RAD( cam.FieldOfView() );
```
Sibras: All the necessary transform information is in the cameras node. So Fbx stores the cameras transformation in the scene graph as well as in the camera settings. This means you can either use one or the other but if you use both then the transform ends up getting applied twice resulting in incorrect values. Assimp already adds the required transformation info to the cameras node so the transform in the camera settings can be ignored.

I tested this with files exported from both blender and maya.

kimkulling: OK thanks for the xplanation.

kimkulling:
Merged!

michaelAdsh:
hey 
what is the diffrent  between
PhysCamera001.Target  and  PhysCamera001 and  Camera001
how can i get my view camera matrix?
thx alot

PR #1135, opened 2017-01-13
Added support for parameters Ni and Tf in OBJ/MTL file format.

hblasins:
These parameters specify the index of refraction and the transmission of a material as described in the [documentation](http://paulbourke.net/dataformats/mtl/).

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -94,7 +94,7 @@ SET( TEST_SRCS
   unit/utIssues.cpp
   unit/utJoinVertices.cpp
   unit/utLimitBoneWeights.cpp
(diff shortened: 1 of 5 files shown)


test/CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ SET( TEST_SRCS
   unit/utIssues.cpp
   unit/utJoinVertices.cpp
   unit/utLimitBoneWeights.cpp
-  unit/utLWSImportExport.cpp
+#  unit/utLWSImportExport.cpp
```
kimkulling: PLease update the latest master, this issue is solved.

PR #1137, opened 2017-01-14
Update Readme.md

simon-heinen:
I would suggest to add https://github.com/makc/assimp2json to the ports of Assimp, also to drive contribution & attention to it since it would be really valuable to have it working in JS.


kimkulling:
Great idea, thanks for the update. I will merge it immediately.

Issue #1084, opened 2016-11-27
COLLADA export texture mapping


agerrius:
When exporting an object to COLLADA and then importing it back with Assimp the texture mappings are corrupted. I tested it with a simple COLLADA objects and the command line tool. Open3DMod does show the texture correctly on the original model, but not on the one that Assimp wrote.

I found at that it is related to this line in the COLLADA file. If I remove the TEXCOORD input in the polylist segment below the texture mapping is correct.

```
<polylist count="6" material="defaultMaterial">
          <input offset="0" semantic="VERTEX" source="#meshId0-vertices" />
          <input offset="0" semantic="TEXCOORD" source="#meshId0-tex0" set="0" />
          <vcount>4 4 4 4 4 4 </vcount>
          <p>0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 </p>
</polylist>

```
This extra TEXCOORD input was introduced in the following commit: https://github.com/assimp/assimp/commit/ad55e769bfc87d684b15ac3804c4d70775b9a9a7

I don't know the COLLADA format well enough to know if the output COLLADA file is incorrect or if the COLLADA reader does not handle this situation well.

agerrius:
Test object (before and after included):
[dice.zip](https://github.com/assimp/assimp/files/614730/dice.zip)

Initial object in Open3DMod
![image2016-11-27 1500 44 404](https://cloud.githubusercontent.com/assets/23453644/20648872/9b0ffcb8-b4b2-11e6-83cd-c933fac5de24.jpg)

Exported object in Open3DMod
![image2016-11-27 1500 47 787](https://cloud.githubusercontent.com/assets/23453644/20648871/9b0f812a-b4b2-11e6-9c53-a6e68a073a69.jpg)

Kvalme:
The problem seems to be in Assimp importer. 
Got same issue but found that applications that do not use assimp for collada input reads file correctly.
In my case:
Import scene to assimp from dae made in blender - ok
Export scene to dae using assimp.
In blender I see correct uv map.
In assimp I see corrupted uv map.

agerrius:
Sorry for the delay, but the solution works fine for me as well. I just tested with a DLL build from the latest source and the issue with exporting and then importing COLLADA is solved. So this issue is fixed for me.

PR #1154, opened 2017-01-27
Fix for COLLADA export texture mapping #1084

Kvalme:
When exporting to COLLADA assimp creates Mesh entry with following parts:
`<vertices id="meshId0-vertices">
          <input semantic="POSITION" source="#meshId0-positions" />
          <input semantic="NORMAL" source="#meshId0-normals" />
          <input semantic="TEXCOORD" source="#meshId0-tex0"  />
</vertices>
<polylist count="12" material="defaultMaterial">
        <input offset="0" semantic="VERTEX" source="#meshId0-vertices" />
          <input offset="0" semantic="TEXCOORD" source="#meshId0-tex0" set="0" />
          <vcount>3 3 3 3 3 3 3 3 3 3 3 3 </vcount>
          <p>0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 0 2 19 3 5 20 6 8 21 9 11 22 12 14 23 15 17 </p>
</polylist>
`
Reading such file with assimp result in duplicated TEXCOORD channels in perVertex and perIndex channels resulting in duplicated texcoord data.

I can't understand from COLLADA spec does this allowed or not, so to fix this issue I propose to move all input semantics (except POSITION) into polylist. Seems that it's a common approach.


code/ColladaExporter.cpp
```diff
@@ -809,8 +809,8 @@ void ColladaExporter::WriteGeometry( size_t pIndex)
     const std::string idstr = GetMeshId( pIndex);
     const std::string idstrEscaped = XMLEscape(idstr);
 
-  if( mesh->mNumFaces == 0 || mesh->mNumVertices == 0 )
```
Kvalme: Fixed alignment since gcc6 provides warning for this line.

acgessler:
Looks good -- thanks!

(as with many changes, we can't really rely on the Collada spec to assess whether it is 'right' or 'wrong' -- in the end, the arbiter is some majority vote of the tools out there. Let's see)

PR #1160, opened 2017-01-31
OBJParser: set material index when changing current material

NolaDonato:
When a new material is parsed from the material library, the name of the current material is set but the old material index remains. This leaves an incorrect material index in the current mesh causing textures to not be associated with the right mesh. This patch updates the material index when changing the material name.

submitted by: Nola Donato nola.donato@samsung.com

### DIFF
--- a/code/ObjFileMtlImporter.cpp
+++ b/code/ObjFileMtlImporter.cpp
@@ -185,7 +185,7 @@ void ObjFileMtlImporter::load()
                     getFloatValue( m_pModel->m_pCurrentMaterial->alpha );
                     m_DataIt = skipLine<DataArrayIt>( m_DataIt, m_DataItEnd, m_uiLine );
                 }
-            }
+            }width * height * 4
             break;
 
         case 'N':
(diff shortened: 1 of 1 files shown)


kimkulling:
Thanks a lot, but there is a crash in the unittests.  I wil take a look.

code/ObjFileMtlImporter.cpp
```diff
@@ -302,6 +302,8 @@ void ObjFileMtlImporter::createMaterial()
         // New Material created
         m_pModel->m_pCurrentMaterial = new ObjFile::Material();
         m_pModel->m_pCurrentMaterial->MaterialName.Set( name );
+        if (m_pModel->m_pCurrentMesh)
+            m_pModel->m_pCurrentMesh->m_uiMaterialIndex = m_pModel->m_MaterialLib.size();
```
kimkulling: I guess you mean MaterialLib.size() -1, right?

kimkulling: Because this will break he unittest excecution.

code/ObjFileMtlImporter.cpp
```diff
@@ -185,7 +185,7 @@ void ObjFileMtlImporter::load()
                     getFloatValue( m_pModel->m_pCurrentMaterial->alpha );
                     m_DataIt = skipLine<DataArrayIt>( m_DataIt, m_DataItEnd, m_uiLine );
                 }
-            }
+            }width * height * 4
```
kimkulling: I do not get this? What the idea behind?

kimkulling:
I guess I will fix that. Thanks a lot for the contribution.

code/ObjFileMtlImporter.cpp
NolaDonato: If the materialLib indices are zero based the index of the one you are adding will be materialLib.size() before it is added. If I set the material index AFTER adding the new material, then the index of that material would be materialLib.size() -1. I think if you use materialLib.size() -1 and you have not added anything yet you will get a negative index.

PR #1161, opened 2017-01-31
OBJParser: Support spaces in texture file names

NolaDonato:
This patch allows embedded spaces in texture filenames found in MTL files. It only affects the OBJ parser.

Submitted by: Nola Donato nola.donato@samsung.com


kimkulling:
Ah, great. Thanks a lot. I will merge it!

PR #1180, opened 2017-02-13
Fix memory leak in Collada importer

JonAllee:
Affects Collada files with morph Controller Elements.


kimkulling:
Oh, nice finding. Thanks a lot for the pull request!

PR #1181, opened 2017-02-21
Issue-1169: fix [] operator of vector3.

kimkulling:
SHould fix compilation error for xcode.


turol:
The second commit is potentially an ABI change and so can't be merged without major version bump. It appears that this issue can't be fixed without doing that.

kimkulling:
It will cause ABI-imcompatibilies, right. people need to recompile against it, when we will release the next version.

kimkulling:
So I would say: assimp 4.0.0 is ahed :-) ...

turol:
You should make a separate issue for people to collect API/ABI changes. I think there was an issue about texture coordinate/color index confusion that also requires an API change. Get all of them at once so you don't end up doing 5.0 immediately after.

kimkulling:
I will do that!

PR #1189, opened 2017-02-27
Create the directory for the asset's extraction on Android

saveman71:
We encountered a problem while using assimp on Android Nougat (but it may also happen with older versions, we don't know), so it's tied to the AndroidJNIIOSystem library. On Android, assimp creates a copy of the models on the filesystem (internal storage / data folder of the app), instead of loading them from the Android asset manager each time. 

When we tried to load a model that is located in a folder (or subfolder), assimp was not able to open the output file for extraction from the Android asset manager because the folder (or subfolder) did not exist yet in the app's data folder.

For example, if the model was `models/foo/bar.obj`, assimp will try to create the file `/data/user/0/com.app.sample/files/models/foo/bar.obj` directly without creating the corresponding directory structure.  

This patch adds the folder creation to the extraction mechanism.


kimkulling:
Thanks a lot. Merged.

PR #1190, opened 2017-02-27
Fix buildsystem minizip

saveman71:
Alright, so this is a little big.

Firstly, I committed various files that were annoying git because they had the wrong line endings, as per the policy in the `.gitattributes` file (4495162).

Secondly, some encoding errors in the `CREDITS` file were fixed.

Thirdly, and this is the main point in the PR, we disabled `pkg-config` when building for Android, because `pkg-config` will return system libraries and not libraries available from the Android ndk, thus resulting in a build failure when minizip is installed on the host system.

I put the check in a macro that calls pkg_config and called this macro in the main CMakeLists instead of the old command. 

This is all in one PR because of the newline problem, it really messed up with my local git so I kinda had to make this one big PR. If you want this split, I could look into it.


kimkulling:
Also merged, thanks a lot for the contribution

PR #1191, opened 2017-02-28
Allow to enable importers in a whitelist manner

saveman71:
In the case when we want to build assimp with only a few importers (let's say less than 10), it would be shorter to specify the importers in a whitelist manner, instead of disabling every other importer but the 10 we want.

The feature works like that:

```sh
cmake \
  -DASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT=OFF \
  -DASSIMP_NO_EXPORT=ON \
  -DASSIMP_BUILD_3DS_IMPORTER=ON \
  -DASSIMP_BUILD_FBX_IMPORTER=ON \
  -DASSIMP_BUILD_OBJ_IMPORTER=ON \
  -DASSIMP_BUILD_COLLADA_IMPORTER=ON;
# that's all!
```

The new flag here is `ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT` which is documented in the `CMakeLists.txt` (by default it is TRUE, which mimics the current behavior).

The change in `code/ImporterRegistry.cpp` was because of a fun edge case when no importer was set, which resulted in missing include directives.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -187,10 +187,16 @@ IF ( ASSIMP_BUILD_NONFREE_C4D_IMPORTER )
   SOURCE_GROUP( C4D FILES ${C4D_SRCS})
 ENDIF ( ASSIMP_BUILD_NONFREE_C4D_IMPORTER )
 
+# if this variable is set to TRUE, the user can manually disable importers by setting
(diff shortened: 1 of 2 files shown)


kimkulling:
Good point and thanks a lot! Merged!

PR #1193, opened 2017-03-01
fix for FreeBSD

murashin:
On my FreeBSD/amd64 12.0-CURRENT machine, I need to modify only 1 line to compile.


kimkulling:
Niced finding. I will check if there are any other places to add this.

PR #1194, opened 2017-03-02
Import FindPkgMacros to main CMake Configuration

Sailsman63:
Commit 125d2ab95581f23908d broke the build process when `-DASSIMP_BUILD_ZLIB=ON`. The main CMake task now uses a macro that is only imported if the FindZlib task is run first. This commit will allow builds with an internally built zlib to progress.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,7 @@ endif()
 
 INCLUDE_DIRECTORIES( include )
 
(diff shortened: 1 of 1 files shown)


kimkulling:

saveman71:
Sorry ^^"

Sailsman63:
It happens. Apparently the travis build does not try to build with this option.

PR #1207, opened 2017-03-16
operators to add and multiply by floats

Celli119:


kimkulling:
Looks good, I will merge it!

PR #1208, opened 2017-03-17
Added parameters to IFC importer to control tessellation

terziman:
Those commits introduce 2 new parameters to the IFC importer for better control of model tessellation during import. In particular, those parameters allow users precise control over model size vs fidelity.

Moreover, those new parameters permitted to exhibit a bug in IFC importer that is now solved (call to front() method of empty vector without check).

Finally, we added a new method to Metadata structure to simplify appending new data to existing metadata structure.


kimkulling:
Thanks a lot for this great contribution, merged!

Issue #1000, opened 2016-09-19
Exporting Collada Animation with Bone ID and weight


winstorm:
Hello, 

Thank you for Assimp !

I want to export 3D model file with animation. 

I think, Collada Exporter has function for exporting Bone, but has not Bone ID and weight.

How to export Bone ID and weight to save Collada file?

Issue #961, opened 2016-08-03
Export file format with Bones and Animation


codetiger:
We are planning to add Assimp to our project (iOS & Android App). In our scenario, Export is as important as Import, as we are building a 3d animation tool for mobile platforms. I see that Collada export is not complete (Animation info is missing). Let me know if there is any other file format that is 100% complete on both Export and Import. 

Otherwise, I am willing to contribute to Assimp in adding Export support for Collada. Let me know if there is already someone working on it.

kimkulling:
The x-fileformat should be ready I guess.

Kim

Harishankar Narayanan <notifications@github.com> schrieb am Mi., 3. Aug.
2016, 12:26:

longht89:
Hi Harishankar and Kim,

I has same this issue.

Do we have any news about your project, Harishankar? Is it running correctly?

At the moment, i has used Maya to build 3D Model. Can i use it to export x-fileformat?

Thank you

codetiger:
Hi, Our project has a unique scenario, we needed a binary format to keep the size as small as possible and loading speed as optimal as possible. So we ended up writing our own format in binary. However, we are using this library to load different formats which are working perfect.

longht89:
Hi Harishankar and Kim,
I'm sorry, this is my bad. This library is perfect.
Thank you for your support.

kimkulling:
Thanks for the compliment, we are doing our really best.

Gondra B Huynh <notifications@github.com> schrieb am Mi., 24. Aug. 2016 um
13:20 Uhr:

Issue #644, opened 2015-08-25
Add support for PMD & PMX & VMD


tuxzz:
`.pmx` amd `.pmd` are 3d model format, `.vmd` is skeletal animation format.
These formats formats are used in MikuMikuDance and MikuMikuMoving, a software that popular in Vocaloid and Japanese anime fans.

MikuMikuMoving:
https://sites.google.com/site/mikumikumovingeng/

PMX & PMD editor:
http://vadim07.deviantart.com/art/PmxEditor-0222-English-520178628

MaddTheSane:
[This](https://gist.github.com/ulrikdamm/8274171) describes the PMX 2.0 file format.

undisputed-seraphim:
I might take a crack at this, but don't hold your breath.

aoowweenn:
[Update](https://github.com/aoowweenn/assimp/tree/umw_dev)
Now we can establish the armature hierarchy and assign bone weights to vertices.
While current DAE exporter is lack of "vertex_weights" tag, I can't verify the weights.

Next step I think is to jump to VMD importer and use Urho3D to verify the weights.

kimkulling:
What about preparing a pull request. We could mark it as experimental.

undisputed-seraphim:
Just pulled v4.0.1. Glad to see that someone worked on this!

I just have a teeny problem: Build fails spectacularly on Visual Studio, due to the Japanese comments in the source files. After stripping out the characters, it builds successfully.

So, what do the maintainers want to do? I don't mind replacing the Japanese comments with English translations if there's any interest, or just strip them out since most of the comments are just explanations of variable names for the Japanese audience.

kimkulling:
I guess a first step would be to temoce these comments.

Tan Li Boon <notifications@github.com> schrieb am So., 6. Aug. 2017, 17:34:

HTTdesu:
Thank you for adding pmx format parser. But I do find some problems in use.

1. File MMDPmxParser.cpp is not using Unicode, and this really caused me a lot of trouble. An interesting thing is that it can still pass the code check but cannot work correctly since line 521.
`stream->read((char*) &version, sizeof(float));`
The `std::istream` cannot read data into `pmx::Model::version` correctly so it will be 0 and then throw an exception.
Please save it in Unicode.

2. Unexpected problems happen in memory recovery process.
[...]

PR #1212, opened 2017-03-21
MikuMikuDance Format

aoowweenn:
From #644, try to import .pmx, .vmd, and .pmd file.

Bug fixed
- Exporter flag should be same with the importer's in the [qt viewer](https://github.com/assimp/assimp/pull/1212/commits/ba357e74d274e7847e31a41e5aba2a976d81d9fd)
- [Assbin loader](https://github.com/assimp/assimp/pull/1212/commits/845d206959993392a08ab756a78246cbae96c9c0) should handle mParent fields in node hierarchy.
  - I faced the problem to import .pmx file to Urho3d since they manually revised the assimp repo, so first convert .pmx to .assbin and import into Urho3d.

Status
- .pmx file: mesh, texture, skeleton
- Create local DevIL cmake module to avoid finding ILUT

TODO
- [#961, #1000] patch COLLADA exportor to verify bone weights
- import animation from .vmd file

Pending
- physics: spring bones, IK, soft bodies
- .pmd file


kimkulling:
Build is broken, could you fix it?

code/MMDVmdParser.h
```diff
@@ -0,0 +1,367 @@
...
+
+	/// IKの有効無効
+	class VmdIkEnable
+	{
+	public:
+		std::string ik_name;
+		bool enable;
+	};
+
+	/// IKフレーム
```
kimkulling: False encoding

kimkulling:
I looked over your pull request: good so far. But there are unreadable comments with a false encoding. Please only englich comments! Thanks!

aoowweenn:
The parser is almost not my work though, I may find some time to revise those Japanese comments...

By the way, morphing is possibly the next one that can be added to the PMX importer,
There are still some problems with the VMD importer since no IK, no full skeleton hierarchy, and position offset against PMD/PMX bind pose.
I've tried to merge a PMX bind pose and a VMD file to .ani format used in Urho3D.
The result is not good but it seems that I got it basically right except:

1. Duplicate bones merging
2. No 2 feet IK
3. Right frame interpolation

Besides, I think the DAE importer is strange,
UTF-8 is XML natively supported, but why filtering out non-ASCII characters?

acgessler:
What's the license status of MMDParser? It doesn't tell and you stated it's not your work. Is it BSD-3 clause compliant?

aoowweenn:
It's under CC0 1.0 so I think there should be no problem with BSD.
Also the author says,

> You can use this library freely. I dont't care about your usages.

[link](https://github.com/oguna/MMDFormats)

I had contributed non-Windows encoding part to his repo,
and in Assimp I used the internal unicode library instead for both Win/*nix.

kimkulling:
Sorry, I completly forgot about this pull-request. When all license info is there we can merge this guy I guess. Could you just merge the latest master into it? Thanks a lot for all the great work and sorry for my carelessness!

kimkulling:
Oh, we can merge this.

aoowweenn:
Thanks,
hope this contribution is useful for others.

PR #1213, opened 2017-03-22
use multimap instead of map where an unsorted_multimap was present

wise86-android:
why are you compiling with c++11 but you remove some c++11 things?

by the way the equivalent of unsorted_multimap is multimap not map..
tell me if I can do something to accelerate the merge of this pull request..


kimkulling:
We had some trouble with VS-2010, that was the reason.

kimkulling:
Merged!

PR #1220, opened 2017-04-01
Fix PDB file installation in RelWithDebInfo configuration

andre-schulz:
When installing a RelWithDebInfo build using Visual Studio, the PDB file passed to install() cannot be found because it is missing the library suffix.


kimkulling:
Thanks a lot, merged!

PR #1221, opened 2017-04-03
Fixed compiler error on clang 4.0 running on OSX

maartenheremans:
glTFExporter.cpp:585:35: error: ordered comparison between pointer and zero ('aiVector3D *'
      (aka 'aiVector3t<float> *') and 'int')
                if(comp_allow && (aim->mNormals > 0)) idx_srcdata_normal = b->byteLength;// Store index of normals array.


kimkulling:
Thanks for the contribution, merged!

PR #1226, opened 2017-04-08
Python wrapping improvements

gbittoun:
- Modifying setup.py and helper.py to be able to install the assimp python wrapping inside a virtual environment

- Adding a fallback on shader version 120 when version 130 could not be compiled in 3d_viewer.py

I wrote an email about these two issues https://sourceforge.net/p/assimp/mailman/message/35775932/ . Feel free to answer and/or validate my pull request :-)


kimkulling:
thanks a lot, merged!

PR #1231, opened 2017-04-11
fix line breakes in obj files

samitc:


kimkulling:
Thanks a lot for the fix! Merged!

PR #1234, opened 2017-04-18
Enable doxygen been properly used from cmake build and install.

heliocastro:
Current documentation lacks a proper directory handling and switch for
Unix like systems.
The option BUILD_DOCS are added as disable by default, even so the
Doxyfile file is generated for a manual build.

Option HTML_OUTPUT are made cached to be properly replaced, as usually
done by some Linux distributions

Microsoft CHM option is enabled if MSVC is detected.


kimkulling:

PR #1246, opened 2017-04-25
- Fix documentation install process. 

heliocastro:
It's nonsense second BUILD_DOCS check on doc/CMakeLists since is already checked on main CMakeLists.
At same time, Doxygen becomes required as no documentation can be built
without it.
Output directory now uses the user defined HTML_OUTPUT instead of
forced dir.
- Added include GNUInstallDirs on same UNIX check, avoiding duplication
of checks


kimkulling:
Looks good, thanks for fixing that! I should add a regression test for that as well I guess!

PR #1247, opened 2017-04-26
Adds texture type enum to pyassimp

rspencer01:
Keeping the same style of `postprocess.py`, this is a port of the aiTextureType enum in [`material.h`](https://github.com/assimp/assimp/blob/master/include/assimp/material.h).


kimkulling:
Thanks a lot, merged!

PR #1248, opened 2017-04-27
Fixed no member named 'atop' in global namespace issue for Android ND…

NcJie:
Fixed compilation error for Android NDK platform (gnu-libstdc++, 4.9) 

> error: no member named 'atof' in the global namespace
>           val = static_cast< float >( ::atof( begin ) );
>           val = static_cast< float >( ::atof( token.c_str() ) );


kimkulling:
Oh, thanks a lot for figuring this out. I will merge it!

PR #1249, opened 2017-04-27
- Apply mechanism to decide use for IrrXML external or internal

heliocastro:
Several distributions usually decide for shared external libraries
instead of an usual embedded, for security reasons, duplicatiion issues.
This change enable the possibility to set SYSTEM_IRRXML=ON for detect
and build against a system installed irrxml.
By default, the internal copy is compiled.

Changes on build:
- Added a FindIrrXML cmake module.
- Moved the source recipe for proper CMakeLists inside contrib directory
- Includes aren't path based anymore, using the provided INCLUDE_DIR
- Enable PIC for non x86 bsed *nix builds

Note: Current internal assimp irrXML is older than upstream irrlicht 1.8.4.
To enable usage of this version, code need to be patched.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,11 @@ OPTION ( ASSIMP_COVERALLS
    OFF
 )
 
+option ( SYSTEM_IRRXML
+    "Use system installed Irrlicht/IrrXML library."
+    OFF
+)
+
 OPTION ( BUILD_DOCS
    "Build documentation using Doxygen."
    OFF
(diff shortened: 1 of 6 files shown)


heliocastro:
Seen like that some of the cmake builders still use older version than 3.0.2, so fPIC can't be enabled through CMAKE_POSITION_INDEPENDENT_CODE.
Since the minimal default on assimp in cmake 2.8.0, i need to revert the request use regular PIC on c/cxx flags

PR #1252, opened 2017-04-29
Shouldn't the configured "config.h" file be going into the binary directory?

a-day-old-bagel:
I use assimp as a submodule in a cmake project that disables in-source changes. Configuring the file back into the source tree (as it does now) breaks that plan, and making this change does not seem to cause any problems.


kimkulling:
Hm, an interesting idea. Let's give it a try,

PR #1253, opened 2017-05-01
Fix static init ordering bug in OpenGEX importer

rdb:
This fixes a bug that causes a crash when querying the list of supported extensions.

The original code attempts to static-initialise a POD structure using a value obtained from an `std::string`, but C++ guarantees that the dynamic initialisation of `std::string` happens *after* `desc` has been initialised in this case, so it would always initialise in the wrong order.

There is absolutely no purpose in keeping `OpenGexExt` as a statically initialised `std::string` object, so I just took it out, which fixed the crashes we were experiencing on load.


kimkulling:
Mergeed, thanks for the contribution!

PR #1254, opened 2017-05-01
Improve performance of obj line break

samitc:


code/ObjFileParser.cpp
```diff
@@ -99,27 +99,24 @@ ObjFile::Model *ObjFileParser::GetModel() const {
...
-			*(copyPosition++) = ' ';
-			do
-			{
-				streamBuffer.getNextLine(buf);
-			} while (buf[0] == '\n');
-			curPosition = buf.cbegin();
-		}
-	} while (*curPosition != '\n');
-	std::copy(buf.cbegin(), curPosition, copyPosition);
+    static std::vector<char> tempBuf;
```
turol: Making this static will break very badly if you're importing things from multiple threads.

samitc: I fixed it by making the vector non static.

kimkulling:
OK, I will merge it and take a look as well.

Issue #1164, opened 2017-02-04
glTF animation export: accessor count mismatch


leerichard42:
When converting this test model: https://www.dropbox.com/sh/i71nlo1v63qre4v/AABAEtMwb_6nOpFC1bublffWa?dl=0
from FBX to glTF, there is a generated animation ("Take 001_17") which references accessors that have counts which do not align with the number of keyframes in the time accessor. 

In the included .glTF file, this can be seen in the accessors "Take 001_17_accessor_1" (scale) and "Take 001_17_accessor_2" (rotation) which only have a count of 1, while "Take 001_17_accessor" (time) and "Take 001_17_accessor_0" (translation) have 509 elements each. 

It seems like this problem only occurred with the "b_Root_$AssimpFbx$_Translation" node.

leerichard42:
Here is a simpler case with the same issue: https://www.dropbox.com/sh/7srtujfkhtx9k4k/AACn0Gf3wzwKKIOTbIXLhEp6a?dl=0

PR #1258, opened 2017-05-08
GLTF exporter: ensure animation accessors have same count

jamesgk:
The goal here is to fix https://github.com/assimp/assimp/issues/1164.

The strategy is to basically stretch the shorter arrays to match the length of the larger arrays. So if [1, 2, 3] needs to be stretched to length 8 it will become [1, 1, 1, 2, 2, 2, 3, 3]. Though this could blow up the size of the exported GLTF, it does seem to be a requirement for all the accessors to have the same length.
(According to https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#animations, "All parameters must have the same number of elements.")

### DIFF
--- a/code/glTFExporter.cpp
+++ b/code/glTFExporter.cpp
@@ -845,30 +845,45 @@ inline void ExtractAnimationData(Asset& mAsset, std::string& animId, Ref<Animati
     //    If yes, then reference the existing corresponding accessor.
     //    Otherwise, add to the buffer and create a new accessor.
 
+    size_t counts[3] = {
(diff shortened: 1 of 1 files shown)


jamesgk:
Perhaps a better way to pad the smaller accessors is to interpolate between the elements that are there based on their corresponding time values. Obviously that would be more work though, and this change's approach works fine for what I'd expect to be the most common case of different lengths: some accessors with length 1 and all others with the same length n.

kimkulling:
Thanks a lot, merged!

PR #1259, opened 2017-05-09
[One-Line-Change] Fix config.h not being found during cmake --build . --target install

Squareys:
Hi everybody!

Thank you for maintaining this amazing project 👍 During compilation I found a tiny little bug, here is my fix. A different way to fix it would be setting a different output of for the configure_file command in the root CMakeLists.txt (I'd say having the generated files separated from the hand-written ones is slightly cleaner, though).

Cheers!
Jonathan.


kimkulling:
Looks good,, thanks for figuring this out and the bugfix as well!

Squareys:
No problem, thanks for merging! :)

PR #1260, opened 2017-05-10
IRR and IRRMesh separated. Added ifndef ASSIMP_BUILD_NO_EXPORT to Ope…

VennGrayDog:
IRR and IRRMesh separated. Added ifndef ASSIMP_BUILD_NO_EXPORT to OpenDDLExport.cpp


VennGrayDog:
already done by someone else

PR #1263, opened 2017-05-11
GLTF exporter: convert animation time from ticks to seconds

jamesgk:
Since time values in GLTF are measured in seconds:
https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#samplerinput-white_check_mark
and time values in assimp are measured in ticks (I think, though I didn't see this explicitly in the documentation), conversion is necessary.


acgessler:
Yes, they're measured in ticks but many importers set ticks/second to 1.
Thanks!

PR #1264, opened 2017-05-11
Add support for reading texture coordinates from PLY meshes with properties named 'texture_u' and 'texture_v'

smokris:
Some PLY meshes (such as [JanMichael.zip](https://github.com/assimp/assimp/files/994204/JanMichael.zip) from [trn.io](http://trn.io/b/blEJaQKsLD/)) use `texture_u` and `texture_v` as their UV property names.  But OAI currently only detects properties named `u`/`v`, `s`/`t`, and `tx`/`ty`.  This patch adds `texture_u` and `texture_v`.


acgessler:
Thank you!

PR #1268, opened 2017-05-15
Add TokensForSearch in BlenderLoader

VennGrayDog:
Add TokensForSearch in BlenderLoader to allow CanRead return true for in-memory files.


acgessler:
Wouldn't we want to look for both BLENDER and blender? Or was the old search for BLENDER simply wrong?

VennGrayDog:
SearchFileHeaderForToken reads chars, converts them to lower case and only then it compares them with token. Old search for BLENDER was wrong.

kimkulling:
Merged!

PR #1274, opened 2017-05-16
Improved obj exporter to export materials to fileName.mtl file instead of fileName.obj.mtl

ab4d:
Usually obj files have material information stored in mtl files - for example file.obj and file.mtl. But the current version of assimp exports the material information into file.obj.mtl file.

This pull request improves this so that file.mtl is created instead of file.obj.mtl.


kimkulling:
Looks good, thanks for the fix.

PR #1275, opened 2017-05-16
Fixed build error in ObjExporter :: GetMaterialLibFileName

ab4d:
The previous pull request had an error because the filename string was const and could not be changed. Sorry for that. This one fix that.


kimkulling:
Thanks, merged.

PR #1255, opened 2017-05-02
Add missing CFLAGS -fPIC.

heliocastro:
This entry was missing due revert of CMAKE_POSITION_INDEPENDENT_CODE
usage. Reported by @Sailsman63

PR #1282, opened 2017-05-23
Fix error when custom CMAKE_C_FLAGS is specified

rdb:
This fixes an error introduced by @heliocastro in #1255.  The CMAKE_C_FLAGS string was not quoted, causing CMake to insert a semicolon instead, which would cause a compiler error when additional CMAKE_C_FLAGS are specified.


kimkulling:
Thanks for the fix, merged!

PR #1293, opened 2017-06-02
PLYImporter refactoring

arkeon7:
PLY :
- optimize memory and speed on ply importer / change parser to use a file stream
- manage texture path in ply import
- manage texture coords on faces in ply import
- correction on point cloud faces generation

IFC :
 - update poly2tri to avoid crash on some IFC files

Collada :
 - manage missing texture reference in collada import


kimkulling:
Great work, thanks a lot for the contribution!

arkeon7:
Thanks :)

PR #1298, opened 2017-06-06
Revert "Installing assimp library as well"

severin-lemaignan:
This reverts commit 33bd5cfcfb0f27794a333273b20b60a7a550d184.

Installing the assimp library from setup.py is *not* a good idea as
it will break every packaging effort.

Besides, the original commit relies on an hard-coded path to find the
library that may not exist.


kimkulling:

PR #1302, opened 2017-06-08
Fix IFC bug of partially imported models

terziman:
For some IFC models, multiple IFCSite elements are defined.

In the current version of Assimp, only one of those elements is loaded, and this element is chosen at random. Therefore, the loaded model would not only be incomplete, but will also differ from successive loading attempts (with same import options and flags).

The proposed fix simply loads all the IFCSite referenced by the root element, or all the IFCSite present in the file if none are referenced by the root element (same behavior as current version, except that now all existing elements are loaded instead of just one).


acgessler:
Nice! LGTM.

PR #1314, opened 2017-06-21
GLTF exporter: ensure buffer view byte offsets are correctly aligned

jamesgk:
According to the GLTF spec, these buffer view offsets must be a multiple of the accessor's attribute type:
https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#bufferview-and-accessor-byte-alignment

This sprung off of discussion at https://github.com/BabylonJS/Babylon.js/issues/2352.


kimkulling:
Merged

PR #1318, opened 2017-06-25
Fixes parameters to pyassimp.load

rspencer01:
Commit 704e57db4ee640262b8b20f4b687c15231d64354 changed the order of the parameters to `pyassimp.load`.  The new order groups `filename` and `file_type` which is preferable, but the samples and other calls to `load` still point to the old order.  This fixes that with named arguments where necessary.


kimkulling:
Looks good, merged. Thans for the contribution!

PR #1323, opened 2017-06-27
Fixed texture bug due simultaneously using 'usemtl' and 'usemap' attr…

vkovalev123:
…ibutes

Signed-off-by: Vitaly Kovalev <vkovalev@gkmsoft.ru>


kimkulling:
Good catch, thanks a lot for the contribution.

PR #1337, opened 2017-07-11
Added support for metadata in assbin format

terziman:
As the title says, assbin format was not handling metadata. It's not the case anymore.


kimkulling:
Looks good, thans a lot for the contribution. I will merge it.

PR #1338, opened 2017-07-11
X3D importer: Support for binary X3D files

pdaehne:
Hello,

I implemented support for binary encoded X3D files. Quite an exotic format, but I need it in my pipeline. Additionally, I fixed some issues with the path handling. Would be nice if you could merge these changes into the main repository. Standard XML encoded X3D files are of course still supported.

Thanks,
Patrick


kimkulling:
Looks good as far as I was able to understand your work. Thanks a lot!

pdaehne:
Thanks for merging!

PR #1339, opened 2017-07-11
new json lib: migrate code first shot.

kimkulling:


turol:
It looks like the new json library needs a newer compiler. It would be a good idea to do the compiler upgrade first as a separate PR instead of mixing it into this one.

PR #1345, opened 2017-07-14
Use unique node names when loading Collada files

jamesgk:
It looks like unique names are important for correct bone-to-node association:
http://assimp.sourceforge.net/lib_html/data.html#bones


kimkulling:
Looks good, I will merge it!

PR #1359, opened 2017-07-22
Elapsed seconds not shown correctly

sacereda:
I was getting this kind of message, instead of seconds when debugging:

Debug, T9224: END postprocess, dt= 11717743 s

Just changed that 'auto' with std::chrono::duration<double>, and it seems to be working:

 Debug, T9772: END   `postprocess`, dt= 1.2176 s

### DIFF
--- a/code/Profiler.h
+++ b/code/Profiler.h
@@ -84,7 +84,7 @@ public:
             return;
         }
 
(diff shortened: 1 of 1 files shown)


acgessler:
Thanks!

Issue #1301, opened 2017-06-07
Support glTF 2.0 Specification


dhritzkiv:
It'd be nice to support importing (and exporting) of glTF 2.0 files.

[Khronos just released](https://www.khronos.org/news/permalink/khronos-releases-gltf-2.0-specification) the finalized [glTF 2.0 Specification.](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0) For the most part, the syntax is changed so now properties are defined as arrays instead of objects/maps, with references defined as indexes instead of keys. The spec also defined PBR-based materials.

kimkulling:
You are right, this would be great!

jamesgk:
I started a branch for this:
https://github.com/jamesgk/assimp/tree/gltf2-export

One issue for now is that, rather than duplicating most of the glTF 1 exporter, I modified it to export glTF 2. So on this branch it's no longer possible to export glTF 1. I also have not implemented glTF 1 importing.

dhritzkiv:
Sweet!

I think it'd be important to still support glTF v1, for legacy reasons, especially on import. I wonder if it'd be easy to sniff the version and then choose the right Parser/Importer.

The default version for glTF export should be v2, but v1 export might be specified via the format option: `--format=gltf1`

realvictorprm:
@dhritzkiv is this in a working state?

dhritzkiv:
Hi @realvictorprm. I'm not working on this feature - I merely suggested it. See @jamesgk's [work on glTF 2 support](#issuecomment-311746189) so far. It likely has some work to go before being integrated into assimp

jamesgk:
Yes, this definitely needs more work. I was just going to push some
fixes to animation exporting but then Daniel's suggestions still need
to be implemented.

jamesgk:
An update: on a new branch I separated glTF1 and 2 into two (largely duplicated) exporters, so now at least the glTF2 exporter should be mergeable: https://github.com/jamesgk/assimp/tree/gltf2.

realvictorprm:
Thank you for your effort. I always appreciate open source work even if I'm not working mainly in the repo. If you need any help because you have no time don't hesitate to tell me.

jamesgk:
@realvictorprm thanks! I probably won't have time in the near future to work on glTF 2 importing. So if there's much demand for that, any help would be welcome.

jcowles:
FYI, glTF 2.0 seems to be the only future for glTF.

[...]

PR #1363, opened 2017-07-26
glTF2 export support

jamesgk:
This is part of https://github.com/assimp/assimp/issues/1301, but certainly not full support since it only includes exporting. I just copied and altered the glTF1 exporter, so there is a lot of duplicate code.


kimkulling:
Whow! What a great commit. Thank you very much for adding export of glFT2.0!

PR #1365, opened 2017-07-26
Feature/custom library loader

jespersmith:
For native libraries in our releases, we use a custom library loader that extracts libraries from the distributed jar file.

The jassimp port directly calls System.loadLibrary() in a static{} block, so we cannot override this. This pull request moves the moment the native library gets loaded to Jassimp.importFile() and provides a helper class to load the native library. The default implementation of this helper class call System.loadLibrary(), but users can set a custom implementation.


acgessler:
Looks good - thank you for contributing this!

PR #1366, opened 2017-07-29
Fix install for builds with MSVC compiler and NMake.

titorgalaxy:
The CMake variable "MSVC" and the related versioned variables indicate that the
Visual Studio compiler is used but when using that compiler and the "NMake
Makefiles JOM" build generator the files it tried to install weren't in the
expected Debug folder.


kimkulling:
Great, thanks a lot for the pull request. I will do a review and merge it, if it is ok.

kimkulling:

PR #1378, opened 2017-08-09
glTF version

dhritzkiv:
Outputs the gltf asset field's version string in the pattern `/[0-9]+.[0-9+]/`, instead of an int, per [the GLTF spec](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/asset.schema.json) and [validator](http://github.khronos.org/glTF-Validator)


kimkulling:
Looks good, merged Thanks for the contribution!

PR #1385, opened 2017-08-15
Fix compiling problems under VS2017.3.

gongminmin:
1. WordIterator's operator== and operator!= must take const WordIterator& as parameter type.
2. OpenMP doesn't work with new two phase lookups. Need to disable twoPhase.


kimkulling:
Merged, thanks for your contributions.

PR #1250, opened 2017-04-29
- Apply mechanism to decide use for IrrXML external or internal

heliocastro:
Several distributions usually decide for shared external libraries
instead of an usual embedded, for security reasons, duplicatiion issues.
This change enable the possibility to set SYSTEM_IRRXML=ON for detect
and build against a system installed irrxml.
By default, the internal copy is compiled.

Changes on build:
- Added a FindIrrXML cmake module.
- Moved the source recipe for proper CMakeLists inside contrib directory
- Includes aren't path based anymore, using the provided INCLUDE_DIR
- Compiler options are grouped in a single entry on main CMakeLists

Note: Current internal assimp irrXML is older than upstream irrlicht 1.8.4.
To enable usage of this version, code need to be patched.

PR #1398, opened 2017-08-20
Fix linking issue with irrXML

Eljay:
Since #1250, IrrXML is built as a separate static library which needs to be linked to, but it isn't currently copied to the build output folder. This patch installs the resultant library to the same folder as the main assimp library.


kimkulling:
Thanks a lot for the fix, merged!

simiula:
It is not fixed, do not create the FAT version of the IrrXML! Assimp cannot run on the simulator.

PR #1400, opened 2017-08-23
Collada importer: Add support for line strip primitives

kebby:
Like the summary says: The COLLADA parser could already read line strips but didn't import them. Added the few lines of code necessary to fix that. :)


kimkulling:
Looks good so far, I will merge it.

kimkulling:
Merged, thanks a lot for this conribution.

PR #1401, opened 2017-08-23
Detect if Anaconda and fixed 3d_viewer for Python 3

melMass:
#### WHY:
- Installing Assimp from conda-forge will install it in the "lib" of either the **main conda** or the **current conda env**. This check will ensure the library gets loaded from there.

#### WHY:
- In python 3 "/" divisions results to float, "//" to int.
- reduce is now deprecated


kimkulling:
Merged, thanks a lot. Haven't tested it until now. Hopefully I will be able to start it at the weekend.

Thanks a lot for the work!

Issue #1175, opened 2017-02-10
FBX imports always have animation duration of 0


domhofmann:
I'm trying to import two different FBX files with bones + animation. Bones seem to load okay but animations only seem to load the first frame and report a duration of 0. Both files import properly in other tools I've tried (FBX Review, Cinema 4D, Unity).

See the attached files:
[fbx_examples.zip](https://github.com/assimp/assimp/files/767982/fbx_examples.zip)

domhofmann:
Strangely, the animations do appear to import properly if I go back to Assimp 3.1.1.

kimkulling:
So I have introduced a bug. Fuxxx ...

Thanks for the report!

domhofmann:
It happens. :) I'll see if I can find anything as well.

prk223:
I'm having the same problem with a different FBX file I'm using.
Appears to have been caused by commit 08ba782bd5fd1d0392b70e675db99702ff349b7f. Specifically, you can just roll back FBXConverter.cpp to the version before to make the problem ago away (might also have to manually remove boost and update includes to get it to build with the rest of 3.3.1).
The FBX file that was added during that commit works for me. My guess is that my FBX does not have the LocalStart/LocalStop populated which is needed by that commit's changes. Anybody know if there's something I can run on FBX files to make them support those attributes?

rickomax:
Got a file with this issue too. Gonna take a look at the source to try fixing soon.

GUNDAM92:
4.0.0,4.0.1,3.3.1and3.1.1.Only 3.1.1 is true。

kimkulling:
Closed by  ce9c8a4 i

PR #1403, opened 2017-08-24
FBX importer: Use actual min/max of anim keys when start/stop time is missing

kebby:
During testing our product I had to find out that some FBX exporters (confirmed with Cinema4D R18) don't fill out the LocalStart/LocalStop properties of an animation, resulting in assimp only importing the key frames at time index zero and discarding all others.

Fix: When LocalStart and LocalStop are both zero, just accept all key frames and calculate offset and duration from the actual min/max time of the imported keys.

### DIFF
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -2368,8 +2368,13 @@ void Converter::ConvertAnimationStack( const AnimationStack& st )
 
     int64_t start_time = st.LocalStart();
     int64_t stop_time = st.LocalStop();
-    double start_timeF = CONVERT_FBX_TIME( start_time );
-    double stop_timeF = CONVERT_FBX_TIME( stop_time );
+    bool has_local_startstop = start_time != 0 || stop_time != 0;
+    if ( !has_local_startstop ) {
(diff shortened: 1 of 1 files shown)


kebby:
File for testing (no license attached, I could whip one of though if needed): https://www.dropbox.com/s/vwq7mcwclw4rlx2/Chained2Objects_RS.fbx?dl=1 - Two cubes, first one tilts, then the other tilts and scales. Before the fix the FBX importer would only import the first key frame.

kimkulling:
Travis is still broken. Just let me know if I should try to fix it on my linux devbox.

kebby:
I'd appreciate it. Seems certain build neither have a good stdint.h nor accept the (admittedly very MS) i64 suffix. I'll do one more try with the ll suffix, and if that doesn't work, we'd need some #ifdefs or something :/

kebby:
Btw, after some looking around, seems that this is a fix for https://github.com/assimp/assimp/issues/1175

kimkulling:
Great, will try this out!

kimkulling:
You are right, this pull requests closes #1175! Just tested both models and their animations are imported in the right way until now!

PR #1405, opened 2017-08-25
Rickomax master

kimkulling:
Just update it.


code/FBXBinaryTokenizer.cpp
```diff
@@ -445,9 +445,10 @@ void TokenizeBinary(TokenList& output_tokens, const char* input, unsigned int le
 	const uint8_t unknown_4 = ReadByte(input, cursor, input + length);
 	const uint8_t unknown_5 = ReadByte(input, cursor, input + length);
 	const uint32_t version = ReadWord(input, cursor, input + length);
+	const bool is64bits = version == 7500;
```
acgessler: Can we please apply the >= here? As suggested previously on some other PR. Unless there is a reason as for why this applies *only* to 7500.

kimkulling:
Sure!

PR #1408, opened 2017-08-29
Update version check in FBX reader to check for version >= 7500 inste…

acgessler:
…ad of exactly 7500.


acgessler:
@kimkulling

kimkulling:
Thanks, merged!

PR #1418, opened 2017-09-03
Debian fixes 4.0.1

umlaeute:
these are fixes from the Debian package:

- spelling fix should be self-explanatory
- pyassimp: in Debian we are patching passimp/setup.py to not include any data, as we are including the data by other means, as the setup.py installs them to the "wrong" place (from Debian's POV). this patch makes the Debian specific patch a little easier to handle.


kimkulling:
Merged, shall I generate a new version for debian?

umlaeute:
no, a new release is not needed. i'm happy if it appears in the next one :-)

Issue #1185, opened 2017-02-24
Mesh name is lost with aiProcess_PreTransformVertices flag


Avithohol:
Hi,

I use **assimp-3.1.1**, and tried to import the following fileformats to the assimp internal structures:
**3ds,obj,dae,ase**

The file is a simple cube, with diffuse texture, and a name.
Export was done from **3d Studio Max 2017**

Used the following flags during import:

**aiProcess_CalcTangentSpace      |
aiProcess_Triangulate           |
aiProcess_JoinIdenticalVertices |
aiProcess_PreTransformVertices**

Apart from the vertex info, I need the mesh name.

I looked for it in
**aiMesh::mName**
and also recursively traversed the scene graph and checked
**aiNode::mName**

No success, mName is always empty.

If I repeat the import without **aiProcess_PreTransformVertices**, I can get the mesh name from **aiMesh::mName.**

Is it a bug, or a feature with that flag?

Thanks for this awesome lib!

Avithohol:
Sample cube to reproduce the issue:
[box.zip](https://github.com/assimp/assimp/files/799959/box.zip)

kimkulling:
Hi,
at first: thanks for the report and the example. It was easy to reproduce the issue.

The obj-names for the meshed will be fixed on the current master. I haven't checked 3Ds, dae and ase until now. I guess I have to do it there as well.

Avithohol:
Thanks for looking into it. Can't wait to grab the next version.

And again many thanks for doing this!

lz199144:
Same error when I import fbx file, aiMesh::mName is null

kimkulling:
Fixed.

PR #1424, opened 2017-09-09
fix name lost when load with PreTransformVertices flag

samitc:
copy the name of old meshs and nodes to the new scene.
this PR fix issue #1185


kimkulling:

PR #1435, opened 2017-09-15
Last remaining build warning emitted from MSVC for x64 builds

jaredmulconry:
This is the last remaining warning emitted in builds on MSVC in the x64 configuration. This fully addressed issue #1065. As long as this change is good to be merged, issue #1065 can be closed.


turol:
Are you also going to enable warning-as-errors on appveyor to prevent introduction of new warnings?

jaredmulconry:
If @kimkulling would like to do that, I think it would be a good thing to do. That said, we could only apply it to builds with MSVC right now. I've no doubt that other compilers are generating their own warnings.

Another consideration is what warning level we want to aim for. MSVC generates hundreds of warnings at /W4. I've no doubt that -Wall -Wpedantic -Wextra on g++ and clang++ would have the same issue.
If we settle on a warning level we want to target, I'd be happy to clean those warnings up.

turol:
GCC and Clang definitely generate their own set of warnings. Enabling MSVC warnings would at least prevent introduction of new ones while we work on the others. Baby steps and all...

jaredmulconry:
I can't argue with that! If you're on gitter, perhaps we can begin a discussion there about what warning level we want to target. :)

turol:
I am but don't have it on this computer. You can start the discussion and I'll get to it this evening.

turol:
I started on the GCC part in #1436.

jaredmulconry:
I think there is still one last build warning on MSVC that is caused by the viewer. That might cause some build failures. I only just spotted it in the output for this pull request because I am not able to build the viewer right now.

kimkulling:
Ops, though I added this defines already. So thanks a lot for the fix! I will merge it

kimkulling:
Merged!

PR #1437, opened 2017-09-15
Added support for SIB models from Silo 2.5

rmitton:
Silo 2.5 upgraded the SIB format with a new version number (no other changes). This PR adds support for it.

Also, this fixes a bug that was introduced as part of the great Unicode rewrite, where material names would get truncated by one character.


kimkulling:
Ok, merged, thanks a lot for the contribution.

PR #1440, opened 2017-09-16
Travis cleanup

turol:
Reduce the number of Travis configurations and other cleanups

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,10 @@ branches:
   only:
     - master
 
+os:
+  - linux
+  - osx
+
 osx_image: xcode8.3
 
 env:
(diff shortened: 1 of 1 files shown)


turol:
Looks like that reduced the jobs from 10 to 7, but not the intended 5.

turol:
Well it sort of works, it reduces the number of builds and the build time. There are still two "extra" builds, one on Linux and one on osx which I can't figure out. Somehow it's adding them to the build matrix in addition to the explicit ones.

There's a problem with the osx builds. They're not actually building anything and don't seem to have done it for some time. I have no idea how osx build systems work so I can't fix it. `travis.sh` should probably have some commands for osx.

kimkulling:
I tried this some weeks ago as well. I didn't get it neither. Special the OSX-build is a big big big pain in the ass! 
Thanks a lot, I will merge it!

kimkulling:
Merged!

PR #1441, opened 2017-09-16
Disable Travis OS X build since it doesn't do anything currently

turol:
It doesn't actually build anything so it will always pass but the jobs are slow to start and delay the whole thing. Disable it for now. Also you should open an issue in case someone actually knows Travis/OSX and wants to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ branches:
 
 os:
   - linux
-  - osx
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


kimkulling:
There is no OSX-build ongoing?

turol:
Look at the Travis logs. It sets up the build environment but then doesn't actually do any *building*.

kimkulling:
You are rigt, crazy! I will merge it. And I will take a look into the source of the c++-json lib. There are using the OXS-build as well.

PR #1442, opened 2017-09-17
Fix glTF 2.0 multi-primitive support

jcowles:
Previously, only one primitive was supported, in fact memory was corrupted
when more than one primitive was found per glTF mesh.

In this change, each primitive is unrolled as a new Assimp Mesh, resulting
in multiple Assimp meshes per node when multiple primitives exist per
glTF mesh. This is required in the general case, since glTF primitives can
have different material bindings and primitive modes.


jcowles:
This addresses the crash & correctness issue raised in #1301

kimkulling:

PR #1444, opened 2017-09-18
Clean up some more GCC warnings

turol:
Mostly simple stuff. The only one which should change behavior is in PLY loader. The previous version could fail on malformed files where the header wasn't complete.


kimkulling:
OK, merged, I will fix the c2.

PR #1445, opened 2017-09-18
Fix glTF2 alphaMode storage and reading

dhritzkiv:
alphaMode is now converted from a std::string to an aiString and back to std::string, since aiString is easier to store and retrieve from aiMaterial properties than std::string

Fixes issues of alphaMode being written out as `\fOPAQUE\0\0\0\0\0\0…`


kimkulling:
Much cleaner code, will merge it!

kimkulling:

PR #1446, opened 2017-09-18
glTF2 primitives fixes

dhritzkiv:
This PR reverts a commit (a0d9750) where I incorrectly assumed that a glTF2's one mesh per node only ever meant one aiMesh per aiNode (when in reality, assimp data structure stores multiple primitives on a mesh as multiple aiMeshes per aiNode). This PR also adds code that takes multiple meshes on a node and merges their primitives into the first (and only) mesh on a node (as per glTF2 spec).

This fixes exporting models that defined multiple primitives per mesh, such as the Brainstem model (in the glTF2 samples repo), and the [random model of the car](https://github.com/assimp/assimp/issues/1301#issuecomment-330001082) that @jcowles found on Sketchfab. It also fixes missing geometry issues I had with many glTF 1 models being exported to glTF2.

---

With this PR, what happens in a glTF2 -> glTF2 export flow is this:

- glTF2 node has at most one mesh (per spec). A mesh has at least one primitive. To convert to assimp's data structure, each primitive is stored as an individual aiMesh on a aiNode (aiNode therefore has multiple aiMeshes)
- When converting from assimp's structure back to glTF, each aiNode's aiMesh is stored in a glTF's node (so multiple meshes per node)
- I then append a node's multiple meshes' primitives into the first mesh's primitives. I then delete the meshes that were merged from the node. This matches the input data.
- The glTF2 is written where each node has at most one mesh, with at least on primitive per mesh.

I couldn't find a cleaner way to achieve this merging behaviour in existing parts of the code, such as `ExportNodeHierarchy` and `ExportMeshes`. There might be a better way to achieve this, but it's beyond my ability and understanding of assimp's data structure. 

This PR also overwrites @jcowles' fix in glTF2Importer.cpp (#1442 - which essentially was a fix for a0d9750), with similar code, but code that's tailored to a node's now multiple meshes. 😞 Hope that's okay, @jcowles!

### DIFF
--- a/code/glTF2Asset.h
+++ b/code/glTF2Asset.h
@@ -781,7 +781,7 @@ namespace glTF2
     struct Node : public Object
     {
         std::vector< Ref<Node> > children;
-        Ref<Mesh> mesh;
+        std::vector< Ref<Mesh> > meshes;
 
         Nullable<mat4> matrix;
         Nullable<vec3> translation;
--- a/code/glTF2Asset.inl
+++ b/code/glTF2Asset.inl
@@ -934,10 +934,7 @@ inline void Node::Read(Value& obj, Asset& r)
     }
 
     if (Value* mesh = FindUInt(obj, "mesh")) {
-        //unsigned numMeshes = (unsigned)meshes->Size();
         unsigned numMeshes = 1;
-
-        //std::vector<unsigned int> meshList;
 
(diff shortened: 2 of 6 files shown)


kimkulling:
Looks good, I will merge it.

kimkulling:

PR #1447, opened 2017-09-18
[gltf1] Don’t ignore rgba(1,1,1,1) color properties

### LINKED ISSUE #1434: gltf loader is ignoring material properties that are all 1s
Is this really expected behavior (according to the gltf specification) ?
code/glTFImporter.cpp, lines 177-180: 

if (col.r != 1.f || col.g != 1.f || col.b != 1.f || col.a != 1.f) {
mat->AddProperty(&col, 1, pKey, type, idx);
}

Thx!

dhritzkiv:
Do not ignore rgba(1,1,1,1) material properties when importing glTF1. While a white diffuse color may be the default value for a default assimp material, `1,1,1,1` is a very explicit color value for ambient/specular/emissive color properties.

Closes #1434

### DIFF
--- a/code/glTFImporter.cpp
+++ b/code/glTFImporter.cpp
@@ -174,9 +174,7 @@ inline void SetMaterialColorProperty(std::vector<int>& embeddedTexIdxs, Asset& r
     else {
         aiColor4D col;
         CopyValue(prop.color, col);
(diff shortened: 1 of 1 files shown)


kimkulling:
Ok, merged! Thanks for the contribution!

Issue #1330, opened 2017-06-30
Too many warnings on both gcc and clang.


volcoma:
Hi, the current state of the library has too many warnings under gcc and clang. And that is with the default flags of the compilers. It floods the compile output

gongminmin:
Actually on msvc, too.

joelteply:
mostly the CMAKE_FORCE_COMPILER stuff on clang for me, yeah

kimkulling:
Ongoing...

jaredmulconry:
@gongminmin Under the default settings, MSVC in x86 and x64 generate very few warnings at HEAD. What is the commit id that you are compiling at?

kimkulling:
I guess I will close this one.

PR #1457, opened 2017-09-24
Partially address issue #1330

jaredmulconry:
These changes address some of the build warnings present when compiling with gcc and all remaining MSVC warnings under x86 and x64. This partly goes towards addressing issue #1330 .

One warnings came from use of struct packing. Packing likely was never having the desired result and should be investigated.


tools/assimp_view/stdafx.h
```diff
@@ -23,7 +23,6 @@
 #define _WIN32_IE 0x0600    // �ndern Sie dies in den geeigneten Wert f�r andere Versionen von IE.
 #endif
 
-#define WIN32_LEAN_AND_MEAN     // Selten verwendete Teile der Windows-Header nicht einbinden.
```
turol: Is this because of a "duplicate definition" warning? If so you should prefer


```
#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif  // WIN32_LEAN_AND_MEAN
```


It includes less stuff and speeds up compile times.

jaredmulconry: WIN32_LEAN_AND_MEAN is defined project-wide already. It appears in the root CMakeLists.txt.
I felt any explicit usage of it in code was redundant.

test/unit/utDefaultIOStream.cpp
```diff
@@ -44,10 +44,20 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 
 #if defined(__GNUC__) || defined(__clang__)
-#define MAKE_TEMP_FILE_NAME mktemp
+#define TMP_PATH "/tmp/"
+void MakeTmpFilePath(char* tmplate)
+{
+    auto err = mkstemp(tmplate);
```
turol: This doesn't do what you think it does. `mkstemp` doesn't change `tmplate`, instead it returns a file descriptor. Throwing away its return value at the end of the function is wrong.

turol: Maybe you should be using [std::tmpfile](http://en.cppreference.com/w/cpp/io/c/tmpfile) instead?

jaredmulconry: I tried to make std::tmpfile work initially, but the point of the test is to get the file's size, which can't be done without a path to the file.
It seemed like mkstemp was filling in the appropriate placeholders into tmplate when I was printing the result. I'll dig deeper tomorrow.

jaredmulconry: Having looked over some more documentation for mkstemp, I can see a better way of implementing that function. I'll make the necessary improvement tomorrow.

jaredmulconry: @turol I'm slightly confused how it ever worked, considering mkstemp opens the file for exclusive use. Maybe fopen is playing some trick.

kimkulling:
Ok, merged! Thanks a lot foe eleminating these warnings!

jaredmulconry:
I'll fix up the IO issues later today. Tests all seem to be passing though, so it's not more broken at least. :p
Discuss the other remaining warnings on gitter.

test/unit/utDefaultIOStream.cpp
turol: it was leaving `fpath` untouched so you `fopen`:ed a file called `"./rndfp.XXXXXX"`.

jaredmulconry: http://manpages.ubuntu.com/manpages/xenial/man3/mkstemp.3.html
This seems to suggest that the template is modified. This seems to be the intended behaviour on Linux, as far as I can work out.
I was wrong about mkstemp opening the file for exclusive use, the common behaviour is only to ensure that the calling process is what creates the file in the first place. That explains why opening it again afterwards wasn't failing immediately. I've improved this function regardless. It will be in my next pull request on this issue. @turol

PR #1461, opened 2017-09-28
Fix possible infinite loop when exporting to gltf2

jfaust:
Some of the postprocessing options generate node names that are crazy long... in the case of one model I was looking at, more than 256 characters. This was causing `glTF2::Asset::FindUniqueID()` to enter an infinite loop.

I can't provide the model unfortunately, but it has a lot of small meshes in it, and the processing flags I used were:
`-cts -jiv -gsn -sbpt -fid -guv -og -om -icl -rrm`


kimkulling:
Ok, merged. Thanks a lot for the contribution.

PR #1464, opened 2017-09-30
Obj: prepare test to reproduce crash on linux.

kimkulling:
Let's see if it crashes here.


test/unit/utObjImportExport.cpp
```diff
@@ -254,3 +254,18 @@ TEST_F( utObjImportExport, issue809_vertex_color_Test ) {
     EXPECT_EQ( aiReturn_SUCCESS, exporter.Export( scene, "obj", ASSIMP_TEST_MODELS_DIR "/OBJ/test.obj" ) );
 #endif // ASSIMP_BUILD_NO_EXPORT
 }
+
+TEST_F( utObjImportExport, issue1453_segfault ) {
+    static const std::string ObjModel =
+        "v  0.0  0.0  0.0"
```
turol: Missing line endings, this file should not be well formed.

PR #1466, opened 2017-09-30
Eliminated warnings when building the project with clang

jaredmulconry:
This is the last remaining thing to be done to address issue #1330 
Only 1 warning remains. This warnings is in the zlib contribution.
`/contrib/zlib/inflate.c:1507:61: warning: 
      shifting a negative signed value is undefined [-Wshift-negative-value]
    if (strm == Z_NULL || strm->state == Z_NULL) return -1L << 16;`
This is a serious issue. We should consider getting an updated copy of this contribution, if we can't fix it ourselves and it has this issue fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,9 +204,11 @@ ELSEIF(MSVC)
   # enable multi-core compilation with MSVC
   add_compile_options(/MP)
 ELSEIF ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" )
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fvisibility=hidden -fPIC -Wall -Wno-long-long -pedantic -std=c++11" )
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fvisibility=hidden -fPIC -Wall -Wno-long-long -std=c++11" )
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 ELSEIF( CMAKE_COMPILER_IS_MINGW )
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -204,9 +204,11 @@ ELSEIF(MSVC)
   # enable multi-core compilation with MSVC
   add_compile_options(/MP)
 ELSEIF ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" )
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fvisibility=hidden -fPIC -Wall -Wno-long-long -pedantic -std=c++11" )
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fvisibility=hidden -fPIC -Wall -Wno-long-long -std=c++11" )
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 ELSEIF( CMAKE_COMPILER_IS_MINGW )
-  SET( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden -Wall -Wno-long-long -pedantic -std=c++11" )
+  SET( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden -Wall -Wno-long-long -std=c++11" )
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```
turol: Not sure this is valid for mingw. Have you tested it?

jaredmulconry: @turol I installed MinGW just now to verify this. It does accept the -fPIC flag.
I've run into unrelated issues that prevent gtest from building with my install of MinGW, but the rest builds fine.

kimkulling:

PR #1469, opened 2017-10-02
Treat warnings as errors on Appveyor

turol:
Treat warnings as errors on Appveyor.I haven't actually tested this, let's see if it works...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,8 +219,12 @@ if (ASSIMP_COVERALLS)
 endif()
 
 if (ASSIMP_WERROR)
+  IF (MSVC)
+    add_compile_options(/WX)
+  ELSE()
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
     SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
+  ENDIF()
 endif()
 
 INCLUDE (FindPkgMacros)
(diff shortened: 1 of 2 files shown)


turol:
Looks like it isn't clean after all. Does someone have access to working Visual Studio build so they can fix that?

jaredmulconry:
I do, though not 2013 at present.
Lemme install the dev tools for it.

jaredmulconry:
Builds fine on 2017 toolset. Seems to be 2013-specific.

jaredmulconry:
@turol After reading through the build logs on AppVeyor, we need to disable warning C4351 for builds on VS2013. It's a noisy warning meant for people upgrading from VS2012.

turol:
Let's see if this does the trick. Once again, completely untested. Also we should probably add MSVC 2017 to the build matrix but that should be a separate PR.

jaredmulconry:
Agreed. Testing on VS 2013 is difficult now, because getting VS 2013 to test locally is a chore.

jaredmulconry:
Command line error D8021: invalid numeric argument '/Wd4351'
RIP

turol:
Nope, looks like wrong parameter name.

turol:
lowercase 'w', not uppercase 'W'...

turol:
Let's try that again

jaredmulconry:
They added the capital W for warnings levels but not for disabling them, because no.

jaredmulconry:
Working now :)

turol:
So far, let's see if the final build succeeds.

turol:
Yay, passed!

jaredmulconry:
Only issue is that it'll immediately break if we try to build the included zlib contrib :'(

jaredmulconry:
That's the only warning I've been able to locate that hasn't been resolved.

turol:
Well don't do that then :)

Or update the included zlib to latest and see if that fixes it.

kimkulling:
Looks good merged.

Issue #1470, opened 2017-10-04
Undefined behaviour in zlib contribution


jaredmulconry:
Building the zlib contribution with the project produces a warning:
`./contrib/zlib/inflate.c:1507:61: error: shifting a negative signed
      value is undefined [-Werror,-Wshift-negative-value]
    if (strm == Z_NULL || strm->state == Z_NULL) return -1L << 16;`

This issue was fixed in the next version of zlib as part of [this issue](https://github.com/madler/zlib/issues/111). We should consider updating zlib on our side to address this.

jaredmulconry:
I have investigated migration to the latest release. It has its own share of warnings that haven't made their way into a release yet. I will look at getting a copy of the project at head and see how that performs.

jaredmulconry:
1.2.11.1 from the develop branch (basically the latest development as of today) builds warning clean across MSVC 2017, gcc and clang. I would like to wait until the work on enabling ASAN is merged in before submitting a pull request for this work. If only, so I can test its cleanliness against ASAN first.

PR #1474, opened 2017-10-05
Update zlib contribution

jaredmulconry:
The previous version of zlib we included had some nasty behaviour within it. This is the latest version and has addressed these issues. No issues were evident from compiling and running tests under ASAN.
This addresses issue #1470 as well as various build warnings.


kimkulling:
Ok, merged, thanks!

PR #1471, opened 2017-10-04
WIP: Enable Travis clang build

turol:
Enable Travis clang build. Expecting this to fail in AddressSanitizer. Will add fixes to this as appropriate.

PR #1475, opened 2017-10-05
Fixed warning on clang 3.9 caused by static data member in class template

jaredmulconry:
Static data members in class templates are a terrible idea. clang 3.9 is the only compiler that managed to identify the obvious issue with this.
This approach is only slightly more code and should also cut down on the number of temporary std::string objects created when logging errors, warnings or general log information.
This ties into the work currently being done to add clang 3.9 to the test matrix.


turol:
Looks good to me.

turol:
This is blocking #1471, please merge ASAP.

kimkulling:
Looks ok, merged!

PR #1479, opened 2017-10-07
AppVeyor cleanup and Addition of VS 2017 and running Tests

jaredmulconry:
This will take some iterating to get right.
I've moved the AppVeyor config away from custom build scripts and towards a more simplified model, including caching of the larger build artifacts for reduced build times in the future.
Addition of MinGW will come in a later update, once I have confirmed that VS 2017 and all else are working correctly.


jaredmulconry:
The work done thus far is worth merging. We can leave MinGW testing for another PR. This can not be merged before #1484, as that PR has fixes that are required in order for this PR to test successfully.

jaredmulconry:
@kimkulling This is ready to be merged. I would like to test AppVeyor Caching first to try and improve the build time. Caching can not be tested in isolation, sadly.

jaredmulconry:
@kimkulling This should be good to merge as of commit dce2be9e09cd8fd14a763f64a345515fdea4800a.
Everything after that was inconsequential changes for testing out the AppVeyor cache, which managed to shave 10 minutes of the build time even despite some build configs doing a full rebuild due to not finding the cache.

jaredmulconry:
VS 2015 and 2017 seem to be noticably slower even with caching enabled. The issue seems to be that running the necessary CMake command is noticably slower on those images. 1:20 for the whole build on VS 2013, it can take up to 2 minutes just to run the cmake command on 2015 and 2017 images.
This may require further investigation, but Windows build and test (which wasn't being done previously) now runs in half the time.

jaredmulconry:
@kimkulling Merge me, plz :)

kimkulling:

PR #1484, opened 2017-10-08
Fixed File Size reported as 0 in tests that use temporary files

jaredmulconry:
This issue only appears on MSVC 2013. This is caused by a temporary file being created and written to, but is slow to be flushed to until it is reopened for reading. This is a blocker for #1479. Once this is merged, that work should be ready to merge.


kimkulling:

PR #1489, opened 2017-10-11
Feature/jassimp IO system

jespersmith:
To distribute our software easily, we load Java resources from the class loader. This pull request adds the ability to define a custom IOSystem in jassimp.

When importing a model, the C++ IOSystem.open() function will call to the Java code and return a java native ByteBuffer (a continuous memory block accessible from the C++ code). The IOStream acts directly on this memory block. 

Using a memory buffer instead of calling back to the Java code for seeking in the file seems the most straightforward solution, especially considering that no seeking is supported in resources loaded from the class loader. 

A implementation of the Java IO system for loading files from the class loader (classpath) is provided.(AiClassLoaderIOSystem)

To provide support for getting the system path separator and checking if a file exists without opening the file I've switched the Jassimp code to use the C++ assimp interface.

### DIFF
--- a/port/jassimp/jassimp-native/src/jassimp.cpp
+++ b/port/jassimp/jassimp-native/src/jassimp.cpp
@@ -471,7 +471,10 @@ public:
 	
 	    return cnt;
     };
-    size_t Write(const void* pvBuffer, size_t pSize, size_t pCount) {};
+    size_t Write(const void* pvBuffer, size_t pSize, size_t pCount) 
+    {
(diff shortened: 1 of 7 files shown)


kimkulling:
Approved and merged. Thanks  a lot for your contribution!

jespersmith:
Thank you for merging. I saw I left our default Apache 2.0 headers instead of the BSD license on the new classes, I'll make a new PR to rectify that in the coming days.

PR #1498, opened 2017-10-17
Add static analysis to Travis build

turol:


turol:
It now passes. There's a warning in MDLLoader.cpp which doesn't trigger a failure for some reason. It doesn't generate a long-form report either. I'll fix it later.

@kimkulling Please take a look at the OpenDDLParser changes. They should preserve the current behavior of the code but are not necessarily what was originally intended. If you think these are not the right fixes, fix them in OpenDDLParser repository and update it in master. I will then rebase this one.

PR #1499, opened 2017-10-17
Add Node metadata to the Jassmip Java API

dljsjr:
This PR exposes the metadata key-value store on Node objects to the Java API.

It is implemented as a Java built-in HashMap from String -> AiMetaDataEntry, an almost 1-to-1 recreation of the native metadata structure as a Java object. The payload is stored as type `Object` and can be cast accordingly by the user at the call site. There are static helper methods to assist in this.

One potential problem is Java's lack of explicit unsigned 64-bit integers. In the case of an AI_LONG metadata entry I just use a Java `Long` wrapper type. This is fine in Java 8 or higher; `uint_64t` can be cast to `jlong` at the native part of the JNI layer and then on the Java side there are static helper methods for doing comparisons/bit ops/division on Longs that treat them as unsigned (https://blogs.oracle.com/darcy/unsigned-integer-arithmetic-api-now-in-jdk-8). But I believe jassimp targets Java 6 to be compatible with as many releases of Android as possible so that may be something to discuss (and possibly lead to rejection of this PR, which would make sense).


kimkulling:

PR #1502, opened 2017-10-19
Fix output of glTF 1 version string

dhritzkiv:
Was writing out “\u0001” instead of “1.0” as the data types were incorrect


kimkulling:
Looks good, I will merge it.

kimkulling:

PR #1503, opened 2017-10-19
glTF2 Material Improvements

dhritzkiv:
TL;DR:
- Diffuse and specular colours and textures are persisted better; and
- glossiness/roughness can be derived from shininess and specularity and vice-versa. 

All of this allows better interoperability between traditional lighting material models and PBR material models.

---

This PR improves importing and exporting some material properties in glTF2. Changes:

- Diffuse Color:
    - Importing:
        - pbrMetallicRoughness' (pbrMR) baseColorFactor is stored as `$clr.diffuse`, as well as a `$mat.gltf.pbrMetallicRoughness.baseColorFactor`.
        - If pbrSpecularGlossiness (pbrSG) is present with a diffuseFactor, it will store and overwrite `$clr.diffuse`
    - Exporting:
        - pbrMR will read from `$mat.gltf.pbrMetallicRoughness.baseColorFactor` first, then fallback to `$clr.diffuse`
        - pbrSG (if present) will read from `$clr.diffuse`
- Diffuse Texture:
   - same behaviour as diffuse colour, but with diffuse textures
- Specular Color:
    - Importing:
        - If pbrSG is present with a specularFactor, it will store `$clr.specular`
    - Exporting:
        - If pbrSG is present, any `$clr.specular` value will be exported to specularFactor
- SpecularGlossiness Texture:
   - similar behaviour as with specular colour
   - Slight issue is that the texture used for specularGlossiness stores specular values in RGB, and glossiness in A channel. This means glossiness in the texture is not accounted for (when not going from pbrSG to pbrSG). One solution would be to store extract the channel groups into two separate texture files, but that seems to be out-of-scope of assimp's responsibilities (maybe).
- Roughness/Glossiness:
    - Importing:
        - the inverse of pbrMR's roughness value (and after being multiplied by 1000†) will be stored in the aiMaterial's shininess slot.
            - This is only a rough approximation of shininess-specularity. But it means that shiny models won't export as flat anymore.
        - pbrSG's glossiness value, times 1000, is stored in the aiMaterial's shininess slot.
    - Exporting:
        - When exporting pbrMR, the roughness value is first retrieved from `AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_ROUGHNESS_FACTOR`, and if that isn't present, the value is derived from the `AI_MATKEY_SHININESS` by dividing it by 1000, square-root-ing, and multiplying by any specular colour's luminance.
        - When exporting pbrSG, the glossiness value is first retrieved from `AI_MATKEY_GLTF_PBRSPECULARGLOSSINESS_GLOSSINESS_FACTOR`, and if that isn't present, the aiMaterial's shininess value (if any) is converted to glossiness by dividing it by 1000.
- alphaMode
    - Use `"BLEND"` over `"MASK"` as the value, for nicer looking results.

†1000 because most shininess exponential values seem to range from 0-1000 (OBJ/MTL spec, and three.js docs). There are other ranges: 0-128, 0-200, 0-500, 0-infinity. But 1000 seems to be most common. The code in this PR also takes the square root of the shininess value, so the difference between larger values (250-1000) aren't so extreme. The code also takes care to clamp the values to [0,1] when exporting to glTF.

### DIFF
--- a/code/glTF2Asset.h
+++ b/code/glTF2Asset.h
@@ -172,9 +172,7 @@ namespace glTF2
 	#define AI_MATKEY_GLTF_ALPHAMODE "$mat.gltf.alphaMode", 0, 0
 	#define AI_MATKEY_GLTF_ALPHACUTOFF "$mat.gltf.alphaCutoff", 0, 0
 	#define AI_MATKEY_GLTF_PBRSPECULARGLOSSINESS "$mat.gltf.pbrSpecularGlossiness", 0, 0
-	#define AI_MATKEY_GLTF_PBRSPECULARGLOSSINESS_SPECULAR_FACTOR "$clr.specular", 0, 1
 	#define AI_MATKEY_GLTF_PBRSPECULARGLOSSINESS_GLOSSINESS_FACTOR "$mat.gltf.pbrMetallicRoughness.glossinessFactor", 0, 0
-	#define AI_MATKEY_GLTF_PBRSPECULARGLOSSINESS_SPECULARGLOSSINESS_TEXTURE aiTextureType_UNKNOWN, 1
 
 	#define _AI_MATKEY_GLTF_TEXTURE_TEXCOORD_BASE "$tex.file.texCoord"
 	#define _AI_MATKEY_GLTF_MAPPINGNAME_BASE "$tex.mappingname"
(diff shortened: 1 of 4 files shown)


kimkulling:

PR #1508, opened 2017-10-21
CMake for MSVC now generates /MT builds instead of /MD as default filename suggests.

### LINKED ISSUE #1507: CMake project file on VS2015 uses /MD while lib name is assimp-vc140-mt.lib
Checking out the latest master and generating files for Visual Studio 2015 [Win64] on Win7x64 with CMake version 3.9.4. results in project files that name the library as assimp-vc140-mt[d].lib, but in Code Generation use /MD and /MDd instead of the assumed /MT and /MTd. 

There should be an option to change between /MT and /MD(Or is there one already?) or at least fix the discrepancy. A simple solution could be changing the default /MD in the configuration of CMake to /MT,

akaltar:
This should fix #1507.

The setting of the flags is done as recommended here:
https://cmake.org/Wiki/CMake_FAQ#How_can_I_build_my_MSVC_application_with_a_static_runtime.3F

This only affects the initial configuration and doesn't interfere with the user changing settings with a GUI.
An alternate is mentioned on the link, but that would override manual settings and introduce another variable that sets wether the bui


akaltar:
I ran the tests in Release mode, not noticing the problems.

After noticing the errors, I tried to debug the issue caused by this configuration change, but I don't know enough about testing to understand what is causing this problem.

kimkulling:
Appveyor nuild is broken.

akaltar:
Just to be clear, please note that the unit tests failed on my computer too in debug mode. And only after the change I made, although I am unsure about how would that cause such an issue. So if there is an issue it is not with the appveyor build but rather with the unit tests on windows in general.

Perhaps some memory allocation checks are failing because the different C runtime version.

Also I just realized that the /MD option is "Multi-threaded DLL" so the "mt" in the dll names may refer to multi-threaded in general, although I haven't seen that used anywhere else. I am unsure if "/MT" was the intention of the person who created the filename.

kimkulling:
I will take a look onto it.

akaltar <notifications@github.com> schrieb am Di., 24. Okt. 2017, 15:00:

kimkulling:
The issue seems to be caused by a not threadsafe runtime. I think we should avoid using /MD. Any other ideas regarding this?

akaltar:
Did you mean avoid using /MT? Originally it was using /MD which didn't cause problems. Introducing /MT caused them.  Also the runtime should be threadsafe AFAIK, as long as you use it correctly. Is it possible that the static runtime was linked multiple times? That does cause issues, but could potentially be resolved. Unfortunately, I'm not experienced with the MSVC runtime, and documentation is vague sometimes.

kimkulling:
You are right, I made a mistake here. Of course it is possible that the runtime was linked twice. The logger is implemented as a singleton so the observed crash could be caused by it!

NebuHiiEjamu:
Would prefer having an option for /MD for Release mode, all the same. Multiple DLLs holding the same opaque MSVC data as opposed to using a runtime DLL seems like unnecessary bloat to me.

kimkulling:
OK, I guess I understand. 

/MD -> dynamic multithreaded runtime will be linked
/MT -> static multithreaded runtime will be linked

If I get it right your fix will configure that the static runtime will be used every time.

akaltar:
Yes you are correct, but that leaves a few question unanswered: Usually most libraries are linked with /MD, I don't quite understand why was the library name postfixed with "mt" (Maybe meaning multithreaded in general?), but I assumed it meant "linked with /MT" and so I thought I would fix the linking by making sure it uses /MT, but my assumption may be wrong.

There is also a test case that fails when linking with /MT maybe due to some memory allocation differences, but I couldn't quite understand why.

AFAIK most people use /MD, but sometimes you need to use /MT, for example if you want to create an exe that has no external dependencies.

Perhaps the best solution would be adding a configuration option for setting this, but that would be intrusive in some ways. See the cmake article linked in my original comment.

I would gladly help resolving this issue, but unfortunately I don't know what should be done here.

kimkulling:
Will not fix the right thing, naming of assimp-dlls is wrong. Thanks a lot for figuring it out. I will close this pull request and generate a new one.

PR #1512, opened 2017-10-21
Raise exception when obj file contains invalid face indice

aavenel:
On OBJ importer, face indices are parsed with std::atoi function.
This function could return 0 on ill-formed input. Since 0 is not a valid value, I think we should raise an exception if this happen.

This should prevent some crashes happening when fuzzing OBJ importer.


kimkulling:
Thanks for the contribution, I will merge it.

PR #1516, opened 2017-10-26
[gltf2 export] Fix shininess to roughness conversion

dhritzkiv:
Was accidentally flipping to value (1 - x) twice, thus negating the effect. This reverses that, and makes the code more readable via comments.


kimkulling:
Looks good, merged. And thanks a lot for the contribution!

PR #1519, opened 2017-10-29
Add two unit tests for OBJ importer

aavenel:
As suggested by @turol in #1512 review, this PR adds an unit test to check that OBJ importer do not crash when importing file with invalid face indice.

It also adds an unit test for homogeneous coordinates as they where not previously tested.

### DIFF
--- a/test/unit/utObjImportExport.cpp
+++ b/test/unit/utObjImportExport.cpp
@@ -305,3 +305,38 @@ TEST_F(utObjImportExport, relative_indices_Test) {
     }
 
 }
+
+TEST_F(utObjImportExport, homogeneous_coordinates_Test) {
+    static const std::string ObjModel =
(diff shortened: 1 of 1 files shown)


turol:
AddressSanitizer test failed because of a memory leak. Looks like it doesn't clean up the Face properly in case of error.

kimkulling:
Just fixed the memory leaks in the unittests in a different pull requests, which is based on this pull-request. So I will close this one and merge the other one ( see https://github.com/assimp/assimp/pull/1538 ).

Thanks a lot for the tests, I just found two bugs by investigating it!

aavenel:
Thanks for your review :) Sorry I didn't have time to investigate the memory leak. I'm glad that it was helpful to find new bugs.

PR #1524, opened 2017-10-31
Invalid texture coordinates cleaning fix

dhritzkiv:
When the `aiProcess_FindInvalidData` post processing step is enabled, line 703 in glTF2 exporter crashes:

https://github.com/assimp/assimp/blob/938c02a3587766d0ebbedebba4cb162445cb397c/code/glTF2Exporter.cpp#L701-L705

I tracked it down to texture coordinates being removed in post processing, but `mNumUVComponents` for the corresponding sets not being set to `0`, causing an access error when flipping texture coordinates.

Is this fix correct?


kimkulling:
Looks right for me. When no uv-coordinates are defined this value needs to be 0. I will merge it. Thanks a lot for the fix.

PR #1525, opened 2017-10-31
[pyassimp] Search for libassimp.so in LD_LIBRARY_PATH if available.

daeyun:
Search for libassimp.so in LD_LIBRARY_PATH if available.


kimkulling:
Looks good, merged. And thanks a lot for the contribution.

Issue #1515, opened 2017-10-26
export .obj without material (.mtl) file


thomas-lemaire:
Hello the assimp's team,

I would like to manipulate a mesh without any material
- since the data aiMesh structure forces to define a material I define an empty one in the aiScene
- and then when I write this mesh in obj I get a .mtl file

Would it be possible
1. either to be able to define no material for a mesh, and then the exporters which support material would respect this definition (no material defined, no exported material)
2. or to have a new *nomtlobj* exporter to prevent from exporting the .mtl file and associated definition in the .obj file

thomas-lemaire:
similar to #1096 

I am planning to implement proposed solution 2. which is much less invasive and I will do a PR...

kimkulling:
Ok, we need to get the pull request running to get this feature request working.

thomas-lemaire:
This is PR #1518 
Indeed there is still "continuous-integration/travis-ci — Waiting for status to be reported" - Is there something I can do ?

thomas-lemaire:
Thanks for merging @kimkulling !

Issue #1096, opened 2016-12-07
OBJ always exports material


johnmaf:
When converting to OBJ, a default material is always used, even when the original file had no such material. Since the OBJ format doesn't require materials, this is arguably incorrect behaviour.

Unfortunately changing this could cause problems for anyone who's currently assuming that OBJ export always produces a `.mtl` file, and could confuse people who are used to seeing the default material.

Is there any interest in removing this behaviour?

kimkulling:
We could add a config switch for the importer / exporter, if anyone is interested?

thomas-lemaire:
I added this issue #1515 and just found this one.

the .mtl file is explicitly referenced in the .obj file, so people should not expect to have always a .mtl, only if it appears in the .obj

PR #1528, opened 2017-11-02
Obj exporter nomtl

thomas-lemaire:
This PR adds a obj exporter which does not export the material file, its name is *objnomtl*
This PR addresses issues #1096 and #1515


kimkulling:
I need to reopen this pull request, because the travis trigger was missing. Stay tuned.

PR #1529, opened 2017-11-02
fix trivial warnings

thomas-lemaire:
mainly unused parameter and unused function
some parameters are indeed used in a debug built, I used the
(void)(param) trick
warnings reported by clang 4


kimkulling:
Doess anyone know how I can retrigger travis for this build again?

PR #1534, opened 2017-11-03
[PLY] fix for TextureFile with number in file name

### LINKED ISSUE #1324: PLY with TextureFile reference that starts with a number breaks parsing
This works:

`$ assimp export laurana500_with-string.ply output.obj`
 
where the ply contains this line:

```ply
comment TextureFile laurana500.png
```

Whereas this does not:

`$ assimp export laurana500_with-number.ply output2.obj`

where the ply contains this line:
```ply
comment TextureFile 5.png
```

The parser begins parsing the `5` as another type of line. I believe this behaviour was introduced in #1293

[I've attached some test files.](https://github.com/assimp/assimp/files/1107057/laurana.zip) (The actual texture files are unimportant, but are included anyway)

dhritzkiv:
In cases where the TextureFile name would start with an integer, `strtoul10` would parse that number and incorrectly set `numOccur` to that number. This caused PLY parsing to fail for vertex positions.

For example, a file name like `512.png` would set `numOccur` to `512`, causing issues when continuing to parse. `foo.png` would  keep `numOccur` at `0`. This PR fixes the issues by always keeping `numOccur` at `0` for TextureFile, by returning early.

Since TextureFile is a single line, and does not have any follow-up lines, it’s okay to return early.

Fixes #1324

### DIFF
--- a/code/PlyParser.cpp
+++ b/code/PlyParser.cpp
@@ -381,6 +381,11 @@ bool PLY::Element::ParseElement(IOStreamBuffer<char> &streamBuffer, std::vector<
   {
     char* endPos = &buffer[0] + (strlen(&buffer[0]) - 1);
     pOut->szName = std::string(&buffer[0], endPos);
+
+    // go to the next line
+    PLY::DOM::SkipSpacesAndLineEnd(buffer);
+
+    return true;
   }
 
   //parse the number of occurrences of this element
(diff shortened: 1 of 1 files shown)


kimkulling:
Ok, thanks a lot for the fix. merged!

PR #1551, opened 2017-11-07
Fix MD2 frames containing garbage

larsjsol:
This PL fixes an issue where animation frames after the first one contains garbage. It does this by using the frame size from the MD2 header when calculating the current frame pointer instead of relying in the size of the Frame struct.

It is quite possible that this issue is only present when compiling for 64 bits.


kimkulling:

PR #1557, opened 2017-11-09
Fix export of deleted meshes; Add LazyDict::Remove method

dhritzkiv:
When I was merging a node’s multiple meshes into that node’s first mesh’s primitives, I was deleting the merged meshes from the node.

However, I wasn’t deleting the merged meshes from the mAsset->meshes Dict, causing the gltf2 export to contain extra unreferenced meshes and duplicate primitives.

This would cause an output like this:

```json
"meshes": [
    {
         "primitives": [
             {
                 "mode": 4,
                 "material": 1,
                 "indices": 4,
                 "attributes": {
                     "POSITION": 1,
                     "NORMAL": 2,
                     "TEXCOORD_0": 3
                 }
             },
             {
                 "mode": 4,
                 "material": 1,
                 "indices": 8,
                 "attributes": {
                     "POSITION": 5,
                     "NORMAL": 6,
                     "TEXCOORD_0": 7
                 }
             }
        ]
    },
    {
        "primitives": [
            {
                 "mode": 4,
                 "material": 1,
                 "indices": 8,
                 "attributes": {
                     "POSITION": 5,
                     "NORMAL": 6,
                     "TEXCOORD_0": 7
                 }
             }
        ]
    }
]
```

where the second mesh's primitive is repeated twice. The fix prevents the second mesh from being output.

---

This new code adds a new method to LazyDict, `Remove`, which removes the object from it, and updates indexes of subsequent meshes in the mObjs vector property. However, the method does not take care of updating the indexes in `Ref`s of `Mesh`es, so that has to be done separately.


kimkulling:

PR #1585, opened 2017-11-20
[glTF2] Moved byteStride from accessor to bufferView

pdaehne:
Bugfix: Moved byteStride member from accessor object to bufferView object. In glTF1, byteStride was part of accessor, but in glTF2, it is part of bufferView. When creating the glTF2 loader/writer by copying the glTF1 loader/writer, someone forgot to move byteStride to the right place.


kimkulling:
Merged, thanks a lot for the contribution,

pdaehne:
Thanks again for merging!

PR #1589, opened 2017-11-24
Fix typo on gltf2 camera parameters

aavenel:
Orthographic cameras are not yet supported by assimp, but this fixes the parsing in case we support them in the future.


kimkulling:

PR #1590, opened 2017-11-24
Copy assimp dll to unit folder on windows

aavenel:
This commit adds the assimp dll in the unit test folder when building on windows, like for assimp_cmd tool.


kimkulling:

PR #1591, opened 2017-11-25
Fix all warnings on MSVC14

aavenel:
These commits should fix all warnings on MSVC14 :)


kimkulling:

PR #1594, opened 2017-11-27
[glTF2] Fix transform matrices multiplication order

### LINKED ISSUE #1568: glTF2 import: Transforms are incorrect (patch attached)
Transforms are incorrect when importing a glTF 2.0 file with a decomposed node transform (translation,  rotation and scale). This is happening because the multiplication order is not to spec, from the glTF 2.0 spec (https://github.com/KhronosGroup/glTF/tree/master/specification/2.0):

"TRS properties are converted to matrices and postmultiplied in the T * R * S order to compose the transformation matrix; first the scale is applied to the vertices, then the rotation, and then the translation."

You can observe the issue when trying to import "Project Polly" from "https://github.com/KhronosGroup/glTF-Blender-Exporter/tree/master/polly". The card box, the name plaque and the styrofoam piece will be in the wrong place.

 I have attached  a patch that fixes the transform calculation.

[gltf2_transform_fix.patch.zip](https://github.com/assimp/assimp/files/1472887/gltf2_transform_fix.patch.zip)

dhritzkiv:
Per spec. This fixes the output of the [Polly model](https://github.com/KhronosGroup/glTF-Blender-Exporter/tree/master/polly)

Closes #1568

This work is @awefers' (via a .patch) and I have attributed it to them as such.

### DIFF
--- a/code/glTF2Importer.cpp
+++ b/code/glTF2Importer.cpp
@@ -515,7 +515,13 @@ aiNode* ImportNode(aiScene* pScene, glTF2::Asset& r, std::vector<unsigned int>& 
             CopyValue(node.translation.value, trans);
             aiMatrix4x4 t;
             aiMatrix4x4::Translation(trans, t);
-            matrix = t * matrix;
+            matrix = matrix * t;
+        }
+
+        if (node.rotation.isPresent) {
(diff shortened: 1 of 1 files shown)


kimkulling:

PR #1595, opened 2017-11-27
[glTF2] Preserve node names when importing

### LINKED ISSUE #1522: Node names are not preserved when importing glTF 2.0 (patch attached)
Node names are not preserved when importing a glTF 2.0 file. Instead new names are generated ("nodes_0", "nodes_1", "nodes_2",etc), shouldn't the original names be preserved like with the other importers?

dhritzkiv:
Closes #1522


kimkulling:
Approved and merged. Thanks a lot for the contribution.

PR #1602, opened 2017-11-29
Restore import of multi mesh binary STLs

dhritzkiv:
Regression introduced in: 9a9f18bbed8938323548335d9225e480a0ea8ca6

This restores the behaviour for binary STLs with multiple bodies.

The code could be improved (to reuse code found at the end of LoadASCIIFile) but for now this is a quick fix

---

Here's a test file: 
[teapot.stl.zip](https://github.com/assimp/assimp/files/1514512/teapot.stl.zip)


kimkulling:
Merged, thanks a lot for the bugfix.

PR #1626, opened 2017-12-09
typo + remove redundant code

elect86:
magic word checked few lines before [here](https://github.com/assimp/assimp/blob/master/code/FBXImporter.cpp#L168)

Also, what does it mean [this](https://github.com/assimp/assimp/blob/master/code/FBXBinaryTokenizer.cpp#L265)?

> // note: do not write cursor += ReadWord(...cursor) as this would be UB


kimkulling:

Issue #1612, opened 2017-12-04
Asset importing failes when special characters are part of the uri


kimkulling:
Try to import an asset in a folder with chinese symbols in it like:

C:\Users\Administrator\AppData\LocalLow\xxx\xxxx\423155567\Assets\海涛\monster.FBX

This will fail. We need file operations to be able to deal with special symbols as well.

loebl:
As far as I know:

On Windows this means either using `char*` with a language-specific locale OR using `wchar_t*` and using UTF16 (and associated `wfstream` classes) independent (more or less) of the selected locale. You might need to extent the interface at that point. For my project I wrote a custom IOSystem class, which translates UTF8 `char*` strings to `wchar_t*` strings (using boost::locale) and used the windows specific `wfopen` function to open files.

As boost isn't part of assimp anymore this is probably not what you want. But utf8cpp also offers a 8 to 16 bit conversion, which should do the same.

On Linux `char*` can be used, as it maps to UTF8, if the locale is set correctly, so UTF-8 file paths work transparently with the existing C-library functions. (That means I do not think a change is necessary; have not tested it)

If you are interested I can modify the default IOSystem and open a pull request with the windows specific changes. The user then has to ensure the `char*` dropped into assimp is UTF8-encoded. Also this problem seems to be related to issue #194

mlavik1:
I'm still having this issue, with a filepath in UTF8 ("D:\Models\UnicodeIssues\テスト.obj").

@kimkulling  in your commit you added `IsTextUnicode` to check if it's Unicode, but in my case it returns 0, even though the filename is UTF8. The [documentation for IsTextUnicode](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-istextunicode) says that it "is not foolproof".
My questions is: Is it necessary to do that check at all? In UTF8 all the ANSI characters are encoded the same as in ANSI, so handling it as unicode (using `MultiByteToWideChar` and `_wstat64`) should still work when it's not Unicode, right?

By the way, I added a breakpoint and checked the encodings to verify that the filename string is actually UTF8, and it definitely is (テ was represented by e38386 in the string).

rmccampbell:
[...]

PR #1629, opened 2017-12-11
added path conversion code for unicode on windows

loebl:
As proposed in #1612, Here is the modified default IOSystem.
Points to note:
- I only tested it locally with the path `C:\projects\examples\_3ds\海涛\🥓.3ds`. In my application this worked fine, if the path was given as UTF8-String
- Function `MakeAbsolutePath` is not tested by me, as it only gets used during batch imports, which i don't use. Although a string might have several valid UTF8-representations (using pre-composed or combining chars) they all get normalized to pre-composed in `MakeAbsolutePath` (so the test in `ComparePaths` should not fail)
- Removed empty ctors to let the compiler create a default one.
- Used `fileno` in `DefaultIOStream` to not have to convert the filename there. If `fileno` is also used for the other platforms, the file name could be dropped from IOStream completely.
- I used the native win32 function instead of utf8cpp, as that would require converting all strings to C++ classes first (it is just more lines, but should work equally well)


loebl:
hm, can't reproduce the failing tests from appveyor on my machine (Win 10 x64, VS 2015 x86 & x64; tests run fine, except the check for assimp revision number)

Is there some way to get more details from the appveyor-build?

PR #1633, opened 2017-12-11
Fix build without 3MF importer

Teybeo:


kimkulling:

PR #1635, opened 2017-12-12
Use unique_ptr in B3DImporter

turol:
Before it crashed if reusing one instance of Importer.

aiNodes should also be stored with unique_ptr but their ownership is complicated. It probably needs a more significant refactor.


kimkulling:

PR #1637, opened 2017-12-13
Clang 5.0 fixes

turol:
Travis updated clang to 5.0 which broke the build. This fixes some warnings and disables static analysis for now to get the builds back on track.

@kimkulling The changes to aiUVTransform and aiVector2 change the ABI so the next release must be 5.0. There's still a problem in [`MDLLoader.cpp`](https://github.com/assimp/assimp/blob/master/code/MDLLoader.cpp#L419), it does really bad things with pointer casting. Any ideas how to fix it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,10 @@ env:
 
 matrix:
   include:
-    - os: linux
-      compiler: clang
-      env: ANALYZE=ON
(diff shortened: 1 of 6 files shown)


kkulling:
Ough! This is really really bad in the MDLLoader. Getting up to 5.0 is not critical IMHO.

kkulling:
I will take a look

PR #1640, opened 2017-12-14
Added support for generating glb2 (binary glTF 2)

Breush:
- Following [specifications](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#glb-file-format-specification) for the binary format ;
- Verified using Khronos Group's glTF validator : http://github.khronos.org/glTF-Validator/


kimkulling:

PR #1643, opened 2017-12-15
Fix incorrect NO_GLTF_IMPORTER define name in glTFExporter.h

mesilliac:
Pretty sure this should be ASSIMP_BUILD_NO_GLTF_EXPORTER


kimkulling:
Merged, Thanks for the contribution.

PR #1644, opened 2017-12-17
Fix non-ascii encoding in comments in FBXMaterial.cpp.

mesilliac:
It looks like it was just saved with a wrong encoding,
as these 0xB4 characters were in place of "'".

Also converted tabs to spaces to match indent in the rest of the file.


kimkulling:
Thanks, just merged it.

PR #1646, opened 2017-12-18
FBX Embedding Fix

rickomax:
FBX files may use a texture reference from an embedded texture that hasn't been loaded yet. This patch fixes this issue, storing all texture filenames, that can be acessed later via "scene::GetEmbeddedTexture", when all textures are already loaded.
Some warnings have been added to other file formats that uses embedded data.


rickomax:
@kimkulling

kimkulling:

PR #1657, opened 2017-12-25
Enable clang static analysis using clang 3.9

turol:
Since 5.0 has some false positives. This way we get at least some static analysis. The only question is whether travis images kept the old clang in addition to the new one.


turol:
Looks like they didn't. Scratch that idea.

PR #1661, opened 2017-12-27
Qt viewer : Improve exports

JLouis-B:
- Add a dialog to select the exporter
- Export based on exporter id instead of extensions (fix issues like for Collada)
- Message box if there is an error


kimkulling:

PR #1666, opened 2017-12-29
Fix some crashes when parsing ill-formed OBJ file

aavenel:


test/unit/utObjImportExport.cpp
```diff
@@ -330,6 +330,18 @@ TEST_F(utObjImportExport, homogeneous_coordinates_Test) {
...
+TEST_F(utObjImportExport, homogeneous_coordinates_divide_by_zero_Test) {
+  static const std::string ObjModel =
+    "v -0.500000 0.000000 0.400000 0.\n"
+    "v -0.500000 0.000000 -0.800000 1.00000\n"
+    "v 0.500000 1.000000 -0.800000 0.5000\n"
+    "f 1 2 3\nB";
+
+  Assimp::Importer myimporter;
+  const aiScene *scene = myimporter.ReadFileFromMemory(ObjModel.c_str(), ObjModel.size(), aiProcess_ValidateDataStructure);
+  EXPECT_EQ(nullptr, scene);
```
turol: I think there's a gtest method which can check that we get the exception we expect. Here we definitely want to use that if possible.

code/ObjFileParser.cpp
```diff
@@ -318,6 +318,8 @@ void ObjFileParser::getVector( std::vector<aiVector3D> &point3d_array ) {
         copyNextWord( m_buffer, Buffersize );
         z = ( ai_real ) fast_atof( m_buffer );
     } else {
+        delete m_pModel;
```
turol: I preser unique_ptr, it gets automatically cleaned up in case of any exception

test/unit/utObjImportExport.cpp
aavenel: Thanks for the review !

I tried to check this using *EXPECT_THROW* but unfortunately, we catch all exceptions in *BaseImporter::ReadFile*

Perhaps we should change this ?

code/ObjFileImporter.cpp
```diff
@@ -414,9 +414,13 @@ void ObjFileImporter::createVertexArray(const ObjFile::Model* pModel,
     // Copy vertices of this mesh instance
     pMesh->mNumVertices = numIndices;
     if (pMesh->mNumVertices == 0) {
+        delete pMesh;
+        pMesh = nullptr;
         throw DeadlyImportError( "OBJ: no vertices" );
-    } else if (pMesh->mNumVertices > AI_MAX_ALLOC(aiVector3D)) {
```
turol: AI_MAX_ALLOC exists because Valgrind and AddressSanitizer have a maximum allocation size and exceeding it makes `new` return `nullptr` which breaks the normal rules. Changing this check breaks running under asan/valgrind. It's not possible to check every return value of `new` against `nullptr` and it would cause warnings from other tools since it can't happen in normal operation. However the value of AI_MAX_ALLOC could be increased or even moved to config.h.

code/ObjFileMtlImporter.cpp
```diff
@@ -89,11 +89,11 @@ static const std::string TypeOption         = "-type";
 // -------------------------------------------------------------------
 //  Constructor
 ObjFileMtlImporter::ObjFileMtlImporter( std::vector<char> &buffer,
-                                       const std::string &,
-                                       ObjFile::Model *pModel ) :
+                                        const std::string &,
```
turol: Unrelated whitespace change.

code/ObjFileParser.cpp
```diff
@@ -642,7 +638,7 @@ void ObjFileParser::getMaterialLib() {
     m_pIO->Close( pFile );
 
     // Importing the material library
-    ObjFileMtlImporter mtlImporter( buffer, strMatName, m_pModel );
+    ObjFileMtlImporter mtlImporter( buffer, strMatName, std::move(m_pModel) );
```
turol: I'm pretty sure this is wrong and the cause of the crash in the test suite. You don't want to pass the ownership of `m_pModel`, just borrow the pointer with ´.get()´. That means that the ObjFileMtlImporter constructor should remain unchanged.

aavenel: Yeah you're right, I have submitted the revision too early :/

code/ObjFileImporter.cpp
aavenel: ok, didn't know about this. Will revert this change.

aavenel:
Made the requested changes. It should be better now :)

turol:
Looks good to me. @kimkulling ?

kimkulling:
For me as well. I will merge the latest changes into the pull request and merge it! Thanks a lot for the contrinution.

Issue #1659, opened 2017-12-27
Bump pyassimp version and distribute


rspencer01:
Certain recent additions (such as the materials package) would be nice to have available on pip.  Can we bump the version number to 3.4 and publish a new package?

kimkulling:
We should increase it to 4.1.0 and of course publish a new package.

PR #1668, opened 2017-12-29
Version bump pyassimp

rspencer01:
What it says on the tin.  Reference #1659


rspencer01:
Is the pyassimp version number to always follow the assimp version number exactly?

kimkulling:
Merged, thanks for the fix.

PR #1676, opened 2018-01-03
Bug fixes for IFC

terziman:
2 differents bug fixes for IFC files:
- Support for optional parameter not defined in IfcSIUnit
- Fix for SwepDiskSolid tessellation when using polylines with only 2 points. The proposed fix increase the reliability of the tesselation and increase the performance by avoiding redundant calls to EstimateSampleCount method.


kimkulling:
Merged, thanks for your contribution for IFC, which I am gonna hate this year I guess.

PR #1681, opened 2018-01-05
[glTF2] Using relative buffers URI

### LINKED ISSUE #1674: [gltf2] when I export other model to gltf2,path of .bin file is not right
The .bin file path will bring the local path, resulting in an error.like this:
image
As you can see, the path to the .bin file is different from .gltf
How can I to do,thanks:)

Breush:
Closes #1674


kimkulling:

makc:
just checking, were there any important gltf2 fixes after this? we can't build from head (the build never ends) but need working gltf2 build

Breush:
@makc pretty sure there's nothing new after this

PR #1682, opened 2018-01-05
[glTF2] Correctly export images with bufferView

Breush:
When exporting, we now export embedded images correctly.


spexial:
Great!

kimkulling:

Issue #1667, opened 2017-12-29
PyAssimp DLL locator misses System32


rspencer01:
Unless I'm reading the code wrong (which I might be, not on a windows machine myself) the code here:

https://github.com/assimp/assimp/blob/dc5726bd0f99e65e1890345d7defd0eb5633bd1f/port/PyAssimp/pyassimp/helper.py#L55

to locate the assimp dll doesn't search through C:\Windows\System32.  Is there a reason that the `assimp` needs to be in the directory name?

kimkulling:
As far as I know Windows/System32 shall be part of the PATH-environment variable, right? But it is ocrrect: when assimp is stored in System32 this line of code is buggy.

rspencer01:
It should definitely be in the path, but won't contain the phrase `assimp`, yes.

kimkulling:
Closed by pull-request https://github.com/assimp/assimp/pull/1685

PR #1685, opened 2018-01-06
Remove check for 'assimp' in pyassimp search path

rspencer01:
Also ignore compiled python files that may be generated.

Reference #1667


kimkulling:
Ok, merged, thanks a lot for fixing the issue!

PR #1688, opened 2018-01-08
Added post-process to embed images

Breush:
Use `aiProcess_EmbedTextures`.

The algorithm goes:
- For each material
  - For each texture
    - Try to load its path directly
    - If it fails
      - try to load using `<imported-asset-directory>/<texture-path>`
      - If it fails
        - try to load using `<imported-asset-directory>/<texture-path-filename>`
        - If it fails, log an error and skip to next texture
    - Add the texture in `scene.mTextures` and set `texture.path = "*XX"` where `XX` is an integer.


kimkulling:
Merged, thanks a lot for this awesome contribution.

PR #1690, opened 2018-01-08
glTF 2.0: Set camera "look at" to (0.0, 0.0, -1.0).

### LINKED ISSUE #1689: glTF 2.0: Incorrect camera look at vector when importing glTF 2.0 file
When importing cameras from a glTF 2.0 file, the "camera look at" vector is incorrectly set. The glTF 2.0 spec specifies that the "look at" vector has to point in negative Z direction, but it is imported with a positive Z direction (the default value of assimp's camera constructor).

awefers:
fixes #1689


kimkulling:
Merged, thanks for the contribution,

PR #1692, opened 2018-01-08
glTF 2.0: Assign default material to meshes with no material reference.

### LINKED ISSUE #1691: glTF 2.0: Default material is not applied to meshes without material reference
If a mesh references no material, it is supposed to use the default material according to the glTF 2.0 specification. However, assimp uses the material with index 0, which is a the first material imported (and not the default material), hence meshes intended to use the default material will use the incorrect material.

awefers:
fixes  #1691 

Creates the default material as last material in the material index and assigns it to nodes without material reference. I decided to use the last index for the default material (the other option would be to use index 0 and shift all material indices by 1), so that the material index from the glTF file is preserved.


kimkulling:

PR #1694, opened 2018-01-08
glTF 2.0: Import scale for normal textures and strength for occlusion…

### LINKED ISSUE #1693: glTF 2.0: Import scale for normal textures and strength for occlusion textures
Assimp should import the scale for normal textures and strength for occlusion textures, otherwise it is not possible to achieve he correct material behavior when rendering an imported mesh. The scale and strength could be stored as glTF 2.0 specific material properties.

awefers:
… textures

resolves #1693


kimkulling:

PR #1695, opened 2018-01-08
Fix for failed import of some glTF 2.0 files.

### LINKED ISSUE #1605: [glTF2] some models don't import properly
See [this comment](https://github.com/assimp/assimp/pull/1603#issuecomment-348142377).

Models that don't import properly (from glTF sample models repo):

- NormalTangentTest
    - GLB variant
        - `ERROR: Failed to load file: OBJ: Invalid face indice`
        - seems to be interpreted as an OBJ?
- MetalRoughSpheres
    - GLB variant
        - `ERROR: Failed to load file: OBJ: Invalid face indice`
        - seems to be interpreted as an OBJ?
    - GLTF and GLTF-embedded variants:
        - `ERROR: Failed to load file: Validation failed: aiScene::mNumMeshes is 0. At least one mesh must be there`

awefers:
fixes #1605

### DIFF
--- a/code/glTF2Importer.cpp
+++ b/code/glTF2Importer.cpp
@@ -117,7 +117,7 @@ bool glTF2Importer::CanRead(const std::string& pFile, IOSystem* pIOHandler, bool
         }
     }
 
-    return false;
+    return true;
 }
 
 


JLouis-B:
An issue with this function is that the GLTF 2.0 files exported by Assimp have .gltf2 and .glb2 extensions.
I don't why, but because of this the GLTF 2.0 files exported by Assimp can't be loaded by Assimp itself.
According to the spec .gltf2 and .glb2 extensions don't exist so they probably have to change for the standard extensions (.gltf and .glb)

Not sure if it's related to your issues and test cases, but it's also probably something to change in the gltf exporters.

PR #1698, opened 2018-01-09
Clarify the matrix layout in the documentation

ombre5733:
The statement in the documentation that Assimp uses the same memory layout for matrices
as OpenGL does is wrong. Actually, both OpenGL and DirectX specify a matrix layout where
the base vectors and the translational part is consecutive in memory.

OpenGL uses post-multiplication of column-vectors and stores the matrix in column-major storage format:

    | X1  Y1  Z1  T1 |  | a |
    | X2  Y2  Z2  T2 |  | b |
    | X3  Y3  Z3  T3 |  | c |
    |  0   0   0   1 |  | 1 |

DirectX on the other hand uses row-major storage format but also pre-multiplication of
row-vectors

    | a b c 1 | | X1  X2  X3  0 |
                | Y1  Y2  Y3  0 |
                | Z1  Z2  Z3  0 |
                | T1  T2  T3  1 |

So a matrix is stored the same way in both frameworks and both times the translational part is consecutive at offsets 12, 13 and 14, which is not the format that Assimp uses (in Assimp the
translation can be found at offsets 3, 7 and 11 in the matrix memory, for example).


kimkulling:
Merged, thanks a lot for figuring this out.

PR #1700, opened 2018-01-10
Add pyassimp code generation script for materials

rspencer01:
In the style of gen/structsgen.py, we construct a similar method for
keeping the constants in materials.py up to date.


kimkulling:
Good idea. Thanks a lot for the contribution. Merged!

PR #1704, opened 2018-01-11
Improve FBX material import

mesilliac:
Some properties (reflectivity, shininess) were being incorrectly interpreted, and by default FBX material import was relying on a legacy system.

I updated the import to always use the modern system, which exists in all recent (2009+) FBX files i could find. Older files won't work anyway, as they follow a different file structure.

I also added a unit test for a simple FBX of a cube with phong material. It has most of the properties set to unique values, and they all load to the expected value with the new code.

I had to tweak PropertyGet to add a flag (defaults false) to query the relevant property template if the value isn't found in the given element. I'm not sure why it wasn't doing this already, but i left it off by default so it shouldn't affect anything else in any case.


kimkulling:

PR #1705, opened 2018-01-11
macOS build fix

### LINKED ISSUE #1684: [Bug] cmake assimp error
wx20180106-175631
when I cmake the new assimp version which version of @Breush solved gltf2's path of .bin file is not right.

awefers:
Requires cast from size_t to uint64_t like in other places, otherwise the implicit call to the GenericValue constructor is ambiguous.

Fixes #1684


kimkulling:

PR #1712, opened 2018-01-16
[glTF2] Fixed buffer alignment

Breush:
Also fixed some warnings emitted by the glTF2 validator about normals not being normalized.


kimkulling:
Looks good, merged! Thanks for the contribution

PR #1732, opened 2018-01-23
Fix thread-safety issue in 3DS and ASE loaders

turol:
3DS::Material creates a default name by incrementing a static non-atomic counter. This is neither thread-safe (even with separate Importer instances) nor reproducible in one thread. Refactor it by always passing in a name.


kimkulling:
Build is broken.

turol:
I know. Working on it. MSVC is a piece of crap and travis seems to have a problem with Ubuntu packages.

turol:
Looks like these weren't the only ones, just the first. There's several more similar bits of code in 3DSHelper.h. I'll fix them...

PR #1734, opened 2018-01-24
GetEmbeddedTexture should be a const function

carmenfan:
As it stands, an assimp scene created from reading a file (returned as a const aiscene*) cannot call GetEmbeddedTexture() as it is not a const member of the object.


kimkulling:
Looks good, merged.

PR #1735, opened 2018-01-24
Interpret IndexToDirect as Direct when the index element is missing

Rohansi:
Paint3D exports FBX like this, which is wrong, but we could still support it. When importing into Blender the vertex colors are missing on some parts of the model but this gives access to it.


Rohansi:
This is probably just Minecraft's 3D exporting at fault. Here's a screenshot comparing to Blender.

https://files.facepunch.com/Rohan/2018/January/25_07-56-56.png

kimkulling:
Ok, lets try it out. I will merge it!

PR #1749, opened 2018-01-29
Warning LNK4221 when not building exporter.

sacereda:
These messages were showing up if building without exporters option:

> AssxmlExporter.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
> AssbinExporter.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library


kimkulling:
Merged, thanks a lot for the contribution,

PR #1756, opened 2018-01-31
including <unzip.h> instead of <contrib/unzip/unzip.h>

sacereda:
contrib/unzip is already included in INCLUDE_DIRECTORIES when UNZIP_FOUND:

https://github.com/assimp/assimp/blob/038457653634040cd3a861b527beba981f93dca2/code/CMakeLists.txt#L844-L849


kimkulling:

PR #1759, opened 2018-02-01
Fix Matrix4x4t Decompose to rotation vector.

mesilliac:
The calculation of the rotation matrix was assuming left-handed coordinates with row-vectors, but assimp uses right-handed coordinates and column-vectors.

Previously this was sometimes giving me incorrect rotations when importing and directly exporting an FBX, but with this new code rotations seem to be correctly decomposed.

### DIFF
--- a/include/assimp/matrix4x4.inl
+++ b/include/assimp/matrix4x4.inl
@@ -424,12 +424,18 @@ inline void aiMatrix4x4t<TReal>::Decompose(aiVector3t<TReal>& pScaling, aiVector
 {
 	ASSIMP_MATRIX4_4_DECOMPOSE_PART;
 
-	/*
-	    |  CE     -CF      D   0 |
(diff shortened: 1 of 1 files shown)


kimkulling:

KakCAT:
Hello,

I just found a problem with Decompose and scale: it seems that scaleY and scaleZ seem to be swapped. I don't know if this fix also fixed the scale, so I'm not opening a new issue.

Unfortunately I can't move to the latest assimp to test, as I'm on assimp.NET (3.1.1 iirc)


This code breaks the decompose scale:

	Matrix4x4 mat=Matrix4x4.Identity;
	mat*=Matrix4x4.FromScaling (new Vector3D (1,1,0.5f));
	mat*=Matrix4x4.FromRotationY (1.2f);

	Vector3D scale;
	mat.Decompose (out scale,out _,out _);

Issue #1196, opened 2017-03-06
LWOB loader appears broken (LWO: Unable to build a valid node graph)


darkf:
When opening any LWOB file, `LWO: Unable to build a valid node graph` is given. To reproduce, try to import (e.g. via `assimp_viewer`) the model `test/models/LWO/LWOB/ConcavePolygon.lwo` (or any other file). You will receive the error.

turol:
The example models all seem to be very simple with only one node each.

@darkf Do you have a more complex LWOB model which you could share?

tellypresence:
I also getting the same error on the 4 bundled LWOB test models
```
test/models/LWO/LWOB/ConcavePolygon.lwo
test/models/LWO/LWOB/MappingModes/bluewithcylindrictexz.lwo
test/models/LWO/LWOB/sphere_with_mat_gloss_10pc.lwo
test/models/LWO/LWOB/sphere_with_mat_gloss_50pc.lwo
```

turol:
They don't work here either. Also they leak memory so they're not currently part of the unit tests and not easy to add. Also it looks like I've tried to fix that at some point with no success...

Did these files ever work? Do they load in some other app? Can you bisect assimp itself? Better yet, can you fix either the memory leaks or the loader itself?

Epsola:
This _still_ appears to be an issue. The test models don't work, nor do any other lightwave 5.5 models I've tried to load. Will this ever be fixed?

tellypresence:
LWOB models loading and rendering now, some screenshots for future reference below.
(Note: [`test/models/LWO/LWOB/ConcavePolygon.lwo`](https://github.com/assimp/assimp/blob/master/test/models/LWO/LWOB/ConcavePolygon.lwo) loaded OK but rendered completely invisible)

<table>
<tr><th>Model</th><th>Result></th></tr>
<tr><td><a href="https://github.com/assimp/assimp/blob/master/test/models-nonbsd/LWO/LWO2/LWSReferences/QuickDraw--Laserbeam.lwo">test/models-nonbsd/LWO/LWO2/LWSReferences/QuickDraw--Laserbeam.lwo</a></td><td><img src="https://user-images.githubusercontent.com/26958898/222350177-8c129b44-4465-4ec7-827c-d1e491d2acff.png" alt="QuickDraw-Laserbeam_lwo_-_assimp_commit_2023-02-27_e9033334" width=160></td></tr>
[...]

PR #1775, opened 2018-02-06
Fix some memory leaks

turol:
The changes in LWO don't actually change anything, I was just trying to understand what's going on there. There's a memory leak in LWOLoader.cpp line 434 which seems to be related to #1196.

This might affect #1767.

There's some real crap zlib decompression code which seems to have been copied to at least Blender and XGL loaders. I added TODO comments about cleaning it up.


kimkulling:

Issue #1776, opened 2018-02-07
ObjectCompare in BlenderImmediate.h doesn't follow strict weak ordering


maxvollmer:
```
    struct ObjectCompare {
        bool operator() (const Object* left, const Object* right) const {
            return ::strncmp(left->id.name, right->id.name, strlen( left->id.name ) ) == 0;
        }
    };
```
This comparator will cause an exception being thrown and possibly crash Assimp, since its implementation violates strict weak ordering. As counter-intuitive as it seems, a comparator must return false for equal values. The C++ standard defines and expects this behavior: `true if lhs < rhs, false otherwise.`
This would be a correct implementation:
```
    struct ObjectCompare {
        bool operator() (const Object* left, const Object* right) const {
            return ::strncmp(left->id.name, right->id.name, strlen( left->id.name ) ) < 0;
        }
    };
```

kimkulling:
Closed by https://github.com/assimp/assimp/pull/1777

PR #1777, opened 2018-02-07
Issue #1776 Fixed potential crash bug in ObjectCompare

maxvollmer:
Issue #1776 Fixed potential crash bug in ObjectCompare, because it didn't follow strict weak ordering. As counter-intuitive as it seems, a comparator must return false for equal values. The C++ standard defines and expects this behavior: true if lhs < rhs, false otherwise.


kimkulling:
Thanks a lot for the fix. Merged!

PR #1789, opened 2018-02-12
Make MemoryIOStream::Seek accept pos=length as valid

### LINKED ISSUE #1781: MemoryIOStream::Seek fails too early
`MemoryIOStream::Seek` looks like this:
```c++
aiReturn Seek(size_t pOffset, aiOrigin pOrigin) {
    if (aiOrigin_SET == pOrigin) {
        if (pOffset >= length) {
            return AI_FAILURE;
        }
        pos = pOffset;
    }
    else if (aiOrigin_END == pOrigin) {
        if (pOffset >= length) {
            return AI_FAILURE;
        }
        pos = length-pOffset;
    }
    else {
        if (pOffset+pos >= length) {
            return AI_FAILURE;
        }
        pos += pOffset;
    }
    return AI_SUCCESS;
}
```

I think, that the comparisons with `pOffset` should be done using `>` instead of `>=`.

Otherwise the following function calls do not behave symmetrical:
* `memoryIOWrapper.Seek(0, aiOrigin_END)`
* `memoryIOWrapper.Seek(length, aiOrigin_SET)`

Both function calls should seek to position "`length`", but only the first function call succeeds.

mabuchner:
Fixes #1781.

### DIFF
--- a/include/assimp/MemoryIOWrapper.h
+++ b/include/assimp/MemoryIOWrapper.h
@@ -99,19 +99,19 @@ public:
     // Seek specific position
     aiReturn Seek(size_t pOffset, aiOrigin pOrigin) {
         if (aiOrigin_SET == pOrigin) {
-            if (pOffset >= length) {
+            if (pOffset > length) {
                 return AI_FAILURE;
             }
             pos = pOffset;
         }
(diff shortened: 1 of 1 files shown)


kimkulling:

PR #1792, opened 2018-02-12
globalScale process not registered

sacereda:
aiProcess_GlobalScale was not being actually applied. Is this a mistake, or meant to be applied in a different way (Importer::RegisterPPStep)?

Feel free to close it if I'm missing something.


kimkulling:
Seems that I forgot to publish this. Thanks a lot for figuring this out. I will merge it!

kimkulling:

PR #1795, opened 2018-02-12
idea: #include <assimp/pbrmaterial.h>

sacereda:
Hi,

This is not a PR itself, but a workaround I'm using to identify PBR parameters and texture channels.
How should we identify those attributes? I had to create this pbrmaterial.h in order to include it from my plugin/exe, outside assimp project.

How is this supposed to be used in the future?

Thanks,

### DIFF
--- a/code/glTF2Asset.h
+++ b/code/glTF2Asset.h
@@ -137,7 +137,7 @@ namespace glTF2
     // Vec/matrix types, as raw float arrays
     typedef float (vec3)[3];
     typedef float (vec4)[4];
-	typedef float (mat4)[16];
+    typedef float (mat4)[16];
 
     namespace Util
     {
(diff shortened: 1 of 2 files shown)


kimkulling:
I like this isea, thanks a lot for the contribution!

sacereda:
Thanks, a thank you for your work on this project.

Regarding this, I'm actually importing some pbr parameters from glTF2 files.
My main concern is the scalability, as glTF2 is becoming a standard, maybe some of these parameters can be more generic and included in default material.h, e.g.:
 
- aiTextureType, adding some new types:
  - aiTextureType_METALLICROUGHNESS
  - aiTextureType_BASECOLOR (not sure this can be use as a diffuse one instead)

- Renaming some pbr material keys:
  - AI_MATKEY_BASE_COLOR_FACTOR
  - AI_MATKEY_METALLIC_FACTOR
  - AI_MATKEY_ROUGHNESS_FACTOR
  - ...

I can try to make those changes if you wish.

sacereda:
@kimkulling Another option would be to create an <assimp/materialextra.h> header, so different  importers can use it as a placeholder for different keys which don't fit in the actual scheme (if any).

In that case, glTF2 can use it to put any pbr-related parameter/texture. Also, any other exporter can place here any extra matkey, so whoever uses assimp can check in their end those keys if they need it. The current assimp process shouldn't be affected, as these keys are optional and importer-specific.

PR #1799, opened 2018-02-16
Some StreamWriter improvements / additions.

mesilliac:
Added the following functions:
 * Flush() to flush the internal StreamWriter buffer to the stream
 * Tell() gives the current stream position
 * Seek() seeks to a given offset in the stream
 * PutString(aiString&) to write an aiString
 * PutString(std::string&) to write a std::string

Seek and Tell were needed for FBX export, as the first thing it writes on each node is the end offset of the node - which might not be easily knowable before writing it.

Flush was a simple way of making sure seek and tell behave reliably, but is probably also useful to keep the buffer size small in certain situations. Otherwise the StreamWriter will buffer forever and only write to the output stream when it is destructed.

PutString is convenient for adding strings, i'm surprised it wasn't already there.

I'm not sure if it's necessary to also flush the output stream... i just made it also flush the output stream to be sure there will be no problems when seeking and telling.


kimkulling:

PR #1803, opened 2018-02-19
Fix aiBone->mOffsetMatrix documentation, which was incorrect.

mesilliac:
Also elaborated a bit on mOffsetMatrix, and slightly clarified the description of mWeights.

The mOffsetMatrix description was saying that it is a matrix transforming from mesh space to bone space, which was in fact the exact opposite of what it is.

I verified that both FBX import and GLTF export treat it as a transform from *bone* space to *mesh* space.

The extra info i added should make it even more clear.


kimkulling:
Thanks a lot for figuring this out. The pull request was merged!

Issue #1306, opened 2017-06-12
Wrong node transformations when importing FBX and DAE files


Trylz:
When trying to inport DAE and FBX files using 3ds max 2016, some nodes seems to have wrong transforms.
Using the aiProcess_PreTransformVertices flag or manually accumulating node transforms give me the same results. In the case of  FBX its look like only the last mesh of the scene is always wrong (i tried others scenes)

OBJ reference:  
![reference-obj](https://user-images.githubusercontent.com/9217999/27017194-b03a1c0e-4ef4-11e7-8cd6-5fc3ed5eece7.png)  

FBX:  
![wrong_node-fbx](https://user-images.githubusercontent.com/9217999/27017200-b39c57ea-4ef4-11e7-83c9-6c9fec3bdfd1.png)  

DAE:  
![wrong_node-dae](https://user-images.githubusercontent.com/9217999/27017201-b572c3c4-4ef4-11e7-8b54-7bd78894d8d5.png)

iKlsR:
If you're going to file an issue, share any file(s) and or specific code snippets used so someone can reproduce and possibly assist. Otherwise this is just noise and will help nobody.

Trylz:
Hey,


Sorry for the missing information. Any Fbx / Dae file from 3ds max should reproduce it. If not it's  not an Assimp issue.

________________________________
De : Ricardo Lovelace <notifications@github.com>
Envoyé : 15 juin 2017 20:43:27
À : assimp/assimp
Cc : Trylz; Author
Objet : Re: [assimp/assimp] Wrong node transformations when importing FBX and DAE files (#1306)


If you're going to file an issue, share any file(s) and or specific code snippets used so someone can reproduce and possibly assist. Otherwise this is just noise and will help nobody.

kimkulling:
Do you have a small example, please?

Trylz:
Sure. Here a zip with the model files corresponding to the screenshots and some code snippet.
The loading is done in BaseModel::loadModel(const std::string& path).
[Files.zip](https://github.com/assimp/assimp/files/1137315/Files.zip)


Thank you :)

rickomax:
Possibly related to:
https://github.com/assimp/assimp/issues/1127

Trylz:
Is there something planned for that ? if not where i could check in the source ? Its look like the problem is shared by collada and fbx models

mesilliac:
[...]

Issue #1297, opened 2017-06-03
Can't load "FBX models with Skeletal animation" properly


papertype:
I'm trying to make a simple model viewer with Assimp to show "FBX models with Skeletal animation".
But results are always broken. Bones are crashed.

Also, neither the latest "assimp_viewer" nor "Open 3D Model Viewer" can load them properly.

This is one of the samples I tested.
https://clara.io/view/2c296545-1204-4fc3-b470-c39bec65e86c

Collada (.dae) format of this model can be loaded somewhat properly:
![gunbot-dae](https://cloud.githubusercontent.com/assets/29165551/26755635/210f494c-48cc-11e7-9b06-d962a7734e92.png)
![viewer-dae](https://cloud.githubusercontent.com/assets/29165551/26755639/35d1d98a-48cc-11e7-9491-69f53e4bdb1b.png)

But FBX format:
![gunbot-fbx](https://cloud.githubusercontent.com/assets/29165551/26755641/3be8a128-48cc-11e7-9525-beead4736031.png)
![viewer-fbx](https://cloud.githubusercontent.com/assets/29165551/26755643/41839aca-48cc-11e7-9b4c-6c8091be0f95.png)

I couldn't find any FBX models with Skeletal animation which can be loaded properly with Assimp.
Is this a bug, or is there a way to fix this?

vexe:
I was able to load mine just fine. I would recommend you look at this example https://github.com/ccxvii/asstools/blob/master/assview.c

papertype:
I'm testing with free FBX models like above (https://clara.io/view/2c296545-1204-4fc3-b470-c39bec65e86c). Were you able to load and play this animation with your program? Open3mod and assimp_viewer can't. Autodesk FBX Review can.

vexe:
Yeah doesn't play correctly with open3viewer. Looks like a corner case in Assimp's importer. Few things to try:

- Import the model in 3dsmax, and try exporting with different settings
- I've had issues like this before and they mostly evolve around export settings, different 3d packages have different X/Y/Z directions, no standards. Plus you could export your model with something crazy like -X being their forward axis! So many cases to handle Assimp probably missed or mis-handled some of them.
- In assview.c, try compiling and running that, load the model, mess with the import flags, in the beginning of the program see if you could find something useful. Then try and manually flip the vertices after you loaded them, swap Z/Y see if it loads any different maybe that will lead you to some clues as to what's actually happening
- If all fails, you'd need the author to debug this.

[...]

Issue #1112, opened 2016-12-15
FBX transformation hierarchy wrong


Nillmo:
Hi everyone,

first of all, thanks for this great library and all the work you put into making it.
We use Assimp to load user provided content, therefore we have a wide variety of input formats and don't have much control over the input files. Most of our test data is loading fine with Assimp 3.3.1, I only have a problem with one FBX file, where the translation of some child meshes is shiftet.
The same model in Collada format loads correctly. I've attached both models in the zip file below.

[crankshaft.zip](https://github.com/assimp/assimp/files/654032/crankshaft.zip)

kimkulling:
Thanks a lot for the report. Hopefully I will find some time to take a look onto it.

Eirenliel:
This model have GeometricTranslation, which should not be inherited by child nodes. Assimp does not support it now.

(I can't fix it myself, cause I don't actually write on C++, but I ate a dog in FBX...)

kimkulling:
Thanks a lot for the hint . At the moment I am trying to understand the code.

rickomax:
Possibily related to:
#1127

mesilliac:
Should be fixed by #1806 (i at least confirmed that converting to OBJ gives something that looks correct)

PR #1806, opened 2018-02-21
FBX geometric transforms fix

### LINKED ISSUE #1127: FBX importer: geometric transformations should not be applied to children
(This is  a duplicate of #671 which has been closed unsolved.)

When the Assimp FBX importer encounters a node with a geometric transformation (GeometricTranslation, GeometricRotation or GeometricScaling) it applies the transformation to all the children nodes, while according to the [FBX Developer Help](http://help.autodesk.com/view/FBX/2017/ENU/?guid=__files_GUID_C35D98CB_5148_4B46_82D1_51077D8970EE_htm) it should be applied only to the mesh attached to the node.

mesilliac:
This applies the inverse of FBX "geometric" transforms to child nodes. The transforms are thus applied to any attached meshes, but correctly not applied to child nodes, as FBX intends.

This should fix #1127, and thus also perhaps #1112, #1297, and #1306 (for FBX).

I tested that the example in #1127 behaves correctly - OBJ export looks identical to the FBX loaded in Maya. But i haven't tested any of the example files in the other issues.

There will be a problem if anybody applies a geometric transform with a scale component of zero (i don't know why anone would do that, but i can't say they won't). In this case children will inherit this zero scale transform and results will be wrong, it won't even try to invert it.

There will also be a problem if the geometric transform is animated. However i think this property should not usually be animatable, so again it probably won't occur in real use.

### DIFF
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -579,6 +579,7 @@ bool Converter::NeedsComplexTransformationChain( const Model& model )
     bool ok;
 
     const float zero_epsilon = 1e-6f;
+    const aiVector3D all_ones(1.0f, 1.0f, 1.0f);
     for ( size_t i = 0; i < TransformationComp_MAXIMUM; ++i ) {
         const TransformationComp comp = static_cast< TransformationComp >( i );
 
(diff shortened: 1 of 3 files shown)


mesilliac:
Oh, right, i also added a --verbose flag to `assimp info` which prints out node transform information (decomposed into scale, rotation, translation).

kimkulling:
Merged, thank you so much for this contribution.

PR #1810, opened 2018-02-25
assimp_cmd info: list meshes and print basic mesh stats.

mesilliac:
just a small addition to the `assimp info` command to print mesh info.


kimkulling:

PR #1818, opened 2018-02-28
Fix material index off-by-one error in some OBJ files

jfaust:
Saw this in a file from Poly (https://poly.google.com/view/5POtMKIT_Ze), which was exported from C4D. This would've been a negative index into the material array if a default material wasn't always created when loading OBJ files. As it was, it'd assign DefaultMaterial to the mesh, rather than the named material.


kimkulling:
Great finding, Just merged it to master. Thanks a lot for the contribution.

PR #1819, opened 2018-02-28
FBX Export: add missing 0 value to file footer.

mesilliac:
This correctly aligns the footer. The FBX SDK doesn't seem to care when importing (neither does Blender, or probably anything else), but this better matches the footer of officially-exported FBX files.


kimkulling:

PR #1820, opened 2018-03-01
Fix default opacity of materials exported to FBX by Blender.

mesilliac:
FBX import was checking for "phong", but Blender was writing "Phong".


kimkulling:

PR #1821, opened 2018-03-03
Collada: add importer property that forces the use of collada names. …

### LINKED ISSUE #1375: DAE Node Names
DAE file nodes have generally three attributes:
**id**, **name** and **sid**.

Assimp returns the **id** attribute value for any node name.

In some engines, like Unity, the **name** field is used instead.

A config can be created to choose between **id** and **name** field
when returning this information.

Maybe I will take this task, just letting it here to remember.

ekpyron:
…Closes #1375.


kimkulling:

PR #1822, opened 2018-03-03
Fix #623 : PLY importer should not automatically create faces

### LINKED ISSUE #623: PLY importer should not automatically create faces (and it does it wrong)
If no faces are found in the header, the PLY importer automatically adds them by simply generating triplets by taking vertex indices triplets sequentially.
The code is in PlyLoader.cpp. from line 207:

```
// if no face list is existing we assume that the vertex
// list is containing a list of triangles
if (avFaces.empty())
{
    if (avPositions.size() < 3)
    {
        throw DeadlyImportError( "Invalid .ply file: Not enough "
            "vertices to build a proper face list. ");
    }

    const unsigned int iNum = (unsigned int)avPositions.size() / 3;
    for (unsigned int i = 0; i< iNum;++i)
    {
        PLY::Face sFace;
        sFace.mIndices.push_back((iNum*3));
        sFace.mIndices.push_back((iNum*3)+1);
        sFace.mIndices.push_back((iNum*3)+2);
        avFaces.push_back(sFace);
    }
}
```

The for loop is bugged: a PLY::Face already has 3 indices (by constructor), so it ends up having 6 indices, first 3 are zero, last ones have constant indices ("iNum" should be "i").

Apart from this bug itself, this prevents the importer to properly detect and load point clouds (i.e., geometry with no faces, just vertices). Even if the whole "if" at line 207 is removed, meshes in later stages will not be created and the scene will be null.

As a temporary fix, I added a new flag to the scene, AI_SCENE_FLAGS_AUTO_FACES: when set, this would report the caller that faces are artificial. The correct behavior is to not create faces at all.

m.

aavenel:
PLY could contain point cloud data, so we shouldn't try to guess how to create faces in this case.


kimkulling:

PR #1823, opened 2018-03-03
FBX Export: minor tweak to footer.

mesilliac:
Should now be identical to those output by the FBX SDK.

Apparently i just had these four bytes in the wrong place previously.


kimkulling:

PR #1825, opened 2018-03-04
Fix #1415 and improve coverage of PLY tests

### LINKED ISSUE #1415: Shipped test file float-color.ply is broken
`assimp info` says:

```Failed to load file: Validation failed: aiMesh::mFaces[0]::mIndices[2] is out of range```

`mIndices[2]` is 200. It looks like the second-to-last line is being misparsed as a triangle instead of a vertex.

The cause seems to be a missing newline at end of file. With the newline added it works. This file seems to be part of some unit test. Why does that unit test work if the file is broken?

### PR DESCRIPTION
**Pull request by aavenel**
Add some tests and fixed the float-color.ply which was broken but not correctly reported.

### CODE REVIEW

**Comment by turol** [comment]
UBsan check is failing because of bad pointer casting around PlyParser.cpp:1072. Can you replace these with memcpy?

**Comment by aavenel** [comment]
Yeah, I've just noticed this, I'll make the changes.

UBsan is really a great tool to catch these bugs.

**Comment by aavenel** [comment]
I tried to do something like this : 
```
  case EDT_UInt:
  {
    uint32_t t;
    memcpy(&t, pCur, sizeof(uint32_t));
    pCur += sizeof(uint32_t);

    // Swap endianness
    if (p_bBE)ByteSwap::Swap(&t);
    out->iUInt = t;
    break;
  }
```

but unfortunately it crashes. Any idea of what I've missed ?

turol:
Looks right to me, not sure where the problem is. You could try
1. Don't use temporary `t`,  just `memcpy` directly to `out->iUInt`
2. Run it with ASAN and see what the problem is. If you don't have Linux yourself you could just push it and see that Travis says. Just be ready to rebase out the broken commit.

aavenel:
Well, looks like I was executing the wrong build ... Everything should be green again ! 
I've rebased as you suggested.
Thanks for your review.

turol:

PR #1827, opened 2018-03-06
Fix import of binary STL files in double-precision builds

JeffH-BMG:
When ASSIMP_DOUBLE_PRECISION is used, the STL loader attempts to read 8-byte double vertex and normal values from the STL file. STL files are written using 4-byte floats, however, and the import will read past the end of the buffer, and possibly crash.


code/STLLoader.cpp
```diff
@@ -449,26 +449,29 @@ bool STLImporter::LoadBinaryFile()
...
-
-        ::memcpy( vp, sz, sizeof( aiVector3D ) );
-        ++vp;
-        sz += sizeof(aiVector3D);
-
-        ::memcpy( vp, sz, sizeof( aiVector3D ) );
-        ++vp;
-        sz += sizeof(aiVector3D);
+
+        // There's one normal for the face in the STL; use it three times
```
kimkulling: This will not work when the memory is not aligned in the right way. That was the reason to remove the c-like casts.

code/STLLoader.cpp
```diff
@@ -449,26 +449,29 @@ bool STLImporter::LoadBinaryFile()
     aiVector3D *vp = pMesh->mVertices = new aiVector3D[pMesh->mNumVertices];
     aiVector3D *vn = pMesh->mNormals = new aiVector3D[pMesh->mNumVertices];
 
+    typedef aiVector3t<float> aiVector3F;
+    aiVector3F* theVec;
```
kimkulling: Thats a good idea. So switching this to double precision will not break the import.

code/STLLoader.cpp
kimkulling: IN general the idea is right. But you need to use this with memcpy even if it's looks not so good.

JeffH-BMG:
Hi Kim -- as I am new to GitHub, is there anything that I need to do except to commit the changes that address your comments? Thanks.

kimkulling:

PR #1828, opened 2018-03-06
STL binary Export should write 4-byte floats for vertex and normal co…

JeffH-BMG:
…ordinates

The STL binary format uses 4-byte floats. When using double-precision builds of Asset Importer, the STL exporter was writing out 8-byte double values instead..


kimkulling:
Merged, thanks a lot for the conrtibution!

JeffH-BMG:
You are most welcome.  I hope to be able to help out in the future (this is my first use of GitHub). Thanks for your work on this library!

PR #1829, opened 2018-03-06
Add support for texture file in PLY exports

JeffH-BMG:
The PLY format has an unofficial way to specify an associated texture, using the "comment TextureFile"  comment line. The PLY loader supports this, but the exporter does not. The change looks for a diffuse texture in the scene's materials, and if it finds one, it adds it to the exported mesh using "comment TextureFile".


kimkulling:

PR #1837, opened 2018-03-16
Support Maya 2018 Collada Export

gonchar:
When blendshapes and bones are applied to a mesh Maya 2018 exports Collada with nested controllers which are not supported in Assimp. I've placed postprocessor which breaks the controllers hierarchy

[torus.zip](https://github.com/assimp/assimp/files/1817949/torus.zip)


kimkulling:

PR #1839, opened 2018-03-16
Expand the current documentation about loading of embedded textures

loebl:
This updates the docs regarding the latest updates (GetEmbeddedTextures). It is based on my experience working with it, so it's not necessarily correct/complete, and might need updates in the future.

I also took the liberty to format the text paragraphs I edited.


kimkulling:

PR #1856, opened 2018-03-26
Address some gcc warnings about signed / unsigned comparison.

mesilliac:
pretty minor warnings, but nonetheless


kimkulling:
Thanks, I will merge it!

PR #1857, opened 2018-03-27
Basic animation support in FBX export

mesilliac:
supports only node-based animation for now.


kimkulling:
Great, thanks a lot for the contribution.

PR #1858, opened 2018-03-29
Ascii FBX Export

mesilliac:
Seems to work for my test models. Should make it easier to debug FBX export in general.

It mostly shares the same code with the binary exporter, just with some small divergences as the file structures differ slightly.

Based off the branch with animation support.


kimkulling:

PR #1859, opened 2018-03-29
assimp_cmd: prettier and better-compressed node hierarchy

mesilliac:
I also removed the hardcoded line and nesting limits. It's better to show all the info than to silently discard some. With the new compressed display, it is far less obtrusive anyway, and the user can always resize their terminal if they want to examine a large hierarchy.

By default it's outputting UTF-8 line-drawing characters to draw the node tree. This is well-supported under linux but i have no idea about other operating systems. I also put an option to use ascii characters. If some platform doesn't support the UTF-8, then adding an `#ifdef` for that platform at the top of `tools/assimp_cmd/Info.cpp` is trivial.


kimkulling:

PR #1863, opened 2018-03-31
Some fixes for MMD PMX file

aoowweenn:
From #644 
The following issues should be solved:
1. [Resolved] Encoding UTF16 to UTF8 by the new utf8 library
2. [Resolved] A material don't need to append texture when texture ID < 0
3. `mMeshes[a]` is already `null` inside aiScene destructor in `Version.cpp` line 146

(3) I can't reproduce the outcome currently.


kimkulling:

Issue #1219, opened 2017-03-31
CalcTangentsProcess::ProcessMesh: division by zero


regrainb:
With Assimp 3.1.1:

In the file CalcTangentsProcess.cpp (line 191), in the function CalcTangentsProcess::ProcessMesh, we test if the 3 points are aligned instead of being at the same place. Because if they are aligned, the tangent and bitangent can't be computed.
Code:
   `if ( 0 == sx && 0 ==sy && 0 == tx && 0 == ty )`
should be replaced with:
   `if ( sx * ty == sy * tx )`

In my case, I had 2 points at the same place instead of 3, but the source of the issue is the same.

kimkulling:
Ok, thanks for the bug report. I will fix that bug.

kimkulling:
Fixed.

PR #1868, opened 2018-04-02
Fixed issue#1219. CalcTangentsProcess::ProcessMesh.

EdwardAndrew:
From #1219.

> In the file CalcTangentsProcess.cpp (line 191), in the function CalcTangentsProcess::ProcessMesh, we test if the 3 points are aligned instead of being at the same place. Because if they are aligned, the tangent and bitangent can't be computed.
Code:
if ( 0 == sx && 0 ==sy && 0 == tx && 0 == ty )
should be replaced with:
if ( sx * ty == sy * tx 

> )

Implemented suggested solution.


kimkulling:
Thanks a lot for integrating this fix. I have completly forgotten to integrate this.

PR #1876, opened 2018-04-03
Updated COB importer plugin to flip normals.

DMJC:
When loading Truespace .cob models into the assimp_QT_viewer the normals are flipped incorrectly. Added a normal flipping call on import to ensure correct orientation of normals.


kimkulling:

PR #1902, opened 2018-04-18
IFC fixes

terziman:
Different fixes for IFC import:
- fixed parser to handle '##' symbols in strings as they should not be interpreted as references
- improved underlying data structure in mesh cache to reduce memory footprint. Indeed, when trying to load big IFC files, the previous implementation would lead to out-of-memory exceptions. This was caused by vectors of indexes of meshes containing lots of duplicated elements (~700 millions elements per object in my case, repeated many times), which resulted in lots of memory wasted. The proposed implementation use sets instead, to avoid keeping duplicated elements. Moreover, it also speed up the import process a little, as we do not need to sort and remove duplicates in those vectors anymore.


kimkulling:

PR #1913, opened 2018-04-24
Changed 'souces' to 'sources'.

Binero:
There was a typo in the `CMakeLists.txt` where if zlib was not found, it would say "Compiling from souces" instead of "Compiling from sources". Added a 'r' to fix the typo.


kimkulling:

PR #1915, opened 2018-04-24
Use CMAKE_INSTALL_PREFIX as ASSIMP_ROOT_DIR.

### LINKED ISSUE #1914: assimp-config.cmake sets incorrect ASSIMP_ROOT_DIR when library path is deeply nested.
This may or may not prove related to #1912 but when ASSIMP_LIB_INSTALL_DIR contains more than one directory level the logic below isn't sufficient to find the root directory producing a broken CMake configuration.

https://github.com/assimp/assimp/blob/620182770da9f851a44040af69cb20ba6c3f6106/assimp-config.cmake.in#L13-L16

This affects the libassimp-dev package on Debian Stretch, Buster, and Sid and Ubuntu Artful and Bionic. Verifiable on these platforms with the following example CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 3.0)
project(asitest)

find_package(assimp)

message(STATUS "ASSIMP_ROOT_DIR: ${ASSIMP_ROOT_DIR}")
message(STATUS "ASSIMP_LIBRARY_DIRS: ${ASSIMP_LIBRARY_DIRS}")
message(STATUS "ASSIMP_INCLUDE_DIRS: ${ASSIMP_INCLUDE_DIRS}")
```

Selected output from above CMakeLists.txt:
```
...
-- ASSIMP_ROOT_DIR: /usr/lib
-- ASSIMP_LIBRARY_DIRS: /usr/lib/lib/x86_64-linux-gnu
-- ASSIMP_INCLUDE_DIRS: /usr/lib/include
...
```

nuclearsandwich:
Since the ASSIMP_LIB_INSTALL_DIR where this CMake config file is  installed to may contain multiple nested directories, we should not use it to back-calculate the location of the ASSIMP_ROOT_DIR unless we can take the directory depth into account. Instead use the
CMAKE_INSTALL_PREFIX as the ASSIMP_ROOT_DIR at configure-time.

This approach is more flexible with respect to install directories but it means that assimp files are not relocatable after installation and I don't think it works if a DESTDIR option is provided when running `make install` after cmake configuration.

Proposed to resolve #1914.

### DIFF
--- a/assimp-config.cmake.in
+++ b/assimp-config.cmake.in
@@ -10,10 +10,7 @@
 # ASSIMP_LIBRARY_DIRS - link directories
 # ASSIMP_LIBRARIES - libraries to link plugins with
 # ASSIMP_Boost_VERSION - the boost version assimp was compiled with
-get_filename_component(_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
-get_filename_component(_PREFIX "${_PREFIX}" PATH)
-get_filename_component(_PREFIX "${_PREFIX}" PATH)
-get_filename_component(ASSIMP_ROOT_DIR "${_PREFIX}" PATH)
+get_filename_component(ASSIMP_ROOT_DIR "@CMAKE_INSTALL_PREFIX@" REALPATH)
 
 if( MSVC )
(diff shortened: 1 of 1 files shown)


kimkulling:
Looks good, I will merge it!

kimkulling:

PR #1916, opened 2018-04-25
Add vertex color support to glTF2 export

LoicFr:
The vertex colors were not exported when choosing the glTF2 format, so I simply add it to the glTF2Exporter.


kimkulling:

PR #1917, opened 2018-04-26
Fix gltf2 export component type error

### LINKED ISSUE #1907: model with animation  to gltf2,   JOINTS_0  componentType is Error
Hi, 
  I have get a bug when  translate a model file which with animation to gltf2.
 The reason is  that attribute JOINTS_0 componentType is 5126
image

image

We changed the source code of assimp,  resloved this bug, but not sure  if it leads to another problem.

The related files is:
ori:
[source.zip](https://github.com/assimp/assimp/files/1934304/source.zip)

gltf:
[gltf.zip](https://github.com/assimp/assimp/files/1934303/gltf.zip)

Thanks!

wuxq:
fix #1907


kimkulling:

PR #1922, opened 2018-04-27
Fix ply exporter to be conformant to spec respecting vertex colors.

### LINKED ISSUE #1167: PLY format color not working
Assimp exports PLY file with float as vertex color property.
However, color is not rendered when visualized with MeshLab or CloudCompare.

Vertex color property should be **uchar** to comply with PLY format (http://paulbourke.net/dataformats/ply )

Attached files are two different PLY files for **uchar** color and **float** color.
Only **uchar** color file works.

[examples.zip](https://github.com/assimp/assimp/files/756984/examples.zip)

kwuerl:
Vertex colors should be defined as red, green, blue, alpha instead of r, g, b, a and should have the type 'uchar' instead of 'float'. This makes assimp compatible with meshlab and blender ply importers.
Also see http://paulbourke.net/dataformats/ply/ .
This fixes #1167 .


kimkulling:

PR #1925, opened 2018-04-28
multi joint has bug  with ReplaceData

wuxq:
multi joint has  a bug  in ReplaceData_joint with "bin" change


wuxq:
Hi, @kimkulling  
  I found a new bug in here.

Issue #1807, opened 2018-02-21
FBX import missing meshes name


Trylz:
When importing FBX files meshes have empty names(aiMesh::mName).
I attached the test scene and my code snippet. The loading is done in BaseModel::loadFromAiModel(const std::string& path, bool importMaterials)

[Files.zip](https://github.com/assimp/assimp/files/1746010/Files.zip)

mesilliac:
FBX files don't usually name the meshes themselves - rather the node that contains the mesh is what is named with the expected name.

In this fbx file also, the mesh geometry entries are actually unamed.

We could name the mesh the same as the node in this case, that might make things clearer.

Trylz:
I see. That probably what other importer does(I tried in Blender). 
The weird thing is i tried myself to assign random name to the MeshGeometry but they are ignored. I am probably doing it wrong. Do you have an idea of where it should be done ?

mesilliac:
Do you mean in the assimp code? The name (if it exists) is assigned in `Converter::SetupEmptyMesh`, in FBXConverter.cpp around line 909.

Trylz:
Yes i tried the following. During import meshes name are still empty.

``    aiMesh* const out_mesh = new aiMesh();

    out_mesh->mName.Set("TestName");

    meshes.push_back( out_mesh );
    meshes_converted[ &mesh ].push_back( static_cast<unsigned int>( meshes.size() - 1 ) );

    return out_mesh;``

mesilliac:
that's weird - i just tried this exact code and the mesh in the imported scene was definitely named "TestName". How are you checking the name?

using test/models/FBX/phong_cube.fbx, i get on assimp master:
```
Meshes:  (name) [vertices / bones / faces | primitive_types]
    0 (): [24 / 0 / 12 | triangle]
```

and using this code above:
```
Meshes:  (name) [vertices / bones / faces | primitive_types]
    0 (TestName): [24 / 0 / 12 | triangle]
```

mesilliac:
(this was using the `assimp info` command but from this PR which isn't merged yet: https://github.com/assimp/assimp/pull/1810 )

Trylz:
Yes thanks, the test name is properly set. The problem was on my side.

kimkulling:
Ok, so I will close the issue.

Trylz:
Well in facts the issue is not fixed.  Meshes name are still missing but my little test pass :)

darksylinc:
[...]

PR #1929, opened 2018-04-29
Fix empty fbx mesh names

Trylz:
Patch provided by darksylinc
This PR fix the bug mentioned in #1807


kimkulling:

PR #1931, opened 2018-04-30
 Fix Issue #1923: OBJ Exporter can't correctly export vertex colors

### LINKED ISSUE #1923: OBJ Exporter can't correctly export vertex colors
What: For example exporting 25 different vertices, all with color 1.0, 1.0, 1.0, 1.0 results in having only one vertex in the resulting obj file. Having 10 different colors will result in only exporting 10 vertices.
Why: Internally the OBJ exporter uses a std::map to avoid exporting the same vertex twice. When adding color information to vertices, the exporter does not include color information as additional keys in the first map, but creates a second map for colors.
Thus, 2 different colors of 10 vertices leads to a color map with 2 elements.
Unfortunately while exporting all distinct vertices the color map is assumed to have the same amount of elements than the vertex map and a linear index is used to access them. If the index gets greater than the color map size, additional vertices are simply omited.
I will probably submit a PR in the next few days to fix this. If I do not, I hope this description helps!

kwuerl:
Exporting 25 different vertices, all with color 1.0, 1.0, 1.0, 1.0 resulted in having only one vertex in the resulting obj file. Having 10 different colors will resulted in only exported 10 vertices.
For more info see Issue #1923 .
This PR fixes #1923 .


kimkulling:

PR #1937, opened 2018-05-02
Fix glTF2 file extensions so that the importer matches the exporter

LoicFr:
The glTF2 importer has "gltf" and "glb" as file extensions while the exporter has "gltf2" and "glb2" as file extensions. Therefore a file exported in glTF2 by Assimp can't be reopened by Assimp. I propose to change the importer extensions to match the exporter extensions because "gltf" and "glb" are already used by glTF 1.0 importer/exporter.

### DIFF
--- a/code/glTF2Importer.cpp
+++ b/code/glTF2Importer.cpp
@@ -80,7 +80,7 @@ static const aiImporterDesc desc = {
     0,
     0,
     0,
-    "gltf glb"
+    "gltf2 glb2"
 };
 
 glTF2Importer::glTF2Importer()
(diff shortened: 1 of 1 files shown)


JLouis-B:
But according to the specification I don't think that gltf2 and glb2 extensions exist (I have neither see theses extensions outside Assimp)
Instead of using gltf2 and glb2 in the exporter, I think that the best solution is to use gltf and glb extensions in the exporters. (the importers can do the diffrence between v1 and v2 by reading the file format version in the header, I think that it's already implemented)

LoicFr:
Thank you for your comment. I checked and it seems that your are completely right. I will cancel this pull request and propose to replace gltf2 and glb2 in the exporter by gltf and glb.

PR #1938, opened 2018-05-02
Change glTF2 file extensions from gltf2/glb2 to gltf/glb in the expor…

LoicFr:
…ter so that it matches the importer and respect the standard specifications

There is a mismatch between importer and exporter file extensions for the glTF2 format. It is not possible to export in glTF2 and reimport after. To respect the standard specfication, the exporter file extensions must be gltf/glb instead of gltf2/glb2. The importer don't have to change to support this modification.


kimkulling:
Nice finding, merged. Thanks a lot for the contribution.

PR #1943, opened 2018-05-05
1740 Issue - Enabling and Disabling importers

Z-EMB:
-ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT will be respected by the macro
--using local variable ASSIMP_IMPORTER_ENABLED to include or exclude
--an importer
-Fixes individual importer ex: set(ASSIMP_BUILD_FBX_IMPORTER OFF)

-ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT is cached
--if assimp is a child cmake project, need to force cache
--from parent project
--ex: set(ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT CACHE INTERNAL FALSE)


kimkulling:
Great, thanks a lot for the fix. I started looking onto that issue but I did not find any time to fix it.

kimkulling:

PR #1948, opened 2018-05-07
Fixes DXF loader false positive on FBX file

Breush:
A binary FBX file can have an header section `FBXHeaderVersion` which starts around the 70th byte.
Therefore, the token check for DXF file was hitting true because the `SearchFileHeaderForToken` is case insensitive.
We limit the scope of the token search to the first 32 bytes.


kimkulling:
Thanks a lot for the contribution, merged.

Issue #952, opened 2016-07-22
AssimpView : Compilation error with MinGW 5.3 


JLouis-B:
Hi,
There are some compilation errors in AssimpView.h :

``` cpp
#ifndef max
#define max(a,b)            (((a) > (b)) ? (a) : (b))
#endif // max
#ifndef min
#define min(a,b)            (((a) < (b)) ? (a) : (b))
#endif // min
```

Give :
C:\Users\Jean-Louis\Desktop\Libs\assimp-3.3.1\tools\assimp_view\assimp_view.h:76:18: error: expected unqualified-id before '(' token
 #define max(a,b) (((a) > (b)) ? (a) : (b))
                  ^
C:\Users\Jean-Louis\Desktop\Libs\assimp-3.3.1\tools\assimp_view\assimp_view.h:79:18: error: expected unqualified-id before '(' token
 #define min(a,b) (((a) < (b)) ? (a) : (b))

kimkulling:
Should be fixed on the latest master.

JLouis-B:
Thanks for your fix, unfortunately there is still some problems on the master :
C:\Users\Jean-Louis\Desktop\Libs\assimp-master\tools\assimp_view\assimp_view.h:186:35: error: 'min' was not declared in this scope

tyduptyler13:
I can confirm that the latest master still has this issue.

```
/home/tyler/workspace/MyEngine/deps/assimp/tools/assimp_view/MessageProc.cpp: In function ‘void AssimpView::DoExport(size_t)’:
/home/tyler/workspace/MyEngine/deps/assimp/tools/assimp_view/MessageProc.cpp:1059:72: error: ‘max’ was not declared in this scope
         char* sz = max(strrchr(szFileName,'\\'),strrchr(szFileName,'/'));
                                                                        ^
/home/tyler/workspace/MyEngine/deps/assimp/tools/assimp_view/MessageProc.cpp:1059:72: note: suggested alternative:
In file included from /usr/lib/gcc/x86_64-w64-mingw32/6.2-posix/include/c++/algorithm:62:0,
                 from /home/tyler/workspace/MyEngine/deps/assimp/include/assimp/matrix3x3.inl:53,
                 from /home/tyler/workspace/MyEngine/deps/assimp/include/assimp/types.h:517,
                 from /home/tyler/workspace/MyEngine/deps/assimp/include/assimp/cimport.h:50,
[...]

PR #1949, opened 2018-05-07
Fix Assimp viewer build with MinGW

JLouis-B:
- Use std::min/max instead of min/max macro in windef.h ( https://github.com/assimp/assimp/issues/952 )
- Use mmsytem.h instead of timeapi.h with MinGW

The changes have been tested with MinGW 5.3 and Visual Studio 2017.


kimkulling:

PR #1951, opened 2018-05-08
_stat64 doesn't seem to exist. use __stat64!

DCubix:
I was trying to build Assimp from sources, by adding it as a submodule of my project. Then for some reason it wasn't compiling.

Then I found out that apparently `_stat64` is wrong, and it should be `__stat64`.

It compiled just fine after that.

🤷‍♂️


kimkulling:

PR #1964, opened 2018-05-15
PyAssimp: Encode filename using file system encoding instead of ASCII

ruxkor:
As of now, `load`ing and `export`ing a file causes a `UnicodeDecodeError` if the filename is not in ASCII format. 

I'm not sure if we really need this restriction here. One possible solution would be to expect a string here, but this would not be intuitive while using python 3. The other solution, proposed here, is to use the encoding `sys.getfilesystemencoding()` returns.

### DIFF
--- a/port/PyAssimp/pyassimp/core.py
+++ b/port/PyAssimp/pyassimp/core.py
@@ -312,7 +312,7 @@ def load(filename,
                                      file_type)
     else:
         # a filename string has been passed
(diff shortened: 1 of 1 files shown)


kimkulling:
Is ok for me. The encoding is already an issue when dealing with chinese filenames, which is unsolved. So thanks a lot for your contribution! I will merge it!

13704017891:
Has the problem been solved?

PR #1965, opened 2018-05-16
Bug/exception on empty mesh after degenerate removal

loebl:
I triggered "code that will hopefully never be called" when importing a 3DS model from one of our customers (It is an exported file from some other software I have no control over). When importing I use `aiProcess_FindDegenerates` in the postprocessing and set it to remove degenerates with `tImporter.SetPropertyBool(AI_CONFIG_PP_FD_REMOVE, true);`

The line that failed is FindDegenerates.cpp:249, it throws an exception if a mesh is empty after removing degenerate primitives. As far as I see, this should not be a case to abort the import, the mesh should simply be removed (maybe it can even be left as it is as empty mesh).

The mesh that triggered the exception was a quad that represented a point (all vertices have the same coordinates, 2 faces (triangles) that form a quad). I can't give you the original file, but I could try to construct a minimal OBJ that shows the same behaviour.

This patch changes the code to remove the mesh in this case. It also updates the scene graph to correct all mesh references. It works for me, but I have only minimal knowledge if assimp internals, so I recommend a second look if the delete is correct. Also I am using the C++11 auto keyword, but that should be easy to correct if it is a problem.

This keeps one edge case open: nodes without mesh references after mesh removal. They are kept as it is for now (they may still contain transformations and child references).

If you want me to correct/modify the patch, just give me a note.

### DIFF
--- a/code/FindDegenerates.cpp
+++ b/code/FindDegenerates.cpp
@@ -54,6 +54,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 using namespace Assimp;
 
+//remove mesh at position 'index' from the scene
+static void removeMesh(aiScene* pScene, unsigned const index);
+//correct node indices to meshes and remove references to deleted mesh
+static void updateSceneGraph(aiNode* pNode, unsigned const index);
+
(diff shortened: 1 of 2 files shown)


kimkulling:
Merged, I guess this shall work as far as I know. Thanks a lot for the contribution.

PR #1966, opened 2018-05-16
Fix build with Qt 5.11.0_beta3, add missing install target

a17r:


acgessler:
Thank you very much for the contribution - merging!

PR #1972, opened 2018-05-19
Fix #1970 STL Ascii, failed managing "empty solid" 

### LINKED ISSUE #1970: STL Ascii, failed managing "empty solid"
Hi,
(assimp is great, thanks for the work)
Here is a small bug.
With a STL ascii, assimp can not read object containing empty solids
The problem have been found exporting from TopSolid, some parts seem empty.

solid 4
endsolid

Manually removing theses "empty solid" is ok and assimp can read the file.

aavenel:
Should fix issue #1970.

Also add more info to assimp_cmd when ValidateDataStructure fails.

### DIFF
--- a/code/STLLoader.cpp
+++ b/code/STLLoader.cpp
@@ -352,7 +352,7 @@ void STLImporter::LoadASCIIFile( aiNode *root ) {
 
         if (positionBuffer.empty())    {
             pMesh->mNumFaces = 0;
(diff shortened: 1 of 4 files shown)


kimkulling:
Looks good, fix the bahaviour that assimp crashes even for small asset-issues. I will merge it.

kimkulling:

PR #1975, opened 2018-05-21
Fix #1587 : add validation to LWS unit test

### LINKED ISSUE #1587: models/LWS/move_x.lws fails validation
```
assimp info models/LWS/move_x.lws 
ERROR: Failed to load file: Validation failed: aiMesh::mFaces[0].mNumIndices is 0
```

This file is used as part of the unit test suite. Adding validation to `test/unit/utLWSImportExport.cpp` triggers test failure. The test probably doesn't test what it should.

aavenel:
It seems that the validation is now OK.
Also add a bunch of tests to try to import each file we have in this
format.


kimkulling:
Merged, thanks a lot for the fix and contribution.

PR #1977, opened 2018-05-22
Fix GenVertexNormals

AdrianAtGoogle:
previous comparison check was failing for very small smoothing angles (fLimit approaching 1.0) because v dot v not guaranteed to be >= 1.0 for all unit vectors


kimkulling:
Merge, thanks a lot for the fix.

Issue #1932, opened 2018-05-01
glTF/2.0 files without a scene index do not load


gstanlo:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SimpleMeshes/glTF

This file does not load because it does not declare a "scene" property. See:
https://github.com/assimp/assimp/blob/bdbad4c64fc618fe3e078e08c5b565bc15e16974/code/glTF2Asset.inl#L1191

The spec says that if no scene is declared, nothing needs to be rendered, but the resulting behavior is that nothing ends up loaded at all and the import ultimately returns a null pointer.

PR #1978, opened 2018-05-23
glTF/2.0: Pick scene 0 as scene to load if no "scene" property is specified

gstanlo:
re: https://github.com/assimp/assimp/issues/1932

This assumes scene 0 if none specified. I believe this behavior is consistent with online viewers.


kimkulling:

PR #1979, opened 2018-05-24
Fix STL unittest in Debug mode

TinyTinni:
Hi,
in [STLLoader.cc Line 353](https://github.com/assimp/assimp/blob/master/code/STLLoader.cpp#L353),
the loader checks, if there is data or not.
Everything is fine, when there is data, but when there is no data, the code block continues and executes the memcpy later in line 368. This is not a real problem, since 0 Bytes are copied, but
we also access memory which is out of the containers bound in the same line via `&positionBuffer[0].x`.

Some Debuggers, i.e. VS in Debug mode, check, whether the program accesses out-of-bound memory or not and throw an exception, if so.
This results in a failed unittest (utSTLImporterExporter.test_with_empty_solid) using a debug build.

This PR uses `std::vector::data()` instead of `&positionBuffer[0].x`, trying to have a solution with minimal changes.


kimkulling:

PR #1980, opened 2018-05-25
Properly reads in glTF/2.0 sampler address modes.

gstanlo:
Assimp was returning glTF/2.0 values as address modes instead of aiTextureMapModes.
Also modified text glTF/2.0 model's sampler uv address modes to mirror/clamp respectively, and tests for them in the unit test.


kimkulling:

PR #1985, opened 2018-05-28
[pyassimp] Minor changes to setup.py to make it slightly more compliant

severin-lemaignan:
While here, I've updated the version available on pypi: https://pypi.org/project/pyassimp/4.1.2/


kimkulling:
Great, merged. Thanks a lot for your contribution.

PR #1988, opened 2018-05-30
FBX: LayeredTextures now work with embedded texture data

kebby:
The code for converting an embedded texture was only in the path for FBX Textures, not LayeredTextures, so files using the latter were missing all embedded images. Factored that code out into a new function and made both paths call it.


kimkulling:
Great, looks fine as far as I can say. I just fixed one minor check.

kimkulling:
Merged

PR #1991, opened 2018-06-01
 FBX animation export

microm8:
Assimp node animation time is already in seconds so I guess it should not be converted from ticks.


kimkulling:
Looks good, I will merge it!

PR #2005, opened 2018-06-09
issue_1973

CwTCwT:
support for multiple (texture) UV mappings.
unit test added.
unit tests ok in 32/64bit.

additionally corrected a bug in STLLoader (crashes on empty model due invalid index usage)


CwTCwT:
finally it passed all test... hard work.
and its amazing how 'lazy' VS2015 toolset was (i use VS2017 with VS2015 toolset).
some findings of travis are 'hard', like a pragma once in cpp as error.
others lead me to some new concepts of newer standards, like subobject initialization.
never stop learning :-D

kimkulling:
Great work, thanks a lot for the contribution!

PR #2013, opened 2018-06-13
fix build for independent build dir (mingw)

cloudwu:
When I generate makefile to a independent build dir by cmake like this
```
cd assimp
mkdir build
cd build
cmake -G "MSYS Makefiles" ..
make
```
It can't build . The error message is
```
E:/assimp/tools/assimp_cmd/assimp_cmd.rc:4:10: fatal error: ../../revision.h: No such file or directory
 #include "../../revision.h"
          ^~~~~~~~~~~~~~~~~~
```
It seems that windres.exe use the source file path rather than current path, so we should add binary directory to include directories.


kimkulling:

PR #2018, opened 2018-06-15
Added check for NULL Compound in Properties70 element (fixes DeadlyImportError on some FBX files)

maxvollmer:
Some FBX files have Null/LimbNode attributes with an empty Properties70 element, where the Element is not NULL, but it's Compound is. The code only checked if the Element itself is NULL, thus causing a DeadlyImportError when trying to instantiate a PropertyTable for an Element that doesn't have a Compound.


kimkulling:
Nice finding, thanks a lot for the contribution. I will merge it!

PR #2020, opened 2018-06-16
Hide commit signature information when fetching commit hash

pierremoreau:
If someone configures git to always show a commit’s signing information,
this breaks the definition of GitVersion in revision.h:

```
#ifndef ASSIMP_REVISION_H_INC
#define ASSIMP_REVISION_H_INC

#define GitVersion 0xgpg: Signature faite le Fri 15 Jun 2018 20:39:53 CEST
gpg:                avec la clef RSA 4AEE18F83AFDEB23
gpg: Impossible de vérifier la signature : Pas de clef publique
878b4b2c
#define GitBranch "master"

#endif // ASSIMP_REVISION_H_INC
```

This can be avoided by passing the `--no-show-signature` to the `git
log` command for retrieving the hash.


kimkulling:
Finally merged, thanks a lot for the contribution.

PR #2022, opened 2018-06-16
issue_2016

CwTCwT:
only add material uv mappings if set, ignore when no uvmapping set
removed unneccessary pad_i1 field (to prevent unnecessary field read exceptions)


kimkulling:

PR #2037, opened 2018-06-23
Fix fbx import regression

aavenel:
Hi Kim,
I noticed that the fbx unit test box.fbx wasn't checking the number of meshes or vertices.
When I added this check, it appeared that this was broken and assimp reported 0 mesh.

Using git bisect, it seems that the regression appeared with PR #1735.

I'm not sure if I understand this code completely, but it seems that the check for `HasElement` has been removed when `isDirect` is true, which doesn't seem intended.

### DIFF
--- a/code/FBXMeshGeometry.cpp
+++ b/code/FBXMeshGeometry.cpp
@@ -437,6 +437,9 @@ void ResolveVertexDataArray(std::vector<T>& data_out, const Scope& source,
     // deal with this more elegantly and with less redundancy, but right
     // now it seems unavoidable.
     if (MappingInformationType == "ByVertice" && isDirect) {
+        if (!HasElement(source, indexDataElementName)) {
+            return;
+        }
         std::vector<T> tempData;
 		ParseVectorDataArray(tempData, GetRequiredElement(source, dataElementName));
 
--- a/test/unit/utFBXImporterExporter.cpp
+++ b/test/unit/utFBXImporterExporter.cpp
@@ -70,6 +70,10 @@ TEST_F( utFBXImporterExporter, importBareBoxWithoutColorsAndTextureCoords ) {
     Assimp::Importer importer;
     const aiScene *scene = importer.ReadFile( ASSIMP_TEST_MODELS_DIR "/FBX/box.fbx", aiProcess_ValidateDataStructure );
     EXPECT_NE( nullptr, scene );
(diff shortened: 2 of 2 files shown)


kimkulling:
Oh, right. Thanks a lot for figuring this out. I will merge it.

kimkulling:

LoicFr:
Hi Alexandre,

I have 2 FBX files coming from 2 different sources and the normals are not imported in Assimp because of this commit. Are you sure that the test you added is necessary ? It seems to be an error because it checked the presence of indexes but it is in direct mode and indexes are only used in indexToDirect mode.

If I comment the test, it works well, even for the fbx unit test box.fbx
Do you think it is ok to make a pull request to cancel your commit ?

Loïc

aavenel:
Hi Loïc,

This commit was made in order to solve a regression, and I was not very confident of this change. 
If you have seen some problem, feel free to make all the changes that you think are necessary!

It'll be perfect if you can add a check to see if normals are correctly imported on unit-tests.

LoicFr:
Ok, thank you for your quick answer. I will make a pull request for that.
Unfortunately this modification has no impact on the already existing fbx files used in unit tests, and I don't have the right to share my fbx files, so I can't add/modify the unit tests to handle this specific issue.

LoicFr:
Due to some building problems I didn't see that my modification has an impact on the unit test. I will investigate...

LoicFr:
Ok I found the problem: the check is necessary but was wrong.
It was HasElement(source, indexDataElementName)
but has to be HasElement(source, dataElementName)

I will create a pull request for that

aavenel:
nice catch !
Perhaps you will need to add the same kind of code ```HasElement(...)``` to the block of code below ```else if (MappingInformationType == "ByVertice" && isIndexToDirect) {```

LoicFr:
The code above already makes this test:
```
// fall-back to direct data if there is no index data element
    if ( isIndexToDirect && !HasElement( source, indexDataElementName ) ) {
        isDirect = true;
        isIndexToDirect = false;
    }
```

but maybe your are right for the block of code below:
`else if (MappingInformationType == "ByPolygonVertex" && isDirect)`
but I don't have the good data to test it. An exception is raised in case of problem, so I think somebody will find it easily if it doesn't work

PR #2039, opened 2018-06-25
Fixed some gltf files being detected as OBJ

maxvollmer:
Added check to BaseImporter::SearchFileHeaderForToken making sure that a detected token is not in fact just a fraction of a longer token. Microsoft exported binary gltf files were detected as OBJ, because the "gltf " in the string "Microsoft GLTF Exporter 2.4.1.7" was detected as the token "f ". I added a new bool parameter to the method enabling this check. It's default false, and only ObjFileImporter sets it to true, so no other code should be affected.


kimkulling:
Looks good. Do you have a small model for a regression test? I will merge the pull request by the way.

maxvollmer:
@kimkulling I don't have any models I can share unfortunately. If possible, you can use models from here for testing: https://www.remix3d.com It's possible to export the models as gltf using Paint 3D under Windows. Otherwise, if you don't have Paint 3D, but have a model for me, I could load it in Paint 3D, export it, and send it back to you.

PR #2044, opened 2018-06-29
Keep small triangles in triangulation

loebl:
This patch changes two things:
1. Possible small triangles generated during triangulation are not rejected (in that post processing step). FindDegenerates can be used for that.
2. Change order of post processing steps, so that FindDegenerates runs after triangulation

This came up as a problem with one of our models, which had rather detailed geometry based on polygons. I also noted that Triangulation and FindDegenerates use different thresholds to reject triangles - Triangulation uses 1e-5 and FindDegenerates 1e-6.


kimkulling:
Sounds like a good idea. I will merge it.

PR #2045, opened 2018-07-01
Enhanced iOS build script + default build type to "Release" + Very minor update

Norod:
Build script

* Auto detect IOS_SDK_VERSION
* Also support building shared libraries (default remains static)
* Autodetect XCODE_ROOT_DIR
* Auto-assign CMAKE_C_COMPILER, CMAKE_CXX_COMPILER and pass it to cmake
* Added optional -d build script parameter for Debug builds

Toolchain cmake files

* Removed deprecated CMAKE_FORCE_C_COMPILER and CMAKE_FORCE_CXX_COMPILER
* Assign SDKVER and DEVROOT using values found at the shell script stage
* Set a CMAKE parameter called IOS as TRUE to be checked against later
* Set IOS_SDK_DEVICE as iPhoneOS or iPhoneSimulator, based on architecture 

CMakeLists

* If building for iOS in Debug, use -Og rather than -O3
* Don't attempt to include qt_viewer in iOS builds

Other / Minor
* minor addition to .gitignore
* TriangulateProcess - Check for NULL


kimkulling:

PR #2048, opened 2018-07-04
Linking fixes when building local minizip

brendanmig:
When building minizip locally assimp was also linking against the system zlib which is redundant and can cause issues if the target system does not contain a compatible zlib (which is why we want to use the local minizip). This patch prevents pkgconfig from looking for zip if building minizip.


brendanmig:
pull request was made off a much older master. going to remake it from a new, up to date branch.

PR #2052, opened 2018-07-05
Feature/join vertices processor kill unused vertices

Nimer-88:
Makes sure we don't add vertices that aren't ever referenced. I stumbled upon such vertices when loading some FBX files, thus the work-around.


kimkulling:
Merged, thanks for the contribution

PR #2053, opened 2018-07-08
Do not link to system zlib/minizip when building zlib from source

brendanmig:
We build assimp, and the application that uses it, using gcc 6 and the '-static-libgcc -static-libstdc++' options. This allows us to use modern C++ but still run on older gcc 4 based systems (like CentOS 6). This also means we need to minimise the system libraries we link to so we build with ASSIMP_BUILD_ZLIB to use the contrib zlib and link it in statically. However, when we do so the linker still uses -lz and -lminizip which makes the system zlib and minizip a dependency.

This patch simply excludes the minizip pkgconfig if building zlib locally so it also uses the contrib minizip instead.


kimkulling:

brendanmig:
No problem! Looking forward to it coming out in the next release

PR #2058, opened 2018-07-13
FBXImporter: Fix GetUniqueName to return names properly

thothonegan:
`FBXImporter` has a function called `GetUniqueName` to try to generate new names for nodes if duplicates are found. Unfortunately instead, it just ends up naming nodes "" (empty) and makes a mess of the tree (transforms tend to get on the empty nodes, which makes things like skeletal animation completely broken). This fixes that and some other bugs with the function, allowing it to generate names properly again.

Example assimp tree before this patch ([empty] was an empty-string named node):
```
Tree: RootNode
├─  - [empty]
│   ├─ Armature
│   │   ├─ [empty]
│   │   │   ├─ RootBone
│   │   │   │   ├─ [empty]
│   │   │   │   │   ├─ Bone.PosPosPos
├─  - [empty]
│   ├─ Cube
├─  - [empty]
│   ├─ Lamp
├─  - [empty]
│   ├─ Camera
```

After
```
Tree: RootNode
├─ Armature
│   ├─ RootBone
│   │   ├─ Bone.PosPosPos
├─ Cube
├─ Lamp
├─ Camera
```

### DIFF
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -411,7 +411,7 @@ void Converter::GetUniqueName( const std::string &name, std::string &uniqueName 
     }
 
     int i( 0 );
-    std::string newName (name);
+    std::string newName( name );
     while ( HasName( mNodeNames, newName ) ) {
         ++i;
         newName.clear();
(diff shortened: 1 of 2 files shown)


kimkulling:
I Gruss this was my distale, Thanks for the fix! I will merged it.

kimkulling:
The CI seems to be broken, fuXXX.

kimkulling:
I reopened a pull request for merging this. Travis has an error with this pull-request. I don't know why. See https://github.com/assimp/assimp/pull/2060

PR #2062, opened 2018-07-19
Add support of 32-bit indices for the glTF2 export

LoicFr:
Hello,

The glTF2 exporter forced the large meshes to be split in small parts (max 65 535 faces/vertices) to use 16-bit indices. I think it is better to allow 32-bit indices because it is possible in the specifications (https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#indicescomponenttype-white_check_mark) and if the user wants the large meshes to be split it is still possible to export with the preprocessing option aiProcess_SplitLargeMeshes.

Forcing 16-bit indices is not efficient for large meshes when the 3D engine can handle 32-bit indices (ex : Unity 2017.3) and the glTF2 exporter in Assimp can be very long when there are a lot of meshes (which is the case when a big mesh is split into many small ones) due to the time complexity in O(n²) of this function called for each mesh:
inline void Buffer::Grow(size_t amount)
{
    if (amount <= 0) return;
    uint8_t* b = new uint8_t[byteLength + amount];
    if (mData) memcpy(b, mData.get(), byteLength);
    mData.reset(b, std::default_delete<uint8_t[]>());
    byteLength += amount;
}

That is why I propose to use 32-bit indices by default and to not force mesh splitting. Do you think it is ok to do that ?

Loïc

### DIFF
--- a/code/glTF2Exporter.cpp
+++ b/code/glTF2Exporter.cpp
@@ -94,19 +94,7 @@ glTF2Exporter::glTF2Exporter(const char* filename, IOSystem* pIOSystem, const ai
     , mIOSystem(pIOSystem)
     , mProperties(pProperties)
 {
-    aiScene* sceneCopy_tmp;
-    SceneCombiner::CopyScene(&sceneCopy_tmp, pScene);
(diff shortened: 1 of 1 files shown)


kimkulling:

Issue #1724, opened 2018-01-20
OBJ loader returns non-white default color 


marek-at-work:
I am loading an OBJ form a location that does not have mtl file, but has PNG texture with same name as the OBJ, like `model.obj`, `model.png`. Assimp actually picks up this texture as a material which is great!

The problem is that at the same time the material's diffuse color is set to `(0.6, 0.6, 0.6, 1.0)` and this is a problem since in our setup we multiply color of model with color of texture and it makes all the models darker.

In this case the `ai_mat->Get(AI_MATKEY_COLOR_DIFFUSE, color)` should return `AI_FAILURE`, instead of arbitrary default color. Or, the default color should be settable and white by default. Now, it is not possible to distinguish between no diffuse color and color equal to (0.6, 0.6, 0.6, 1.0).

Other color types such as ambient and missive are also defined and set to (0, 0, 0, 1).

I have found some Collada loader code that has these defaults: https://github.com/assimp/assimp/blob/f0bed87e0787c9f3bcd654ae84e990889f7e59c9/code/ColladaHelper.h#L559

mvidiassov:
What is the name of the ai_mat?
There is AI_DEFAULT_MATERIAL_NAME constant, that looks like it is intended to help detect the case of a default material being used when there is no material information in the imported file.

MarekAtGoogle:
Thanks for the answer. The material name seems to be the name of model/texture, in my example it would be just `model`. Since the material has a texture I think it cannot be simply default material.

If there is a way of setting default material color I would love to know how!

kimkulling:
Ok, thanks for the report. Would be an adaption to be able to define a default value for the default material enought?

MarekAtGoogle:
Being able to configure default color would certainly solve the problem I am having. In my case I would like to have white color as default.

In general case, it would be nice to actually know whether any color was specified or not. It seems that the current API already supports this by returning non-ok status from the `Get` function. To help with  backwards compatibility, one could still set a default color even if `Get` returns `AI_FAILURE` so people who do not check the return value would not see any difference in behavior.

Thanks!

kimkulling:
I will adapt the material interface like this:

`
// ---------------------------------------------------------------------------
/** @brief  Will set the default material.
 *
 *  @param[in] mat Pointer to the default material. May not be NULL
 *  @return AI_SUCCESS on success, otherwise something else. Have fun.
 */
// ---------------------------------------------------------------------------
[...]

PR #2069, opened 2018-07-25
closes https://github.com/assimp/assimp/issues/1724: add default mate…

kimkulling:
…rial access to the material API.

### DIFF
--- a/code/MaterialSystem.cpp
+++ b/code/MaterialSystem.cpp
@@ -63,9 +63,9 @@ aiReturn aiGetMaterialProperty(const aiMaterial* pMat,
     unsigned int index,
     const aiMaterialProperty** pPropOut)
 {
-    ai_assert (pMat != NULL);
-    ai_assert (pKey != NULL);
-    ai_assert (pPropOut != NULL);
+    ai_assert( pMat != NULL );
+    ai_assert( pKey != NULL );
+    ai_assert( pPropOut != NULL );
 
     /*  Just search for a property with exactly this name ..
      *  could be improved by hashing, but it's possibly
@@ -76,7 +76,7 @@ aiReturn aiGetMaterialProperty(const aiMaterial* pMat,
 
         if (prop /* just for safety ... */
             && 0 == strcmp( prop->mKey.data, pKey )
-            && (UINT_MAX == type  || prop->mSemantic == type) /* UINT_MAX is a wildcard, but this is undocumented :-) */
+            && (UINT_MAX == type  || prop->mSemantic == type) /* UINT_MAX is a wild-card, but this is undocumented :-) */
             && (UINT_MAX == index || prop->mIndex == index))
         {
             *pPropOut = pMat->mProperties[i];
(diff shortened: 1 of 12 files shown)


JekaS178RUS:
Could you explain a usage scenario a littlebit?

I expect it to be like setter method setDefaultMaterial( aiMaterial* material ).
Instead it's looks like I should call aiCreateAndRegisterDefaultMaterial to get/create material instance, set my parameters to it and then call Importer::read.... 
Also in Importer dtor I see reseting DefaultMaterial to nullptr, but not deleting. When and where it will be deleted?

If I will use different Importer instances will I force to fill default material every time instead of reusing signle instance?

code/STLLoader.cpp
```diff
@@ -211,10 +214,11 @@ void STLImporter::InternReadFile( const std::string& pFile, aiScene* pScene, IOS
 
     // create a single default material, using a white diffuse color for consistency with
     // other geometric types (e.g., PLY).
-    aiMaterial* pcMat = new aiMaterial();
+    aiMaterial* pcMat = aiCreateAndRegisterDefaultMaterial();
```
turol: I think this is wrong. The code assumes that the pointer it puts into materials is unique and owned by the scene. Later in the destructor `~aiScene` in Version.cpp it tries to free it, leading to double free and use-after-free bugs when using multiple (entirely separate) importers.

code/MaterialSystem.cpp
```diff
@@ -381,15 +383,37 @@ aiReturn aiGetMaterialTexture(const C_STRUCT aiMaterial* mat,
     if (flags){
         aiGetMaterialInteger(mat,AI_MATKEY_TEXFLAGS(type,index),(int*)flags);
     }
+
     return AI_SUCCESS;
 }
 
+static aiMaterial *DefaultMaterial = nullptr;
```
turol: This will cause bad concurrency bugs when using multiple independent importers.

code/STLLoader.cpp
JekaS178RUS: What about call to aiReleaseDefaultMaterial() in void Importer::FreeScene( ) and Importer dtor?

turol: That only NULLs out the DefaultMaterial, it doesn't free it or break the link in scene `mMaterials`.

While this may seem like it works in single thread it breaks badly in multithreaded scenarios where the previous code worked. We can easily end up with multiple different scenes pointing at the same material and attempting to free it.

JekaS178RUS: Yes, now I see, if I will create 2 importers they will use same material twice. Or 2 scenes created from same importer will also use same material. And aiScene is responsible for deleting material?

turol: Yes. `DefaultMaterial` should be in the `Importer` and managed by it instead of `MaterialSystem`.

JekaS178RUS: But what about Importer::GetOrphanedScene in such case?

code/MaterialSystem.cpp
kimkulling: Indeed

PR #2071, opened 2018-07-26
Use a more accurate way of checking if the build is 64 bit

Maxqia:
My 64 bit Visual Studio builds were failing because I was missing /bigobj. The Ninja CMake generator I was using doesn't have Win64 in the generator name, so even though I was doing a 64 bit build, the option wasn't getting added.


kimkulling:
Nice feasture, I did not know that this is possible. Thanks a lot for the pull-request.

kimkulling:

PR #2076, opened 2018-07-30
Optimisation of FBX node name uniqueness

LoicFr:
Hello,

When importing a FBX file with a huge scene tree, the node names are checked to be unique with a time complexity in O(n²). This is the most time consuming task for importing such FBX files, so I rewrote this with a time complexity in O(n*log(n)).

Loïc


kimkulling:
Looks good, I will merge it.

kimkulling:
merged, thanks a lot for your contribution.

PR #2081, opened 2018-08-04
Fix #2077 - add a lot of unit-tests for gltf2 importer

### LINKED ISSUE #2077: Segfault when importing GLTF1/2 with triangle strip or triangle fan
gltf1 and gltf2 with Triangle Strip and Triangle Fan primitive generates a segfault.

You can reproduce this issue using models here:
https://github.com/bghgary/glTF-Asset-Generator/tree/master/Output/Mesh_PrimitiveMode

Indeed, we access faces[2].mIndices[2] on the first iteration, which is undefined.
```
                    case PrimitiveMode_TRIANGLE_FAN:
                        nFaces = count - 2;
                        faces = new aiFace[nFaces];
                        SetFace(faces[0], data.GetUInt(0), data.GetUInt(1), data.GetUInt(2));
                        for (unsigned int i = 3; i < count; ++i) {
                            SetFace(faces[i - 2], faces[0].mIndices[0], faces[i - 1].mIndices[2], data.GetUInt(i));
                        }
                        break;
```

aavenel:
Hi Kim,

I added some unit-tests for gltf2 importer, in order to investigate and solve #2077 segfault.

Since this pull request has grown larger than expected, here are a few highlights that need to be reviewed carefully :)

I have added a set of tests file from https://github.com/bghgary/glTF-Asset-Generator
License is MIT, so it seems OK to have it in this repo, but please tell me if this causes any problems.
Only tests files about `PrimitiveMode` have been added, but we could add other tests later.

Removed the `try - catch(...)` on glTF2 importer. This seems to be the only assimp importer that have this, and it seems better to throw exception than hide it.

glTF importer executed a `MakeVerboseFormatProcess` postprocess, which splitted vertices. It seems strange to split vertices in the importer and then merge them on `JoinIdenticalVertices` postprocess.
I removed this bit of code.

glTF1 importer is not yet updated. Let me know if you prefer to add this in this pull request or if this is OK to do this later.


kimkulling:

PR #2083, opened 2018-08-04
Fix .obj displacement texture parsing

eddiejames:
The string for an opacity texture is a substring of the displacement
texture string. Due to the nature of the string comparison in the
material texture parsing, any displacement textures will be incorrectly
assigned as opacity textures. Fix this by simply performing the check
for displacement texture before checking for opacity texture.


kimkulling:

PR #2089, opened 2018-08-08
Support out of order channels in BVH

griffin2000:
The BVH standard states that rotation and position components can be in any order.  But the Assimp BVH reader assumes the will always be ordered with position first, followed by rotation.  

This causes BVHs like the one attached to crash.

This change fixes the issue by using a std::map to ensure correct ordering.

[take1.zip](https://github.com/assimp/assimp/files/2272216/take1.zip)


griffin2000:
Thanks for the GCC @kimkulling. Was about to look at that.

kimkulling:

Issue #254, opened 2014-04-15
Assimp shall import of pointcoulds


kimkulling:
Yes, it would be nice if Assimp could read point clouds. I think this would become a nice feature since point clouds are used more and more with 3d printers etc...
Do you think this is easy to implement in a near future version? Or maybe provide me with a patch at the moment?

Thanks for your assistance!

MarekAtGoogle:
Hi, I am just curious if point clouds are already supported? I was trying to load a ply but I was not able to get the points out in any way.

Thanks!

(As a side effect, option aiProcess_Triangulate will take up to minutes to finish with large point clouds.)

chrisidefix:
I would certainly be interested in getting this to work as well, if it isn't behaving as expected at the moment. PLY has multiple and flexible definitions, though - but I guess we can agree that no face elements shouldn't assume triangle (that's stl). @kimkulling I am going to give this a shot unless you already have something?

kimkulling:
I do not have any implementation at the moment. So if someone wants to go for it: have fun :-).

swomack:
Hi @chrisidefix !

Is there any update regarding this?

kimkulling:
There were some other requests, yes!

chrisidefix:
Hello @swomack - I haven't been working on this and am not planning on
importing / exporting point clouds with assimp - seems like this wouldn't
be difficult to implement, but probably time consuming.

Frooxius:
Hello,

I would  also like to join this issue. Assimp 4.0.1 still seems to import point clouds incorrectly, in several different formats.

I have exported a point cloud from Agisoft PhotoScan in several formats that Assimp supports, but each one of them fails in a different way:

OBJ - Produces generic "Failed to read file"
PLY - Generates nonsensical triangles, rather than point primitives (I believe that's the same issue as originally reported)
DXF - Produces "This file contains no 3D data" message

I have tested each with Blender and it imports them correctly.

If it helps, here are the test files: https://cp.sync.com/dl/390004720#bnybycbr-nrz6nkf7-66m78zgz-b9bc9xpn

It includes some extra formats that the software can export, adding support for those might also be potentially interesting in the future, but tangential to this issue.

kimkulling:
OK, good to know! Have you also tried to disable model validation?

JekaS178RUS:
Any update on this? I am will be very happy to be able read and write point clouds with assimp in OBJ and PLY

JekaS178RUS:
[...]

PR #2094, opened 2018-08-12
Add point cloud support for OBJ importer

JekaS178RUS:
Hello!
I added point cloud support for OBJ format import. Logic is simple: if imported file has only vertex attributes it is point cloud. Such files can be created by MeshLab for example.
New mesh added directly to root node, no child nodes will be created.

Relates to https://github.com/assimp/assimp/issues/254

[points.zip](https://github.com/assimp/assimp/files/2281356/points.zip)


JekaS178RUS:
Damn... Memory leak

kimkulling:
Great, thanks a lot for the contribution! I will merge it.

PR #2099, opened 2018-08-17
Fixes crash when importing invalid glTF/2.0 files

gstanlo:
Skips some glTF/2.0 uv processing if the count of uvs in the attribute stream doesn't match the vertex count.

This happens with some malformed glTF/2.0 files, and the change will allow them to be processed properly.
Without the change, an access violation will occur several lines below if uv count is less than the position count.


gstanlo:
An example of a file that renders in some viewers properly but crashes assimp:
https://poly.google.com/view/52s3JpUSjmX

Note that it renders correctly, and note line 57 of the glTF file in the glTF download link in the third accessor: `"count": 0,`

kimkulling:

PR #2103, opened 2018-08-24
Pyassimp update

vjf:
1) The first commit fixes this problem:

>>> from pyassimp import *
>>> s = load('tetra.obj', processing = None)
>>> print(s.mRootNode)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "python3.5/site-packages/pyassimp/core.py", line 128, in <lambda>
    target.__class__.__str__ = lambda x: x.name
AttributeError: 'LP_Node' object has no attribute 'name'

2) The second commit updates 'structs.py' with missing metadata fields that are in the C++ header files already.


kimkulling:

PR #2114, opened 2018-09-01
Native Python interface for accessing metadata in PyAssimp

wmatyjewicz:
This PR is related only to PyAssimp. It builds Python representation of scene & nodes metadata to make accessing the metadata easier.


kimkulling:

PR #2117, opened 2018-09-03
Fix expensive memory allocation and memory copying occurring in obj…

schuttejoe:
…files with a large number of meshes.

This changes the time it takes to import the Disney Moana Island Scene from around 10 hours to about 2 minutes.


kimkulling:
Great, thanks a lot for the contribution.

PR #2122, opened 2018-09-05
Correct 3MF displaycolor import

loebl:
Was testing out 3mf import and this occurred to me:

The displaycolor property was imported with the values directly converted from hex:
```
#00FF00 -> {0.0, 255.0, 0.0}
```
Whereas other importers use a normalized range:
```
#00FF00 -> {0.0, 1.0, 0.0}
```

Also added option of "#RRGGBB" string in addition to assumed "#RRGGBBAA" string (see chapter 5.1.1 of https://github.com/3MFConsortium/spec_core/blob/master/3MF%20Core%20Specification.md)

### DIFF
--- a/code/D3MFImporter.cpp
+++ b/code/D3MFImporter.cpp
@@ -297,8 +297,9 @@ private:
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


kimkulling:
Looks good, thanks a lot for the contribution.

PR #2133, opened 2018-09-11
FBX import: fix import of normal data + unify node renaming strategy

LoicFr:
Hi,

The normals are not imported for some FBX files because of a check which seems not necessary (checking of unecessary indexed data). The check was added in pull request "Fix fbx import regression" #2037 and it has been discussed that maybe it was a mistake because the unit tests are still functional even without this check.

Also I found that the nodes are renaming to be unique at FBX import. It is done at two different steps in the code but not in the same way, so I propose to unify this. In fact, the node renaming is maybe not a good idea because it can be very time consuming for FBX files with a huge scene tree with many duplicate node names, and it is supported by the FBX specifications to have duplicate names, but I guess if it has been implemented it is for a good reason so I didn't remove it.

Loïc

### DIFF
--- a/code/FBXConverter.cpp
+++ b/code/FBXConverter.cpp
@@ -185,12 +185,8 @@ void FBXConverter::ConvertNodes( uint64_t id, aiNode& parent, const aiMatrix4x4&
                 }
 
                 if ( !name_carrier ) {
-                    NodeNameCache::const_iterator it = mNodeNames.find(original_name);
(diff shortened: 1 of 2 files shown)


LoicFr:
It is very strange: unit tests don't pass with the continuous integration but it passes with no errors on my side. I don't know how to fix this.

LoicFr:
Ok I rebuilt everything and I can see an issue. This pull request must be ignored.

PR #2140, opened 2018-09-13
Fix blend test files

turol:
a8a1ca9894a38164f3111f57fddb887b3549ea60 broke the blend test files. It might have broken something else too, didn't check. This just reverts the previous version of the broken files.


kimkulling:
Thanks a lot for the contribution, merged.

PR #2141, opened 2018-09-14
Updated all of struct.py to be in sync with assimp header files

vjf:


kimkulling:

PR #2143, opened 2018-09-17
Disable strict aliasing optimization

### LINKED ISSUE #1587: models/LWS/move_x.lws fails validation

turol:
Our code is not clean and breaks on later GCCs.

Fixes #1587.


turol:
I'm not entirely sure about the second one. It's required for this to work:

`cmake -DCMAKE_CXX_FLAGS="-O2 -fstrict-aliasing -Wstrict-aliasing=1"`

but is there a better way?

kimkulling:
I do not know anything better. I will merge it. Let's see.

PR #2148, opened 2018-09-20
Allow findDegenerate and SplitLargeMesh to pass point clouds models.

arkeon7:
Otherwise the model is removed or crash


kimkulling:
Merged and build fixed, thanks a lot for the contribution.

PR #2156, opened 2018-09-24
Add unit test for subdivision modifier on Blender importer

aavenel:
Blend file is composed of default cube with subdivision modifier applied
and same cube with subdivision modifier.


aavenel:
Well, UBSAN is not happy and the code doesn't look very safe ...
```
// NOTE: this cast is potentially unsafe by design, so we need to perform type checks before
    // we're allowed to dereference the pointers without risking to crash. We might still be
    // invoking UB btw - we're assuming that the ModifierData member of the respective modifier
    // structures is at offset sizeof(vftable) with no padding.
    const SharedModifierData* cur = static_cast<const SharedModifierData *> ( orig_object.modifiers.first.get() );
for (; cur; cur = static_cast<const SharedModifierData *> ( cur->modifier.next.get() ), ++ful) {
```
I'll try to fix this!

Issue #2138, opened 2018-09-13
glTF 2.0 orthographic cameras cause import errors


ardenpm:
Even though AssImp itself does not support them, if an orthographic camera is defined such as in [this](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Cameras/glTF-Embedded) example, AssImp throws and error.

I tracked this to the Camera::Read() method in glTF2Asset.inl. The call:

`type = MemberOrDefault(obj, "type", Camera::Perspective);`

Is looking for a number in the glTF data since Camera::Perspective is an enum value, since it tries to get a number if fails and always returns the default, if the camera is correctly defined and has an orthographic property with the settings then an import error is thrown.

Working on a pull request for this one. Looks like basically need to use a string instead of an enum to retrieve the value then check the value of the string.

PR #2158, opened 2018-09-26
Merge fix to prevent errors when orthographic cameras encountered in glTF 2.0

ardenpm:
Fixes issue #2138 which caused problems when an orthographic camera was defined in glTF 2.0. Note that this does not add support for orthographic cameras, it just prevents issues when they are encountered.


kimkulling:

PR #2167, opened 2018-10-05
Update AssimpNet redirect

Starnick:
After the exchange on twitter, I noticed that this readme never actually had a link pointing to the bitbucket repository that I maintain (originally the googlecode one, and then a fork of the googlecode one that I didn't know about).

This is very confusing for people who report issues, and I would like them to make sure they can report on a tracker that I keep tabs on.


kimkulling:
Good point, merged. Thanks for the contribution.

PR #2173, opened 2018-10-07
Fix Blender FOV loading (fixes #1854)

### LINKED ISSUE #1854: field of view computation for blender files is off by a factor of 2
https://github.com/assimp/assimp/blob/d81078edddb815031cf6072dc11d98a4db18392f/code/BlenderLoader.cpp#L1150

I think the calculation for the field of view form focal length and sensor width isn't correct. The relationship between the two is d/2f = tan fov/2, so the computation is off by a factor of two.

Gargaj:
Fixes #1854


kimkulling:

PR #2174, opened 2018-10-07
ReadFieldPtrVector should return true when successful

Wasabi375:


kimkulling:

PR #2185, opened 2018-10-17
Make gltf2's roughnessAsShininess matches between importer and exporter.

gongminmin:
In glTF2Exporter, roughnessFactor = 1 - normalizedShininess = 1 - someFunc(sqrt(shininess / 1000)). But in gltf2Importer, roughnessAsShininess = (1 - roughnessFactor) * 1000. No sqr there. So if save a shininess to a gltf2 and load it back, the value changed.

This fix add a square to importer to make sure the consistency.


kimkulling:

Issue #1492, opened 2017-10-13
AI_MATKEY_TWOSIDED always zero when importing Collada file


awefers:
When importing Collada files, aiGetMaterialInteger(pMaterial, AI_MATKEY_TWOSIDED, &isTwoSided) always returns zero instead of the actual value as set in the Collada file via the "double_sided" tag.

I have attached a Collada file with a single plane that is set to double sided.

[plane.dae.zip](https://github.com/assimp/assimp/files/1384034/plane.dae.zip)

kimkulling:
Ok, thanks a lot for the report and the example to reproduce the issue.

awefers:
I noticed that you can retrieve the correct value when using the cpp interface: aiMaterial::Get() with a reference to a bool for the value. Unfortunately the C interface is unable to retrieve the correct value.

smumryak:
It is also incorrect for glTF materials. for example this model: https://sketchfab.com/models/38aedc02c0b2412babdc4d0eac7c6803

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2189

PR #2189, opened 2018-10-21
Fix the wrong results when calling aiGetMaterialInteger on bool type properties, such as AI_MATKEY_TWOSIDED

gongminmin:
It only happens on C interface, not aiMaterial::Get(). If a return buffer length is < sizeof(int32_t), it won't be copied to the output parameter.

Fix issue #1492 .


kimkulling:
Merged, thats a lot for the contribution.

Issue #1983, opened 2018-05-26
glTF2 Importer: animations


corporateshark:
There is a `LazyDict<Animation> animations` member field in `glTF2Asset.h`. However, it is never used in `glTF2Importer`.

PixelClear:
Any one working one this issue?
I am also looking for same fix?

ne0ndrag0n:
Does assimp 4.1.0 support importing glTF2 animations yet? gltf 2.0 export has both channels and samplers fields but no animations are found.

corporateshark:
@ne0ndrag0n Nope, it does not support importing of animations. It does not even read glTF2 animation data into the internal data structs.

kimkulling:
That is bad. So I will mark this as a bug.

corporateshark:
Both skeletal and node animations are necessary.

PixelClear:
@kimkulling Any clues when 5.0 will be out?

EvilPudding:
This seems really easy to implement, is anyone on it or could I look into it?

corporateshark:
@EvilPudding It seems no one is working on it right now. But having it fixed would be just awesome!

PR #2191, opened 2018-10-26
Add support for importing GLTF2 animations.

haroonq:
Refactors the glTF2 internal classes to more closely reflect the structure
of the actual GLTF2 file format. Adds implementations for reading skins
and animations from GLTF2 files into those structures.

Also provides implementations for converting skins and animations from GLTF
into assimp data structures.  Special handling is required for bone weights
since assimp stores vertex-weights-per-bone whereas GLTF2 stores
bone-weights-per-vertex.  Only supports keyframed LINEAR animation data;
STEP and CUBICSPLINE is not currently supported.  Animating morph targets
is also not currently supported.


corporateshark:
#1983

kimkulling:

PR #2208, opened 2018-11-08
added DropFaceNormalsProcess to out vector

daef:
Still untested, but I guess it won't do anything when never referenced...


kimkulling:

And bad, I haven't saw it during my code-review.

Issue #2206, opened 2018-11-07
Add more info for bone validation


sagaceilo:
https://github.com/assimp/assimp/blob/bca41ceba5a76936807ea4d23d14a98444ed89b0/code/ValidateDataStructure.cpp#L494

Usefull would be to add actual bone name

`                    ReportError("aiMesh::mBones[%i] has the same name as "
                        "aiMesh::mBones[%i] -> \"%s\"",i,a,pMesh->mBones[i]->mName.C_Str());`

PR #2211, opened 2018-11-08
closes https://github.com/assimp/assimp/issues/2206: make bone error …

kimkulling:
…in verification more verbose.


kimkulling:
Good point, I just integrated your version. Thanks a lot for the feedback.

PR #2215, opened 2018-11-11
glTF2 importer multiple primitives and 16-bit index buffer skinmesh support.

ruyo:


kimkulling:

PR #2236, opened 2018-11-23
Fix static init ordering issue in 3mf importer

rdb:
Same issue as #1253 but for D3MFImporter.  The 3mf importer has its `aiImporterDesc` completely unnecessarily initialized from the value of an `std::string`, even though `std::string` cannot be initialized at constant initialization time.  This fixes a crash for us.


kimkulling:

PR #2242, opened 2018-11-26
Fix compressed assbin load/export, 3DS loader hardening, ...

mjerabek:
I rediscovered a bunch of fixes I have once done when using assimp for my uni graphics course and rebased them on current master. 

Among them is:
* fix loading/exporting the Assimp binary compressed format
* Exception safety improvements (less leaks on errors), throw exceptions instead of asserts on malformed input, ...

I also played a bit with fuzzing (afl / libfuzzer with asan), which finds problems easily. Some of them were addressed in the patches, but I guess there's a lot left. It is unfortunate that the code has to keep C compatibility and it's impractical to employ C++ smart pointers / vector in the structs, so it's super easy to miss some cleanup, especially when exceptions get involved.

I'm not using assimp any more, but I hope the patches will serve at least as an inspiration.

### DIFF
--- a/code/glTF2Asset.inl
+++ b/code/glTF2Asset.inl
@@ -1357,6 +1357,13 @@ inline void Asset::Load(const std::string& pFile, bool isBinary)
         }
     }
 
+    // Force reading of skins since they're not always directly referenced
+    if (Value* skinsArray = FindArray(doc, "skins")) {
+        for (unsigned int i = 0; i < skinsArray->Size(); ++i) {
+            skins.Retrieve(i);
+        }
+    }
+
     if (Value* animsArray = FindArray(doc, "animations")) {
         for (unsigned int i = 0; i < animsArray->Size(); ++i) {
             animations.Retrieve(i);


kimkulling:
Thanks a lot for the pull request. I will do the review tomorrow.

include/assimp/MemoryIOWrapper.h
```diff
@@ -80,7 +80,8 @@ class MemoryIOStream : public IOStream
     // -------------------------------------------------------------------
     // Read from stream
     size_t Read(void* pvBuffer, size_t pSize, size_t pCount)    {
-        const size_t cnt = std::min(pCount,(length-pos)/pSize),ofs = pSize*cnt;
+        ai_assert(pvBuffer && pSize);
```
kimkulling: Please do not check two statements in one asserrt. Which one will trigger the assert?

include/assimp/color4.h
```diff
@@ -63,8 +63,7 @@ class aiColor4t
     aiColor4t (TReal _r, TReal _g, TReal _b, TReal _a)
         : r(_r), g(_g), b(_b), a(_a) {}
     explicit aiColor4t (TReal _r) : r(_r), g(_r), b(_r), a(_r) {}
-    aiColor4t (const aiColor4t& o)
-        : r(o.r), g(o.g), b(o.b), a(o.a) {}
+    // aiColor4t (const aiColor4t& o) = default;
```
kimkulling: Dead code.

include/assimp/vector2.h
```diff
@@ -69,7 +69,7 @@ class aiVector2t
     aiVector2t () : x(), y() {}
     aiVector2t (TReal _x, TReal _y) : x(_x), y(_y) {}
     explicit aiVector2t (TReal _xyz) : x(_xyz), y(_xyz) {}
-    aiVector2t (const aiVector2t& o) : x(o.x), y(o.y) {}
+    // aiVector2t (const aiVector2t& o) = default;
```
kimkulling: Dead code

include/assimp/vector3.h
```diff
@@ -69,7 +69,7 @@ class aiVector3t
     aiVector3t() AI_NO_EXCEPT : x(), y(), z() {}
     aiVector3t(TReal _x, TReal _y, TReal _z) : x(_x), y(_y), z(_z) {}
     explicit aiVector3t (TReal _xyz ) : x(_xyz), y(_xyz), z(_xyz) {}
-    aiVector3t( const aiVector3t& o ) : x(o.x), y(o.y), z(o.z) {}
+    // aiVector3t( const aiVector3t& o ) = default;
```
kimkulling: Dead code

include/assimp/color4.h
mjerabek: Do you prefer the copy constructor to be explicitly default or implicitly default? It can be either removed or uncommented.

include/assimp/MemoryIOWrapper.h
mjerabek: Both conditions must always pass, otherwise we have undefined behavior (division by zero for pSize and write access to nullptr for pvBuffer). It's just a sanity test.
```suggestion
        ai_assert(pvBuffer);
        ai_assert(pSize);
```

include/assimp/color4.h
turol: I think either C++17 or C++20 will change it so that if we don't explicitly default the copy constructor it will no longer be implicitly emitted. GCC already has a warning about this. We should add either default or explicitly write all the constructors for future proofing.

mjerabek: Ok, I didn't know that. The whole point of this is to make these classes trivially copiable, so I'd go the explicitly defaulted way.
```suggestion
    aiColor4t (const aiColor4t& o) = default;
```

include/assimp/vector2.h
mjerabek:
```suggestion
    aiVector2t (const aiVector2t& o) = default;
```

include/assimp/vector3.h
mjerabek:
```suggestion
    aiVector3t( const aiVector3t& o ) = default;
```

PR #2244, opened 2018-11-27
Calculate the mMaxDuration property based off of keyframes for GLTF2

bubba:
It just uses the longest keyframe (not sure if this is the best way to go about this)


kimkulling:
We will see. Thanks a lot for the contribution.

PR #2249, opened 2018-11-30
Update SMDImporter

wxyu247:


kimkulling:
Merged, thanks a lot for the great work.

PR #2252, opened 2018-12-03
Fix CMake target alias typo

sfalexrog:
Currently the file [code/CMakeLists.txt](https://github.com/assimp/assimp/blob/master/code/CMakeLists.txt#L926) creates an alias target ```assimp::asimp``` for target ```assimp```, whereas ```find_package(assimp CONFIG)``` should (if I understand correctly) provide a target ```assimp::assimp```. This pull request renames the alias target to ```assimp::assimp```.


kimkulling:

PR #2257, opened 2018-12-04
collada export: Use Camera local coordinate system

### LINKED ISSUE #2255: Collada export: camera node must convert from specified mLookAt to (0, 0, -1)
Imported collada files specify mLookAt to be -Z. Other imported files set mLookAt to be another value (usually +Z).

The Collada exporter should use whatever mLookAt is specified on the aiCamera to change the exported node transform to:

(a) take account of any Camera lookat / up vector transforms.
(b) output the node in the correct orientation to match the Collada specification.

e.g:
case A: input Camera mLookAt == (0, 0, -1), pass the node transform through unchanged
case B: input Camera mLookAt == (0, 0, 1), convert the node transform to flip the z direction.
case C: input Camera mLookAt == (1, 0, 0), convert the node transform to rotate +X to -Z

(similarly with mUp)

twhittock:
Fixes #2255

### DIFF
--- a/code/ColladaExporter.cpp
+++ b/code/ColladaExporter.cpp
@@ -1533,7 +1533,23 @@ void ColladaExporter::WriteNode( const aiScene* pScene, aiNode* pNode)
 
     // write transformation - we can directly put the matrix there
     // TODO: (thom) decompose into scale - rot - quad to allow addressing it by animations afterwards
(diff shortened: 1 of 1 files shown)


kimkulling:
Good catch, thanks a lot for the contribution.

Issue #2179, opened 2018-10-11
MinGW build problem: "string table overflow at offset"


JekaS178RUS:
Hello! Recently I tried to build assimp with mingw, again...
And found another issue:

$ cmake .. -G "Ninja" # or "MSYS Makefiles" 
$ ninja

> [ 19%] Building CXX object code/CMakeFiles/assimp.dir/Importer/StepFile/StepFileGen1.cpp.obj
> D:/Programming/tools/msys2/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/7.2.0/../../../../x86_64-w64-mingw32/bin/as.exe: CMakeFiles/assimp.dir/Importer/StepFile/StepFileGen1.cpp.obj: section .xdata$_ZNSt10_Head_baseILy0EPN6Assimp8StepFile29directed_dimensional_locationELb0EE7_M_headERKS4_: string table overflow at offset 10000029
> D:\Programming\tools\msys2\tmp\ccJ7OgCw.s: Assembler messages:
> D:\Programming\tools\msys2\tmp\ccJ7OgCw.s: Fatal error: can't close CMakeFiles/assimp.dir/Importer/StepFile/StepFileGen1.cpp.obj: File too big
> mingw32-make[2]: *** [code/CMakeFiles/assimp.dir/build.make:1431: code/CMakeFiles/assimp.dir/Importer/StepFile/StepFileGen1.cpp.obj] Error 1
> mingw32-make[1]: *** [CMakeFiles/Makefile2:191: code/CMakeFiles/assimp.dir/all] Error 2
> mingw32-make: *** [Makefile:130: all] Error 2
> 
> 
> FAILED: code/CMakeFiles/assimp.dir/Importer/StepFile/StepFileGen1.cpp.obj
[...]

PR #2259, opened 2018-12-04
Use ADD_ASSIMP_IMPORTER for STEP importer

sfalexrog:
Right now the STEP importer is added unconditionally. This PR should allow users to set ASSIMP_BUILD_NO_STEP_IMPORTER to disable building the importer/exporter.

As a side note, building STEP importer alone in debug configuration triggers bug #2179. This should allow at least a partial workaround for those kinds of problems.


kimkulling:

PR #2270, opened 2018-12-15
Add silent flag to assimp cmd

aavenel:
Add a new command line flag to assimp cmd in order to only display minimal
info when loading a file.

This is especially useful for quick profiling/fuzzing.


kimkulling:

PR #2277, opened 2018-12-21
iOS support update

byzyn4ik:
Added makefile for new arm64e  (iPhone XS Max, XS , XR, iPad 2018).
Enabled bitcode support by default on iOS Platform.


kimkulling:

PR #2288, opened 2018-12-30
Fix: except `SyntaxError` for py3 viewer

carasuca:
This was causing `SyntaxError` in python 3.7 (Windows 10)
And should now conform to the new syntax: https://www.python.org/dev/peps/pep-3110/


kimkulling:
Thanks for the pull request, I didn't know this. I will merge it.

kimkulling:

PR #2295, opened 2019-01-08
Add progresshandler support jassimp

kurtzmarc:
Add ProgressHandler support for progress callback while importing models through jassimp.


kimkulling:

PR #2306, opened 2019-01-19
MemoryIOSystem delegates to existing IOSystem

AdrianAtGoogle:
Goal is to support cases where the initial file has already been loaded (hence a desire to use `ReadFileFromMemory`) but there may be other files necessary to resolve.

### DIFF
--- a/code/Importer.cpp
+++ b/code/Importer.cpp
@@ -483,7 +483,7 @@ const aiScene* Importer::ReadFileFromMemory( const void* pBuffer,
     IOSystem* io = pimpl->mIOHandler;
     pimpl->mIOHandler = NULL;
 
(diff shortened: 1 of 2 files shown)


kimkulling:

DDMichael:
Hello, I currently doing a project that use ASSIMP to load obj models on Android platform. As I'm using the Android AssetManager to handle the asset importing, I need to use ReadFileFromMemory to load the content in memory buffer to a aiScene object. I can now successfully load all the vertex, but not the texture. I believe this is caused by not reading the .mtl file properly. Hence it seems like this MemoryIOWrapper will help me getting around. However, I'm not quite sure how to use it. Could you please letting me know how to parse the extra files?

Thanks in advance
Daoming

PR #2312, opened 2019-01-28
Fix glTF2 export with no texture coordinates

LoicFr:
Hello,

The glTF2 export crashed when there were no texture coordinates, so I added a test before reading texture coordinates.
I also fixed a warning (explicit cast from size_t to unsigned int).

Loïc


kimkulling:

PR #2319, opened 2019-02-04
glTF/glTF2 Buffer grow changes and large files support

muxanick:
Buffer grow changes:
The exporting of relatively large data (aprox. 7.2Gb) could take a few days, because of reallocation of a buffer every time for a few new bytes is overkill. I've introduced standard capacity member for the buffer and growth it by 1.5 times every time it needs to grow. That helps a lot, decreases exporting to a minute (from a few days).

Large file support:
glTF is a json file, all lengths and offsets don't have a type, just numbers, but code was always reading it as uint32, this doesn't work for files bigger than int32 (almost all files we have as an example). So, I've changed it to be reading as size_t. Specification doesn't specify the type for it, so it's not against spec.

### DIFF
--- a/code/glTF2Asset.h
+++ b/code/glTF2Asset.h
@@ -430,9 +430,9 @@ namespace glTF2
     struct Accessor : public Object
     {
         Ref<BufferView> bufferView;  //!< The ID of the bufferView. (required)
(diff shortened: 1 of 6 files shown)


kimkulling:
Right, thanks a lot for the fix. Just merged!

Issue #2034, opened 2018-06-21
FBX import: unexpected type code: c


ZimmerA:
I tried to load a fbx file and got the following error:

> ERROR::ASSIMP::FBX-Tokenize (offset 0x3e3) cannot ReadData, unexpected type code: c

From what i understood by looking at the binary tokenizer (https://github.com/assimp/assimp/blob/master/code/FBXBinaryTokenizer.cpp) it checks for uppercase 'C' on line 236. I dont know why this file seems to contain a lowercase 'c'.

aavenel:
Hi ZimmerA,

This seems strange, could you provide the FBX file ?
Thanks!

PR #2326, opened 2019-02-06
Fix for FBX binary tokenization of arrays of type 'c'

MalcolmTyrrell:
This fixes issue #2034 and allows the tokenizer to handle arrays of type 'c'.

We encountered some FBXs which used this type of array and therefore could be loaded. I believe the FBXs were generated from a photogrammetry application called "RealityCapture".

It seems the array is used to carry a thumbnail. In the wild, we've only seen compressed arrays but I synthetically generated an uncompressed array to determine the correct stride value to use in that case.

I've included some synthetically generated FBXs which carry arrays of this type. (These are in a recent FBX version, but assimp can load them successfully.)

Please note that this commit only addresses the tokenization problem. The files load, but no attempt is made to interpret the data.
[db.zip](https://github.com/assimp/assimp/files/2837021/db.zip)


kimkulling:

Issue #2327, opened 2019-02-06
master does not build on clang-7.0.1


lerppana:
Master does not currently build on my build machine:
macos 10.14.3
Clang 7.0.1

seems to be a problem with gltf2importer and rapidjson:
```
In file included from .../assimp/code/glTF2Importer.cpp:61:
In file included from .../assimp/code/glTF2Asset.h:64:
.../assimp/code/../contrib/rapidjson/include/rapidjson/document.h:1314:22: error: call to constructor of 'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >' is ambiguous
        GenericValue v(value);
                     ^ ~~~~~
.../assimp/code/../contrib/rapidjson/include/rapidjson/document.h:1385:16: note: in instantiation of function template specialization 'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >::AddMember<unsigned long>' requested here
        return AddMember(n, value, allocator);
               ^
.../assimp/code/glTF2AssetWriter.inl:100:13: note: in instantiation of function template specialization 'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >::AddMember<unsigned long>' requested here
        obj.AddMember("byteOffset", a.byteOffset, w.mAl);
            ^
... other stuff
```

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2328

PR #2328, opened 2019-02-06
Bugfix issue: #2327 Build broken on clang-7.0.1

lerppana:
This pull request fixes issue #2327 (master was not building on clang-7.0.1)

It seems that calling addMember with second parameter of type size_t is ambiguous, changed to unsigned int


kimkulling:
Merged, thanks a lot for the fixes.

PR #2353, opened 2019-03-01
Fix automatic name assignment for ColladaLoader when using name based assignment

phr34k:
In ColladaLoader in version 3.2.0 (release est. 2015) the name assignment was based on name attribute (xml) with optional fallbacks to id and sid. Over time this was changed to fix a bug and support for names was removed, only to be later (re)added using an optional `useColladaName` setting. We discovered a problem that with when using the name based assignment it doesn't assign automatically generated names like what it did in the logic in v3.2.0 and on the id based approach on current master.

We discovered this causes problems with matching the aiCamera and aiLight. So we rectified this problem by auto-generating names also on the name based method.


kimkulling:
Thanks a lot for the pull request. Merged!

phr34k:
Absolutely, you're welcome thanks for checking it out and to merge it.

PR #2354, opened 2019-03-01
Ensure our include directories get added in the correct order

asmaloney:
If you have assimp installed already and in the include path (e.g. I have it via homebrew), it can pick up the wrong headers.

This forces the include order so our local ones are found first when building assimp.


kimkulling:
Thanks a lot for the pull request. I din't know that keyword BEFORE. I will merge it!

asmaloney:
Thanks Kim. I wasn't aware of `BEFORE` before either. I just found it when trying to solve my issue :-)

My use-case is that I'm including assimp as a git submodule and adding it in cmake using `add_subdirectory()`.

One suggestion to make this cleaner would be to not add certain options (building tests and tools for example) if you are using it this way. i.e. check if we are "included in another project" or "building ourselves" and only show those options in the latter case.

If that sounds reasonable I can make the changes and another PR.

(I'm also a bit confused by the merge above: `Merge branch 'master' into include_fix`. What was merged where? :-) The PR wasn't merged into master.)

kimkulling:
Hi,

at first I merged the latest master into your branch. Afterwards I merged your branch back to master. So now all is on the master. Thanks a lot for the contribution.

Kim

asmaloney:
Ah - I didn't know you could push to my branch... I've never seen anyone do it that way.

Any thoughts on my other suggestion?

PR #2372, opened 2019-03-14
fix: change ScaleProcess priority

Aidmx:
When using aiProcess_PreTransformVertices and aiProcess_GlobalScale  at the same time, you can directly get the desired value without transforming twice.


kimkulling:
Great finding, thanks a lot for firuring this out. I will merge it!

kimkulling:

PR #2376, opened 2019-03-15
fix loading 3D uvs from obj

malytomas:
mNumUVComponents used to be set to 2 irrespective of loaded data. This fix tracks the number of loaded uv coordinates and propagates it.


kimkulling:

Issue #1760, opened 2018-02-01
AMD64 switch breaks builds


MauriceH1:
If you use the "Visual Studio 15 2017 Win64" generator in CMake together with the AMD64 switch you get the following error in when trying to build in Visual Studio:
`4>   Creating library C:/Users/Youkai/Development/Tools/Assimp/build/contrib/zlib/Debug/zlibd.lib and object C:/Users/Youkai/Development/Tools/Assimp/build/contrib/zlib/Debug/zlibd.exp
4>inflate.obj : error LNK2019: unresolved external symbol inflate_fast referenced in function inflate
4>infback.obj : error LNK2001: unresolved external symbol inflate_fast
4>C:\Users\Youkai\Development\Tools\Assimp\build\contrib\zlib\Debug\zlibd.dll : fatal error LNK1120: 1 unresolved externals
4>Done building project "zlib.vcxproj" -- FAILED.`

I've also noticed that AMD64 switch breaks the build when using the "Visual Studio 15 2017" generator, although this shouldn't be too much of an issue as I imagine people won't use the x64 implementation in a 32bit project.

Using the 32bit assembly with a 64bit project works as expected though.

kimkulling:
OK, thanks a lot for the report.

kimkulling:
Shall be fixed now.

asmaloney:
@kimkulling Is it possible this problem has reappeared? You didn't reference your fix, so I can't see what you did.

Someone trying to build one of my plugins using assimp has [run into this](https://github.com/asmaloney/MeshIO/issues/1). (I am not set up on Windows to build any of this, so I can't attack it myself.)

As I mention there, there's an extensive analysis of the issue on [StackOverflow](https://stackoverflow.com/questions/29505121/cmake-zlib-build-on-windows).

PR #2382, opened 2019-03-19
{cmake} Explicitly turn off ASM686 & AMD64 options when ASSIMP_BUILD_ZLIB is on

asmaloney:
The AMD64 option causes a build failure on MSVC (#1760) and the ASM builds seem to have problems:

   https://github.com/madler/zlib/issues/41#issuecomment-125848075

This change also prevents these options from "polluting" the cmake options if assimp is being included as a submodule.


kimkulling:

PR #2395, opened 2019-03-29
added casts for rapidjson in gltf asset writer, converted std::to_string to internal to_string in fbx.

dibenedetto:


kimkulling:

Issue #2393, opened 2019-03-28
FBX: "TransparencyFactor" does not affect the alpha of the material


mlavik1:
On FBX models that have "TransparencyFactor" defined but don't have an "Opacity" property, surfaces that are supposed to be transparent become opaque.
I'm not exactly sure about what the difference between "Opacity" and "TransparencyFactor" is. 
According to [FBXSDK's documentation](http://docs.autodesk.com/FBX/2014/ENU/FBX-SDK-Documentation/cpp_ref/class_fbx_surface_lambert.html#ac47de888da8afd942fcfb6ccfbe28dda) it " is used to make a surface more or less opaque (0 = opaque, 1 = transparent)", which is basically the same as "Opacity".
However, in Assimp it's currently used as a multiplier for the "TransparencyColor"'s RGB-values - and does not affect the alpha value at all. Shouldn't it instead be used to modify the colour's alpha value?

In FBXSDK the "TransparencyColor" is exposed to the user [through this property](http://docs.autodesk.com/FBX/2014/ENU/FBX-SDK-Documentation/index.html?url=cpp_ref/class_fbx_surface_lambert.html,topicNumber=cpp_ref_class_fbx_surface_lambert_htmldea88f87-afee-49de-9ef5-73a4ac9477b2,hash=ac47de888da8afd942fcfb6ccfbe28dda), and you can decide what to do with it yourself.
In our sofware we used it to modify the alpha. Now we are switching from FBXSDK to Assimp and we need the same in Assimp.
**Should we maybe add this as a material porperty?**

The particular model I was having an issue with had a material that looks like this:
![image](https://user-images.githubusercontent.com/6906872/55159862-4c48e000-5162-11e9-8eec-3c5fa9834f90.png)
(I do not have the rights to share the model, so I'm adding a screenshot)
As you can see it's not a standard material, and [Corona's plugin](https://corona-renderer.com/) was used to make it.
This material is supposed to be semi-transparent. However, it does not have an "opacity" property - only "TransparencyFactor".
We tried creating a simple test-FBX with a 50% transparent material, and the result was an FBX material where both "opacity" and "TransparencyFactor" was set to 0.5.

My suggestion is to expose this property to users of Assimp.
I'm not that familiar with FBX though, so I'm hoping for some feedback here.

mlavik1:
[...]

PR #2400, opened 2019-04-03
issue_2393: Added AI_MATKEY_TRANSPARENCYFACTOR

mlavik1:
We are switching from FBXSDK to Assimp. FBXSDK already has a way of fetching the "TransparencyFactor" from a material, which we need to support some models that don't use standard materials (exported from CAD software). We needed the same in Assimp, and there might be other users who need it too. I think it's good to expose this property since it's something FBX SDK already does.

See the description here: https://github.com/assimp/assimp/issues/2393


kimkulling:

Issue #2043, opened 2018-06-29
FBX import - texture UV transform


warkid:
Importing code for FBX file format assumes that UV transform is stored within texture's ModelUVTranslation/ModelUVScaling. 

But FBX exported 3DS MAX 2017, for example, has texmap Offset/Tiling stored in Properties70's Scaling and Translation. See [texmapuvtiling_ascii.FBX.zip](https://github.com/assimp/assimp/files/2148947/texmapuvtiling_ascii.FBX.zip)

Is there a reason why FBX converter chooses to use those parameters instead of Props'?

kimkulling:
I guess the main reason was that the author didn't had any files exported from Maya.

mlavik1:
What is the difference between "ModelUVScaling" and "Scaling"? (and "ModelUVTranslation" vs "Translation")?

mlavik1:
I have done some testing, and as far as I can see "ModelUVScaling" and "ModelUVTranslation" have no effect on 3DS Max. When exporting from 3DS Max these are always set to the default values (scale: 1,1; offset: 0,0). 
So all models I export from 3DS Max get the wrong UV scale when using Assimp. However, when using FBX SDK I can fetch the correct UV scale using FbxTexture::GetUVScaling(). The value returned is the one stored in "ModelUVScaling".

**My guess**: 3DS Max uses "ModelUVScaling" and "ModelUVTranslation", while some other software use "Scaling" and "Translation".

**Suggested solution**: Use "ModelUVScaling" and "ModelUVTranslation" if available - if not, use "Scaling" and "Translation".
Alternatively we can leave it as it is and expose all 4 properties (as _AI_MATKEY_...) and let the user decide what do do in these cases.

kimkulling:
Good point, I merged the pull request. Thanks a lot for figuring this out.

PR #2404, opened 2019-04-04
Suggested fix for issue 2043

mlavik1:
If available, use "Scaling" and "Translation" instead of "ModelUVScaling" and "ModelUVTranslation". This seems to be what 3DS Max and FBX SDK use.

See comment here: https://github.com/assimp/assimp/issues/2043

> I have done some testing, and as far as I can see "ModelUVScaling" and "ModelUVTranslation" have no effect on 3DS Max. When exporting from 3DS Max these are always set to the default values (scale: 1,1; offset: 0,0).
So all models I export from 3DS Max get the wrong UV scale when using Assimp. However, when using FBX SDK I can fetch the correct UV scale using FbxTexture::GetUVScaling(). The value returned is the one stored in "Scaling".

So far I haven't seen any models where  "ModelUVScaling" and "ModelUVTranslation" are used.
Please let me know if you have any information and feedback on this.

Alternatively we can also expose all these properties as (_AI_MATKEY_...).

### DIFF
--- a/code/FBXMaterial.cpp
+++ b/code/FBXMaterial.cpp
@@ -206,6 +206,20 @@ Texture::Texture(uint64_t id, const Element& element, const Document& doc, const
 
     props = GetPropertyTable(doc,"Texture.FbxFileTexture",element,sc);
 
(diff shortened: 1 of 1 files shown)


kimkulling:

Issue #2302, opened 2019-01-17
Assimp::Exporter::ExportToBlob ignores PreProcessing flags and ExportProperties settings


RichardTea:
The PreProcessing and ExportProperties parameters are both ignored:

Exporter.cpp, Line 290:
const aiExportDataBlob* Exporter::ExportToBlob( const aiScene* pScene, const char* pFormatId,
                                                unsigned int, const ExportProperties* /*pProperties*/ ) {

They should both be passed on to Exporter::Export()

RichardTea:
Fixed by the above pull

PR #2412, opened 2019-04-08
Some MSVC/Windows corrections and updates, issue 2302

RichardTea:
Fixes issue #2302 

Nicer human-name for Assimp Binary Importer

Also makes some corrections to the Windows/MSVC platform:
- Embed version info into the DLL
- Set MSVC projects to default to Unicode (recommended since XP/Vista)


RichardTea:
Apologies, the cmakelists changes for Unicode seem to have broken the Qt-based viewer.
Will look into this.

RichardTea:
I cannot build the assimp_view application at all, as it's looking for the DirectX SDK in the wrong place.

Microsoft have moved DirectX directly into the Windows 10 SDKs, it's no longer a separate install.

RichardTea:
I've now done a simple workaround for assimp_viewer, however as the official builds will not run on my Windows 10 PC #2422 I cannot tell whether or not this is a real fix.
Will create a new pull request for this.

PR #2417, opened 2019-04-09
Implement Collada 1.4 scene metadata load and improve save

### LINKED ISSUE #2413: Collada importer ignores author metadata
It ignores all the metadata, however the author and copyright metadata is probably the most important and are set by the Collada Exporter

RichardTea:
Fixes #2413 by adding Collada 1.4 <asset/> metadata to the Scene metadata.

Behaviour Change:
The Collada Exporter now exports the scene metadata if any is present.
If there is no Scene metadata, it uses the Root Node metadata as before.


kimkulling:
Will continue working on different branch because I need ot fix the scripts for innosetup.

PR #2419, opened 2019-04-10
Fix FBX face materials not being properly loaded if the face count mi…

merwaaan:
…smatches the material count

Assimp currently crashes when loading ill-formed FBX files that contain a per-face material that has a mismatch between the face count and the material count. The crash occurs [here](https://github.com/merwaaan/assimp/blob/cea0f5dca9d63a3d604abf89d3fd9be0ca7c4f53/code/FBXConverter.cpp#L1234).

There's actually code to deal with such case when loading the file, in [FBXMeshGeometry](https://github.com/merwaaan/assimp/blob/cea0f5dca9d63a3d604abf89d3fd9be0ca7c4f53/code/FBXMeshGeometry.cpp#L610) (it resizes the material list to match the geometry).
However, there's a bug: the material list that got adjusted is then swapped out by the calling code and never used again, [here](https://github.com/merwaaan/assimp/blob/cea0f5dca9d63a3d604abf89d3fd9be0ca7c4f53/code/FBXMeshGeometry.cpp#L370).

This commit fixes this by adjusting the material list being constructed instead of the "global" material list.


kimkulling:
Merged, thanks a lot for the contribution

Issue #2339, opened 2019-02-15
PyAssimp can't load simple OBJ


treyenelson:
Here's the error trace:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pyassimp/core.py", line 321, in load
    scene = _init(model.contents)
  File "/usr/local/lib/python3.7/site-packages/pyassimp/core.py", line 216, in _init
    call_init(obj, target)
  File "/usr/local/lib/python3.7/site-packages/pyassimp/core.py", line 84, in call_init
    _init(obj,parent=caller)
  File "/usr/local/lib/python3.7/site-packages/pyassimp/core.py", line 212, in _init
    setattr(target, name, obj)
AttributeError: 'bytes' object has no attribute 'aketrans'
```

This happens even on a simple cube:
```
# Blender v2.80 (sub 44) OBJ File: ''
# www.blender.org
mtllib cube.mtl
o Cube
v -1.000000 -1.000000 1.000000
v -1.000000 1.000000 1.000000
v -1.000000 -1.000000 -1.000000
v -1.000000 1.000000 -1.000000
v 1.000000 -1.000000 1.000000
v 1.000000 1.000000 1.000000
v 1.000000 -1.000000 -1.000000
v 1.000000 1.000000 -1.000000
vt 0.375000 0.000000
vt 0.625000 0.000000
vt 0.625000 0.250000
vt 0.375000 0.250000
vt 0.625000 0.500000
vt 0.375000 0.500000
vt 0.625000 0.750000
vt 0.375000 0.750000
vt 0.625000 1.000000
vt 0.375000 1.000000
vt 0.125000 0.500000
vt 0.125000 0.750000
vt 0.875000 0.500000
vt 0.875000 0.750000
vn -1.0000 0.0000 0.0000
vn 0.0000 0.0000 -1.0000
vn 1.0000 0.0000 0.0000
vn 0.0000 0.0000 1.0000
vn 0.0000 -1.0000 0.0000
vn 0.0000 1.0000 0.0000
usemtl None
s off
f 1/1/1 2/2/1 4/3/1 3/4/1
f 3/4/2 4/3/2 8/5/2 7/6/2
f 7/6/3 8/5/3 6/7/3 5/8/3
f 5/8/4 6/7/4 2/9/4 1/10/4
f 3/11/5 7/6/5 5/8/5 1/12/5
[...]

PR #2426, opened 2019-04-17
Fix "bytes aketrans" issues in Python>=3.1

Epsirom:
Issue https://github.com/assimp/assimp/issues/2339

In Python>=3.1, pyassimp.load('filepath') will raise AttributeError: 'bytes' object has no attribute 'aketrans' because the field `mPrivate` is of type `bytes` and Python 3.1 introduced `bytes.maketrans` method (ref: https://docs.python.org/3/whatsnew/3.1.html#other-language-changes ).

Solution: in core.py, modify last condition of `_is_init_type(obj)` from `not isinstance(obj, int)` to `not isinstance(obj, (int, str, bytes))`.


kimkulling:

kimkulling:

Issue #2428, opened 2019-04-17
OBJ files containing "nan" or "infinity" values for vertex texture coords cannot be opened


RichardTea:
Some exporters have been found to do this, notably Vectorworks 2016, 2017, 2018 and 2019.

Attempting to import a file like this results in the fatal error `OBJ: Invalid number of components`

This is caused by `ObjFileParser::getNumComponentsInDataDefinition()`, which doesn't accept NaN or Infinity as being a component value, only using assimp::IsNumeric()

JekaS178RUS:
Yes, I also hit this problem. But what we should do if value is nan?

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2429

PR #2429, opened 2019-04-17
OBJ: Coerce texture coords of nan or infinity to zero

RichardTea:
Proposed fix for #2428


kimkulling:

PR #2430, opened 2019-04-18
[FBX Export] fixed ordering of skin indices and weights, to be consistent between systems

Hugo-Pereira:
Currently, the FBX exporter may generate different files across different systems, because the way the bone influences (skin weights + indices) are ordered relies on the pointer address to the aiNode of each bone. The model representation, visually, will be the same, so this might not be an issue for most.
I simply created a custom comparator to be used in the std::set that stores the references to these nodes, that relies in the name of the nodes instead. I tested on a few different machines, and the only difference on the resulting FBX is now the "CreationTimeStamp".

Please let me know if I should better organize the way I implemented this change :)


kimkulling:
Really good catch. Just merged it and many thanks for the contribution.

PR #2433, opened 2019-04-24
Fix inconsistency between animation and node name in glTF2 Importer

hshakula:
1. Take [animated model](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedTriangle) from gltf repo
2. Import via assimp:
    - rootNode name - "nodes_0"
    - animation channel (which supposeed to animate root node) name - ""
3. In result model is not animated as expected

As I found out the problem was caused by the fact that node and animation names are handled differently. Looks like all previous tests with animated gltf2 objects was done in case when animated node name is not empty


kimkulling:
Thanks a lot for  figuring this out. Also in other importers this issue seems to be there. I will check!

kimkulling:

PR #2434, opened 2019-04-24
Generate attenuation constants if non are provided in the Blender file…

mvorbrodt:
…. Using: https://imdoingitwrong.wordpress.com/2011/01/31/light-attenuation/


kimkulling:
Great work, thanks a lot for the contribution.

PR #2435, opened 2019-04-26
bad OBJ file that can still be read

ov:
I've got a bad OBJ file that has the following structure:

```obj
# vertices
v ...
v ...
v ... 

# normals
vn ...
vn ...
vn ...

# faces
f 1/1 2/2 3/3
f 4/4 5/5 6/6
```

The file obviously misses texture coordinates, but the "face" lines seem addressing normals instead. The parser fails on this by skipping the first digit after the slash, which leads to various harmful consequences...

It is discussable if the second number of each pair should be ignored or treated as a normal index instead of UV index, but the parser may (and so should) read the file at least partially.

This pull request sorts it out by checking if the currently read number should be a UV index, but the mesh has only normals and uses the number as the normal index instead. It passes all the tests and I added another one representing the file I have. Unfortunately I can't provide the file itself because of its size and unknown copyright status.


kimkulling:

PR #2438, opened 2019-04-28
Updated FBXConverter to set material shading mode when possible

b1skit:
Currently, aiMaterial.Get() always fails when attempting to find the shading mode of a material loaded from an FBX file, as the shading mode property is never added by the FBXConverter.

I suspect this is because Autodesk's FBX documentation only mentions Lambert and Phong surface shaders, and Assimp's aiShadingMode enum does not contain an entry for Lambert shading. So, Phong shading is the only time we ever have a perfect match between an FBX's listed shading model and an Assimp aiShadingMode enum value.

The FBXConverter has been updated to set the material shading mode when a Phong shader is detected.
This is the best we can do without adding a "aiShadingMode_Lambert" value to Assimp's aiShadingMode enum, but still improves support and is better than simply never being able to find a valid shading mode.


kimkulling:

PR #2441, opened 2019-04-29
Fix element access operator of aiColor4D

jslee02:
The element access operator of `aiColor4D` doesn't point to the alpha component by index 3. Instead, it points to the red component, which is not the expected behavior I believe. This is because [the operator is missing a case for index 3](https://github.com/assimp/assimp/blob/55cdf1d37c4ecd973c2a2e134e213cf32a772635/include/assimp/color4.inl#L94-L107). This PR fixes it by adding the missing case.


kimkulling:
Ops, thanks a lot for the fix. I will merge it.

kimkulling:

PR #2450, opened 2019-05-09
Fix for crash in StreamWriter::PutString when exporting ASCII FBX

mlavik1:
**The fix:**
Don't call PutString with an empty string. Both DumpChildrenAscii and EndAscii can return without modifyting the string, so we need to check the string before calling PutString.
This used to cause a crash.


kimkulling:

PR #2458, opened 2019-05-11
Set directory and name properties for installing static lib PDB

jackoalan:
These properties allow assimp to be installed as a static library along with its PDB. The dynamic PDB will continue to take precedence unless the builder has disabled BUILD_SHARED_LIBS.


kimkulling:

PR #2460, opened 2019-05-12
Fix Windows build with external zlib

jackoalan:
assimp.rc contains an include with a double-parent header search.
In a default build configuration, this is resolved via the
configuration header path generated for in-tree zlib.

When external zlib is used, this header search path is not provided
to the RC compiler, therefore fails to find revision.h.

This is solved by simply including "revision.h" since the root of
the binary directory is already used as a search path.


kimkulling:

PR #2463, opened 2019-05-14
Degenerate triangles with small area removing fix

muxanickms:
Triangles with small area had never removed, if there is no other type of degenerates `deg` never been updated


kimkulling:

PR #2464, opened 2019-05-14
Exception spam fix for FBXMaterial

muxanickms:
It throws an exception too many times just 3 lines below
`const Token& token = GetRequiredToken(*Content, 0);`
though we can easily avoid it to check if there is any token to process.


kimkulling:

PR #2466, opened 2019-05-14
Fragmented FBX ASCII emdedded resource

muxanickms:
Base64 encoded data in FBX ASCII format could be split into multiple chunks by 512Kb size, when decoding it should be accumulated in one buffer.
Make code little bit faster in terms of allocations, added unit test for it as well.
Sample FBX looks like that:
```
		Filename: "U:/Some/Absolute/Path/paper.png"
		RelativeFilename: "paper.png"
		Content: ,
 "iVBORw0KGgoAAAANSUhEUgAABAAAAAQACAIAAADwf7zUAAAACXB...
 "N7eDUphJ7gxutqfsCDkA+NQ2APoZsmjfiC2KidzqmOGYMzoHZaa....
	}
	Texture: 1826343863344, "Texture::Map #2", "" {
```


kimkulling:

PR #2468, opened 2019-05-14
Support for FBX file sizes more than 4GB

muxanickms:
Well, it's self-explanatory, there were places where uint32 was used, I fixed it


kimkulling:
I guess we need to switch to size_t instead of unsigned int.

kimkulling:

PR #2469, opened 2019-05-14
Node names optimization and fixing non-unique name

muxanickms:
Make sure all nodes in the scene have unique names and if there is a chain for a node it will have same name as a node that the chain belongs to.
If node name is empty it traverses the hierarchy up until it found node with name (RootNode always has a name). For real word examples coming from CAD Systems content creation tools most of the leaf sub-graphs don't have names and it will use name of closest parent.
This also dramatically reduces the run time of the function `FBXConverter::GetUniqueName` because there are not that many duplicated prefixes.
I added unit tests to reflect that change.


kimkulling:
Merged, thanks a lot for the contribution. Great!

muxanickms:
@kimkulling Hey, thank you for accepting my PRs!
May I ask, will they be a part of next Release Candidate? It would be great if they are.

PR #2472, opened 2019-05-17
Add mesh names to imported ASCII STL

bnara:
Currently, `assimp` stores the names, when available, of imported ASCII STL `solid` entries into nodes but not into corresponding meshes.  This PR stores the names into the mesh entries as well.


kimkulling:

PR #2473, opened 2019-05-20
Speed up loading of XML-based formats

RichardTea:
The irrXMLWrapper removes all NUL characters from the loaded data.

This is very slow (especially in debug builds) and is nearly always unnecessary.
This PR replaces the tight loop with std::find(), which is nearly always far better optimised.


aavenel:
Hi!

I was just passing by and came across your commit.
Nice catch! However, I think your commit will introduce quadratic behavior when file contains several NULL characters, which might sometimes lead to a large import time regression.
The original code might be a bit slow due to the data dependency on the inner loop, but it still has a worst case in o(n).
Since using `std::find` will be faster on the best case scenario, maybe you can use `std::find` the first time, and fall back to the original code if you ever find a `\0` character.

RichardTea:
Optimising for illegal files is a pessimisation.
This is a situation where big-O is highly misleading, because both the 'n' and the constant multipliers are very different.

The bytewise search and copy of the original are three or more orders of magnitudes slower than the STL-find and DMA copy of the patch (due to cache invalidation).
The only case where this patch is slower is when the file contains a very large number of NUL characters (hundreds to thousands).
This will not occur for real input.

An XML file containing any NUL characters is not well-formed in the first place, so the 'RFC-correct' behaviour would be to either reject them completely, or crop the input at the first NUL.

kimkulling:

PR #2475, opened 2019-05-21
Revert e45d12: Refers erroneously to ctype instead of ctypes

ruxkor:
PyAssimp's `helper.try_load_functions` should be referring to the [`ctypes`](https://docs.python.org/2.7/library/ctypes.html) module, not `ctype`.


kimkulling:
Why?

kimkulling:

Cgettys:
@ruxkor nice catch - sloppy mistake on my part. Sorry about that

PR #2480, opened 2019-05-21
Parse post process arguments when using info tool

aavenel:
Parse post-process arguments when using `info` in the same way as `export`.


kimkulling:

PR #2484, opened 2019-05-22
Optim FindInstancesProcess

aavenel:
ComputePositionEpsilon() is a costly function which was called in the inner
loop although it only uses data from upper loop.


kimkulling:

PR #2485, opened 2019-05-23
Export static libaries as CMake package

jackoalan:
When built with BUILD_SHARED_LIBS disabled, the exported CMake
targets are modified to reference the static library location.

This also fixes version detection for VS2017 and beyond within
the target import script.


kimkulling:

Issue #1946, opened 2018-05-06
Integrate with Hunter


Binero:
[Hunter](https://github.com/ruslo/hunter) is a package manager for cmake. It automatically downloads dependencies and builds them, while the interface to packages is the same as find_package. 

At the moment there is a [fork](https://github.com/hunter-packages/assimp) of assimp that integrates with hunter, but it is quite dated.

I propose to support hunter in-tree, to prevent it from falling behind in the future. Package managing in C++ sucks (just look at this very project's CMakeLists.txt file), and hunter takes a bit of that pain away.

NukeBird:
+1!

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2488

PR #2488, opened 2019-05-24
Use Hunter for pulling in dependencies

rbsheth:
Assimp has a lot of dependencies, copied into the repo at `contrib/`. The Hunter package manager (https://github.com/ruslo/hunter) helps manage such a dependency tree, as well as potential integration into other projects.

As mentioned in https://github.com/assimp/assimp/issues/1946, there is a fork of Assimp in Hunter but it is quite dated: https://github.com/hunter-packages/assimp.

This PR adds support for the Hunter package manager in Assimp, using it to resolve and pull in dependencies **only when the `HUNTER_ENABLED` CMake argument is set**. By default, `HUNTER_ENABLED` is set to `OFF` and the current CMake behavior is preserved.

This PR does a few things besides adding the requisite Hunter-related CMake files if `HUNTER_ENABLED=ON`:
1) Disable setting of important toolchain-wide flags (like `-std=cxx11` and `-fPIC`) as they will be set externally
2) Download and link against dependent packages through Hunter
3) Define `ASSIMP_USE_HUNTER` and rewrite include paths for dependencies if set
4) Create a Hunter-compatible CMake config file for easy library consumption

What's missing:
Open3DGC - I haven't added this to Hunter and disabled it in Assimp for Hunter builds. For glTF 2.0, perhaps Draco is more widely supported and I have added that to Hunter so it would be pretty easy to add to Assimp.
Testing of tools/samples/tests - I focused on incorporating the library build for use in the package manager.

As a sanity check, I was able to build the library after completely removing the `contrib/` directory.


rbsheth:
@kimkulling Any thoughts on this?

kimkulling:
For me it looks fine as a prototype test, which could get enabled on master. I also played around with Conan, but there aren't any working results out there until now.

Let me do the code review and then I will merge it, if no one else has any concerns.

mlfarrell:
Few notes from myself. Personally, I don’t update my assimp version as much as I should.  Namely because when I do, I have to (re)build it so many times for so many different systems. iOS, Android, Windows store (arm,x86,x64), mac, windows desktop, etc.

 So, I’m on board with a new dependency manager as long as it doesn’t impede my ability to use CMake to build assimp again for the above systems. 

Just my two cents.

kimkulling:
Merged, let's see!

PR #2496, opened 2019-05-30
FBX Importer double precision fix

muxanickms:
It casts to float and loses precision for large coordinates even if ASSIMP_DOUBLE_PRECISION defined


kimkulling:

PR #2501, opened 2019-06-05
glTF importer - clean all member arrays

petrmohelnik:
When importing more than one scene using the same gtlf importer loading fails with error 'bad array new length'. This happens because 'meshOffsets' are not cleared and the importer continues to push_back values at the end. Adjacent values are then used to calculate the length of 'mMeshes' array. This calculation expects that each value is higher than the previous otherwise we get negative length. But when pushin at the end of uncleared array we get contents like this: 0,1,2,3,0,1,2. Then when calculating 0-3 we try to allocate array of length -3 and get this exception.


petrmohelnik:
Why is the Continuous Integration still pending?

kimkulling:
Merged, thanks a lot for the contribution. And sorry for the delay, was too busy.

PR #2502, opened 2019-06-07
Prevent crash when Ogre skeleton file is missing

arkeon7:
When the /skeleton file is missing, the loader crash.
This correct this behavior.


kimkulling:

PR #2508, opened 2019-06-11
fixed MSVC toolset versions >140

pkoller:
according to https://cmake.org/cmake/help/latest/variable/MSVC14.html the use of the MSVC** flags is deprecated, this change uses the encouraged MSVC_VERSION variable.


kimkulling:

PR #2509, opened 2019-06-11
additional changes in cmake template to find correct toolset version

pkoller:


kimkulling:

Issue #2456, opened 2019-05-11
Text formats precision


JekaS178RUS:
Hello!

I am noticed that assimp exporters for text formats (Obj, Dae, maybe some other) uses quite high precision: 16. For float it is redundantly high value, 7-8 would be enogh for my taste. I don't know if assimp uses double somewhere.

My suggestion is to lower this value a bit or make it configurable, in order to produce smaller files. For example my model in OBJ format is 100 Mb, when saved using assimp it becomes around 150 Mb.

One good plase to do so is ExportProperties class like:

```
// Worker function for exporting a scene to Wavefront OBJ. Prototyped and registered in Exporter.cpp
void ExportSceneObj(const char* pFile,IOSystem* pIOSystem, const aiScene* pScene, const ExportProperties* /*pProperties*/) {
    // invoke the exporter
    ObjExporter exporter(pFile, pScene);
    exporter.mOutput.precision( getValueFromProperties(pProperties) );
    ...
```

I didn't find any ExportProperties  documentation or use cases, so I am not sure how to use properly.

kimkulling:
Thanks for the hint, you are right. This could help to reduce the size for exported files.

I guess we do not have a precision export property until now. So we need to create one!

JekaS178RUS:
@kimkulling  I can prepare PR during next week for this

RichardTea:
This doesn't need a property, it should simply key off the size of `ai_real`.
It's either float (default) or double (if the macro ASSIMP_DOUBLE_PRECISION is defined in config.h)

After all, if it's a float internally then there's no point in storing precision that doesn't exist.

The necessary format sizes are:
float:
```printf("%1.8e",d); // always with exponent
printf("%.9g",d); // shortest possible```

double:
```printf("%1.16e",d); // always with exponent
printf("%.9g",d); // shortest possible```

I'd say Assimp should use shortest possible, unless there are formats where stringified floating point are required to have exponents?

JekaS178RUS:
I like @RichardTea approach, it is much simpler.

PR #2517, opened 2019-06-20
Auto-detect text format export precision

RichardTea:
Adds a define to `defs.h` to specify the precision for issue #2456 

Can be overridden if desired


kimkulling:

Issue #2518, opened 2019-06-24
API to determine which Importer was used


RichardTea:
The various importers have a variety of different features, and so applications may wish to apply different post processing depending on which importer was used for the file/in-memory blob.
(Or perhaps attempt to use a different importer should the default-detected one be unable to import.)

However, it is not currently possible to determine which importer was used.

To avoid applications re-writing the format-detection heuristics of all importers, Importer::ReadFile should expose the chosen importer.

To ensure backwards binary compatibility, this could be via an Importer Property, similar to how sourceFilePath is already set.

PR #2519, opened 2019-06-24
API: Add active importer identification

RichardTea:
Issue #2518 

Adds an Import property "importerIndex" that contains the index of the importer that was selected, or -1 if none of the importers matched the provided file/blob.


kimkulling:

PR #2523, opened 2019-06-27
All textures use relative path except embedded textures, this is a fix for it

muxanickms:
`FBXConverter::GetTexturePath` returns RelativeFilename() for all textures (this is stored in the material $tex.name), but `FBXConverter::ConvertVideo` returns FileName() as a texture name, so it's very hard to match both strings in run-time to figure out which embedded texture it is.
I propose to use relative path everywhere, that helps a lot.


kimkulling:

Issue #2525, opened 2019-06-28
Bug leads to LWO Node name repeats


rconde01:
In LWOLoader.cpp LWOImporter::LoadLWO2File there is an iUnnamed local which should be incremented to give unique names to unnamed nodes. However it is defined inside of the loop so each time it is used its value is zero. Therefore all names for empty node names are "Layer_0".

rconde01:
Added PR:

https://github.com/assimp/assimp/pull/2526

PR #2526, opened 2019-06-28
initialize unnamed node uniqueness index outside of loop

rconde01:
Fix for issue:

https://github.com/assimp/assimp/issues/2525


kimkulling:

PR #2536, opened 2019-07-10
Fix Issue: group node in fbx being exported as bone node

thomasbiang:
Issue #2537: using Assimp, simply import an fbx file then export it without any modification, the output fbx file could be different from the original one: group node (being a parent of any bone node) in original fbx is exported as bone node in the output fbx. 

Solution: for any node, if it is not a bone node in the scene, and if its transformation is identity matrix (means: does not contribute to the world transformation computation for its children), these nodes should not be exported as bones.  

Code changes: minimal. Use a set to record all bone names in scene. When exporting, check each node, if its name is not among these bone names and its transformation is identity matrix; then it will not be exported as bone node.


code/FBX/FBXExporter.cpp
```diff
@@ -1617,6 +1617,17 @@ void FBXExporter::WriteObjects ()
     // at the same time we can build a list of all the skeleton nodes,
     // which will be used later to mark them as type "limbNode".
     std::unordered_set<const aiNode*> limbnodes;
+    
+    //actual bone nodes in fbx, without parenting-up
+    std::set<std::string> setAllBoneNamesInScene;
```
turol: Prefer `std::unordered_set` where ordering is not required.

code/FBX/FBXExporter.cpp
```diff
@@ -1617,6 +1617,17 @@ void FBXExporter::WriteObjects ()
...
+    for(int m = 0; m < mScene->mNumMeshes; ++ m)
```
turol: Signed vs unsigned compare warning.

code/FBX/FBXExporter.cpp
```diff
@@ -1617,6 +1617,17 @@ void FBXExporter::WriteObjects ()
...
+    {
+        aiMesh* pMesh = mScene->mMeshes[m];
+        for(int b = 0; b < pMesh->mNumBones; ++ b)
```
turol: Same, signed vs unsigned.

thomasbiang:
Hi, 

I just updated code changes according to the comments. 

Thank you for reviewing

kimkulling:
Finally merged!

Issue #2538, opened 2019-07-10
Long skeletal chain failed be exported 


thomasbiang:
I have a fbx file containing skin mesh bound to a skeleton structure. I tested Assimp with the simplest process: import this fbx then export it without any extra modification. Somehow the export fails and no output fbx is generated. 
<img width="398" alt="Screen Shot 2019-07-10 at 10 23 46 AM" src="https://user-images.githubusercontent.com/37556510/60991200-e9d5b300-a2fe-11e9-95a9-25ea954e5d2d.png">

I traced this problem and find the cause of this issue is: in FBXExporter.cpp, Assimp checks bone node needs to be in bind pose with error threshold 1e-4f, otherwise no fbx is exported. In my case, bones are in bind pose, but due to long skeleton chain of matrix multiplication, the error is more than 1e-4f.
```
const float epsilon = 1e-4f; // some error is to be expected
bool bone_xform_okay = true;
if (b && ! tr.Equal(b->mOffsetMatrix, epsilon)) {
```
I think 1e-4f is okay as a default value, while user should be able to specify customized error rate, otherwise it is too restrictive for certain cases.

PR #2543, opened 2019-07-12
Fix export custom bindpose error

thomasbiang:
This pull request is to fix issue #2538. 

I have a fbx file containing skin mesh bound to a skeleton structure. I tested Assimp with the simplest process: import this fbx then export it without any extra modification. Somehow the export fails and no output fbx is generated.

I traced this problem and find the cause of this issue is: in FBXExporter.cpp, Assimp checks bone node needs to be in bind pose with error threshold 1e-4f, otherwise no fbx is exported. In my case, bones are in bind pose, but due to long skeleton chain of matrix multiplication, the error is more than 1e-4f.

```
const float epsilon = 1e-4f; // some error is to be expected
bool bone_xform_okay = true;
if (b && ! tr.Equal(b->mOffsetMatrix, epsilon)) {
```
To fix it, 1e-4f will be the default value, and user can still customize the error rate through export property "BINDPOSE_EPSILON". Thus for a skeleton with long bone chains, it can still be exported.


kimkulling:
Looks good, I will merge it!

kimkulling:

Issue #2539, opened 2019-07-11
Add support for Collada ZAE packaging


RichardTea:
Collada 1.5.0 mandates support of a new packaging format "ZAE".

This is simply a ZIP containing the Collada DAE file, texture resources and a single "manifest.xml" file.

The manifest file contains an element `<dae_root>%colladafilepath.dae%</dae_root>`, eg:
```
<?xml version="1.0" encoding="utf-8"?>
<dae_root>./some_file.dae</dae_root>
```

Texture paths inside the Collada DAE are relative to the top of the ZIP, not the location of the DAE.

kimkulling:
OK, thanks a lot for the information. Do you have an example file next to you :-) ?

RichardTea:
I've started work on this, hopefully will have a pull request 'soon'.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2545

PR #2545, opened 2019-07-12
Add Collada zae import support

RichardTea:
Resolves feature request #2539

Collada importer will now attempt to load ZAE, and generic zip files that contain a DAE.

Zip archives that do not contain a manifest.xml will instead attempt to load the first DAE found in the zip.
If there is more than one DAE inside then it is not easily predictable which one will be loaded.

API change:
Added a new Assimp::ZipArchiveIOSystem, as this functionality is useful for library users.
This changes the DLL public API, but it is backwards compatible and the new class is pimpl'd to ensure binary compatibility. Currently read-only support.


kimkulling:
Looks good, I will merge it!

kimkulling:

PR #2546, opened 2019-07-13
Some FBX multi-material mesh fixes

Frooxius:
Thank you @kimkulling so much for making the blendshapes work with the multi-material meshes, your work is amazing and has helped us tremendously! I found a few minor issues, but I've been able to fix these myself.

- Fixed anim meshes generated from blendshapes not being copied to output for multi-material meshes
- Fixed first vertex of each blendshape on a multi-material mesh having all unmapped vertice offsets being added to it
- Fixed blendshapes not importing for multi-material FBX meshes with no bones


kimkulling:
Thanks a lot for the great contribution, merged!

Issue #2186, opened 2018-10-18
OBJ file parsing error (2)


rickomax:
Hi.

There is a strange parameter in a method call in 'ObjFileParser' which isn't allowing Assimp to load certain OBJ files.

This is the line:
[https://github.com/assimp/assimp/blob/4d3ce0fed2859576e63c2eb5b67f1e5f2913be89/code/ObjFileParser.cpp#L122](https://github.com/assimp/assimp/blob/4d3ce0fed2859576e63c2eb5b67f1e5f2913be89/code/ObjFileParser.cpp#L122)

When loading OBJ files that have relative paths in 'usemtl' lines, the above code will cause the model to stop loading prematurly, as the relative path may contain the backslash character, like this one:
`usemtl BMCD2AR3\Solid_Materials_\Matte\Normal\Brown_medium`

Can anyone explain why we have that backslash there? Since there isn't such thing defined as token or anything else in any OBJ File Format specification.

rickomax:
@kimkulling I've fixed that on TriLib repo. I'll make a pull-request as soon as I have sorted out some stuff.

jfaust:
This appears to have gotten changed via https://github.com/assimp/assimp/commit/35a044bda37b2bad50a41746622c746acd139631

But it has broken parsing of a large number of OBJ files. The most complete specification I know of (http://paulbourke.net/dataformats/obj/) has many uses of "\" as a line-continuation character.

A model I just ran into has a bunch of lines like this, amongst other uses of "\" as a continuation:
```
vn 0.9999989867210388 -3.294318593957679e-12 \
 0.001427346956916153
vn 0.9999989867210388 -3.294318593957679e-12 \
 0.001427346956916153
vn 0.9999989867210388 -5.843482707261483e-13 \
 0.001427346956916153
vn 0.9999989867210388 -5.843482707261483e-13 \
 0.001427346956916153
vn 0.9999989867210388 -1.939333459446968e-12 \
 0.001427346956916153
vn 0.9999989867210388 -1.939333459446968e-12 \
 0.001427346956916153
```

@rickomax @kimkulling

rickomax:
@jfaust Never seen such thing on OBJ files!
And I work with a lot of these...
We should parse it differently then, as if this character be found on a group name, it will break the parser as it was doing before.

jfaust:
It does seem like a silly thing to have as part of the format. Seems to be mostly designed for matrix data, but at least one exporter appears to use it to limit line lengths.

[...]

PR #2550, opened 2019-07-16
Fix line continuations in OBJ files

merwaaan:
Hi,
This fix to the OBJ parser reverts a faulty line from a January [commit](https://github.com/assimp/assimp/commit/35a044bda37b2bad50a41746622c746acd139631).

Some OBJ files have backslashes for line continuations, like in [this documentation](http://paulbourke.net/dataformats/obj/) or in OBJ files exported from Rhino for instance.

In such cases

`v 0.00000000001 0.00000000002 \`
`0.00000000003`

is equivalent to

`v 0.00000000001 0.00000000002 0.00000000003`

Assimp's OBJ parser dealt successfully with this case via the [getNextDataLine()](https://github.com/assimp/assimp/blob/475e51d157e7826bdcbc0d88e6a73b1764960e0a/include/assimp/IOStreamBuffer.h#L244) method which role is specifically to join lines separated by a continuation character (`\` in our case).

However, its usage was changed in the aforementioned commit. From [this issue](https://github.com/assimp/assimp/issues/2186) by the same author, it seems that they wanted to fix another bug related to material names and changed the continuation character but it actually broke OBJ parsing (the final data is wrong in most cases, and it crashes in other cases).

Tested on various obj files with continuations and now it works well, as before. The unit tests all pass.

### DIFF
--- a/test/unit/utObjImportExport.cpp
+++ b/test/unit/utObjImportExport.cpp
@@ -448,3 +448,37 @@ TEST_F(utObjImportExport, import_without_linend) {
     const aiScene *scene = myImporter.ReadFile(ASSIMP_TEST_MODELS_DIR "/OBJ/box_without_lineending.obj", 0);
     ASSERT_NE(nullptr, scene);
 }
+
+TEST_F(utObjImportExport, import_with_line_continuations) {
+    static const char *ObjModel =
+        "v -0.5 -0.5 0.5\n"
+        "v -0.5 \\\n"
+        "  -0.5 -0.5\n"
+        "v -0.5 \\\n"
+        "   0.5 \\\n"
+        "   -0.5\n"
+        "f 1 2 3\n";
+
+    Assimp::Importer myImporter;
+    const aiScene *scene = myImporter.ReadFileFromMemory(ObjModel, strlen(ObjModel), 0);
+    EXPECT_NE(nullptr, scene);
+
+    EXPECT_EQ(scene->mNumMeshes, 1U);
+    EXPECT_EQ(scene->mMeshes[0]->mNumVertices, 3U);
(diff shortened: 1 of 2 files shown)


turol:
The build is broken but it appears unrelated to your changes. Strange...

Can you also add a testcase for this?

merwaaan:
@turol, yeah I noticed that other PRs also fail some checks, something must be wrong.

I added a unit test, as requested.

test/unit/utObjImportExport.cpp
```diff
@@ -448,3 +448,33 @@ TEST_F(utObjImportExport, import_without_linend) {
...
+    EXPECT_NEAR(vertices[0].z, 0.5f, threshold);
+
+    EXPECT_NEAR(vertices[1].x, -0.5f, threshold);
+    EXPECT_NEAR(vertices[1].y, -0.5f, threshold);
+    EXPECT_NEAR(vertices[1].z, -0.5f, threshold);
+
+    EXPECT_NEAR(vertices[2].x, -0.5f, threshold);
+    EXPECT_NEAR(vertices[2].y, 0.5f, threshold);
+    EXPECT_NEAR(vertices[2].z, -0.5f, threshold);
+}
```
turol: Also add checks for number of vertices and primitives.

merwaaan: Done.

turol:
Looks good to me, now we just need someone to fix the build. @kimkulling ?

kimkulling:
The broken build seems to get caused by a new compiler warning / error. Maybe they updated the clang / g++ version? I will prepare a fix for that!

kimkulling:

Issue #2059, opened 2018-07-16
STL files have broken default material (full white ambient)


dangelog:
Hi,

f84851e8931fb04e7074d487c91c3b3faa566994 refactored the default material provided by the STL importer. For some reason now both the diffuse and the ambient are set to full white. 

Having a full white ambient of course breaks literally *all* appliations using the default materials provided by assimp. 

The commit also changes PLY's default material to the same values, and the STL seems to have changed for consistency reason. This however is inconsistent with OBJ files, whose default material has no ambient, no specular and (0.6, 0.6, 0.6) diffuse.

kimkulling:
Ok, thanks for the report.

Megidd:
Just as a demonstration:

When importing STL file with assimp 3.1: 

![screenshot from 2018-07-21 11-40-14](https://user-images.githubusercontent.com/17475482/43033190-09e6f176-8cdb-11e8-9d31-a366f39717d9.png)

When importing STL file with assimp 4.1:

![screenshot from 2018-07-21 11-44-03](https://user-images.githubusercontent.com/17475482/43033218-80e02072-8cdb-11e8-9ce8-c7b2502297ec.png)

JekaS178RUS:
Is this issue also closed by last [PR](https://github.com/assimp/assimp/pull/2069)

kimkulling:
On the latest master you now can get the default-material via aigetDefaultMaterial() and set the colors as you want to have them. So this issue should be closed as well.

JekaS178RUS:
I think this issue should be reopened. Even with new API aiCreateAndRegisterDefaultMaterial STL loader overwrite any parameters setted previously for that material, which is pointless.

Only usefull usage of aiCreateAndRegisterDefaultMaterial is a way to detect if material is default. But not to use your own parameters.

All because of this peace of code from STLLoader.cpp:
```
   // create a single default material, using a white diffuse color for consistency with
    // other geometric types (e.g., PLY).
    aiMaterial* pcMat = aiCreateAndRegisterDefaultMaterial();
    /*aiMaterial* pcMat = new aiMaterial();
    aiString s;
    s.Set(AI_DEFAULT_MATERIAL_NAME);
    pcMat->AddProperty(&s, AI_MATKEY_NAME);*/

    aiColor4D clrDiffuse(ai_real(1.0),ai_real(1.0),ai_real(1.0),ai_real(1.0));
    if (bMatClr) {
        clrDiffuse = clrColorDefault;
    }
[...]

PR #2563, opened 2019-07-25
Fix white ambient in STL loader

mosra:
A problem that sneaked in from #1293 (f84851e8931fb04e7074d487c91c3b3faa566994) due to insufficient code review, later reported in #2059 but never fixed properly. Having a white ambient practically means all other color information is ignored and the model stays bright white no matter how you set up your lighting. Application code then has to check for this case and patch it during model import (see [here](https://github.com/mosra/magnum-plugins/commit/6444d543d943a7c60e2263f6417517b0b699d231#diff-bd6ab8f40c2db2c93f2d75de9e4997ed), for example). Putting it back to how it was before the commit above so workarounds on app-side can be removed again.

Thanks for merging.


kimkulling:
OK, merged. Now the old behavious shall be there again. Thanks a lt for the contribution.

PR #2564, opened 2019-07-25
Add Export Unit Test output to gitignore

RichardTea:
The unit tests for Exporters create files, those shouldn't get committed to the repo by mistake.

Added a gitignore rule to ignore all files with names ending "_out." in the Test folder, and updated the unit tests that didn't already use that.


kimkulling:

Issue #2557, opened 2019-07-22
Undefined reference to Assimp::ExportAssimp2Json


panquez:
Hi,

If I call 
 ```Assimp::Exporter exporter;```
I got an undefined reference to « Assimp::ExportAssimp2Json(char const*, Assimp::IOSystem*, aiScene const*, Assimp::ExportProperties const*) ».

I configure assimp with these options :
ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT:BOOL=FALSE
ASSIMP_BUILD_OBJ_IMPORTER:BOOL=TRUE
ASSIMP_BUILD_PLY_IMPORTER:BOOL=TRUE
ASSIMP_BUILD_STL_IMPORTER:BOOL=TRUE
ASSIMP_BUILD_ALL_EXPORTERS_BY_DEFAULT:BOOL=FALSE
ASSIMP_BUILD_OBJ_EXPORTER:BOOL=TRUE
ASSIMP_BUILD_PLY_EXPORTER:BOOL=TRUE
ASSIMP_BUILD_STL_EXPORTER:BOOL=TRUE

Is there something I've missed.

Thanks for helping

panquez:
For information if a comment l. 105 and 187 in file code/Common/Exporter.cpp (those lines correspond to prototype of ExportAssimp2Json(...)), the linking issue does not here anymore.

PR #2566, opened 2019-07-25
Test Exporter Descriptions and IDs

RichardTea:
Add unit test to ensure all registered Exporters have a description that makes sense

Also changes the Exporter ID of Assimp JSON format from "json" to "assjson", as json is a rather common extension that is almost certainly going to be used by something else.


kimkulling:
Good catch, I was unsure if I shall use assjson or just json. So I started with json. Makes sense to swithc to assimpjson from my point of view.

RichardTea:
Just realised that this reverted panquez' fix for issue #2557 
I've now done the consistent fix for that issue by capitalising in the CMakeLists

kimkulling:

Issue #2571, opened 2019-07-25
Multiple UV sets export problem, only one exported


thomasbiang:
I have a scene with mesh that has two uv sets. When loaded using aiImport, it shows correctly. Somehow when I use aiExport and re-import again use aiImport, there is one uv map left. 

I think the aiImport is working correctly as originally the two uv sets are all imported. But after aiExport, one uv gets lost.

kimkulling:
Alrerady fixed.

PR #2572, opened 2019-07-25
Fix issue # 2571: extra layer for multi uv sets

thomasbiang:
This pull request is to fix issue 2571 https://github.com/assimp/assimp/issues/2571

A scene with mesh that has two uv sets. When loaded using aiImport, it shows correctly. Somehow after aiExport and re-import again use aiImport, there is one uv map left. I think the aiImport is working correctly as originally the two uv sets are all imported. But after aiExport, one uv gets lost.

After diving into the FBXExport.cpp code,  it seems multi uv sets is a TODO task (comments in code). To fix that, just add extra LayerElementUV for extra uv map.


kimkulling:

PR #2574, opened 2019-07-30
Remove the usage of std::iterator

gongminmin:
std::iterator is deprecated in C++17. When using /std:c++17 in MSVC, WordIterator generates some warnings. Remove the usage of std::iterator to avoid future compilation problems.


kimkulling:

PR #2585, opened 2019-08-07
FBX importer armature fixes and root bone fixes - animations should now work for more models.

RevoluPowered:
This also fixes issue with root bone overwritten - which should not happen as it is a hack.

Before the behaviour would create an additional bone which would end up breaking animations completely on import.

Additionally this fixes required bones being removed.

Co-authored-by: K. S. Ernest (iFire) Lee <ernest.lee@chibifire.com>


kimkulling:

PR #2588, opened 2019-08-08
Scale Process overhaul 

RevoluPowered:
This now scales element transforms rather than purely the root transform. Which allows you to handle conversion from CM into M or whatever unit of measurement you want.

Added arguments --gs to assimp command line option to enable global scaling. (scaling specified by file format will be used from meta information

For an artist this means that if they scale their model 100x100x50 it will keep that meaningful information in scale, and not convert it (the positions will be converted instead; which is more transparent IMO).

### DIFF
--- a/code/PostProcessing/ScaleProcess.cpp
+++ b/code/PostProcessing/ScaleProcess.cpp
@@ -39,19 +39,17 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 ----------------------------------------------------------------------
 */
-#ifndef ASSIMP_BUILD_NO_GLOBALSCALE_PROCESS
-
 #include "ScaleProcess.h"
 
 #include <assimp/scene.h>
 #include <assimp/postprocess.h>
+
 
 namespace Assimp {
 
 ScaleProcess::ScaleProcess()
 : BaseProcess()
 , mScale( AI_CONFIG_GLOBAL_SCALE_FACTOR_DEFAULT ) {
-    // empty
 }
 
 ScaleProcess::~ScaleProcess() {
@@ -71,10 +69,15 @@ bool ScaleProcess::IsActive( unsigned int pFlags ) const {
 }
 
 void ScaleProcess::SetupProperties( const Importer* pImp ) {
-    mScale = pImp->GetPropertyFloat( AI_CONFIG_GLOBAL_SCALE_FACTOR_KEY, 0 );
+    mScale = pImp->GetPropertyFloat( AI_CONFIG_GLOBAL_SCALE_FACTOR_KEY, 1.0f );
 }
 
 void ScaleProcess::Execute( aiScene* pScene ) {
+    if(mScale == 1.0f) return; // nothing to scale
+    
+    ai_assert(mScale != 0);
+    ai_assert(nullptr != pScene && nullptr != pScene->mRootNode);
+
     if ( nullptr == pScene ) {
         return;
     }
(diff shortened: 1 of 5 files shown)


RevoluPowered:
Please note I am still working on this

RevoluPowered:
Please don't merge this yet, not ready for release.

RevoluPowered:
@kimkulling can you please review this change?

Thanks,
I'm just waiting on the CI to complete.

code/FBX/FBXConverter.cpp
```diff
@@ -3522,54 +3549,27 @@ void FBXConverter::SetShadingPropertiesRaw(aiMaterial* out_mat, const PropertyTa
...
-                    scale = (ai_real)100000.0;
-                } else if (unit == FbxUnit::m) {
-                    scale = (ai_real)1000.0;
-                }
+        ai_real FBXConverter::GetUnitScale( FbxUnit unit )
+        {
+            double scale;
+            if( out->mMetaData->Get("UnitScaleFactor", scale) )
+            {
+                printf("unit scale factor from file: %f\n", scale);
```
fire: Logging without error.

code/FBX/FBXConverter.cpp
```diff
@@ -3522,54 +3549,27 @@ void FBXConverter::SetShadingPropertiesRaw(aiMaterial* out_mat, const PropertyTa
...
-
-                if (mesh->HasPositions()) {
-                    for (unsigned int i = 0; i < mesh->mNumVertices; ++i) {
-                        aiVector3D &pos = mesh->mVertices[i];
-                        pos *= scale;
-                    }
-                }
+            else
+            {
+                printf("no file scale detected!");
```
fire: Logging without error. Isn't this common?

code/FBX/FBXImporter.h
```diff
@@ -89,10 +89,11 @@ class FBXImporter : public BaseImporter, public LogFunctions<FBXImporter>
     void InternReadFile( const std::string& pFile,
         aiScene* pScene,
         IOSystem* pIOHandler
-    );
+    );
```
fire: Whitespace.

code/FBX/FBXConverter.cpp
RevoluPowered: If no file scale exists, doesn't make sense to error and not read the file, we can assume the scale is 1.0.

RevoluPowered: @kimkulling do you have any other way of providing feedback to user which isn't an error, or should we just remove the prints?

kimkulling: Hi, sure, just use **ASSIMP_LOG_DEBUG** for this instead of printf.

RevoluPowered: Thanks!

RevoluPowered:
Changed printf to ASSIMP_LOG_DEBUG

RichardTea:
While I like the overall idea, I think the API is backwards. We wouldn't be able to use this.

When importing a model, the application doesn't know what units the model was drawn in until _after_ the import completes.

So, the application cannot set an appropriate import scale factor override until after it's imported the model, by which time it's too late.

It would be better if the application could set the absolute units it requires, and then each importer can calculate the appropriate scale factors to resize the model to those units.
Formats that don't specify any real-world units would be untouched.

For example, an application that requires metres should set the value "1.0".
An application requiring cm should set "0.01", one that wants inches could set "0.0254"

As to why this is unsuitable:
FBX files can specify which units were used. This pull request erases that data for all models, and so if the modeller exported using metres or millimetres the result will be wildly off.

Can you split out the ScaleProcess part of this pull request into another?

RevoluPowered:
Thanks for the feedback. 

> While I like the overall idea, I think the API is backwards. We wouldn't be able to use this.

I dont understand what you mean by this can you clarify, as FBX during import knows the file scale from the meta data, provided you place the scaling code after the meta data being read it should be fine to inform scale process of the units used in the file, provided you put it after that read meta data call?

On the second point:

> FBX (and some other formats) can specify which units were used.
> This pull request erases that data for all models, and so if the modeller exported using metres or millimetres the result will be wildly off with no way for an application to determine the cause.

I understand what you mean, I will implement a better way of dealing with this, which doesn't overwrite the scale of the original format, good point.

I'll make it so that the scale process handles the unit conversion AND any custom import setting.
Right now it's basically overridden by scale process which I agree is not the correct behavior. 

This isn't actually hard to achieve.

> Can you split out the ScaleProcess part of this pull request into another?

Yes, no problem.

RevoluPowered:
Going to rebase soon after running tests.

RevoluPowered:
@fire @RichardTea @kimkulling what do you think about importers implementing an extra function 'GetFormatScale', so that in the post process we can call that, so that it can't be ignored by mistake?

if it's not declared it just returns 1.

RichardTea:
Let me explain in another way:

The Godot engine desires model data to be in meters.

The FBX files you've loaded recently happened to be in cm, so you apply a scale factor of 0.01 and all is well.

However, that only works for FBX files. In fact, it only works for thoses _specific_ FBX files.
Other FBX files will be in meters, mm or even inches.
Collada files can be in metres (default!), cm, inches or in fact any arbitrary unit (double-decker-buses? Linguine?)

What does the Godot engine tell the Collada importer?

In short, telling the Assimp importer "Apply a scale factor of 0.01" only works for the small subset of FBX files you've tried recently. It doesn't work for all FBX files, and doesn't work at all for _any_ other importer!

Instead, the application should tell Assimp the _units_ it wants.

If Godot tells Assimp "I desire the model in Meters", each Importer can apply whatever post-processing scale factors are appropriate _for that specific model_.

(Assuming it is possible to determine)

RevoluPowered:
Sounds like a good addition. I'll work on a proposal for this change and submit a new PR as it's definitely required going forward to deal with this.

I have updated this PR to only include the overhauled ScaleProcess and completely removed FBX conversion / scaling.

PR #2589, opened 2019-08-09
FBX node chain assert fix

MalcolmTyrrell:
This fixes an inconsistency in how the FBXConverter::GenerateTransformationNodeChain handles epsilons. 

The chain is populated based on a square length test against 1e-6f. When it is later processed, the test aiMatrix4x4t<TReal>::IsIdentity() is used on the contents of the chain element. This is a component-wise comparison against a different epsilon of 10e-3f. I had an asset where a transform would pass the first test but fail the second one. Every loop iteration would just continue, and ai_assert(output_nodes.size()) would be hit. I've added a unit test that hits this assert without my fix.

The fix is to explicitly store which transforms are intended to be used in a bit field, and query that in the loop.

(Aside: While I'm here, I'll comment that 10e-3f is looked biggish for an epsilon, and is also a slightly strange way of writing 1e-2f. Is it possible it was intended to be 1e-3f? In any case, changing an epsilon would not fix the potential inconsistency).


kimkulling:
If I do remember right this epsilon to check the Identity-matrix against was introduced some years ago. I have to fix that!

kimkulling:
See https://github.com/assimp/assimp/issues/2598

MalcolmTyrrell:
Investigating the epsilons would be a good idea. However, I don't think you'll be able to adjust them to prevent this assert firing; you'll just make it less likely. The two epsilons which affect this code are used in difference calculations, and there will always be edge cases which expose the difference. 

The fix in this PR avoids the issue by allowing only one of the epsilon-based checks to influence the behaviour. I think this or something similar is required to ensure the assert never fires.

kimkulling:

PR #2592, opened 2019-08-13
Fix issue: FBXExporter.cpp merges identical vertices disrespect export flag 0x0

ywangsc:
This pull request is to fix issue #2593 

I have scenes that prefer the vertex number and order untouched, thus I particularly avoid specifying aiJoinIdenticalVertices. However, after exported as fbx, my scenes are messed up.

After looking into this issue, it is FBXExporter.cpp will always merge identical vertices, regardless my specified flag.

Here I fix this issue by checking if user specified aiJoinIdenticalVertices when exporting scenes, if not, no need to merge the vertices.


kimkulling:
Need to merge it on a different pull-requesz ( see https://github.com/assimp/assimp/pull/2643 ).

Issue #2599, opened 2019-08-16
multi-config debug postfix no longer working


RichardTea:
Platform is Windows/MSVC.
(note that most other platforms don't support multi-config cmake)

This appears to have been broken by #2568 or #2086 

The addition of the bool INJECT_DEBUG_POSTFIX for #2075 was not necessary, and doesn't work either.

It is easier to simply explicitly set CMAKE_DEBUG_POSTFIX to empty when a debug postfix is not desired.

RichardTea:
This was broken in commit 9dce8e40 ("Update CMakeLists.txt", 2019-07-25)

PR #2600, opened 2019-08-16
Issue 2599 multiconfig debug postfix

RichardTea:
Re-enable the debug suffix with multi-config generators. #2599 

Can we please not break this again?


kimkulling:
Will try my really best!

kimkulling:
Merged, thanks a lot for the contrbution ( or let*s say repairing it again )

PR #2601, opened 2019-08-17
fix for div by zero reported by address sanitizer

GBhutra:
Got a runtime undefined behavior error when running with the address sanitizer


kimkulling:

PR #2605, opened 2019-08-18
Update CMakeLists.txt

anandvsingh:
Issue was on Windows MSBuild since the debugger was unable to expand symbols inline.


kimkulling:

PR #2610, opened 2019-08-21
Update DLL PE details: Copyright, git commit hash and original filename

RichardTea:
Looks like this:
image

Windows only of course.


kimkulling:

PR #2617, opened 2019-08-24
Better reinterpret_cast'ing

alexeys85:
Since type of copy.length and aliased type of s are different types than reinterpret_cast'ing to uint32_t is violation of strict aliasing rule. But casting to uint8_t is permittable.


kimkulling:
Sorry, but this pull request will break all unittests on x64 platforms. I tried my really best to fix this as well, but it didn't work. Maybe switching to uint32_t is much better?

Issue #2614, opened 2019-08-22
[Patch] FBX Crash on import


RevoluPowered:
Just putting this here so I don't forget to fix this happening and segfault happening on NVIDIA models.

```cpp
---------------- thirdparty/assimp/code/FBX/FBXMeshGeometry.cpp ----------------
index e60fc25e24..874cee52e0 100644
@@ -610,7 +610,11 @@ void MeshGeometry::ReadVertexDataMaterials(std::vector<int>& materials_out, cons
     const std::string& ReferenceInformationType)
 {
     const size_t face_count = m_faces.size();
-    ai_assert(face_count);
+    if(face_count == 0)
+    {
+        return;
+    }
```

kimkulling:
Ok, I will take care! Thanks for the bug report.

RevoluPowered:
Patch created :)

PR #2621, opened 2019-08-26
Fix segfault #2614

RevoluPowered:
Fixes segfault with models with missing faces.

This can be a common thing if a modeller makes a mistake so unfortunately we must not crash the entire importer because this means you can't import some models.

This ensures compatibility even with bad modelling practices because humans aren't perfect.


alexeys85:
size_t can't be less than zero, just equal or greater.

RevoluPowered:
Good point, I will in future not do this check.

kimkulling:

PR #2623, opened 2019-08-27
FIX missing File Scale and Application scale fixes for all conversions 

RevoluPowered:
This is hotfix for FBX scaling issues we were experiencing in godot. This is another urgent fix.

Before the compiled scale was not processed properly.

I also removed depreciated scaling code from the FBX system.

I have tested this code with 5 models and seems to be okay with CM to M scaling, and various other units.


kimkulling:

PR #2626, opened 2019-08-28
A crash in the aiMesh destructor

muxanickms:
When Exporter::Export method called, it creates a temporal copy of the scene in the scope, the issue is that aiAnimMesh instances are not copied and on descructor it tries to delete it twice hence the crash.
I added a unit test for it as well.
Small addition to the debug flags is to be able to debug unit tests in the Visual Studio using CMake feature.


kimkulling:
Looks good, I will merge it.

kimkulling:

PR #2630, opened 2019-08-30
Make IsVerbose accessible outside the exporter

MalcolmTyrrell:
This trivial change allows IsVerboseCheck to be called outside the exporter, and means that it's possible for other code to check for verbose format without having to reimplement the logic.


kimkulling:
Merged, thanks alot for the contribution.

Issue #2513, opened 2019-06-17
Importer unicode support broken


rmccampbell:
This issue https://github.com/assimp/assimp/issues/1612 addressed unicode support on windows. However although it was closed, the solution is broken as mentioned in the follow up responses. The IsTextUnicode check should be removed and the proper unicode functions always used on windows. (This check is not useful for efficiency anyway, since windows will translate legacy ANSI function calls to unicode behind the scenes regardless).

rmccampbell:
Also it seems that MB_PRECOMPOSED is deprecated. And the fixed sized buffers should ideally be replaced with dynamically sized.

kimkulling:
You are right, this issue is still there. The issue #1612 was duplicated. That was the reason that I haven't reopen it again.

rmccampbell:
Where's the duplicate?

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2633

PR #2633, opened 2019-09-03
Unicode fix

rmccampbell:
This is a fix to #2513. It removes the IsTextUnicode functions and always uses unicode functions on windows, as discussed in this issue and #1612. I also while I was at it removed the the fixed string lengths and made all the strings safely dynamically allocated, while cleaning up the code and removing most redundancy. 
I left in the WindowsStore checks, but I'm not sure these are necessary (it may have been the IsTextUnicode functions that necessitated them in the first place?). Ideally the windows store code should also be unicode-compatible, so if it could be tested without the checks then that code could be removed as well.


rmccampbell:
I tested the code on windows store platform with the unicode support enabled and it built fine so I removed the checks. I confirmed the IsTextUnicode function was the culprit (it seems to be removed in UWP), so since this function is removed they are no longer necessary. I don't see any other uses of this macro so it may be removable altogether but I haven't done so.

kimkulling:
Thank you so mzuch for the fix, I will merge it!

kimkulling:

PR #2640, opened 2019-09-06
FBX Implemented basic PBR material textures

RevoluPowered:
This is to support Maya materials properly from FBX Importer.

This adds the following texture types:
- BASE_COLOR
- NORMAL_CAMERA
- EMISSION_COLOR
- METALNESS
- DIFFUSE_ROUGHNESS
- AMBIENT_OCCLUSION

Here is an example of the work in use:
https://github.com/godotengine/godot/pull/32019

Reuses existing API as it is fairly simple to use.

The existing 'ambient', 'normal' maps are not the same shading method, therefore I opted to implement proper enums for them, as they're for the legacy format only, which is easier to support than the PBR / PBS route.


RevoluPowered:
Got some time over the weekend so I decided to add stingray support too for Maya.

Same type of code so it was worth expanding it.

RevoluPowered:
Rebased on master - should be OK to merge once CI completes.

kimkulling:

DrNoodles:
Do you have an example Maya file, or tutorial, showing how this works?

1. Created a mesh

pplied a StingrayPBS material, added a TEX Color Map, TEX Normal Map, TEX Roughness Map and TEX Metallic Map. I've exported the FBX. When import

PR #2644, opened 2019-09-08
Fix Sample_SimpleOpenGL.c

FabienPean:
Extra keyword `struct` causing problems


kimkulling:
But only when you have to use it as c++, for pure c-compilers this struct still works.

kimkulling:
Just introduced the assimp-struct compatibility statement for older c-compilers.

kimkulling:
Mergws,

PR #2646, opened 2019-09-09
Build Binaries example batch

RumbleballTheReal:
I'm still not sure what about the config.h file that is generated from cmake. Is it important when copying the includes?


RichardTea:
Yes, config.h is required.
The enums and macros in there are used by applications that consume the library.

The deploy section of my Assimp Windows batch looks like this:

```
SET OPT_ROBOCOPY=/np /log+:"%PROJECT_DEPLOY_LOG%"

:deploy
:: Replace old includes
robocopy "%SOURCE_INCLUDE_DIR%" "%DEST_INCLUDE_DIR%" %OPT_ROBOCOPY% /mir /xf *.in
xcopy "%BUILD_DIR%\include\assimp\*.h" "%DEST_INCLUDE_DIR%" /i /y

:: Copy DLLs and EXEs
robocopy "%BUILD_DIR%\code\Debug" "%DEST_BIN_DIR%" *.dll *.exe %OPT_ROBOCOPY%
robocopy "%BUILD_DIR%\code\RelWithDebInfo" "%DEST_BIN_DIR%" *.dll *.exe %OPT_ROBOCOPY%
robocopy "%BUILD_DIR%\tools\assimp_cmd\RelWithDebInfo" "%DEST_BIN_DIR%" *.dll *.exe %OPT_ROBOCOPY%

:: Copy LIBs and PDBs
robocopy "%BUILD_DIR%\code\Debug" "%DEST_LIB_DIR%" *.lib *.pdb %OPT_ROBOCOPY%
robocopy "%BUILD_DIR%\code\RelWithDebInfo" "%DEST_LIB_DIR%" *.lib *.pdb %OPT_ROBOCOPY%
robocopy "%BUILD_DIR%\tools\assimp_cmd\RelWithDebInfo" "%DEST_LIB_DIR%" *.pdb %OPT_ROBOCOPY%
```

As a general rule, on Windows you should use `RelWithDebInfo` instead of `Release`.
RelWithDegInfo is the same as Release but also creates PDBs so you can debug the library afterwards.

My build batch also embeds source indexing into the PDBs and publishes everything to an artifact server, but that kind of thing doesn't feel like it should be in a basic example.

RumbleballTheReal:
Thanks @RichardTea, I added more information to the example.

kimkulling:
Hi, looks nice. We have some scripts to build the installer as well. Maybe this would be interesting for you as well?

kimkulling:

RumbleballTheReal:
Thanks Kim. Only intrested in the .dlls. Though you might add it for others.

PR #2655, opened 2019-09-15
Issue 2603

### LINKED ISSUE #2603: corrupted normals loaded from .x file
I get corrupted normals from .x files using current assimp master (1cb23134b8f6f806ba3302e03c942cefe8eb4d69)
It used to load the same files correctly in the past (2f6aa2460ac24d553c5114d4dde2bd8542e01963).
I tried git-bisect, but there were too many unbuildable commits in between and I failed to locate the problem.

Here is an example problematic file:
[cylinder.zip](https://github.com/assimp/assimp/files/3512587/cylinder.zip)
The archive contains .blend file with simple model and .x file that I exported from the blender.

PascalGilbraith:
Closes #2603 
Normal face indices added to end of already resized vector, so normal faces did not match position faces.
Use vector index rather than push.

### DIFF
--- a/code/X/XFileParser.cpp
+++ b/code/X/XFileParser.cpp
@@ -596,11 +596,11 @@ void XFileParser::ParseDataObjectMeshNormals( Mesh* pMesh)
     // do not crah when no face definitions are there
     if (numFaces > 0) {
         // normal face creation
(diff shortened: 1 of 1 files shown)


kimkulling:

Issue #2551, opened 2019-07-17
colada output path is worng when that is exported.


copycd:
assimp/code/Collada/ColladaExporter.cpp
line number 320:            std::unique_ptr<IOStream> outfile(mIOSystem->Open(mPath + name, "wb"));

mPath + name  =>  It is combined without splash.

PR #2659, opened 2019-09-18
fix issue #2551

Bruno361:
This PR fixes issues #2551 , where a separator is missing in the textures export path.


kimkulling:

PR #2660, opened 2019-09-19
[WIP - do not merge] FBX animation refactor

RevoluPowered:
This fixes some bugs which caused advanced models to break.

This also refactors the code to be clearer and simpler in terms of processing the keyframes

It is my intention if I get some time to work on the blend shapes but this is functional for now.
I have tested for regressions and would welcome testing before this is merged.


RevoluPowered:
Closing found a regression in keyframe time handling with some meshes which needs resolved.

PR #2663, opened 2019-09-21
fix issue textureCoords are messed up for multiple uvset 

thomasbiang:
Issue:
When having multiple set of uvs, the number of uv channels are exported correctly. However the exact uv coordination is wrong. 
After some debug, I find the issue is caused by the wrong index used for the uv set. 
To fix it, simply update the uv set index correctly, instead of always using constant 1 as the index. 
The code change is minimal.


kimkulling:
Good catch!

kimkulling:

PR #2666, opened 2019-09-22
Fix: Wrong aiAnimation::mTicksPerSecond for gltf2 imports (fixes #2662)

### LINKED ISSUE #2662: Incorrect aiAnimation::mTicksPerSecond value for gltf2 imports
When importing gltf 2 files, the value of `aiAnimation::mTicksPerSecond` is incorrect (at least in some cases).

According to the documentation, the value of aiAnimation::mDuration is in ticks (http://assimp.sourceforge.net/lib_html/structai_animation.html#ae2bbc49320b4b75c05e23e0ab704ece7) and the value of aiAnimation::mTicksPerSecond should be the number of ticks per second (http://assimp.sourceforge.net/lib_html/structai_animation.html#afd26a40c3f16b6aa6e150effa3eaeab1).

This means the duration of an animation in seconds should be `mDuration * mTicksPerSecond`. Unfortunately I'm getting an `mDuration` value of `4000.0` and a `mTicksPerSecond` value of `1.25` when I try to import the attached GLB file.

[object.tar.gz](https://github.com/assimp/assimp/files/3632426/object.tar.gz)

(I'm importing the GLB file, but I've also included the original Blender 2.80 file from which the GLB file was exported).

The animation in reality is 4s (48 frames at 12fps), but according to Assimp it's 4000*1.25=5000s!

I think the problem is with these two lines in `code/glTF2/glTF2Importer.cpp`:

image

In the `CreateNodeAnim` method, the keyframe times are being specified in milliseconds since they're all being multiplied by `kMillisecondsFromSeconds`. That means `mTicksPerSecond` should always be `0.001`.

(On a separate note, what is the duration of an animation if `mTicksPerSecond` is 0? The documentation notes that this is possible, so does that mean the length of the animation in seconds is just `mDuration` when `mTicksPerSecond` is 0?)

theakman2:
Fixes #2662.

### DIFF
--- a/code/glTF2/glTF2Importer.cpp
+++ b/code/glTF2/glTF2Importer.cpp
@@ -1176,7 +1176,7 @@ void glTF2Importer::ImportAnimations(glTF2::Asset& r)
             }
         }
         ai_anim->mDuration = maxDuration;
-        ai_anim->mTicksPerSecond = 0.001;
+        ai_anim->mTicksPerSecond = 1000.0;
 
         mScene->mAnimations[i] = ai_anim;
     }
--- a/code/glTF2/glTF2Importer.cpp
+++ b/code/glTF2/glTF2Importer.cpp
@@ -1177,7 +1177,7 @@ void glTF2Importer::ImportAnimations(glTF2::Asset& r)
             }
         }
         ai_anim->mDuration = maxDuration;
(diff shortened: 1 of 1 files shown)


kimkulling:

PR #2671, opened 2019-09-23
Fix UV texture coords generation along Z Axis

JenniferBuehler:
The projection of textures along z axis is currently projecting along the x axis.


kimkulling:

PR #2677, opened 2019-09-25
Support Apple naming conventions - shared library

mdinim:
On OSX CMake does not append the `SOVERSION` target property after the `dylib` file, but creates a dylib named `libassimp.5.dylib`. This PR is to respect that behaviour.


kimkulling:

PR #2683, opened 2019-09-29
Code cleanup

smalcom:
gcc: 8.3.0
Built with: "-DASSIMP_COVERALLS:BOOL=ON -DASSIMP_WERROR:BOOL=ON"

### DIFF
--- a/code/Common/SceneCombiner.cpp
+++ b/code/Common/SceneCombiner.cpp
@@ -45,7 +45,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // possible as new fields are added to assimp structures.
 
 // ----------------------------------------------------------------------------
-/** 
+/**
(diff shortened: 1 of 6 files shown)


turol:
Casting to `void *` just hides the issue. Either fix it correctly (by not using `memcpy`) or detect the compiler version and silence the warning.

smalcom:
> Casting to void *
But
man memcpy:
> void *memcpy(void *dest, const void *src, size_t n);


> detect the compiler version and silence the warning.
I think its a bad idea; step to macrohell.

turol:
Yes, it's theoretically correct to cast to `void *` but that doesn't fix the problem, only hides it. If someone actually wants to fix it later they're going have to find and undo your changes before they can do it.

And checking the compiler doesn't take any macros in this case, it's purely a CMake thing. Check if the compiler supports '-Wno-class-memaccess' and turn it on if it does, then file an issue so it's not forgotten.

smalcom:
Understood.

PR #2690, opened 2019-10-01
ColladaExporter: improve name/id handling

### LINKED ISSUE #2687: Material name is altered after saving to Collada
Hi.
I'am using Assimp for loading  and saving Collada files and I found what Assimp is altering material names during export. E.g. material named _some-thing_ is renamed to _some_thing_.

This happens because every symbol in the material name is validated with the function isalnum_C, which allows only digits and letters. Any other symbols are replaced with the underscore symbol.
ColladaExporter.cpp:
```c++
for( std::string::iterator it = materials[a].name.begin(); it != materials[a].name.end(); ++it ) {
    if( !isalnum_C( *it ) ) {
        *it = '_';
    }
}
```
I think some other symbols should be allowed in material names.
Also another usage of isalnum_C in the same file has exceptions for some special symbols:
```c++
for( std::string::const_iterator it = pSurface.texture.begin(); it != pSurface.texture.end(); ++it )
{
    if( isalnum_C( (unsigned char) *it) || *it == ':' || *it == '_' || *it == '-' || *it == '.' || *it == '/' || *it == '\\' )
        imageUrlEncoded << *it;
    else
        imageUrlEncoded << '%' << std::hex << size_t( (unsigned char) *it) << std::dec;
}
```

TGEnigma:
Improve handling of names used in the scene by properly encoding the names as XML ids when used as such, while preserving the original formatting of the name when used in a 'name' tag.
In simple terms, this PR allows names with symbols to be used.
Fixes bugs like the previously broken weights when using node names with spaces in them, due to the bone name array being space terminated.
Fixes #2687


kimkulling:

PR #2698, opened 2019-10-06
Fix #2685 - library compiled by MinGW is unusable in MinGW.

### LINKED ISSUE #2685: Building v5.0.0 as a shared lib using MinGW on Windows creates .dll without exports and a broken cmake configuration
Assimp 5.0.0 seemingly builds on Windows with MinGW compiler but the installed library is unusable. This can be fixed ad-hoc by following these instructions:

1. In the extracted source code directory navigate to `include/assimp/defs.h` and remove `#ifdef _MSC_VER` that is right before `#undef ASSIMP_API` and all of its `#elif`s and `else`s. Keep the body of `#ifdef _MSC_VER`, but remove `#pragma warning (disable : 4251)` or else compiler will spam.
2. Configure the project as usual picking your preferred build options. Ensure `BUILD_SHARED_LIBS` is enabled. I also disabled other stuff: `ASSIMP_BUILD_ALL_EXPORTERS_BY_DEFAULT`, `ASSIMP_BUILD_ASSIMP_TOOLS`, `ASSIMP_BUILD_ASSIMP_VIEW`.
3. Compile (`cmake --build .`), then install to your preferred `CMAKE_INSTALL_PREFIX` (`cmake --build . --target install`).
4. Statements `cmake_policy(VERSION 2.6)` and `if(ON)` are used in assimpTargets.cmake resulting in a CMake `CMP0012` policy warning that cannot be squelched. This is annoying as it produces a barrage of red text in user's project, but doesn't block the compilation and can be ignored by the user. The fix here could be to remove the `if(ON)`/`else` statements and simply substitute the proper `SHARED` or `STATIC` word in the `add_library` statement that the `if(ON)` encompasses.
5. Linux library naming is mixed with Windows, therefore stuff will break because the library name is `.dll.5` (a mix of `.dll` and `.so.5`?) and also there's an assumption that corresponding `.a` file is not needed. Find `.dll.5` text in `assimpTargets-release.cmake` in the `${CMAKE_INSTALL_PREFIX}/lib/cmake/` directory. Replace it with `.dll.a`.

Zalewa:


kimkulling:

PR #2704, opened 2019-10-08
Fix gltf importer crash

MalcolmTyrrell:
Invalid GLTF files which had an incorrect number of vertices for their mode were causing buffer overruns, leading to mysterious delayed crashes.

For example, if the number of vertices in Triangles mode was not divisible by 3, the loop would write beyond its allocation.

This PR fixes the known cases and provides unit tests. Unfortunately, since it's a corruption bug, the unit tests typically pass even without the fix. Nevertheless, it's I'd say it's useful to have these files pass through the system.

Right now, this particular GLTF problem is considered a warning by the GLTF validator, but there's a request to make it an failure:
https://github.com/KhronosGroup/glTF/issues/1674


MalcolmTyrrell:
@turol Thanks for the suggestion. It would be nice to know that the unit tests fail prior to the fix.

Do you know if its viable to get the AddressSanitizer working on Windows?

turol:
I don't know, I've only used it on Linux. Windows version of clang might support it. There's been talk about porting it to MSVC, no idea what the status is there. You could either 1) install Linux or 2) commit just your test case and make a PR of that, it should then fail in CI.

MalcolmTyrrell:
Thanks @turol, your proposal worked. The build failed on the PR with the test but without the fix.

MalcolmTyrrell:
We're keen to get a fix for this issue into the repo. If there are any issues blocking this, please let me know :)

kimkulling:
Looks good, will merge it after CI is finished!

kimkulling:

Issue #2693, opened 2019-10-02
v5.0.0, aiGetVersionMajor() returns 4


Opioid:
According to git I am building v5.0.0-15-g09d80ff4, but aiGetVersionMajor() still returns 4. Am I not actually building the newest version or is this number just out of sync?

mosra:
@kimkulling I just ran into this as well -- in [magnum](https://magnum.graphics) I have to do various workarounds / data patching based on Assimp version used and this omission effectively makes version 5 indistinguishable from 4.1.

PR #2709, opened 2019-10-11
Update assimp legal and version

RichardTea:
Issue #2693 was fixed in a rather brute-force manner.

This PR changes to report the major and minor versions specified in the cmakelists file and thus used for dylib/solib and DLL PE headers.

Also updates Legal Info to match the new website (sourceforge is way out of date!)


kimkulling:
That was just the older way, this variant is much better. Thanks!

kimkulling:

PR #2711, opened 2019-10-14
Collada ZAE texture loading

### LINKED ISSUE #2710: Collada Importer: Embedded textures are not always loaded from ZAE archive
Referenced texture files inside the archive should be extracted into an aiTexture during import

RichardTea:
Closes #2710 
Manifest and Image URIs now passed through the UriDecode function

Note: In the future the XML Encode/Decode and URI Encode/Decode should be moved into Common, as there are several XML-based formats.


kimkulling:

Issue #2678, opened 2019-09-26
VS2013 failed to build v5.0.0


ardenpm:
The Visual Studio 2013 build currently fails with the 5.0.0 release. This seems to be due to the lack of implementation bracket list initialises since Visual Studio 2013 Update 3 (Update 4 is current). Here is an example of the errors emitted (there are 114 of them, all similar):

```
Error   1   error C2797: 'Assimp::BaseImporter::importerUnits': list initialization inside member initializer list or non-static data member initializer is not implemented (C:\Temp\assimp-5.0.0\code\Common\ImporterRegistry.cpp) C:\Temp\assimp-5.0.0\include\assimp\BaseImporter.h  203
```

This is the problematic part of the file.

```
    std::map<ImporterUnits, double> importerUnits = {
        {ImporterUnits::M, 1},
        {ImporterUnits::CM, 0.01},
        {ImporterUnits::MM, 0.001},
        {ImporterUnits::INCHES, 0.0254},
        {ImporterUnits::FEET, 0.3048}
    };
```

I've attempted using `std::make_pair` on the inner list instead but that doesn't help. I'm sure someone else already knows the proper fix.

[Here](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2797?view=vs-2019) Microsoft describe what the issue is.

kimkulling:
Thanks a lot for the pull reuqest, I will try to merge them as fast as possible.

PR #2715, opened 2019-10-16
fix vs2013 build

ardenpm:
Fixes broken vs2013 build which was due to a compiler bug in vs2013 which resulted in Microsoft disabling brace initialisers in vs2013 update 3 and later. Not ideal since the initialisers needed to move to the constructor, if vs2013 support is dropped at some point can be changed back. This pull request should fix issue #2678.


kimkulling:

Issue #2721, opened 2019-10-18
CMake configuration broken for windows


jherico:
The `assimpTargets-release.cmake` and `assimpTargets-debug.cmake` configuration files generated in a windows build for a shared library configuration are broken.  

When I use the CMake find and target functions

    find_package(assimp CONFIG REQUIRED)
    target_link_libraries(${TARGET_NAME} PRIVATE assimp::assimp)

It causes the linker to use the DLL filename as the link input, not the import filename.  See image below:

![image](https://user-images.githubusercontent.com/1533642/67115097-3f3d1e00-f192-11e9-934e-050f77ae505c.png)

This in turn causes the link to fail

> 2>D:\ezvcpkg\fa7a3eaec35409d60dbc456d5229b54c192f11de\installed\x64-windows\bin\assimp-vc142-mt.dll : fatal error LNK1107: invalid or corrupt file: cannot read at 0x300

I'm unable to determine why this is happening.

kimkulling:
Ok, I will close the issue.

PR #2722, opened 2019-10-18
Fix CMake import

jherico:
Using CMake 3.15 & Visual Studio 2019 to build on windows produced a cmake configuration which was unable to be used by clients.  Investigating this I was eventually able to determine this was caused by a mismatch in the target properties.  Specifically, it was setting the TYPE property to STATIC_LIBRARY even though it was building as a SHARED_LIBRARY.  

This in turn was because the if clause on line `if(@BUILD_SHARED_LIBS@)` of this file was evaluating to `if (ON)` which was then not properly evaluated while under the CMake 2.6 policy.  Explicitly setting the `CMP0012` policy to NEW fixes the problem.

Fixes bug #2721


kimkulling:

Issue #2724, opened 2019-10-25
[glTF 2] KHR_materials_pbrSpecularGlossiness glossinessFactor in wrong object


ardenpm:
Currently the glossinessFactor in glTF 2 PBR materials using the `KHR_materials_pbrSpecularGlossiness` extension is being written directly to the main material object instead of the `extensions.KHR_materials_pbrSpecularGlossiness` object. So you get this:

```
{
    "name": "test",
    "pbrMetallicRoughness": {
        "baseColorFactor": [ 1.0, 1.0, 1.0, 0.0 ],
        "metallicFactor": 0.0
    },
    "glossinessFactor": 0.8,
    "extensions": {
        "KHR_materials_pbrSpecularGlossiness": {
            "diffuseFactor": [ 1.0, 1.0, 1.0, 0.0 ]
        }
    }
}
```

Instead of this:

```
{
    "name": "test",
    "pbrMetallicRoughness": {
        "baseColorFactor": [ 1.0, 1.0, 1.0, 0.0 ],
        "metallicFactor": 0.0
    },
    "extensions": {
        "KHR_materials_pbrSpecularGlossiness": {
            "diffuseFactor": [ 1.0, 1.0, 1.0, 0.0 ],
            "glossinessFactor": 0.8,
        }
    }
}
```

Looks like a simple fix, I will submit a pull request shortly.

PR #2725, opened 2019-10-25
glTF2 fix glossinessFactor being put into the wrong object on export

ardenpm:
Fixes issue where glossinessFactor for KHR_materials_pbrSpecularGlossiness was put in the main material object instead of the extension sub-object. Fixes issue #2724


kimkulling:

PR #2727, opened 2019-10-25
ColladaParser: fix handling of empty XML-elements

da-raf:
Empty XML-tags like `<revision/>` in collada files crashed the importer. This commit fixes that.


kimkulling:

Issue #2652, opened 2019-09-11
Clang formatting


RevoluPowered:
Any chance we can start using clang-format?

I don't need it for every file, but it would be nice if new contributions could optionally include files formatted using clang-format.

It really cleans up files, we use it in godot and it's quite simple.

kimkulling:
Do you measn include it into the CI?

RevoluPowered:
I was considering that we could add it to the CI for only changed files, so that anything edited must be updated but anything which is not touched can remain the same, so we don't get huge commits with thousands of files changed.

It would be up to you but I would really like to enforce it because we can get much cleaner code in the long run.

RevoluPowered:
the only hard part to this is picking a style guide

https://clang.llvm.org/docs/ClangFormatStyleOptions.html

The LLVM standard is probably what I would use, this is what godot uses and it's pretty simple. Most tools support clang-format.

RevoluPowered:
I would also recommend we update the C++03 to C++11.

kimkulling:
This should be the case already!

RevoluPowered:
do you mind if I add it an PR it after I do the update?

kimkulling:
For the clang-formatter? Sure

kimkulling:
Closed and fixed.

PR #2728, opened 2019-10-26
Clang format added for code reformatting

RevoluPowered:
This is the clang formatting file which was requested in a previous issue.

This will eventually be required for new code, but old code will remain with previous formatting. Meaning if you update somefile.c and someheader.h you must clang-format for the CI to pass in future versions.

Includes .o and .a in project git ignore too as this is such a tiny change.


RevoluPowered:
Resolves #2652

kimkulling:

PR #2731, opened 2019-10-26
Implemented armature lookup and updated FBX importer to properly support this

RevoluPowered:
This lets you directly retrieve the node a bone links to and informs you of the armature and node transform directly. This should work in all formats which support aiBone's :)

This also fixes a bug with bone name being made unique which causes them to become not 1:1 what the modeller has imported.

**This means bones tell you:**
- the inverse bind matrix
- the armature the bone is for
- the node which the bone has in the hierarchy (used for rest and skin matrix calculation, quick and easy) 
- the vertex weights

I have added a post process to handle this, and unit tests are provided too.


code/PostProcessing/ArmaturePopulate.h
```diff
@@ -0,0 +1,107 @@
...
+#include <map>
+
+
+struct aiNode;
+struct aiBone;
+
+namespace Assimp {
+
+// ---------------------------------------------------------------------------
+/** ScaleProcess: Class to rescale the whole model.
```
kimkulling: Wrong doc!

kimkulling:

include/assimp/mesh.h
```diff
@@ -268,6 +271,16 @@ struct aiBone {
...
+    // you must enable aiProcess_PopulateArmatureData to populate this
+    C_STRUCT aiNode* mArmature;
+
+    // The bone node in the scene - used for skeleton conversion
+    // you must enable aiProcess_PopulateArmatureData to populate this
+    C_STRUCT aiNode* mNode;
+
+#endif
     //! The influence weights of this bone, by vertex index.
     C_STRUCT aiVertexWeight* mWeights;
```
3b: Compile-time ABI changes without also changing version or something in `aiGetCompileFlags` make shared libs harder to use safely. Could these pointers be included always, even if the extra processing isn't included?

Also, adding to the end of structures instead of the middle makes it slightly easier to support multiple versions of the library at once, since they can share a common prefix (or even whole thing if not included directly in another structure).

PR #2736, opened 2019-10-29
Added M3D format support

bztsrc:
Dear Assimp developers,

I've added a new file format to your great library. It is a clean merge. Importing works perfectly (meshes, materials, bone structure, animations etc.), but exporter only supports static meshes for now.

Thanks,
bzt


bztsrc:
Hi,

I don't think my patch caused this, the log says

> The command "cmake . -DASSIMP_ENABLE_BOOST_WORKAROUND=YES" failed and exited with 1 during .

I've check this with a vanilla assimp, it fails too with a warning "CMake Warning: Manually-specified variables were not used by the project: ASSIMP_ENABLE_BOOST_WORKAROUND".

Cheers,
bzt

bztsrc:
About the PR Quality Review, it's 5 scope could be reduced messages. It was intentional to declare the loop variables as soon as possible and reuse them for speed up. Let me know if this is an issue and I'll move them inside the outer loops.

Cheers,
bzt

turol:
Enlarging variable scope does nothing for speed, the compiler SSA transformation will just break them up again. However it can lead to bugs when you accidentally reuse a value when you should have reinitialized or shadow a variable you shouldn't have. Always put everything in the minimal scope possible. If you want to know how it actually works search for "Static Single Assignment".

code/M3D/M3DMaterials.h
```diff
@@ -0,0 +1,106 @@
...
+ * only once. D.R.Y. and K.I.S.S.
+ */
+typedef struct {
+    const char *pKey;
+    unsigned int type;
+    unsigned int index;
+} aiMatProp;
+
+/* --- Scalar Properties ---        !!!!! must match m3d_propertytypes !!!!! */
+static aiMatProp aiProps[] = {
```
turol: These should be const so the compiler knows they can't change. Also they should be in a .cpp file, otherwise the compiler will emit then in every compilation unit. Worse, if they're not const the linker will not de-duplicate them.

code/M3D/m3d.h
```diff
@@ -0,0 +1,4544 @@
...
+#define M3D_INDEXMAX 0xfffe
+#endif
+#ifndef M3D_NUMBONE
+#define M3D_NUMBONE 4
+#endif
+#ifndef M3D_BONEMAXLEVEL
+#define M3D_BONEMAXLEVEL 8
+#endif
+#ifndef _MSC_VER
+#define _inline __inline__
```
turol: Identifiers starting with underscore are reserved for the implementation and we must not define them.

code/M3D/m3d.h
```diff
@@ -0,0 +1,4544 @@
...
+#define _pack __attribute__((packed))
+#define _unused __attribute__((unused))
+#else
+#define _inline
+#define _pack
+#define _unused
+#endif
+#ifndef  __cplusplus
+#define _register register
```
turol: Using `register` in modern code is a bad idea. The compiler will most likely just ignore them and if it doesn't it will still probably not do what you want. Also underscore trouble like above.

code/M3D/m3d.h
```diff
@@ -0,0 +1,4544 @@
...
+/* private prototypes used by both importer and exporter */
+m3ds_t *_m3d_addskin(m3ds_t *skin, uint32_t *numskin, m3ds_t *s, uint32_t *idx);
+m3dv_t *_m3d_addnorm(m3dv_t *vrtx, uint32_t *numvrtx, m3dv_t *v, uint32_t *idx);
+char *_m3d_safestr(char *in, int morelines);
+
+/*** C implementation ***/
+#ifdef M3D_IMPLEMENTATION
+#if !defined(M3D_NOIMPORTER) || defined(M3D_EXPORTER)
+/* material property definitions */
+static m3dpd_t m3d_propertytypes[] = {
```
turol: Another non-const array.

code/M3D/m3d.h
```diff
@@ -0,0 +1,4544 @@
...
+};
+#endif
+
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(M3D_NOIMPORTER) && !defined(STBI_INCLUDE_STB_IMAGE_H)
+/* PNG decompressor from
+
+   stb_image - v2.23 - public domain image loader - http://nothings.org/stb_image.h
```
turol: Did you really need to copy-paste the header? Why not include it as a separate file? At least it would make it easier to update. Also, do we export these symbols from the dynamic object? What happens if user code also uses stb_image?

code/M3D/m3d.h
```diff
@@ -0,0 +1,4544 @@
...
+#define stbiw__zlib_huff3(n)  stbiw__zlib_huffa(0 + (n)-256,7)
+#define stbiw__zlib_huff4(n)  stbiw__zlib_huffa(0xc0 + (n)-280,8)
+#define stbiw__zlib_huff(n)  ((n) <= 143 ? stbiw__zlib_huff1(n) : (n) <= 255 ? stbiw__zlib_huff2(n) : (n) <= 279 ? stbiw__zlib_huff3(n) : stbiw__zlib_huff4(n))
+#define stbiw__zlib_huffb(n) ((n) <= 143 ? stbiw__zlib_huff1(n) : stbiw__zlib_huff2(n))
+
+#define stbiw__ZHASH   16384
+
+unsigned char * stbi_zlib_compress(unsigned char *data, int data_len, int *out_len, int quality)
+{
+   static unsigned short lengthc[] = { 3,4,5,6,7,8,9,10,11,13,15,17,19,23,27,31,35,43,51,59,67,83,99,115,131,163,195,227,258, 259 };
```
turol: Avoid function-scoped static variables, they might be delay-initialized which will cause trouble. Also should const.

test/models/M3D/README.md
```diff
@@ -0,0 +1,14 @@
+Model 3D Samples
+================
+
+ aliveai_character.m3d - from Minetest aliveai mod (textures, animations, original 47k, m3d 2.5k)
```
turol: License?

bztsrc:
Hi turol,

Thanks for checking out.

About the speed up, yes there is, read https://en.wikipedia.org/wiki/Red_zone_%28computing%29

About the changes you requested. Please realize that M3DImporter and M3DExporter are wrappers around a C API.

code/M3D/M3DExporter.cpp
```diff
@@ -0,0 +1,394 @@
...
+    (void)mProperties;
+
+    // open the indicated file for writing (in binary / ASCII mode)
+    outfile.reset(pIOSystem->Open(pFile, toAscii ? "wt" : "wb"));
+    if (!outfile) {
+        throw DeadlyExportError( "could not open output .m3d file: " + std::string(pFile) );
+    }
+
+    // use malloc() here because m3d_free() will call free()
+    m3d = (m3d_t*)calloc(1, sizeof(m3d_t));
```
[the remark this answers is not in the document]
bztsrc: M3DExporter is actually a wrapper around a C API, therefore what you ask is impossible. Calloc is the proper way, there's no such thing new or delete in C.

code/M3D/M3DExporter.cpp
```diff
@@ -0,0 +1,394 @@
...
+    // explicitly release file pointer,
+    // so we don't have to rely on class destruction.
+    outfile.reset();
+}
+
+// ------------------------------------------------------------------------------------------------
+// recursive node walker
+void M3DExporter::NodeWalk(const aiNode* pNode, aiMatrix4x4 m)
+{
+    unsigned int i, j, k, l, n, idx;
```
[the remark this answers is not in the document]
bztsrc: I've minimized already, and the code passes the Quality Review. If they are initialized there and the Quality Review said it's fine that's because it is fine. Just saying.

code/M3D/M3DMaterials.h
bztsrc: The M3DImport and M3DExport are not the only files that include this conversion table. The C part also needs it, it can't be a .cpp file. Defining it const causes compilation errors, I've tried. (Then I accidentally removed the const keyword from the pKey too, which travis spotted and I've fixed.)

code/M3D/m3d.h
bztsrc: Ehem, these are implementation specific defines...

code/M3D/m3d.h
bztsrc: Please read it again. This define is about NOT to use the register storage class in C++ code. However in the C API it eliminates the entire stack frame in certain quite often called functions resulting in drastic speed up.

code/M3D/m3d.h
bztsrc: Another C definition, which would not compile correctly with const. There's a reason for static, and the ifdef guards. Remember, this code has to be compiled with C++ and ANSI C compilers too. Assimp only checks for C++.

code/M3D/m3d.h
bztsrc: As the comment says, it is not the entire header, just part of it (which by the way is also needed by this format, not just for PNG). And about including, the purpose of dependency-free single file headers is exactly that, "not to include as a separate file". Don't be afraid, the ifdef guards are there, so if it is included in a separate file it will work as expected.

code/M3D/m3d.h
bztsrc: I'm afraid you have to consult this with Sean Barrett. I'm not in a position to review or fix his code. But this particular code is used by many people in the game industry for its speed, I think they would have noticed is something is wrong with it.

test/models/M3D/README.md
bztsrc: They are from Open Source games. Should I only keep the models converted from Assimp samples? I can do that.

code/M3D/M3DImporter.cpp
```diff
@@ -0,0 +1,737 @@
...
+    ai_assert(orientid != -1U && orientid < m3d->numvertex);
+    m3dv_t *p = &m3d->vertex[posid];
+    m3dv_t *q = &m3d->vertex[orientid];
+
+    /* quaternion to matrix. Do NOT use aiQuaternion to aiMatrix3x3, gives bad results */
+    if(q->x == 0.0 && q->y == 0.0 && q->z >= 0.7071065 && q->z <= 0.7071075 && q->w == 0.0) {
+        m->a2 = m->a3 = m->b1 = m->b3 = m->c1 = m->c2 = 0.0;
+        m->a1 = m->b2 = m->c3 = -1.0;
+    } else {
+        m->a1 = 1 - 2 * (q->y * q->y + q->z * q->z); if(m->a1 > -1e-7 && m->a1 < 1e-7) m->a1 = 0.0;
```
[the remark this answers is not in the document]
bztsrc: OK

bztsrc:
Hi turol,

I've commited the changes, added unit test, clicked resolved buttons, but github is still showing "changes requested". What did I miss? Where should I click to let github know?

Cheers,
bzt

code/M3D/M3DImporter.cpp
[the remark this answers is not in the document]
kimkulling: Or you can use the assimp-global epsilon getter,

kimkulling:
Merged, thanks so much for this great work!

code/M3D/m3d.h
rbsheth: @turol Just ran into your exact hypothetical scenario. My code uses `stb_image` and then conflicts with these symbols here. Why does it matter if we include the entire header or not? You're only supposed to define `STB_IMAGE_IMPLEMENTATION` once. #2766

turol: Because at that point two symbols (functions) exist with the same name. The user code defines one set (with `STB_IMAGE_IMPLEMENTATION`, used exactly once as you're supposed to) and your copy-pasted code defines the other. If the dynamic library doesn't export these symbols (on windows it shouldn't, on Linux and OS X it's possible to do it but I don't know if that's been done here) then no problem. If does or if using static library then duplicate symbols error happens.

Two ways to fix this:
1. Rename your copy-pasted symbols so none of them clash. Leads to duplicate code, makes it hard to update it and can bloat the executable.
2. Replace your code with the real `stb_image.h`, include one object file where you define `STB_IMAGE_IMPLEMENTATION `. Document that it's in there so users aren't caught unaware. Provide CMake option to disable it in case users want to use theirs.

rbsheth: @turol Correct, we are defining it once in our user code and linking Assimp statically through Hunter, resulting in the duplicate symbol issue. To be clear, we are just consuming the latest version of assimp and are not contributing the M3D code. It looks like #2766 is implementing solution (1), which I agree is adding bloat and less maintainable. (2) would be preferred. Thoughts @bztsrc?

PR #2741, opened 2019-11-01
FBX orphan embedded textures fix

### LINKED ISSUE #2742: FBX orphan embedded textures are not imported
Embedded textures in FBX could be connected to nothing but to itself,
for instance a Texture -> Video connection only but no connection to the main graph,

muxanickms:
Embedded textures in FBX could be connected to nothing but to itself,
for instance a `Texture` -> `Video` connection only but no connection to the main graph,
It is fixes #2742
the idea here is to traverse all objects to find these Textures and convert them,
so later during material conversion it will find converted texture in the textures_converted array.
I added unit test as well.

Bonus change: I replaced some `std::map` to `std::unordered_map` that significantly speed up the import stage.


kimkulling:

PR #2744, opened 2019-11-01
Update CXMLReaderImpl.h

tanolino:
Issues with German locale number converter expecting German numbers.

In Germany we have numbers like #,## instead of #.## . Thus a unit conversion in COLLADA of 0.01 (centimeter) turned out to be 0.00.


kimkulling:

PR #2746, opened 2019-11-04
glTF2 code doesn't work when double precision enabled

### LINKED ISSUE #2745: glTF2 code doesn't work when double precision enabled
The data loads/saves corrupted because it assumes float type with 4 bytes size and glTF Spec 2.0 claims that double precision is not supported, it has to be a float type.

muxanickms:
As the title says, there are massive issues with double precision with other formats, the data loads/saves corrupted because it assumes float type with 4 bytes size.

I am trying to fix #2745 glTF2 format. The idea is to cast doubles to floats on a save and back on a load.
glTF Spec 2.0 claims that double precision is not supported, so we have to do that.
I added a unit test which is little bit useless if double precision is not enabled, but it is not completely useless.

It would be good to have double precision added to matrix on Travis and Appveyor, but it would be much bigger PR with all formats issue fixed or massive unit test disabling.


muxanickms:
@kimkulling Is any comments on this PR please?

coryf:
I believe that the min and max values for Accessor need to be stored as doubles. The glTF spec supports full uint32 values which can't be accurately represented by a single precision 32bit float (the maximum accurately represented integer value is 2^24 + 1). This is one of the issues I addressed in my PR before I realized that this PR existed. Converting uint32 values to 32bit floats and back will cause precision issues.

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#accessorcomponenttype-white_check_mark

muxanickms:
@coryf I am not sure I follow your idea, I am trying to address the crash when it writes .bin file using doubles which is not supported by glTF spec 2.0.
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#floating-point-data
```
Floating-Point Data
Data of 5126 (FLOAT) componentType must use IEEE-754 single precision format.

Values of NaN, +Infinity, and -Infinity are not allowed.
```

coryf:
@muxanickms Sorry for adding confusion to the discussion. I was describing an additional issue related to precision problems that is specific to the Accessor.min and Accessor.max values. I believe that what you have in this PR is also necessary to fix precision issues specific to the ExportData function. The fixes for Accessor.min and Accessor.max are in: https://github.com/assimp/assimp/pull/2779

kimkulling:
One question from my side: does it make sense to support double-precision for the glTF-expoter in common? For me this seems to be a hack to make it work for it. Maybe we should just enable it for a dedicated list of importers / exporters before trying to build such hacks.

So my feedback: maybe we shall only support double precision when the format provides it in a usefule way and when there are some real requirements for it.

Any comments?

muxanickms:
@kimkulling The scenario here is having FBX file with double precision positions and I want to convert it to glTF file. But glTF 2.0 Spec explicitly says that double precision doesn't supported, so one must use floats as positions. After importing FBX file, aiScene has all positions in aiVector3D as doubles and to write them to glTF we need to convert it to float somewhere (not using aiVector because it's in doubles). So I do it on writing buffers, so for text glTF it doesn't matter because float/doubles are okay there, it only matters for binary data.
It could be very useful to have different precision in the same scene, like explicitly aiVector3Dd, aiVector3Df or something similar.

muxanickms:
@coryf should be min/max be of the same type as data? float for float, uint for uint?

muxanickms:
@kimkulling One thing to say here, if double precision define is not enabled then code works as before, no changes.

kkulling:
Ok, let's close this pull request. I will add a new issue to disable double-precision export for glTF at all.

kkulling:
See: https://github.com/assimp/assimp/issues/2826

And thanks a lot for your work. Helped me a lot to work on a deeper understanding of the glTF and glTF2.0 format.

PR #2751, opened 2019-11-05
Fix for exporting fbx bigger than 2GB

### LINKED ISSUE #2749: It is not possible to export aiScene to file bigger than 2GB on Windows with Visual Studio
Implementation of `DefaultIOStream` uses `long` type which is 4 bytes on Windows, so it should be fixed.

### LINKED ISSUE #2750: It is not possible to export aiScene to FBX that is bigger than 2GB
By default FBXExporter uses FBX SDK 7.4 and it's 32-bit, starting from 7.5 it uses 64-bit sizes and offsets. It would be good to move defaults to 7.5 because it's very little overhead with great benefits of large files.

muxanickms:
Here are two issues fixed:

1. It fixes #2749 , on Windows with 64-bit it now uses `ftelli64` and `fseeki64` to fix a `long` type is being 4 bytes on Windows

2. it fixes #2750 , FBX Exporter uses a version 7.4 by default that not supports 64-bit offsets, I changed the version to 7.5 and fixed relevant code, so now it's possible to export FBX bigger than 2GB and it is very small overhead to use 64-bit offsets in general.

### DIFF
--- a/code/Common/DefaultIOStream.cpp
+++ b/code/Common/DefaultIOStream.cpp
@@ -52,6 +52,35 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 using namespace Assimp;
 
+namespace
+{
+    template<size_t sizeOfPointer>
+    size_t select_ftell(FILE* file)
+    {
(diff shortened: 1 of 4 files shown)


kimkulling:

PR #2752, opened 2019-11-06
avoid weighting vertex repeatedly when joining identical vertices

thomasbiang:
When export fbx with flag aiProcess_JoinIdenticalVertices, multiple original vertices will be combined into one final vertex. When these multiple original vertices are bind to a bone, the final combined vertex should be weighted similarly and it should NOT be weighted multiple times, due to the many vertices referring it. 

This pull request is to fix this problem that the combined vertex being weighted more than once and caused bad shapes in result scene.


rbsheth:
@kimkulling Thoughts on this PR?

kimkulling:
Looks good for me. Merge will be performed, when CI is ready :-).

kimkulling:

Issue #2753, opened 2019-11-06
m3d.h redefines stb image functions. 


concubicycle:
Around like 1022 of m3d.h, the file checks for the stb image header gurad and defines some stb image functions. This breaks my build. Shouldn't it be checking for the STB_IMAGE_IMPLEMENTATION define? 

In general, sticking stb functions into an h file like that without doing whatever preprocessor plumbing that stb does seems like a bad idea.

quantumsheep:
Same issue! If M3D requires implementations from `stb_image.h`, maybe including the whole header or whole `stb` project in assimp would a better solution.

`stb_image.h` doesn't have a specific pre-processor to tell if the PNG compressor is already implemented so we can't fix it that way.

bztsrc:
Could you please copy the exact error message? It works for me.

m3d.h does not require stb_image.h, and it does check for the STBI_INCLUDE_STB_IMAGE_H define, meaning if you include stb_image.h it won't. Checking for the STB_IMAGE_IMPLEMENTATION define would result in incorrect behavior, that's why it checks for the include's define.

Just for the records, this file format uses the same compression as PNG, therefore it needs a few functions also used for PNG, and it needs those without stb_image.h.

bztsrc:
I was wondering, when and how do you get these errors.

Assimp compiles without a problem as it does not include stb_image.h, and your code (which includes stb_image.h) should be _linked_ with libassimp, but in linking C header files does not play at all. So how exactly can a shared/dynamic library have a conflict with a header file?

Are you trying to compile some patched and modified version of Assimp inlined directly in your project? If so, simply include stb_image.h _before_ the Assimp headers, and the ifdef guards will take care of the rest.

quantumsheep:
I'm building and linking assimp with CMake. Building's done like that:
```cmake
set(ASSIMP_BUILD_NO_EXPORT ON CACHE BOOL "" FORCE)
set(ASSIMP_BUILD_TESTS OFF CACHE BOOL "" FORCE)
add_subdirectory(assimp)
```
and `stb_image` by including it in the `target_include_directories` directive.
```cmake
target_include_directories(${PROJECT_NAME} PUBLIC
    "${CMAKE_CURRENT_SOURCE_DIR}/include"
    "${CMAKE_CURRENT_SOURCE_DIR}/stb"
)
```

All the sources from `assimp` and `stb` are not modified.

bztsrc:
[...]

PR #2766, opened 2019-11-18
Updated M3D SDK and some fixes

bztsrc:
M3DImporter / M3DExporter wrappers updated for the latest SDK. Improved texture import, and fixed a minor texture related bug in materials. URL also updated to the new, final homepage.


rbsheth:
@bztsrc Curious why you are copying the `stb_image` functions into your header here? Without your prefix commit, I get linker errors when trying to link the latest `assimp` since I use `stb_image.h` in my library, not `assimp`. It isn't just for those who do not know how to use include. 😀 

Do you really need to copy those functions into this header?

bztsrc:
Hi @rbsheth,

Because the M3D SDK is an stb-style dependency-free single header file. It is supposed to be used by its own.

Including it into Assimp is a convenience so that you don't have to manage two code bases. Even if the M3D SDK changes, it's interface that was M3DImport built upon won't. Unless there's a bug (hopefully this was the last), no need to change the Assimp wrappers.

About the commit's note, that's because M3D SDK is only using it's built-in zlib if you don't include stb_image.h (and Assimp btw don't). If you would include stb_image then the SDK would rely on those, and would not use it's built-in version.

Cheers,
bzt

bztsrc:
Dear @kimkulling,

This patch also fixes the issue https://github.com/assimp/assimp/issues/2753 some people are having. The issue only exists if Assimp is statically linked with a project that also contains stb_image. (I honestly don't understand why to statically link against Assimp, as a big and rapidly evolving file format library always better to be linked dynamically. Their projects are using other DLLs anyway, one more wouldn't be a problem). There's no issue if one link Assimp dynamically, or if they compile Assimp directly into their project and they include stb_image before the Assimp headers, I've tested. But this fixes that particular static link issue too, just to be sure.

Cheers,
bzt

kimkulling:
Using a static-linked lib is quiet normal if you want to deploy one executable into production. So doing this is in some companies the default way. 
Thanks a lot for the pull request, I will try to merge it in the evening.

rbsheth:
Yes, we use static linking for all libraries in Hunter by default for this reason. Thanks for the explanation and fix @bztsrc

bztsrc:
Hi,

Thank you very much!

I see, I just honestly don't understand. This comment is going to be totally unrelated to this particular patch, I'm just trying to figure out what's the point in statically linking Assimp.

I think there are basically two kinds of applications that need 3d models. First group is the ones that need to support as many different formats as possible, like modellers, editors, viewers, converters etc. The second group want performance, therefore specializes in exactly one model format, like game engines.
Now for the first group, there's no point in using a statically linked Assimp, because  dynamic linkage means replacing a single DLL (or even better, upgrading a single package with the package manager) will make all the installed applications to support new formats automatically.
For the second group, there's also no point in a statically linked Assimp, because if you limit it to one format only, you'll still have to deal with a universal in-memory structure, which is never as performant as a specialized format that maps the model's structures to the engine's structures directly.

So either way, statically linked Assimp will involve all the disadvantages of handling multiple file formats  (you have to deal with a structure which most probably not aligns with your engine one-to-one), while you essentially cut yourself off from all the advantages (you must maintain, keep track of versioning, recompile and reinstall ALL your applications to support new file formats or include fixes).

Therefore I was wondering, is there any advantage of statically linked Assimp what I'm missing, or is it just "my stupid boss said so"? (I imagine not much employee would admit the latter :-))

Thanks,
bzt

kimkulling:
No, it is just:
- I do not want to deal with incompatible ABI-versions of different assimp-dlls, because Kim broke this stuff
- I do not want to deal with ABI-impompatible runtimes
- I do not want to take care about deploy all 1000 dll's when moving this to another installation folder ( ever deployed a QT-based aplication to an embedded platform, that really sucks! )
- I do not wat to loose performance by relocated calls into a dll during critical runtime issues.

kimkulling:

bztsrc:
@kimkulling: you're welcome!

Oh I see, thanks for the info! So most of the reasons leads back to workaround the lack of proper package management. Sounds terrible, but understandable. I haven't thought of that indeed, but makes sense now.

No, I haven't deployed any QT-based app on embedded platform, but I had to maintain mission critical HA firewalls with the same rulesets and I also had to deploy webpages to load-balanced servers dynamically, so I guess I've a clue how difficult your work could be.

Anyway, thank you for this wonderful library!

Cheers,
bzt

rbsheth:
@bztsrc I think you're being a little naive. There are many valid cases for linking statically, and not every project would be using Assimp for the two use cases you described. What if I want to deploy a single binary to consumers, as was mentioned earlier?

It is also funny that you mention proper package management. That is exactly what our project https://github.com/cpp-pm/hunter is supposed to do, and I added Hunter support to Assimp myself so that users of the package manager would not have to deal with dependency conflicts. Quite simply, the M3D header needed to be changed to maintain compatibility with other libraries using `stb_image`. Happy to chat more to find alternative solutions if you'd like.

bztsrc:
Hi @rbsheth,

Thanks for your answer!

I'm everything but naive. I've spent 20 years of my life to work with badly designed software. Then I got fed up, and 10 years ago I've decided to tell them if they are wrong, and for Open Source, I even fix that. Not everybody likes the truth, sure, but I'm convinced that's the right way, and this is more benefitial on the long run.

Your argument about deploying a single binary to customers is just a necessary evil because the OS lacks proper package management. There's no doubt.

Good for you that Hunter has package management, but the truth is, it is just another workaround, as it is definitely not your project's duty, it should be handled by the operating system. Ask Tannenbaum if you don't believe me. If the OS could keep the libraries up-to-date, then there would be absolutely no need for your project to do that.

I'm also perfectly aware how stb_image works, and how the C pre-compiler works in general. I've only modified M3D because YOU haven't included stb_image. If you were, then there would be no conflict. But you haven't. I decided to provide a workaround by prefixing stbi_zlib, because many others haven't included stb_image as they should with a header-only library.

I'm not interested in alternative solutions thank you. The M3D SDK works exactly as it supposed to. The fact that if you include it in a library without proper C-precompiler directives could cause conflict was expected, and worked around anyway.

Let's be perfectly clear: this is not an M3D SDK issue, this is an issue with C-header libraries in general, you would face exactly the same error if you'd include stb_image implementation in ANY two object files and you'd try to link them together. Only one of them supposed to have the implementation define, not both. It is a well known fact, I don't need alternatives to that.

Cheers,
bzt

PR #2769, opened 2019-11-18
Error string of Importer should contain a message in case of an exception

muxanickms:
It's currently empty because it swallows the exception at the end of reading file, I would like to preserve it to process it further in my code.
I added unit test as well.


code/glTF2/glTF2Asset.inl
```diff
@@ -270,13 +270,14 @@ Ref<T> LazyDict<T>::Retrieve(unsigned int i)
         throw DeadlyImportError("GLTF: Object at index \"" + to_string(i) + "\" is not a JSON object");
     }
 
-    T* inst = new T();
+    // In case Read method throws an exception this will not leak
+    auto inst = std::make_unique<T>();
```
turol: `std::make_unique` is C++14 and this project is still C++11.

You seem to be drifting away from the initial purpose of the PR. Either make a separate PR of gltf improvements or use another format for your unit tests.

muxanickms: Oh, my bad, I can do separate PR for this leak.
I really need to use glTF here to be sure it will not be broken in the future.
Do you have something in mind how can I fix this leak? I can introduce simple struct that emulate unique_ptr, deletes pointer of destructor.

muxanickms: I also could construct `unique_ptr` without helper, it is C++11

muxanickms: I made #2770 PR

kimkulling:
Looks good so far. Let's start to work on the CI to make this flying.

kimkulling:

PR #2770, opened 2019-11-19
Fix for memory leak in glTF2 Importer if an exception has been thrown

muxanickms:
Raw pointer wasn't correctly deleted if Read method throws an exception.
Unique ptr constructor is used here instead of `make_unique` because of C++11 support.


code/glTF2/glTF2Asset.inl
```diff
@@ -270,13 +270,14 @@ Ref<T> LazyDict<T>::Retrieve(unsigned int i)
         throw DeadlyImportError("GLTF: Object at index \"" + to_string(i) + "\" is not a JSON object");
     }
 
-    T* inst = new T();
+    // Unique ptr prevents memory leak in case of Read throws an exception
+    auto inst = std::unique_ptr<T>();
```
turol: Don't you need to actually put something in `unique_ptr`? It should default-construct as empty (effectively nullptr).

muxanickms: you're right, thank you!

kimkulling:
Looks good, will merge it after CI is ready!

kimkulling:

I will test if some glTF2.0 issue, which had already issues with memory-leaks, are still there.

PR #2776, opened 2019-11-21
Added CMake option to set the compiler warning to max (-Wall, /W4). Off by default

dylankenneally:
# On this PR
- I've added an option to set the compiler warning level to `-Wall` / `/W4`. 
- The option is named `ASSIMP_ERROR_MAX`
- It's turned **off** by default (same as the existing `ASSIMP_WERROR`)

## Motivation
- I tend to use all compiler warnings in my projects, in particular on Windows.
- This above step allows me to turn them on
- I intend to review & rectify the warnings this generates (if appropriate)

## Things to note
- `/W4` (MSVC) will generate compiler warnings with this - so if you use it with `ASSIMP_ERROR_MAX`, the build will fail...


kimkulling:
This one is good. I guess we shall enable this as the default otion to have all warnings enable per defaut in the CI-buildchain. Bu this is a lot of work to complete

kimkulling:

PR #2779, opened 2019-11-22
Fix glTF validation error related to accessor min and max values

coryf:
This fixes two issues with the accessors in glTF files:
1. The specification states that max and min values for integer types should not write a fractional part to the file.
2. The accessor max values were capped at `65535.0` when the accessor type was `ComponentType_UNSIGNED_INT`, which can represent any value `< 4294967295`.

Issue 2 was causing validation errors with models that had vertex indices with values greater than 65535.


coryf:
This duplicates some of the effort in https://github.com/assimp/assimp/pull/2746/files. I can merge master back into this PR after that one is merged.

RevoluPowered:
@kimkulling can you please take a quick look to see if you think this is OK?

kimkulling:
Looks good for me. First I was a little bit confused by the fact that the minimum and maximum will be moved to double. As far as I know glTF2 is not supporting any double coordinates. BUt I just saw that the list of provided components includes double.

And doing this with templates makes sense from my point of view. Thanks a lot!

code/glTF/glTFExporter.cpp
```diff
@@ -173,6 +174,62 @@ static void IdentityMatrix4(glTF::mat4& o)
...
+template<typename T>
+void SetAccessorRange(Ref<Accessor> acc, void* data, unsigned int count,
+	unsigned int numCompsIn, unsigned int numCompsOut)
+{
+	ai_assert(numCompsOut <= numCompsIn);
+
+	// Allocate and initialize with large values.
+	for (unsigned int i = 0 ; i < numCompsOut ; i++) {
+		acc->min.push_back( std::numeric_limits<double>::max());
+		acc->max.push_back(-std::numeric_limits<double>::max());
```
muxanickms: There is also `std::numeric_limits<double>::min()`, could be better than `-`

code/glTF2/glTF2Exporter.cpp
```diff
@@ -152,6 +153,62 @@ static void IdentityMatrix4(mat4& o) {
...
+template<typename T>
+void SetAccessorRange(Ref<Accessor> acc, void* data, size_t count,
+	unsigned int numCompsIn, unsigned int numCompsOut)
+{
+	ai_assert(numCompsOut <= numCompsIn);
+
+	// Allocate and initialize with large values.
+	for (unsigned int i = 0 ; i < numCompsOut ; i++) {
+		acc->min.push_back( std::numeric_limits<double>::max());
+		acc->max.push_back(-std::numeric_limits<double>::max());
```
muxanickms: There is also `std::numeric_limits<double>::min()`

muxanickms:
I agree with your change, it seems like casting everything to double and then back on writing is little bit non-optimal solution, but I can see that it's only for min/max, so not much performance decreased.
Would be good to see unit test which works with your code but doesn't work without your change, is it possible to add?
I am going to merge your change with some modifications in my PR and I would be glad to have a unit test to be confident that I preserve the behavior.

code/glTF/glTFExporter.cpp
coryf: `float` and `double` `min` is different from integer mins. It's actually the "smallest" accurately represented non-zero number and its not negative. It is super confusion and a bad design choice in C/C++ stdlib, but that's the way it is.

code/glTF2/glTF2Exporter.cpp
coryf: `double` min isn't negative.

code/glTF/glTFExporter.cpp
muxanickms: you're right , it's `lowest`

coryf:
@muxanickms I agree that using a double is not necessarily optimal, but it's better than using a float, which couldn't accurately represent a 32 uint. It seemed like a smaller modification to the code to switch to double. I think the better solution for both of our PRs would be to have a glTF variant type wrapper class that could be used to maintain the original type throughout the code, and only convert at the edges. It would also clean up the C++ templating so that it could be isolated only in the glTF type class.

coryf:
@muxanickms I will attempt to add some unit tests, but I'm not certain when I'll have enough time to look into it.

code/glTF/glTFExporter.cpp
coryf: `lowest` would work, but it was added in C++11. I'm not sure what version of C++ this project is targeting actually.

PR #2780, opened 2019-11-22
Add function aiGetVersionPatch() to be able to display Assimp version as in Git tags

LoicFr:
Hello,

Currently it is only possible to display "Assimp 5.0 (commit XXXXXXX)". The patch version (3rd number) is missing. This pull request just adds a function to be able to display "Assimp 5.0.0 (commit XXXXXXX)" (or "Assimp 4.0.1 (commit eb8639d)" like the Git tag of 28th July 2017).

Loïc


kimkulling:
Good idea, thanks a lot for the contribution.

Issue #2718, opened 2019-10-18
glTF UV flipping and corruption in 5.x.x vs 4.x.x


ardenpm:
After switching to 5.0.0 I'm seeing the V component of UV coordindates in glTF files are getting flipped. When creating a test case to reproduce I also seem to see more general UV corruption.

I've attached a simple OBJ file with a UV test pattern. To reproduce convert to glTF using the assimp.exe program like so:

`assimp export box.obj box.gltf`

In 4.1.0 I see the correct result, like this:

![image](https://user-images.githubusercontent.com/5742686/67066561-d6b86600-f1be-11e9-9b38-87f682700d19.png)

However in 5.0.0 I am getting this:

![image](https://user-images.githubusercontent.com/5742686/67066575-e3d55500-f1be-11e9-9227-c8ae39197939.png)

I'm not sure why this one shows corruption rather than a simple flip of the V component which is what I see on more complex models.

There is explicit code for flipping the V component in the exporter, which is correct since that is required for glTF, however that code has not changed between the versions tested above and a quick look through the diffs doesn't reveal anything obvious that would cause this.

[box.zip](https://github.com/assimp/assimp/files/3742351/box.zip)

ardenpm:
It seems my test case is actually exposing a different bug in the OBJ importer. I did a `git bisect` session and tracked that to commit 9a533d826ad6f880ded45c801d2a022f6d9e4f68 which changes the way UVs are read from OBJ files. I believe that should be filed as a separte bug. In the meantime I will now devise a new test to try and reproduce this issue so I can run another bisect and find it.

ardenpm:
Ok, another bisect session now and I have isolated the issue to commit 3d0775eebc0f2a86b87752385782e9b90c83279f and it seems to be related to vertex joining somehow. Will dig deeper now that I have an exact commit.

ardenpm:
This issue is that commit 3d0775eebc0f2a86b87752385782e9b90c83279f introduced a duplicate call to the exporter, so basically the whole exporter was being called twice. This seems to create issues due to the complex handling of pre-processing flags, verbose scenes and other aspects which I don't really understand. Either way I assume the exporter should be be being called twice so I will submit a pull request to remove that.

Looks like this may have came about due to mismatching indentation so it wasn't so easy to spot the duplicate line.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/2782

Issue #2119, opened 2018-09-04
GLTF KHR_texture_transform


Danny-Kint:
Any plans to support this [feature](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_texture_transform)?

ardenpm:
I just came across this also testing our app against the 2.0 sample model repository in the TextureTransformTest. Would be good to at least store the properties. I guess they would need to be stored as material properties related to textures similar to the wrapping modes?

Had a quick look at what might be involved in implementing it and since most of the extension detection stuff is there it looks like first the properties would need to be added to the glTF2Asset and then converted into material properties by flattening the offset, rotation and scale into something storable in the existing AI_MATKEY_UVTRANSFORM_BASE key.

I wouldn't like to get too far into it before hearing from someone if this might be the right approach.

Starnick:
Sounds like a good feature to support IMO

sayduck-daniel:
@kimkulling I am indeed interested in this feature.

Silvaire:
I also believe this extension is a very good one, with most glTF applications supporting it already.

silvainSayduck:
I also believe this extension is a very good one, with most glTF applications supporting it already.

corporateshark:
@kimkulling I'm interested as well. 

It seems `AI_MATKEY_UVTRANSFORM_BASE` is unused in the glTF2 loader. Is it possible to put the combined UV matrix in there?

kkulling:
I am currently working on this feature

ardenpm:
Excellent! Just a heads up, I imagine it will be tricky to test this feature until issue #2718 is also addressed since the underlying UVs won't be correct. I'm not sure if people are reproducing that issue as no one else commented on it yet.

ardenpm:
Unfortuantely there was a small issue with this. The y translation was taking its value from the X translation. I will submit a pull request with a change to fix that.

ardenpm:
Have been doing further testing and unfortunately the rotation values are not coming through. I haven't been able to determine why yet and will need to continue investigating. The code looks right, I tried some different ways to do the same thing but they don't make any difference.

ardenpm:
[...]

PR #2782, opened 2019-11-25
Migenius fix doubleexport

ardenpm:
Remove duplicate call to the exporter function in the Exporter base class. This fixes issue #2718 though it is not clear why, the organisation of pre-processing steps, verbose scenes and other factors and the fact that the scene is duplicated during export seem to affect this.

This should ideally be merged prior to the release of 5.0.1 since basically from what I can tell this means the full export is being called twice, for large files that would significantly increase export time and as the linked issue shows, created at least one unforseen issue. It was hard to track down.

I don't believe this was intentional, the line prior to the offending line was indented differently and it wasn't clear until the indentation was unified that the issue was there. It was introduced by commit 3d0775eebc0f2a86b87752385782e9b90c83279f and the commit messages by @ywangsc did not mention a need to export twice so I assume it was an error.


ardenpm:
Added fix for y UV translation to this PR which fixes a problem with the fix for #2119.

ardenpm:
Added another commit here to fix issues retrieving aiUVTransform values.

Issue #2781, opened 2019-11-24
M3D/m3d.h: Comparison always false


tellypresence:
Clang warning about comparison of -1 with char type in 2 locations in m3d.h:
```
In file included from assimp/code/M3D\M3DImporter.cpp:59:
In file included from assimp/code/M3D/M3DImporter.h:51:
assimp/code/M3D/m3d.h:2574:51: warning: result of comparison of constant -1 
    with expression of type 'char' is always false 
    [-Werror,-Wtautological-constant-out-of-range-compare]
        if(model->errcode == M3D_ERR_ALLOC) { M3D_FREE(pe); goto memerr; }
assimp/code/M3D/m3d.h:3228:47: warning: result of comparison of constant -1 
    with expression of type 'char' is always false 
    [-Wtautological-constant-out-of-range-compare]
        if(model->errcode == M3D_ERR_ALLOC) goto memerr;
```

bztsrc:
You can safely ignore that warning.

Clang is wrong, the C type "char" is signed (POSIX limits.h defines it to be), therefore "errcode == -1" is perfectly valid. It could be an issue if it were unsigned (not really, as the code emitter would generate exactly the same x86 instructions for both, cmpb does not care about signess), but in m3d.h line 441 you can see that the errcode is declared without the "unsigned" keyword, therefore ranges from -128 to 127 (aka. -1 is inside the range). By the way, I don't get that warning with gcc, g++ nor with Clang, and I've used "-ansi -Wall -Wextra -pedantic".

Regardless thanks for the report, I'll add the "signed" keyword.

Cheers,
bzt

kimkulling:
So I will close this issue.

PR #2783, opened 2019-11-25
MinGW support, profiling and signed error code

bztsrc:
Hi,

Updated m3d.h to the latest, most changes are irrelevant to Assimp (as M3DImport uses Assimp's post-processing, the minor performance tweaks in the SDK's doesn't matter), but this patch also solves issue https://github.com/assimp/assimp/issues/2781 about a Clang warning.

Cheers,
bzt


kimkulling:

PR #2792, opened 2019-11-27
Migenius fix dracocrash

### LINKED ISSUE #2791: [glTF 2.0] Crash when loading Draco compressed files
Currently Assimp will very likely crash or encounter undefined behaviour if a glTF has Draco compression enabled and required. While Draco is not supported (hopefully at some point we'll have that see #2195), in the meantime Assimp should likely check if `KHR_draco_mesh_compression` is present in the `extensionsRequired` array.

The specification states that if only Draco data is provided (it is possible to have a file with both compressed and uncompressed data) then the extension must be listed in `extensionsRequired`. So looks like it would make sense to explicitly check for this and throw and error if encountered for now to avoid crashes.

It may be worth exposing the `extensionsRequired` array in some fashion so callers can determine for themselves for other extensiosn whether they are able to support them, since it may depend on the engine it is coming into.

ardenpm:
Add reading of extensionsRequired so that we can check if Draco support is required to load the file at all. Fixes #2791.


kimkulling:

Issue #2785, opened 2019-11-26
Optimisation is enabled in debug build targets on Visual Studio


ardenpm:
I was having all sorts of odd issues debugging and it took me a while to work out the debug builds in Visual Studio (I tested 2013 and 2015) from solutions generated with CMake and using the 'Debug' build target have optimisation turned on, which completely breaks debugging.

Not sure if this is related to #2625. I am not very experienced with CMake so am not sure of why this occurs, a quick look through the CMake files showed options to disable optimisation for debug builds but it seems they are not being applied.

Unclear whether this affects Linux or not as I did not test there.

RevoluPowered:
When you build in MSVC can you please do a clean rebuild and send across the build log file?

ardenpm:
[build.log](https://github.com/assimp/assimp/files/3894061/build.log)

![image](https://user-images.githubusercontent.com/5742686/69673788-3de70580-10ef-11ea-9d00-938c42f148a0.png)

kimkulling:
I guess this is just a bug in the debug-specific cppflags setting in the CMake-configuration.

I will check that.

PR #2801, opened 2019-12-02
MSVC: Disable optimisations in debug build

### LINKED ISSUE #2799: MSVC: D9002 warning in Debug builds
/O0 is a GCC flag. It is not an MSVC flag!
https://docs.microsoft.com/en-us/cpp/build/reference/o-options-optimize-code?view=vs-2019

This warning was introduced by commit 4754d2d in a failed attempt to correct issue #2785

RichardTea:
In MSVC debug builds:
Use the correct /Od flag to disable all optimizations under MSVC and remove the /Ob2 inline-functions flag.
Note: /Od implicitly sets /Ob0, however /Ob1 (inline only explicitly inline) would also make sense.

Fixes #2799 and #2785


kimkulling:

PR #2805, opened 2019-12-03
Some cleanup of M3D support

RichardTea:
This is some of the cleanup mentioned in #2803 

Wrap the m3d.h header
Note: C++11 support required to use in a threaded environment due to the lack of a cookie.
Delete the m3d buffer after saving to file.
Apply clangformat.

Other issues need upstream API and implementation changes to the "m3d.h" header.


bztsrc:
Thanks for these!
Let's talk about the upstream modifications, but keep in mind, Assimp is not the only library that uses m3d.h, there's a much bigger picture here.

bzt

bztsrc:
I'm trying to review your changes, but since you have changed the identation's size, everything is marked as changed...

You have basically created a wrapper around a wrapper, is that correct? I don't really care as long as it works!

Cheers,
bzt

RichardTea:
Apologies for the indentation change.
The CI applies the .clangformat for this project to apply the Assimp project style.

This is only a small improvement that hides the actual M3D header away from the rest of the code, and should prevent the non-re-entrancy crash.

The m3d.h header out-of-bounds buffer crash and the warnings remain, as those are SDK/API issues that need to be corrected upstream.

This wrapper is still creating multiple definitions of the M3D functions (but in different translation units) so may create linker errors in some configurations.
The M3D implementation should be solely in the M3DWrapper.cpp, so that the exporter and importer can both use a single copy of the implementation.

I could not clean that up as I don't understand the meaning of each of the M3D macros and there is only a single (very basic) test.

kimkulling:
Merged! Thanks a lot for the contribution.

PR #2810, opened 2019-12-05
Modeller meta data

MalcolmTyrrell:
Many asset formats carry informative metadata, such as a record of the software used to generate them. In the same spirit, this change adds metadata to the aiScene about the format the file was imported from, and also preserves the generator information, when available.

The "SourceAsset_Format" metadata is the name of the importer. Since that is available in one common place, I set it for all formats. The other two metadata values need to be populated by the importers themselves. I've only done this for FBX and GLTF (version 1 and 2). Support for other formats can be added as needed.

Unit tests for FBX and GLTF are included in this PR.


RevoluPowered:
please can you rebase and squash history into a single commit :)

kimkulling:

Issue #2788, opened 2019-11-27
Assimp 5 crash on pyassimp.load function


LocutusOfBorg:
Hello, looks like some files are crashing when tried to be opened, I opened a report on Debian BTS
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=944742
python3 -c 'import pyassimp;pyassimp.load("/usr/share/assimp/models/Collada/cameras.dae", processing=pyassimp.postprocess.aiProcess_Triangulate)'
Segmentation fault (core dumped)

Same happens with
python -c 'import pyassimp;pyassimp.load("/usr/share/assimp/models/Collada/cameras.dae", processing=pyassimp.postprocess.aiProcess_Triangulate)'
Segmentation fault (core dumped)

Can you please have a look?

kimkulling:
Thanks for the report. Could you provide a call-stack from the crash? Or do you have a dump-file from that crash? As far as I know the camera.dae file will be imported from the build chain for every CI-run. But I will check that!

kimkulling:
I just opened a pull-request to reproduce the issue:
https://github.com/assimp/assimp/pull/2790

umlaeute:
here's some info:

~~~
$ gdb --args python3.7 -c 'import pyassimp;pyassimp.load("/usr/share/assimp/models/Collada/cameras.dae", processing=pyassimp.postprocess.aiProcess_Triangulate)'
[...]
(gdb) run
Starting program: /usr/bin/python3.7 -c import\ pyassimp\;pyassimp.load\(\"/usr/share/assimp/models/Collada/cameras.dae\",\ processing=pyassimp.postprocess.aiProcess_Triangulate\)
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7645f3f in I_get (ptr=0x800000003f800000, size=4) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34
warning: Source file is more recent than executable.
34	  return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));

(gdb) py-bt
Traceback (most recent call first):
  <built-in method getattr of module object at remote 0x7ffff78d5d10>
  File "/<ASSIMP>/port/PyAssimp/pyassimp/core.py", line 136, in _init
    obj = getattr(self, m)
[...]

PR #2813, opened 2019-12-07
Python fixes: ctypes declarations and cosmetics

umlaeute:
this fixes the crashing #2788, where the ctypes sctruct definition used a different type than the C-declaration (`c_size_t` vs `ai_uint32`; they are the same on 32bit systems, but not on 64bit systems...)

it also renames a field `Mesh.mAnimMesh` to `Mesh.mAnimMeshes`, which makes the field-name consistent with the C-declaration (but break the API!). include  630f013  at your own discretion!

finally there's also some cosmetic changes (removal of trailing whitespace in the touched file, lowercasing the encoding declaration to keep emacs happy)


umlaeute:
@kimkulling did you intend to merge `master` into this PR (e.g. for testing), or the PR into `master`?

kimkulling:

PR #2815, opened 2019-12-08
[WIP] FBX Implemented full pivot support without fake transforms

RevoluPowered:
Pivot support was previously quite hard to maintain so we opted to improve and rewrite it the way that should work the best for assimp.

This method uses the standard pivot formula and apply's this to all node transforms as per the specifications.

This implements many standard things we were missing in FBX too relating to FBXPose, cluster handling and fixes a lot of bugs regarding importing pivots.

Previously we could not import pivots properly, now we can. This will need cleaned up and is a proof of concept for the 'production ready version'

we have known issues but we are working on them.

In future re-sampling can be removed as not required all of the time. 

todo: add code to interpolate keyframes and apply pivot xform data so that we can do this during animation import (this already happens for nodes)
todo: remove cout debugging and replace with assimp debugging.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -603,6 +603,7 @@ ADD_ASSIMP_IMPORTER( FBX
   FBX/FBXUtil.h
   FBX/FBXUtil.cpp
   FBX/FBXDocument.h
+  FBX/FBXPose.cpp
   FBX/FBXDocument.cpp
   FBX/FBXProperties.h
   FBX/FBXProperties.cpp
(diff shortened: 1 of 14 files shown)


RevoluPowered:
AppVeyor is down for me :/

RevoluPowered:
meta data test failing locally not sure why yet

RevoluPowered:
More info on pivots here:
https://github.com/assimp/assimp/pull/2815

rickomax:
Wait. Does this PR removes the "PreRotation, PostRotation, etc" extra nodes?

RevoluPowered:
> Wait. Does this PR removes the "PreRotation, PostRotation, etc" extra nodes?

It doesn't remove them, it implements them the way the FBX SDK requires rather than making extra nodes, which is not required.

RevoluPowered:
In new version which will be pushed in a while we have also got the resampling out I have moved the pivot sampling into the import_animation handler in the FBXConverter.cpp but this is not final yet as the original code there is messy.

PR #2818, opened 2019-12-09
Backfacing odd negative scale 2383

### LINKED ISSUE #2383: aiProcess_PreTransformVertices cant inverse face order ,if has Odd Negative Scale.
if there is an odd negative scale, then must inverse the face's mIndices. if not,the face will become inward.

RichardTea:
Fixes #2383 for both OptimizeGraph and PretransformVertices


RevoluPowered:
Thank you for your work :)

PR #2819, opened 2019-12-09
Update M3D ASCII exporter ident and extension

RichardTea:
Changed to match existing Assimp style.
Set extension as per:
https://gitlab.com/bztsrc/model3d/blob/master/docs/a3d_format.md


kimkulling:
Merged, thanks a lot fo rthe contribution.

PR #2825, opened 2019-12-10
MSVC workarounds

bztsrc:
Workarounds for https://github.com/assimp/assimp/issues/2803

> warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data

all of these because of "int len = strlen(str)". The return value of the strlen function now explicitly casted. (How come that only MSVC complained about this?)

> warning C4244: '=': conversion from 'M3D_FLOAT' to 'int', possible loss of data

Yes, that conversion from float to int was intentional as part of compressing the data. There's no data loss because the input has already been normalized (how can't you see that MSVC? Sarcasm). Double casted to silence MSVC.

> warning C4146: unary minus operator applied to unsigned type, result still unsigned

DA, that was the purpose! -1U must be interpreted as a binary full of 1s for the architecture's native word size, because ints are stored in 2s complement on all modern CPUs. Replaced by defines which makes the code less maintainable and less platform-independent: now instead of relying on the compiler knowing the native integer size, the code maintainer has to take care of using the right define if index size configured differently than 32 it. Clap! Clap! Well done MSVC! (yes, sarcasm)

> warning C4244: '=': conversion from 'uintptr_t' to 'unsigned int', possible loss of data

Again, intentional, exactly that was the purpose. Memory address differences bigger than 4G are not possible with valid data; it would be VERY BAD not to truncate bigger chunks than 4G. Double explicit casting to silence MSVC.

That was all, no other warning types. Let's see if this patch brings more MSVC compiler bugs to the surface!

bzt


code/M3D/m3d.h
```diff
@@ -70,11 +70,14 @@ typedef double M3D_FLOAT;
 #endif
 #if !defined(M3D_SMALLINDEX)
 typedef uint32_t M3D_INDEX;
+#define M3D_UNDEF 0xffffffff
 #define M3D_INDEXMAX 0xfffffffe
 #else
 typedef uint16_t M3D_INDEX;
+#define M3D_UNDEF 0xffff
```
kimkulling: Double defined symbol, please just define one!

code/M3D/m3d.h
```diff
@@ -70,11 +70,14 @@ typedef double M3D_FLOAT;
...
 #define M3D_INDEXMAX 0xfffe
 #endif
+#define M3D_NOTDEFINED 0xffffffff
```
kimkulling: Where is the differece between the tw others?

code/M3D/m3d.h
```diff
@@ -88,7 +91,7 @@ typedef uint16_t M3D_INDEX;
 #else
 #define _inline
 #define _pack
-#define _unused
+#define _unused __pragma(warning(suppress:4100))
```
kimkulling: One hint: you can also get aware of this warning by this:
`void foo(int i) {}`

to

`void foo(int ) {}`

code/M3D/m3d.h
```diff
@@ -2199,6 +2202,9 @@ void _m3d_getpr(m3d_t *model, _unused m3dread_t readfilecb, _unused  m3dfree_t f
     }
     if(freecb && buff) (*freecb)(buff);
 #else
+    (void)readfilecb;
```
kimkulling: Same here!

code/M3D/m3d.h
turol: This is only valid in C++ but not in C.

code/M3D/m3d.h
bztsrc: Wrong, they are two different symbols, with different values.

code/M3D/m3d.h
bztsrc: Should you have copied a few more lines from the code and you'd see the ifdef which defines the difference. You can configure the index size with a define.

code/M3D/m3d.h
bztsrc: turol is right, that wouldn't compile under C.

code/M3D/m3d.h
bztsrc: Uhm, same what? It is a way to tell the compiler that the argument is not used if you haven't configured the interpreter with the define. Not an issue with gcc and clang which understand __attribute__, and that's the reason why _unused is defined above for MSVC.

bztsrc:
The requested changes would break the code.
As I wrote in the comment, so far -1U relied on the compiler knowing the size, now there's two different defined, one is constant 32 bit in length, the other depends on the actual configuration, either 16 or 32 bit in size. Without -1U, you must do this distinction and use the proper define.

Cheers,
bzt

code/M3D/m3d.h
kimkulling: Ok, thanks for the update.

code/M3D/m3d.h
kimkulling: Ok, thanks for the update.

code/M3D/m3d.h
kimkulling: We can add just a cpp preprocessor step to make sure that it works only when we are compiling this as c++. And then this issue is resolved from my point of view.

code/M3D/m3d.h
kimkulling: Maybe just a typo?

code/M3D/m3d.h
bztsrc: no need for the ifdef, works in ANSI C too.
https://docs.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=vs-2019&viewFallbackFrom=vs-2013

code/M3D/m3d.h
bztsrc: This is for compilers that don't understand the attribute nor the pragma. Casting to void without using the result is the last resort to silence the unused argument warning. Note that casting without using the result has no side effects, and it was used commonly in early C compilers.

code/M3D/M3DWrapper.h
```diff
@@ -53,8 +53,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 
 // Assimp specific M3D configuration. Comment out these defines to remove functionality
-#define ASSIMP_USE_M3D_READFILECB
-#define M3D_ASCII
+//#define ASSIMP_USE_M3D_READFILECB
```
turol: It should be possible to set these from CMake without editing the source

bztsrc: I don't think so. If they are really as hazardous as you claim them to be, then better if there's no option in Assimp to enable them.

code/M3D/M3DImporter.cpp
```diff
@@ -557,8 +585,8 @@ aiColor4D M3DImporter::mkColor(uint32_t c) {
 void M3DImporter::convertPose(const M3DWrapper &m3d, aiMatrix4x4 *m, unsigned int posid, unsigned int orientid) {
 	ai_assert(m != nullptr);
 	ai_assert(m3d);
-	ai_assert(posid != -1U && posid < m3d->numvertex);
-	ai_assert(orientid != -1U && orientid < m3d->numvertex);
+	ai_assert(posid != M3D_UNDEF && posid < m3d->numvertex);
```
kimkulling: Please one statement per assertion test. It will get hard if we try to understand which statement has caused the failure. Thanks!

kimkulling:

PR #2827, opened 2019-12-10
Add a support for 3DSMax Physically Based Materials for FBX format

muxanickms:
Small addition to material import from FBX, to add support for 3DSMax


kimkulling:

PR #2829, opened 2019-12-12
Fix texcoord

MalcolmTyrrell:
Commit 75204e20bdc9a2a71cc1c183d14c549576e3d1f4 ("fix invalid setup for texture enum") dropped support for the matching textures with texture coordinates in GLTF2 import.

This change puts it back in, possibly in a more correct form, and adds a unit test.


kimkulling:
Seems to be a merge-issue I guess. Thanks a lot for the fix!

kimkulling:

PR #2839, opened 2019-12-18
Made the CMake config more submodule friendly

apanteleev:
One shouldn't override the /MT vs /MD setting of the parent project.


kimkulling:

PR #2843, opened 2019-12-22
Update SMDLoader.cpp

9739654:
GCC compilation on windows was failing. Better to check for Visual c++ than Win32.


kimkulling:

PR #2863, opened 2020-01-06
Fix PlyExporter to support faces with 0 vertices

Dunni:
The Ply format standard doesn't preclude Ply files from containing faces with 0 vertices. In the current existing code, faces with 0 vertices are being written incorrectly because the for loop would never write the endline character if the number of indices was 0. 
This change fixes that behavior by writing a space character before each index and then adding the endline character outside of the loop.


kimkulling:

PR #2865, opened 2020-01-07
Fix possible null pointer exception on scene metadata when exporting a glTF2 file

LoicFr:
Hello,

The glTF2 exporter crashes if the scene has no metadata (since commit 986b6780 on 9th december). I have just added a check to avoid that.

Loïc


kimkulling:

PR #2884, opened 2020-01-14
Fix Assimp patch version to match the last bug fix release

LoicFr:
Hello,

I am not sure to understand how Assimp versions are numbered. The last release is numbered 5.0.1 (https://github.com/assimp/assimp/releases) with the comment "Fix wrong version" but in the code it is 5.0.0. I propose this commit to fix the issue but I am maybe wrong.

Regards,
Loïc


LoicFr:
Maybe the 3rd digit in Assimp version number 5.0.1 must be ignored, because I see that in exported file metadata there are only the 1st, the 2nd digit and the git revision number (ex: 5.0.9a944665)

PR #2905, opened 2020-01-16
Suppressed MSVC++ warnings C4244 and C4267

matt77hias:


kimkulling:
From my point of view you haven't supressed these warnings, you have fixed them!

matt77hias:
Just wanted to be cautious with regard to the maximum allowed name sizes on 64-bit architectures. But I think those casts in these use cases are safe enough. 🙂

In other use cases, I use something explicit similar to [GSL's narrow](https://github.com/microsoft/GSL/blob/master/include/gsl/gsl_util#L108):

```c++
// without checking by the caller
template< typename ToT, typename FromT >
[[nodiscard]]
constexpr ToT narrow_cast(FromT&& from) noexcept
{
	return static_cast< ToT >(std::forward< FromT >(from));
}

// with checking by the caller
template< typename ToT, typename FromT >
[[nodiscard]]
constexpr std::optional< ToT > narrow(FromT from) noexcept
{
	const auto to = narrow_cast< ToT >(from);

	// Check whether the types have different signedness.
	if constexpr (!has_equal_signedness_v< ToT, FromT >)
	{
		// Check whether the values have different signs (while ensuring 
		// +0.0 == -0.0 for floating point types).
		if ((to < ToT{}) != (from < FromT{}))
		{
			return {};
		}
	}
	
	// Check whether the values are equal.
	if (from != static_cast< FromT >(to))
	{
		return {};
	}

	return to;
} 
```

kimkulling:
Good idea, works with c++17?

kimkulling:

matt77hias:
> Good idea, works with c++17?

Yes, `std::optional` requires C++17, but there are existing libraries (e.g., EASTL, optional-lite, etc.) that provide a pre-C++17 implementation. 

With regard to the narrow function, I prefer my optional-variant over GSL's terminating or throwing variant as it still allows the caller to handle issues using whatever strategy he/she prefers (e.g., asserts, exceptions, etc.).

PR #2906, opened 2020-01-16
[MDL] Updated header to use when checking file size.

malortie:
[Link to issue](https://github.com/assimp/assimp/issues/2898). 

Read this [post](https://github.com/assimp/assimp/issues/2898#issuecomment-575336842) in the linked issue for a detailed explanation about this issue.

Thanks to users @tellypresence and @turol for their help and improvements.

PR #2908, opened 2020-01-17
[MDL] Updated header to use when checking file size. (Redone)

malortie:
This is needed since it's possible to have an MDL sequence group file that has a size less than the Quake 1 MDL header.


malortie:
[Link to original PR](https://github.com/assimp/assimp/pull/2906).

kimkulling:

Issue #2898, opened 2020-01-16
Problems loading bundled .mdl model files


tellypresence:
Have two different errors loading bundled .mdl model files:

Model file
`test/models/MDL/MDL (HL1)/man01.mdl`
fails to load with error
`File is too small.`

Model file
`test/models/MDL/MDL (HL1)/manT.mdl`
fails to load with error
`std::exception` (that's the entire error message)

turol:
The code doesn'd handle exceptions correctly, after some tweaking I got a better error message:
```
[Half-Life 1 MDL] Model has no bodyparts
```
Looks like these models have always been broken. They were introduced by @malortie, any comments on whether they should be broken or not?

malortie:
@tellypresence @turol 

Thanks for reporting. 

I am posting this to let everyone know I have been investigating this issue and found the root cause, which I explain in my next post. The error about the exception message will also be explained in the next post.

Starting with the 2 files you mentioned, **man01.mdl** and **manT.mdl**. The files are not broken. See below.

First, the entire model "man" is made of 3 files.

* man.mdl
* man01.mdl
* manT.mdl

**man.mdl** is the main MDL file. It contains meshes; bone info; sequence description eye position; model bounds. It also contains animation data for animations that are paired with the "default" sequence group, which is always the first sequence group.

**man01.mdl** is a demand loaded sequence group file. Demand loaded sequence group files store animation data such as bone position; rotation; compressed scale.

**manT.mdl** is an external MDL texture file. StudioMDL may store textures either in the main MDL file (man.mdl) or in an external MDL file, in this case (manT.mdl). A texture file contains texture pixels, texture info - width, height, palette.

For more information about the MDL file formats, please refer to this [page](https://github.com/malortie/assimp/wiki/MDL:-Half-Life-1-file-format).

----
Secondly, as for both files **man01.mdl** and **manT.mdl**.

**man01.mdl** is a demand loaded sequence group file and this file is split in small chunks containing the rotation and position for each frame of the animation, per bone. Please see the [MDL file format](https://github.com/malortie/assimp/wiki/MDL:-Half-Life-1-file-format#Animation-frame-offsets) for more information.

[...]

PR #2909, opened 2020-01-17
Re-throw exception

matt77hias:
Fixes MSVC++ warning C4101: 'e': unreferenced local variable

### DIFF
--- a/code/FBX/FBXConverter.cpp
+++ b/code/FBX/FBXConverter.cpp
@@ -1566,7 +1566,7 @@ namespace Assimp {
             }
             catch (std::exception&e) {
                 std::for_each(bones.begin(), bones.end(), Util::delete_fun<aiBone>());
-                throw;
+                throw e;
             }
 
             if (bones.empty()) {


code/FBX/FBXConverter.cpp
```diff
@@ -1566,7 +1566,7 @@ namespace Assimp {
             }
             catch (std::exception&e) {
                 std::for_each(bones.begin(), bones.end(), Util::delete_fun<aiBone>());
-                throw;
+                throw e;
```
turol: This is wrong. It will cause any exception which has been derived from `std::exception` (like our `DeadlyImportError` to be "sliced" back into `std::exception`, losing important information about it.

The correct way to fix this is to remove the named variable from `catch` :

```
catch (std::exception& /* e */) {
``` 

or not care about the exception type at all, like this:

```
catch (...) {
```

matt77hias: One still has the exception message which is stored in `std::exception` compared to ignoring the complete exception. ¯\_(ツ)_/¯

turol: I though so too but it doesn't seem to work for some reason. See #2898

matt77hias: Thanks, it is the first time I heard about this pattern to correctly rethrow in C++. 👍

PR #2927, opened 2020-01-18
Fixed memory leak in MDL importer.

malortie:
If an exception is propagated from one of the MDL implementations and is uncaught, or because of an unknown MDL format, the memory allocated at **mBuffer** may never be freed.


kimkulling:
Merged, thanka a lot for the contribution.

PR #2928, opened 2020-01-19
[MDL HL1] Fixed children nodes not deleted.

malortie:
As the HL1 MDL loader reads the file data, it creates and stores child nodes for the scene's root node.
At the end, it transfers every child contained in the list to the scene's root node. This avoids constantly resizing root node's **mChildren** array.

A problem occurs when an exception is thrown before node children from the list are added to the scene's root node. This results in a memory leak.


code/MDL/HalfLife/HL1MDLLoader.cpp
```diff
@@ -125,12 +125,14 @@ void HL1MDLLoader::release_resources() {
         anim_headers_ = nullptr;
     }
 
-    if (rootnode_children_.size()) {
+    // Root has some children ndoes. so let's proceed them
```
malortie: **ndoes** should be "nodes".

kimkulling: Fixed!

code/MDL/HalfLife/HL1MDLLoader.cpp
```diff
@@ -374,7 +374,7 @@ void HL1MDLLoader::read_texture(const Texture_HL1 *ptexture,
     for (outheight = 1; outheight < ptexture->height; outheight <<= 1)
         ;
 
-    if (outheight > BuffenLen) {
+    if (static_cast<size_t>(outheight) > BuffenLen) {
         outheight = BuffenLen;
```
malortie: There should probably be a **static_cast<int>**, otherwise this could raise a warning when building on x64.

malortie:
@kimkulling 

There are a lot of failures when performing checks for code quality, as well as build errors. Is it possible to know why these errors were not raised in the earlier commits of HL1MDLLoader.cpp?

malortie:
@kimkulling 

If you plan on making further changes in the method read_texture() in HL1MDLLoader.cpp, please review these PR before https://github.com/assimp/assimp/pull/2942  https://github.com/assimp/assimp/pull/2943. Thanks in advance.

kimkulling:
Will do, thanks for the tipp!

kimkulling:

PR #2941, opened 2020-01-22
Fix version revision formatting in glTF metadata

LoicFr:
Hello,

In glTF and glTF2 headers there is a small error for the writing of the version revision which is not considered as an hexadecimal number. For example it is written "assimp v5.0.-1701558683" instead of "assimp v5.0.9a944665".

Loïc


kimkulling:

PR #2942, opened 2020-01-22
[MDL HL1] Removed texture dimensions limitation.

malortie:
In the original Half-Life model viewer found in the SDK, the code clamps the dimensions of each texture to ensure neither the width and the height gets > 256. [See this link](
https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/utils/mdlviewer/studio_utils.cpp#L43-L53).

This code has been used in the HL1 MDL importer to try and replicate the state of imported models. But going back on the decision, I believe this limitation should be removed, as the importer should only provide the texture to the developer and allow it to make the appropriate changes if needed.

Of note, several mods feature models that use higher resolution textures. I tested several of these models in the game engine and it appears the game engine allows textures > 256x256 to be rendered. Therefore, I currently see no reason to keep this limitation.


kimkulling:

PR #2943, opened 2020-01-22
[MDL HL1] Fixed wrong texture format used.

malortie:
In BMP files, pixel colors are stored as blue, green, red.

When compiling a Half-Life 1 model, the model compiler reads each BMP and saves each pixel color in the order RGB. [See this file](https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/utils/common/lbmlib.c#L307-L315). As seen in the link, the code stores the palette to a temporary  [RGBQUAD](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-rgbquad) array where the order of the color component bytes is BGR. The code then reads the colors using the members of the data structure in the order RGB.

Therefore, it would be more appropriate to use **RGBA8888** instead of **BGRA8888** since we are referring to the texture data stored in the MDL.


kimkulling:
Looks good, CI is on its way!

kimkulling:

PR #2947, opened 2020-01-23
Updated places where read/write for achFormatHint.

malortie:
References to `achFormatHint` have been adjusted to take the latest size of this array.

### DIFF
--- a/code/MDL/HalfLife/HL1MDLLoader.cpp
+++ b/code/MDL/HalfLife/HL1MDLLoader.cpp
@@ -367,7 +367,7 @@ void HL1MDLLoader::read_texture(const Texture_HL1 *ptexture,
     for (outwidth = 1; outwidth < ptexture->width; outwidth <<= 1)
         ;
 
-    if ((size_t)outwidth > BuffenLen) {
+    if ( outwidth > static_cast<int>(BuffenLen)) {
         outwidth = BuffenLen;
     }
 


code/Assbin/AssbinExporter.cpp
```diff
@@ -413,7 +413,8 @@ class AssbinExport
 
         Write<unsigned int>(&chunk,tex->mWidth);
         Write<unsigned int>(&chunk,tex->mHeight);
-        chunk.Write( tex->achFormatHint, sizeof(char), 4 );
+        // Write the texture format, but don't include the null terminator.
+        chunk.Write( tex->achFormatHint, sizeof(char), HINTMAXTEXTURELEN - 1 );
```
turol: I think this would be a format change. Not sure how stable the binary format is supposed to be but this probably needs at least a major version bump.

code/Assbin/AssbinLoader.cpp
```diff
@@ -535,7 +535,7 @@ void AssbinImporter::ReadBinaryTexture(IOStream * stream, aiTexture* tex) {
 
     tex->mWidth = Read<unsigned int>(stream);
     tex->mHeight = Read<unsigned int>(stream);
-    stream->Read( tex->achFormatHint, sizeof(char), 4 );
+    stream->Read( tex->achFormatHint, sizeof(char), HINTMAXTEXTURELEN - 1 );
```
turol: Same here.

malortie:
@turol 
[Link to new PR](https://github.com/assimp/assimp/pull/2948).

I will reply further on to the comments you made in the current review.

code/Assbin/AssbinExporter.cpp
malortie: @turol It looks like the changes have been [merged](https://github.com/assimp/assimp/pull/2948). I am afraid I do not know if backward compatibility is important to Assimp devs/maintainers but this could be discussed further.

PR #2958, opened 2020-01-26
Uniformized error codes (return values) in assimp_cmd.

malortie:
This could probably help with maintenance but would consequently break the original error code values. 

At the exception of 0 (success) and 1 (Invalid number of arguments),  the original error code values seem randomly chosen, but please correct me if I am wrong.


kimkulling:
Good idea, I will merge it. But before this let's take a deeper look onto the CI-builds.

malortie:
@kimkulling I will fix it.

kimkulling:

PR #2962, opened 2020-01-27
Remove explicit setting of macos install_name

### LINKED ISSUE #2961: MacOS: CMAKE_INSTALL_NAME_DIR/INSTALL_NAME_DIR is deprecated
CMake 2.8.12 added the ability to use "@rpath" for library install names via CMAKE_MACOS_RPATH.
https://blog.kitware.com/upcoming-in-cmake-2-8-12-osx-rpath-support/

The @RPATH approach is generally recommended as it allows the consuming application to specify the loader search paths. It's also easier to swap out after-the-fact via install_name_tool as the prefix is a known string.

This was enabled by default policy (CMP0042) in CMake 3.0, which is the current minimum version.
https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042

However, setting INSTALL_NAME_DIR or CMAKE_INSTALL_NAME_DIR overrides this behaviour.
Assimp shouldn't set either of these by default.

A user requiring the old behaviour can explicitly set CMAKE_INSTALL_NAME_DIR as desired.

RichardTea:
Use CMake 3.0 default install_name behaviour for libassimp

Fixes #2961


RichardTea:
I don't know if this affects PkgConfig on macOS. Does anyone use that?

kimkulling:

PR #2969, opened 2020-01-29
Import/export of embedded texture names for the glTF/glTF2 format

LoicFr:
Hi,

I noticed that the texture names of embedded textures are not read/written during import/export of the glTF/glTF2 format. For example, Blender stores such data so I propose this pull request to have the same behaviour in Assimp.

Loïc


kimkulling:

PR #2970, opened 2020-01-29
Added missing checks for tempData and uvIndices sizes in all cases

ms-maxvollmer:
Original code checked sizes only in two places. Invalid model files could cause data being written beyond allocated memory.

I also made the log messages consistent. (There were two different messages, one that logged sizes and one that didn't. I made all messages log the sizes.)


turol:
Can you also add unit tests for these?

ms-maxvollmer:
The file that caused us to find this issue is an internal model file we fuzzed, so I can't publish that. I'll try to recreate a test case from a model file I can share.

kimkulling:
Looks good, I will merge it!

kimkulling:

ms-maxvollmer:
@kimkulling We actually noticed today that two of the four checks I added are wrong (tempData in those special cases does not need to be the same size as number of vertices, instead there is already an out-of-bounds check in the for-loop that accesses tempData) I will shortly create a new PR to remove those faulty checks, as they can create false positives for perfectly valid models. Sorry!

PR #2971, opened 2020-01-29
Check input token length before copy

ms-maxvollmer:
In ParseTokenAsFloat: Added check if length of input token is actually within expected range before copying the token into buffer. If length is too big, ignore the input and simply return 0.


turol:
Can you add a unit test for this?

turol:
If we ever hit this the model is probably corrupt. Would it be better to throw an exception instead?

ms-maxvollmer:
The file that caused us to find this issue is an internal model file we fuzzed, so I can't publish that. I'll try to recreate a test case from a model file I can share.

An exception is fine as well. The important bit is that the check is there. Let me know if you want me to change it to throw an exception.

turol:
Did you use AFL? You could try if its testcase minimification tool makes the file sufficiently unrecognizable to be included.

I'd prefer an exception from a corrupt model rather than appearing to load it but producing incorrect results. @kimkulling your opinion?

kimkulling:

PR #2974, opened 2020-01-30
Renamed WriteDumb.cpp to WriteDump.cpp

malortie:
This includes as well changes to places referencing WriteDumb.cpp.


kimkulling:

PR #2978, opened 2020-01-31
Fixed mValues allocated twice in SceneCombiner.cpp.

malortie:
`dest` is allocated from `aiMetadata::Alloc`, which returns an `aiMetadata` instance will all members allocated, including `dest->mValues`. See the following code below.

https://github.com/assimp/assimp/blob/78b074ac5aa3ff3573675d49d154b358ba8e9700/code/Common/SceneCombiner.cpp#L1312-L1315

`dest->mValues` is allocated again, which makes it lose track of the previously allocated memory and results in a memory leak.


kimkulling:

PR #2984, opened 2020-02-05
Removed unnecessary checks that may result in false positives rejecting valid models.

ms-maxvollmer:
Follow-up to previous PR https://github.com/assimp/assimp/pull/2970

We actually noticed today that two of the four checks I added are wrong. (tempData in those special cases does not need to be the same size as number of vertices, instead there is already an out-of-bounds check in the for-loop that accesses tempData.) This PR rectifies this. Sorry!


kimkulling:
Not to worry, I have merged it. Thanks a lot for your contribution.

PR #2985, opened 2020-02-05
Fixed size check to use correct value

ms-maxvollmer:
In the _"ByVertice"_ case, `tempData.size()` needs to be equal to `mapping_offsets.size()`, not equal to `vertex_count`.

This is hopefully the last PR on this topic.


kimkulling:
Merged, thans a lot for your contribution.

PR #2986, opened 2020-02-05
Use the translation matrix in gltf2 cameras for aiCamera.mPosition

bubba:
In gltf2, camera positions are specified in nodes by transforms, whilst other data such as projection and perspective are stored in the camera part itself. This sets the `aiCamera.mPosition` to be the translation specified in the camera's node, if any. An example of a file which might do this can be seen [here](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Cameras/glTF/Cameras.gltf)


kimkulling:

PR #2987, opened 2020-02-06
Fixed population of bone armature and node fields for some meshes

Frooxius:
I've had some users import an FBX file that would look mangled after import. The issue was due to the mesh nodes having same name as some of the bones (joints), so they would get mapped as a joint for that mesh.

I've added an exclusion for any nodes that contain meshes from being considered a potential joint for the mapping, assuming that a single node cannot be both a mesh node and a joint (I'm not 100 % sure if this assumption is correct, but it seems so). This fixes the import issue.

If this is not the right way to fix it, I can look for alternate solution.


kimkulling:

PR #2994, opened 2020-02-09
Small changes in the CMake file

iamAdrianIusca:
- don't include HunterGate.cmake if you don't enable HUNTER package manager
- don't check for DirectX if you don't build the assimp tools or samples


kimkulling:
Merged, thanks a lot fro your contribution.

Issue #2982, opened 2020-02-04
The output of gltf is not correct.


runforu:
1. import gltf file using assimp.
2. export gltf file using assimp.
3. check the difference between exported gltf file and the original gltf file. Some parts are not in correct position and animation is not correct.

runforu:
The test model is at https://sketchfab.com/3d-models/willy-805be6f9445e42cebdc5563b571a7b96

runforu:
Or uploaded test gltf 3d model.
[willy.zip](https://github.com/assimp/assimp/files/4156263/willy.zip)

runforu:
I think it is caused by animation.

runforu:
For more information: https://github.com/KhronosGroup/glTF/issues/892

runforu:
https://d3cw3dd2w32x2b.cloudfront.net/wp-content/uploads/2015/01/matrix-to-quat.pdf

runforu:
Please review and merge my commit for this bug. https://github.com/assimp/assimp/pull/2995

runforu:
Please review and merge my commit for this bug. #2995

kimkulling:
Closed, thanks for your help.

PR #2995, opened 2020-02-10
Fix: GLTF animation works on RTS not matrix; fix matrix related bug; …

runforu:
related issue #2982


kimkulling:
Thanks for the fix. I just saw that we have a lot of duplicate code in there. So I guess we need to work on these topics as well.

kimkulling:

PR #3003, opened 2020-02-12
minor code improvements for the obj code

iamAdrianIusca:
- make use of range for loops
- make use of empty() method for the std::vector
- removed unnecessary std::string initialization
- use ' ' instead of " " for the find methods (should be "faster")
- simplified some function
- make use of emplace_back instead of push_back (should be "faster")
- other small changes


kimkulling:

iamAdrianIusca:
Your welcome. I will be back with another one.

PR #3009, opened 2020-02-14
A bug when importing multiple gltf files

inhosens:
meshOffsets are not cleared when importing multiple gltf files
In some cases it makes an exception or accessing array with wrong values.


kimkulling:

PR #3015, opened 2020-02-15
Fixed SimpleTexturedOpenGL sample.

malortie:
Several places in the sample's code were calling Unicode versions of Win32 functions with "multibyte" strings. A few changes were required to fix it.

I added a class **UTFConverter**, which handles calls to unicode/multibyte string conversions. This should help minimize the impacts on code change in case C++'s codecvt_utf8 ever changes.

In addition, seveal memory leaks have been found, but these fixes will come in another PR because it goes beyond the scope of this PR.

DevIL.lib was removed in CMakeFiles.txt, as it is unused in the sample.

Here is a list of the changes:

- Fixed MB string calls to Unicode functions.
- Added class UTFConverter to handle string conversions.
- Removed reference to DevIL.lib.
- Fixed compile warnings.


kimkulling:

PR #3029, opened 2020-02-25
Fix to read orthographic camera data

### LINKED ISSUE #3028: glTF orthographic cameras properties are never read
Currently the glTF orthographic camera properties will never be read even though the check for an orthographic camera is performed. This is because the top level `obj` variable is used to search for the properties instead if `*it` where the actual properties are found.

Will fix and do a pull request.

ardenpm:
Merge to fix missing orthographic camera data in glTF2. Fixes #3028.

### DIFF
--- a/code/glTF2/glTF2Asset.inl
+++ b/code/glTF2/glTF2Asset.inl
@@ -1087,10 +1087,10 @@ inline void Camera::Read(Value& obj, Asset& /*r*/)
         cameraProperties.perspective.znear       = MemberOrDefault(*it, "znear", 0.01f);
     }
     else {
-        cameraProperties.ortographic.xmag  = MemberOrDefault(obj, "xmag", 1.f);
-        cameraProperties.ortographic.ymag  = MemberOrDefault(obj, "ymag", 1.f);
-        cameraProperties.ortographic.zfar  = MemberOrDefault(obj, "zfar", 100.f);
-        cameraProperties.ortographic.znear = MemberOrDefault(obj, "znear", 0.01f);
+        cameraProperties.ortographic.xmag  = MemberOrDefault(*it, "xmag", 1.f);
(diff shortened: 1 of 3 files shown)


ardenpm:
Just a head up, might want to wait before merging this in as there may be one additional commit. It seems there is confusion as to whether glTF orthographic views are specified in half width/height or full width/height. Currently the code assumes half width, however the glTF reference viewer and Three.js loaders assume full width/height. I'm trying to get that clarified. If it turns out it should be full width/height this will require one additional change.

kimkulling:
See https://github.com/assimp/assimp/pull/3164

ardenpm:
I have been unable to get a clear answer on whether ortho cameras should be half width/height or not. I've pushed again to try and get a more complete answer on this.

ardenpm:
I now have confirmation that it definitely is the half width/height. Any way this request can be restored or do I need to do another?

kimkulling:
We can reopen this one. But you have to merge the master to it. That was my problem with your pull-request.

kimkulling:
This one:


Conflicting files
code/glTF2/glTF2Asset.inl

ardenpm:
I didn't get a change to fix that overnight but it looks like this has been merged, do I still need to take any action?

kimkulling:
Just check if its al right. And thanks a lot for your contribution.

ardenpm:
It looks like this commit: https://github.com/migenius/assimp/commit/fb5b01958a448e1c3dea9b6b2796b8bf9f98a5a7

Got lost somewhere in the process. This was to fix the issue where the actual data wasn't being read (`obj` should be `*it`).

kimkulling:
Right, I was not able to update your branch with the current master, so I created my own pull-request based on your input and merged that instead of yours. Then I added a link from your PR to my based on your work and the other way around. 

Thanks for your feedback on this. I am not sure why it was not possible to merge master into your work.

PR #3024, opened 2020-02-23
Removed uneeded expression in else() and endif() constructs.

malortie:
[Link to issue](https://github.com/assimp/assimp/issues/3007).

The following changes were tested against building with a master copy.

The flags that were checked were the ones checked by default + the following:

**ASSIMP_BUILD_ASSIMP_VIEW**
**ASSIMP_BUILD_SAMPLES**

The list of compilable projects in Visual Studio were the following:

- assimp
- assimp_cmd
- assimp_simpleogl
- assimp_simpletextureddirectx11
- assimp_simpletexturedogl
- assimp_viewer
- INSTALL
- IrrXML
- uninstall
- unit
- UpdateAssimpLibsDebugSymbolsAndDLLs
- ZERO_CHECK
- zlib
- zlibstatic

The solution that was generated with the new changes comprised the same number of projects as the one from the master, unmodified. All projects except **UpdateAssimpLibsDebugSymbolsAndDLLs** were built without any error. All tests were run.

**UpdateAssimpLibsDebugSymbolsAndDLLs** could not be built because of the reasons explained in issue https://github.com/assimp/assimp/issues/3004.

PR #3033, opened 2020-02-26
Raised minimum CMake version to 3.0 for assimp_cmd and assimp_view.

malortie:
This is a follow-up to PR https://github.com/assimp/assimp/pull/3024, in which I explained that I would change the minimum version of CMake to 3.0 in several files to match the changes that were originally made by OP in PR https://github.com/assimp/assimp/pull/3008.

The minimum CMake version have been raised to 3.0 in both projects **assimp_cmd** and **assimp_view**.

The same CMake configuration as in PR https://github.com/assimp/assimp/pull/3024 was used. All projects from PR https://github.com/assimp/assimp/pull/3024 are found in the solution with the current changes. All projects built. All tests were run.


kimkulling:
Finally! Thanks a lot for your contribution.

PR #3034, opened 2020-02-26
cmake: double quotes around the <path>

maquefel:
see https://cmake.org/cmake/help/latest/command/file.html#to-native-path


kimkulling:

Issue #2719, opened 2019-10-18
assimpTargets.cmake issue on Linux


umlaeute:
the cmake-configuration for 3rd-party consumers is somewhat broken on Linux.

I'm currently preparing the Debian packages for assimp-5.0.0, and test-compiling 3rd-party software (`ros-*`) that uses assimp/cmake now fails.

I narrowed it down to cmake to being able to find the `libassimp.so.5` file (which of course is present on the system).

- assimp's cmake configuration is installed into `/usr/lib/x86_64-linux-gnu/cmake/assimp-5.0/`
- assimp's shared library  is installed as `/usr/lib/x86_64-linux-gnu/libassimp.so.5`

- `assimpTargets.cmake` [calculates](https://github.com/assimp/assimp/blob/v5.0.0/assimpTargets.cmake.in#L44-L48) the `_IMPORT_PREFIX` as `/usr/lib/x86_64-linux-gnu/cmake/assimp-5.0/../../../` which in the end resolves to `/usr/lib`: https://github.com/assimp/assimp/blob/v5.0.0/assimpTargets.cmake.in#L44-L48

- for shared-lib builds on non-windows systems, a file-check is added for `${_IMPORT_PREFIX}/lib/${sharedLibraryName}`, which really resolves to `/usr/lib/lib/libassimp.so.5`: https://github.com/assimp/assimp/blob/v5.0.0/assimpTargets-release.cmake.in#L79

- since there is no `/usr/lib/lib/libassimp.so.5` file (esp. note the `/lib/lib`) , any call to `find_package(ASSIMP)` fails :-(

umlaeute:
I don't really know much CMake, but it seems to me that the proper path to search for `libassimp.so.5` would be `ASSIMP_LIBRARY_DIRS` instead.
So for now, I will just change https://github.com/assimp/assimp/blob/v5.0.0/assimpTargets-release.cmake.in#L79 to:

~~~cmake
list(APPEND _IMPORT_CHECK_FILES_FOR_assimp::assimp "${ASSIMP_ROOT_DIR}/@ASSIMP_LIB_INSTALL_DIR@/${sharedLibraryName}" )
~~~

umlaeute:
see https://salsa.debian.org/debian/assimp/blob/master/debian/patches/cmake.patch for the full patch.

The patch is of course a bit hack-and-slash, as we only need to care about "proper" installations.

vmatare:
[...]

PR #3035, opened 2020-02-26
[RFC] cmake: targets: check lib or lib64 path

maquefel:
Linkage against lib or lib64 should be taken into assumption.

Without it we get:

The imported target "assimp::assimp" references the file

     "/usr/lib/libassimp.so.5"

When compiling against x86_64 target. The library of couse exits in
/usr/lib64.

see:
https://cmake.org/cmake/help/v3.17/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html

As i am not a master of cmake this should be double checked if
it doesn't break anything.

Signed-off-by: Nikita Shubin <NShubin@topcon.com>


maquefel:
Might be related to https://github.com/assimp/assimp/issues/2719

vmatare:
> When compiling against x86_64 target. The library of couse exits in
> /usr/lib64.
> 
> see:
> https://cmake.org/cmake/help/v3.17/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html

It is stated nowhere that libs always live in `/usr/lib64` on a 64-bit system. In fact, the FHS makes pretty explicit the fact that this is "commonly used", but in no way part of the standard: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s10.html

umlaeute:
esp. Debian based systems (Ubuntu, Mint,...) use a [multi-arch approach](https://wiki.debian.org/Multiarch) that goes way beyond supporting a "32bit and 64bit variants of the same CPU" and does not use the `lib64/` folders

kimkulling:

vmatare:
@kimkulling I don't think merging this was a good idea. Just guessing that libs will be in `lib64` is plain wrong, as I've stated before and others have confirmed. The whole stuff is not just superseded by #3115, but also incompatible, in that it uses variables that aren't defined any more. That is something that should have been visible in the merge conflicts in https://github.com/assimp/assimp/pull/3035/commits/2d4eab6db087d67726da46d2bf2b4ef1cbf2b501

PR #3036, opened 2020-02-26
Minor fixes and improvements in sample SimpleOpenGL.

malortie:
The changes are the following:

- Removed references to test models with relative path.
- Fixed a compile warning.
- Added usage message.

----

In addition, the calls to functions `aiReleaseImport` and `aiDetachAllLogStreams` after  `glutmainloop` never occur because `glutmainloop` never returns, the memory allocated for the scene, as well as the logger, cannot be freed, and results in a memory leak.

One way to solve the memory leak would be to use Freeglut and then use the following:

```c
glutSetOption(GLUT_ACTION_ON_WINDOW_CLOSE, GLUT_ACTION_GLUTMAINLOOP_RETURNS);
```

With the change above, this sample would no longer be able to run on the original Glut. I have no idea how many developers still use the original Glut. A discussion about this could be made in another issue since it goes beyond the scope of the current pull request.


malortie:
Update:

According to assimp/samples/README, see below:

https://github.com/assimp/assimp/blob/0201fc57dca48910ca7f9fdf7457534ea6a57efc/samples/README#L15

OpenGL based samples are expected to use FreeGLUT. I will be able to commit changes in order to fix the memory leak I explained previously in this issue. To maintainers, please wait until I am finished committing the changes in question.

malortie:
Changes have been made to fix the memory leak. You may now review.

kimkulling:

Issue #3037, opened 2020-02-27
[FATAL] SplitByBoneCountProcess::SplitMesh goes into infinite loop


Nimer-88:
When loading glTF file SplitByBoneCountProcess goes into infinite loop.

The importer flags I'm using are `aiProcess_CalcTangentSpace | aiProcess_Triangulate | aiProcess_SortByPType | aiProcess_GenNormals | aiProcess_SplitLargeMeshes | aiProcess_FlipUVs | aiProcess_RemoveComponent | aiProcess_SplitByBoneCount | aiProcess_RemoveRedundantMaterials | aiProcess_JoinIdenticalVertices`. The model that causes the infinite loop is https://sketchfab.com/3d-models/kako-059ddf8d773748a0aa32c778897e711e.

NukeBird:
https://github.com/assimp/assimp/issues/2837

NukeBird:
actually problem is worse, it causes memory leak too

Nimer-88:
Note: it's highly likely that it's long standing issue in `SplitByBoneCountProcess` which mentioned glTF just unearthed. Since I added the test sample, it should be quite easy to repro.

kimkulling:
Thanks a lot for that.

Nimer-88:
#3038 contains the fix.

Nimer-88:
#3039 fix for glTF2 importer causing this issue

Nimer-88:
#3040 fix for the split by bone count process infinite loop

NukeBird:
Wow! You did a great job, thank you!

kimkulling:
@Nimer-88 Thanka a lot for your support. I will close it, because https://github.com/assimp/assimp/pull/3040 has closed it

PR #3040, opened 2020-02-29
Fix for #3037 [FATAL] SplitByBoneCountProcess::SplitMesh goes into infinite loop

Nimer-88:
Fix for #3037 infinite loop.

Basically if single face requires more than `mMaxBoneCount` then splitting cannot be performed.


kimkulling:
Looks good, I will merge it.

kimkulling:

PR #3048, opened 2020-03-05
Fix for memory leak in Importer and Exporter code

Nimer-88:


kimkulling:
Hi,
could you please verify what has cause the crashes which we see in the tests?

Thanks, Kim

PR #3049, opened 2020-03-05
ifdef the exporters as specifying stricter linker flags than what's in default CMake causes linking issues

Nimer-88:


iamAdrianIusca:
good catch!

code/Common/Exporter.cpp
```diff
@@ -80,32 +80,61 @@ namespace Assimp {
...
 void ExportSceneXFile(const char*,IOSystem*, const aiScene*, const ExportProperties*);
+#endif
+#ifndef ASSIMP_BUILD_NO_STEP_EXPORTER
 void ExportSceneStep(const char*,IOSystem*, const aiScene*, const ExportProperties*);
+#endif
+#ifndef ASSIMP_BUILD_NO_OBJ_EXPORTER
 void ExportSceneObj(const char*,IOSystem*, const aiScene*, const ExportProperties*);
 void ExportSceneObjNoMtl(const char*,IOSystem*, const aiScene*, const ExportProperties*);
+#endif
+#ifdef ASSIMP_BUILD_NO_STL_EXPORTER
```
turol: This one is wrong, you accidentally typed `#ifdef`

kimkulling: Fixed

kimkulling:

PR #3053, opened 2020-03-06
Fixed SimpleTexturedDirectX11 sample.

malortie:
- Removed OpenGL/Glut references in SimpleTexturedDirectX11 CMakeFile
- Moved UTFConverter from SimpleTexturedOpenGL to its own file to be able to reuse it.
- Added compile definition SHADER_PATH to allow to locate the shader files in CMakeFile (**1**)
- Fixed compile warnings
- Made global pointers null to help prevent dangling references.
- Added missing members initialization in class constructors.
- Removed references to missing model 'Models/mymodel.fbx'
- Fixed error when extracting model file directory (**2**)
- Added missing device context assignment in ModelLoader Load method (**3**)
- Fixed memory leak caused by variable 'ourModel' not deleted.
- Removed call to dev->Release() in ModelLoader.cpp
- Adjusted Release() calls in reverse order when cleaning up D3D (**4**)
- Made Throwanerror implementation throw an error instead of displaying a message box
- Fixed leaking D3D resources (**5**)
- Added a pointer to an ID3D11Debug to dump live objects.

For more information about issues, please read the following. Each number (**N**) refers to the number at the end of each entry in the list above.

(**1**) D3DCompileFromFile would return 'File not found'. I added a compile definition to allow to get the actual location where shader files are located.

(**2**) When having a Windows formatted path with double slashes, the last position for '/' could not be found, and would return the path itself rather than the path without the file name.

(**3**) The variable `devcon` in ModelLoader::Load method was never assigned, this would leave us with a dangling pointer. This caused the program to crash later because of an invalid object.

(**4**) According to this [page](https://docs.microsoft.com/en-us/windows/win32/dxtecharts/directx-9-frequently-asked-questions) - *Does it matter in which order I release DirectX interfaces?*, it is recommended to release resources in reverse order of creation. I have not yet found a place where it says otherwise for DirectX 11, but this would logically be the appropriate way.

(**5**) Several DirectX resources were not released. This has been verified by using the `ID3D11Debug` interface to dump live objects.


kimkulling:
Thanks for the fixes. I just applied the clang code style to it.

kimkulling:

PR #3055, opened 2020-03-08
Added missing std namespace prefix to std types.

malortie:
It seems that changes in https://github.com/assimp/assimp/pull/3053/commits/31c6f0db928add0e4ff22877bd4c91b5418aa2e4  break the compilation of SimpleTexturedDirectX11 sample. The instruction `using namespace std;` has been removed, but many types defined farer in the file do use standard library types without prefixing them with `std::`.

The following changes simply add the missing std prefix to variables using std types. Removing the 
`using namespace std;` instruction was a good move, since it is usually preferable to avoid importing an entire namespace.


kimkulling:

PR #3056, opened 2020-03-08
Fixed wrong matrix type used in aiMatrix3x3t comparison operators.

malortie:
The class `aiMatrix3x3t` has 2 operator overloading methods as well as an `Equal` method, which all take an `aiMatrix4x4t` as the input parameter. Please correct me if I am wrong, but I believe it was rather meant to be an `aiMatrix3x3t`.

See the files below:

https://github.com/assimp/assimp/blob/1e2c3c9c016d7a557a364161ac3cd8fbee16c72d/include/assimp/matrix3x3.h#L96-L99

https://github.com/assimp/assimp/blob/1e2c3c9c016d7a557a364161ac3cd8fbee16c72d/include/assimp/matrix3x3.inl#L144-L148

The following changes were tested with the following projects:

* assimp
* assimp_cmd
* assimp_simpleogl
* assimp_simpletextureddirectx11
* assimp_simpletexturedogl
* assimp_viewer
* unit

All tests were run.

### DIFF
--- a/include/assimp/matrix3x3.h
+++ b/include/assimp/matrix3x3.h
@@ -93,10 +93,10 @@ public:
     const TReal* operator[] (unsigned int p_iIndex) const;
 
     // comparison operators
-    bool operator== (const aiMatrix4x4t<TReal>& m) const;
-    bool operator!= (const aiMatrix4x4t<TReal>& m) const;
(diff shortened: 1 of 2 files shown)


kimkulling:

PR #3060, opened 2020-03-10
Minor changes in CMakeLists files.

malortie:
- Fixed condition preventing **ASSIMP_BUILD_ASSIMP_VIEW** showing up on first time (**1**)
- Set **ASSIMP_BUILD_ASSIMP_VIEW** to off by default.
- Removed unnecessary reference to DirectX 9 library. (**2**)

(**1**) No searches for DirectX were made before the following condition:
```cmake
IF ( WIN32 AND DirectX_D3DX9_LIBRARY )
```
The variable `DirectX_D3DX9_LIBRARY` would never be defined, and the option to build Assimp viewer could never be displayed unless you would choose to build samples, in which SimpleTexturedDirectx11 would trigger a search for DirectX.

(**2**) The variable `DirectX_LIBRARY` was referencing **d3d9.lib** which is unnecessary because the sample uses directx11.

The following projects were tested with the changes above:

- assimp
- assimp_cmd
- assimp_simpleogl
- assimp_simpletextureddirectx11
- assimp_simpletexturedogl
- assimp_viewer
- unit

All tests were run.


kimkulling:

PR #3068, opened 2020-03-16
GLTF2: ExtractData now throws exception instead of returning false if data is invalid

ms-maxvollmer:
Explanation: The return value of ExtractData is never checked anywhere in code. However if it returns false, outData remains uninitialized. All code using ExtractData assumes outData is initialized and proceeds to using it. I haven't encountered a real-life case where this goes wrong - but the simple fact that it can go wrong is a red flag. Instead of relying on every bit of code checking the return value and handling this properly, I think it makes much more sense to have ExtractData throw an exception. It obviously is an exceptional situation, and throwing makes sure that no code that doesn't explicitly handle such a scenario continues running and potentially causing harm.


ms-maxvollmer:
@kimkulling Sorry for the ping - do you think you can have a look at this PR soon? No pressure, just helpful for us internally to know when this goes in :) Thanks!

kimkulling:

PR #3069, opened 2020-03-16
GLTF2: Fixed behavior of glTF2Importer::ImportNodes

ms-maxvollmer:
Assimp guarantees that a root node is present after successfully loading a scene.

glTF2Importer::ImportNodes leaves the root node as nullptr without any error in 2 distinct scenarios. Thus it's possible to load a gltf file "successfully" but end up with a scene without a root node.

1. scenario: An invalid gltf file without a scene. I made ImportNodes throw a DeadlyImportError.

2. scenario: A valid gltf file with a scene without nodes. I made ImportNodes create an empty root node.

I added test cases with test model files for both scenarios.


ms-maxvollmer:
@kimkulling Sorry for the ping - do you think you can have a look at this PR soon? No pressure, just helpful for us internally to know when this goes in :) Thanks!

kimkulling:
Just finished the code review. Sorry for all the delays. Corona makes my days a little bit more complicated. 

Stay healthy and thanks a lot for your help!

kimkulling:

PR #3070, opened 2020-03-16
GLTF2: Detect and abort recursive references

ms-maxvollmer:
It is possible to create a GLTF2 file with recursive references. I added such a file (with a unit test).

Assimp didn't detect this and would run into an endless recursion, causing a crash.

I added a set of indices to LazyDict that keeps a record of all indices currently being retrieved. If it encounters the same index again, it will throw a DeadlyImportError, thus breaking the recursion.


ms-maxvollmer:
@kimkulling Sorry for the ping - do you think you can merge this PR soon? No pressure, just helpful for us internally to know when this goes in :) Thanks!

kimkulling:

PR #3077, opened 2020-03-21
Make file formatting comply with POSIX and Unix standards

aaronfranke:
UTF-8, LF, and no BOM are all recommended for file formatting. These settings are the most compatible with various tools and also take up the least amount of space (CRLF -> LF in a file means saving one byte per line, and no BOM saves two bytes per file). If you see an addition and removal on the first line, that's a BOM removal. If the whole file has diffs, that's likely CRLF -> LF.

The POSIX standard for text files says that each text file should end with a newline, otherwise it is by definition not a text file. GitHub shows a warning when a file is committed without a terminating newline (see the "before" diffs of this PR), and GCC shows warnings when compiling files that don't end with newlines. See [this SO question](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/25322168) and [this Unix SE question](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file) for more information. 

There are a *lot* of different file types in this repo, so I only changed it in places where I'm sure it's a good change, mostly C++ files, Java files, HTML files, markdown files, and other text files. This could likely be done for many more file types, but I don't want to break anything.


kimkulling:
I resolved the merge conflicts here: https://github.com/assimp/assimp/pull/3204

PR #3079, opened 2020-03-22
Optimized LimitBoneWeightsProcess. 

napina:
Added SmallVector to reduce heap allocations. Simplified algorithm and removed unnecessary copying.


napina:
Looks like quality review fails because parser can't find AI_LMW_MAX_WEIGHTS macro. I set default SmallVector capasity to AI_LMW_MAX_WEIGHTS * 2 to cover usual counts.

napina:
In my testing with character with 163 bones and 40k triangles this change reduces step from 16ms to 4ms.

PR #3083, opened 2020-03-23
{cmake} Prefix options to avoid pollution when included as a submodule

asmaloney:
When libraries are included as submodules in large projects, having an option with a generic name like "BUILD_DOCS" is not very helpful. (e.g. one project I work on includes many libraries as submodules)

It can also clash with options from other libraries which can break things.

So prefix externally-visible options with "ASSIMP_".

(Also fixes a small capitalization error - "DIsable" - in ASSIMP_INSTALL.)


kimkulling:

PR #3085, opened 2020-03-25
Optimized LimitBoneWeightsProcess.

napina:
Added SmallVector to reduce heap allocations. Simplified algorithm and removed unnecessary copying. In my testing with character with 163 bones and 40k triangles this change reduces step from 16ms to 4ms.


kimkulling:
What is the advantage of adding a new class calles sall-vector, which main-content is a reimplementation of the std::vector growing strategy.

kimkulling:
Just saw, that the inplace capacity in already in use for smaller vector data. Now I do understand it. Thanks for the great idea!

kimkulling:

napina:
Thanks. This technique is from https://www.youtube.com/watch?v=vElZc6zSIXM if you want to know more.

PR #3087, opened 2020-03-25
GitHub Actions and Misc. Changes

rickomax:
-Added CProgressHandlerWrapper to wrap progress-handling calls for C code
-Added aiImportFileP method, to import a file on C using a custom progress-handler
-Added aiImportFileIP method, to import a file on C using a custom file-io and a custom progress-handler
-Added aiImportFileIPP method, to import a file on C using a custom file-io, a custom progress-handler and a custom property-store
-Added aiImportFileFromMemoryP method, to import from memory on C using a custom progress-handler
-Added aiImportFileFromMemoryIP method, to import from memory on C using a custom file-io and a custom progress-handler
-Added aiImportFileFromMemoryIPP method, to import from memory on C using a custom file-io, a custom progress-handler and a custom property-store
-Removed the blend-shape weight normalization on fbx converter (weights can have values higher than 1.0)
-Added disableDiffuseFactor setting for fbx parser, to disable multiplication with the diffuse factor, since some models will have a black color when multplied with the diffuse color, in cases where the diffuse factor is 0
-Commented-out shinniness exponent reading on fbx converter, as it was being interpreted as specular factor, and it has nothing to do with the foremention value, specular info doesn't come as shinniness exponent!
-Commented on the transparency factor calculation on fbx parser. I've made many tests and this information breaks compatibility with a lot of files, also, there is no mention on this TransparencyFactor being used on any FBX importer I've used
-Replaced code declared twice to get a texture path on FBX converter by single calls to GetTexturePath
-I've set the removeEmptyBones option default value to false on fbx importer, since some empty nodes that were important for the models were being removed in some instances
-I've added a condition to the fbx scale conversion, which is only applied now when the convertToMeters option is true. That was done to prevent compatibility issues with users who werent using this feature before
-Fixed pmd parser
-Added helper to load right characters on pmd/mmd node names
-Added the material-alpha value loading for obj files, and fixed the transmission value loading
-There was an old issue on obj reader that identifies the '' character as a new-line character. Fixed it by adding a character to identify when the parser is inside a string, to avoid detecting a line-break inside it
-Added a cmakelists flag called SKIP_MSVC_UPDATE, to avoid adding suffix to generated msvc libraries
-Fixed fbx config name: IMPORT_FBX_EMBEDDED_TEXTURES_LEGACY_NAMING
-Added emscripten libraries finding on cmakelists
-Added actions to automatically build cross-platform MinSizRel libraries
-Added lightweight_build.sh script to iOS port and changed the CMAKE files that aren't generating 1gb libraries anymore, and generates smaller (<100mb) libraries instead.


RichardTea:
There's a lot of unrelated changes here, can you split it up into separate PRs?

Really big PRs that do a lot of things are hard to check over and are much less likely to apply cleanly.

It also makes it very difficult for downstream projects to decide whether they also want to accept the changes.

rickomax:
> There's a lot of unrelated changes here, can you split it up into separate PRs?
> 
> Really big PRs that do a lot of things are hard to check over and are much less likely to apply cleanly.
> 
> It also makes it very difficult for downstream projects to decide whether they also want to accept the changes.

It's compiling fine @RichardTea. I've added GitHub Actions to my branch and everything is working fine. What are these unrelated changes?

RichardTea:
It's not about whether it merges or builds, it's that it changes too many different things at once.

Can you please divide it up into a separate PR for each related set of changes?

In particular, the C API changes are a compatibility break which will require version number changes. Whether that's manor or minor remains to be determined as the level of binary compatibility will need to be checked on all platforms.

From the list you posted, this PR:
- Adds GitHub Actions
- Changes to the C API. Causes a binary compatibility break, not sure how significant.
- Changes to FBX import, in both functionality and configuration.
- Changes to OBJ import
- Changes to TriLib
- Changes to cmake configuration requirements

And probably some others that I haven't noticed because it's so large.

I expect that we will all want most of this, just not all at once!

rickomax:
> The problem is that it changes too many different things at once.
> 
> It's not about whether it merges or even whether it compiles.
> 
> At second glance there appear to be some places where existing fixes have been rolled back, and I don't know whether that's intentional or accidental.
> 
> Can you please divide it up into a separate PR for each related set of changes?
> 
> From the list you posted, this PR:
> 
> * Adds GitHub Actions
> * Changes the public C API.
> * Changes the public C++ API.
> * Changes FBX import, in both functionality and configuration. Binary compatibility break.
> * Changes OBJ import
> * Changes MMD format (is that a new encoding of the format? Cool!)
> * Changes cmake configuration requirements
> * Changes cmake on iOS, hardware and simulator
> * Adds cmake support for Windows UWP platform (nice!)
> 
> There might be some other changes that I haven't noticed because it's so large.
> 
> I expect that we will all want most of this, just not all at once!
> 
> Finally, some of these are public C and C++ API changes.
> Changing the public API always requires a version number change so applications can properly use the library without breaking.
> 
> Some of these API and importer config changes will require applications to be recompiled for them to be able to load and use it properly, and a few will also require source changes in applications using the changed APIs.
> 
> Whether the version number changes are major or minor remains to be determined as the level of binary compatibility will need to be checked on all platforms.

Richard, can we talk on Gitter.im about these changes so I can explain what everything does?
I'm 99% sure it doesn't break compatibility with anything.

napina:
> The problem is that it changes too many different things at once.
> 
> It's not about whether it merges or even whether it compiles
> ...
> I expect that we will all want most of this, just not all at once!

I'm not maintainer of this project but you really follow these lines with pull requests. Pushing multiple changes to different things is not good practice. For example all those format changes should be separate pull requests so that those can be more easily compared and removed if there is some issue.

rickomax:
You guys are overcomplicating things, but ok. I will follow the rules.

RichardTea:
Thanks, I look forward to these upcoming PRs. A lot of this looks very interesting, thank you very much!

As to the API changes, there's two things to consider:

1. It is very easy to accidentally break binary compatibility with applications compiled for previous versions, by altering the layout of function stubs and structs in the DLL/dylib/so. In this case applications suffer undefined behaviour. Usually they will simply crash, but not always...

2. A new application using the "new" C APIs you're adding obviously cannot use an older version of the library, because the functions didn't exist.

The version numbering is used to ensure an application's dynamic loader won't try to load an incompatible version, but will keep searching for a compatible one.

So the version number needs to change any time the public API changes.

rickomax:
> Thanks, I look forward to these upcoming PRs. A lot of this looks very interesting, thank you very much!
> 
> As to the API changes, there's two things to consider:
> 
> 1. It is very easy to accidentally break binary compatibility with applications compiled for previous versions, by altering the layout of function stubs and structs in the DLL/dylib/so. In this case applications suffer undefined behaviour. Usually they will simply crash, but not always...
> 2. A new application using the "new" C APIs you're adding obviously cannot use an older version of the library, because the functions didn't exist.
> 
> The version numbering is used to ensure an application's dynamic loader won't try to load an incompatible version, but will keep searching for a compatible one.
> 
> So the version number needs to change any time the public API changes.

The C API haven't changed at all, only new methods have been added.
I myself work on a project that uses Assimp version earlier than 4.0 release, so I took care to don't break the compatibility.

Issue #3092, opened 2020-03-27
Pertinence of target UpdateAssimpLibsDebugSymbolsAndDLLs.


malortie:
In the main CMakeFiles.txt, at the [end](https://github.com/assimp/assimp/blob/bd5b7ecab9016b50913d17abc71921ba0ea85132/CMakeLists.txt#L687) of the file, the target `UpdateAssimpLibsDebugSymbolsAndDLLs` copies both compiled static and shared Assimp libraries to respective directories - either `lib32/bin32` or `lib64/bin64` based on the target architecture.

About a month and a half ago, [changes](https://github.com/assimp/assimp/pull/3013) have been made to output compiled projects binaries and static libraries to a common output folder. Reasons behind theses changes are explained in https://github.com/assimp/assimp/issues/3006. 

For these reasons, I am not sure that  `UpdateAssimpLibsDebugSymbolsAndDLLs` is still pertinent. More precisely, I currently do not see any advantage/benefit of having it.

The changes mentioned above that were made could be adapted to set the output directory compiled binaries/static libraries to be  `lib32/bin32` or `lib64/bin64`. This would allow us to remove `UpdateAssimpLibsDebugSymbolsAndDLLs`, as it would no longer be needed.

What are your thoughts on this?

tellypresence:
@malortie Is this still a problem?

malortie:
I currently don't have time to look back into it, so I'll just close this issue.

This can be reopened later if needed.

PR #3097, opened 2020-03-29
Small changes in CMakeLists.txt.

malortie:
[Link to issue](https://github.com/assimp/assimp/issues/3092).

Changes:
- Removed target UpdateAssimpLibsDebugSymbolsAndDLLs
- Added a new macro `USE_COMMON_OUTPUT_DIRECTORY` to set the correct output directories for a single target/project
- Added calls to `USE_COMMON_OUTPUT_DIRECTORY` in CMakeLists.txt of each target/project that is part of Assimp (**1**)

(**1**) This is necessary to prevent contrib static libraries from being output to the common output directory. Without these changes, libs such as **zlib** and **irrXML** output to /lib output folder.

Projects now output to `lib/platform` and `bin/platform`, where `platform` is the value in `CMAKE_GENERATOR_PLATFORM`. If the platform could not be identified, the name will be chosen based on whether we are actually compiling in 32 or 64 bit - `lib/x86` and `bin/x86`, or  `lib/x64` and `bin/x64`.

The changes have been tested with platforms Win32, x64, and unspecified (which defaults to x64). For each platform, each of the projects listed further have been built with the following build configurations:

- Debug
- MinSizeRel
- Release
- RelWithDbgInfo

The projects that were built are the following:

- assimp
- assimp_cmd
- assimp_simpleogl
- assimp_simpletextureddirectx11
- assimp_simpletexturedogl
- assimp_viewer
- unit

In addition, tests have been run with each build configuration.


malortie:
After looking in the log, it appears that the test 'unit' has been correctly built:
```
[100%] Linking CXX executable ../bin/x64/unit
[100%] Built target unit
```

The problem seems to occur because the file not found.

```
Run cd bin && ./unit
/home/runner/work/_temp/e61e8782-d26c-46d5-a512-b514823cefb4.sh: line 1: ./unit: No such file or directory
##[error]Process completed with exit code 127.
```

`cd bin` would have to be replaced by `cd bin/x64`.

I could make changes to simply output to `bin` folder, but keeping it in separate folders 'Win32', 'x86', 'x64' might be preferable since people may want to have builds of different configurations. @kimkulling Please, let me know what would be the best and I will adjust it.

kimkulling:
It would be great if you can investigate how to adress this issue. And sorry for my late response, I did not find a timeslot for this until now. 

I will try to solve the conflicts.

malortie:
@kimkulling 

According to the build logs, the targeted architecture seems to be 64 bit (x64).

linux & mac:
`
[100%] Linking CXX executable ../bin/x64/unit
`

windows:
`
unit.vcxproj -> D:\a\assimp\assimp\bin\x64\Release\unit.exe
`

Assuming the following [file](https://github.com/assimp/assimp/blob/master/.github/workflows/ccpp.yml) (ccpp.yml) is used for C/C++ CI, the simplest way to solve the build issue would be to change the following lines in [ccpp.yml](https://github.com/assimp/assimp/blob/master/.github/workflows/ccpp.yml):

linux & mac:
`cd bin && ./unit` should be changed to `cd bin/x64 && ./unit`
  
windows:
```
cd bin\Release
.\unit
```
should be changed to 
```
cd bin\x64\Release
.\unit
```

------

Please note that I have few knowledge of Github action workflow, but after reading the official [Github workflow article](https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow) as well as many tutorial websites, I have not yet found a way to retrieve the target architecture name. At this time, I am afraid i do not know if it is feasible.

If you are ok with the solution above, I could make the changes to the ccpp.yml  - though, it's not the ideal. Otherwise, I could make changes to CMakeLists.txt to set back the common output folder to `bin/` and `lib/` so it would make the build pass.

Please let me know what you would prefer, and I will make the desired changes. I would encourage others as well to share their thoughts and options on these changes.

malortie:
I am closing this pull request, as there are a few incorrect things.

The variables set inside macro `USE_COMMON_OUTPUT_DIRECTORY` override the global output directory, which was an issue in https://github.com/assimp/assimp/issues/3239.

As for the removal of target `UpdateAssimpLibsDebugSymbolsAndDLLs`, I will make a separate PR for this when I will have time.

PR #3105, opened 2020-03-30
Fixed bone splitting with excessive amount of bones with 0 weight

rudybear:
We have an issue with exporting a model and I noticed a problem in bone splitter code. The issue is that code pollutes newBonesAtCurrentFace and never clears it when face fails the check for bones amount. Besides that no need to process bones with 0 weight. Also I replaced vector to set to avoid linear search. Plus moved the bones count check outside the main loop.


kimkulling:

Issue #3047, opened 2020-03-04
gltf2/glb2 exporter crash (memory related)


thomasbiang:
Hi,
We have an application that loads fbx file and export to gltf2/glb2. It has a memory related crash. 
It crashes inside Buffer::Grow.When I change 
`capacity = std::max(capacity + (capacity >> 1), byteLength + amount);`
to 
`capacity=byteLength + amount;`
The crash goes away.  But I am not sure this is the right way to fix it.
 I paste the screen shot of the crash here.  Please take a look at.

![assimpCrash](https://user-images.githubusercontent.com/37556510/75930011-8f0d4080-5e26-11ea-9f48-194d3d19d741.png)

turol:
This is almost certainly a symptom of the old buffer having been overrun and corrupted the memory manager's internal data. Can you share the FBX file or some other case which reproduces this?

thomasbiang:
> This is almost certainly a symptom of the old buffer having been overrun and corrupted the memory manager's internal data. Can you share the FBX file or some other case which reproduces this?

Hi, 
Here is more detailed information. In our application, we load multiple fbx files, and edit and consolidate them into one aiScene and export. Thus currently I have difficulty providing the exact fbx file as the aiScene being exported is in memory.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3113

PR #3113, opened 2020-03-31
Fix gltf2 exporter memory crash

thomasbiang:
This pr is just to address https://github.com/assimp/assimp/issues/3047.
When I slightly change the buffer expansion code as in this pr, the memory crash goes away.
But I hardly think this is the right fix, but may provide some clue for diagnosing the issue. 

Thank you.


turol:
This is almost certainly the wrong thing and just hides the problem. I don't recommend merging it.

kimkulling:

Issue #3102, opened 2020-03-30
GitHub "Sponsor" button is broken


jakrams:
I would expect this button to do something other than this:

![image](https://user-images.githubusercontent.com/36191331/77933407-0fe91c00-72af-11ea-9b95-aacf7876ee21.png)

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3114

jankrassnigg:
Thanks for fixing. Clicking on the Patreon link I noticed a typo, in case you want to fix it:

![image](https://user-images.githubusercontent.com/6001174/78105336-3a42f280-73f1-11ea-9fb7-5984e34d3780.png)

kimkulling:
Thanks fpr the hint, I just have fixed the typo.

PR #3114, opened 2020-03-31
Fix funding

kimkulling:


kimkulling:
closes https://github.com/assimp/assimp/issues/3102

PR #3115, opened 2020-03-31
use GNUInstallDirs where possible

vmatare:
Emulate the `CMAKE_INSTALL_FULL_*` variables on non-Unix systems and disable redefining FHS-mandated install locations via user-editable `ASSIMP_*_INSTALL_DIR` variables. Instead, if it REALLY proves necessary, Unix users can edit the advanced, canonical `CMAKE_INSTALL_*` variables.


vmatare:
This fixes #2719, but without the fragile guesswork in #3035 .

vmatare:
I have no experience with hunter, and can't test on Windows or Apple platforms. So I'm pretty confident this is the right way to do it on Unix (which should apply to MacOS, too...?), but if this project is serious about Windows support, it should be tested there as well.

kimkulling:
The new git workflow shall be able to test it. But it is under development at the moment and not stable.
You have used not the master? Why?

maquefel:
Yes @vmatare solution looks more correct to me.

vmatare:
> You have used not the master? Why?
I thought of it as a bugfix to the 5.0 release, but I can make a PR against master as well of you like.

kimkulling:
Would be great.

PR #3117, opened 2020-03-31
Removed unneeded SceneDiffer.h includes.

malortie:
The header [SceneDiffer.h](https://github.com/assimp/assimp/blob/master/test/unit/SceneDiffer.h) is included in several unit test files, but these tests do not make use of the **SceneDiffer** class.


kimkulling:

Issue #3123, opened 2020-04-02
Sample projects and Assimp viewer unable to build due to warnings treated as errors.


malortie:
Recent changes have set warning level to **/W4**, and **/WX** (treat warnings as errors).

The following projects would fail to build because of additional warnings treated as errors:

- ~~SimpleOpenGL~~ (Fixed in https://github.com/assimp/assimp/pull/3124)
- ~~SimpleTexturedDirectx11~~ (Fixed in https://github.com/assimp/assimp/pull/3127)
- ~~SimpleTexturedOpenGL~~(Fixed in https://github.com/assimp/assimp/pull/3124)
- ~~Assimp viewer~~ (Fixed in https://github.com/assimp/assimp/pull/3129)

kimkulling:
Thanks for the update. Why I haven't checked them? Sorry for that!

malortie:
@kimkulling 

I currently have a fix for each sample project, as well as for the Assimp viewer. You can view these changes [here](https://github.com/malortie/assimp/commit/6b15586e192889d0fd32ca3ce50650817f83d040). I will submit PRs per project in order to keep them small and easier to review.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3124

malortie:
@kimkulling 

Please, would it be possible to reopen the issue? I was thinking of closing the issue once all projects would have been fixed. I apologize, as I should have explained it in my previous answer.

PR #3129, opened 2020-04-04
Fixed /W4 compile warnings in Assimp viewer.

malortie:
[Link to issue](https://github.com/assimp/assimp/issues/3123).

The following changes fix compile warnings for project Assimp viewer.

These changes have been tested in 32 bit and 64 bit builds, and compared against a build from master. Both appear to show similar results.

The following models were used for testing:

- test/models/OBJ/spider.obj
- test/models-nonbsd/MD5/Bob.md5mesh

The following functionalities were tested:

**Menu / Toolbar**

[Viewer]

*Note: Import settings were left to default.*

- Open asset
- Close asset
- Reload
- Reset view
- Memory consumption
- Clear history
- Save Screenshot

[Tools]
- Log window
- Clear log
- Original normals
- Hard normals
- Smooth normals
- Flip normals
- Stereo view

[Background]
- Set color
- Load texture
- Clear background

**Control panel**

[Display]

- Multisampling
- Wireframe
- No materials
- Display normals
- Low quality
- No specular
- Show skeleton
- Auto rotate
- Zoom/Rotate
- Rotate lights
- Two lights
- Backface culling
- No transparency

[Statistics]

- Vertices
- Nodes
- Shaders
- Load Time (arbitrary)
- Faces
- Materials
- Meshes
- FPS (arbitrary)

[Colors]
- Primary
- Secondary
- Ambient
- Colors reset button

Each animation has been verified in each model, including default bind pose.

All tests were run.

Additional information about the changes:

- SColor anonymous struct was given the name `_data` to prevent compiler warning:
`warning C4201: nonstandard extension used: nameless struct/union`

- The function [SafeRelease](https://github.com/assimp/assimp/blob/085212e58ab0cf6dc96256e1577e9e31d0a8f3e2/tools/assimp_view/assimp_view.cpp#L805-L810) was always releasing `g_piPassThroughEffect` instead of
`ptr`. This has been changed to `ptr`. In addition, the function attempts to set `ptr` to null, but it cannot be changed because `ptr` is neither passed as a double pointer or by reference. Therefore, a reference operator **&** has been added.

### DIFF
--- a/tools/assimp_view/Material.cpp
+++ b/tools/assimp_view/Material.cpp
@@ -287,7 +287,7 @@ bool CMaterialManager::TryLongerPath(char* szTemp,aiString* p_szString)
                             size_t iLen2 = iLen+1;
                             iLen2 = iLen2 > MAXLEN ? MAXLEN : iLen2;
                             memcpy(p_szString->data,szTempB,iLen2);
-                            p_szString->length = iLen;
+                            p_szString->length = static_cast<ai_uint32>(iLen);
                             return true;
                         }
                     }
@@ -301,7 +301,7 @@ bool CMaterialManager::TryLongerPath(char* szTemp,aiString* p_szString)
                         size_t iLen2 = iLen+1;
                         iLen2 = iLen2 > MAXLEN ? MAXLEN : iLen2;
                         memcpy(p_szString->data,szTempB,iLen2);
-                        p_szString->length = iLen;
+                        p_szString->length = static_cast<ai_uint32>(iLen);
                         return true;
                     }
                 }
@@ -389,10 +389,10 @@ int CMaterialManager::FindValidPath(aiString* p_szString)
                     if( !q ) q=strrchr( tmp2,'\\' );
                     if( q ){
                         strcpy( q+1,p+1 );
-                        if((pFile=fopen( tmp2,"r" ))){
+                        if((pFile=fopen( tmp2,"r" )) != nullptr){
                             fclose( pFile );
                             strcpy(p_szString->data,tmp2);
-                            p_szString->length = strlen(tmp2);
+                            p_szString->length = static_cast<ai_uint32>(strlen(tmp2));
                             return 1;
                         }
                     }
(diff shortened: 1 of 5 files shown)


tools/assimp_view/Material.cpp
```diff
@@ -627,7 +627,7 @@ void CMaterialManager::HMtoNMIfNecessary(
     {
         union
         {
-            struct {unsigned char b,g,r,a;};
+            struct {unsigned char b,g,r,a;} _data;
```
turol, line 630: Don't start identifiers with underscore, some of those are reserved for implementation only.

tools/assimp_view/MessageProc.cpp
```diff
@@ -107,7 +107,7 @@ void MakeFileAssociations() {
         RegCreateKeyEx(HKEY_CURRENT_USER,buf,0,NULL,0,KEY_ALL_ACCESS, NULL, &hRegistry,NULL);
         RegSetValueEx(hRegistry,"",0,REG_SZ,(const BYTE*)"ASSIMPVIEW_CLASS",(DWORD)strlen("ASSIMPVIEW_CLASS")+1);
         RegCloseKey(hRegistry);
-    } while ((sz = strtok(NULL,";")));
+    } while ((sz = strtok(NULL,";")) != nullptr);
```
turol, line 110: Change the other `NULL` too.

tools/assimp_view/Material.cpp
malortie: @turol Thanks. I changed it to data.

tools/assimp_view/MessageProc.cpp
malortie: @turol 

Do you mean, to change only the three **NULL** in the `RegCreateKeyEx` call above the highlighted line, or in the entire project?

Changing every **NULL** reference in the project would add 326 +- changes to the diff. I would recommend to make such changes in a separate PR because it goes beyond the scope of the current PR, which is to simply fix the compile warnings.

turol:
Separate PR for all `NULL` -> `nullptr` is ok by me.

malortie:
@turol @kimkulling Thanks. I will look onto replacing `NULL` -> `nullptr` after the current PR is merged. I might as well do it for the other samples, (one PR for each).

kimkulling:

PR #3140, opened 2020-04-07
Fix SimpleOGL-warnings.

kimkulling:
closes https://github.com/assimp/assimp/issues/3123


turol:
`stb_image.h` is a third-party dependency. Your changes (reformatting it apparently?) will be lost when it's updated from upstream so please don't do that.

PR #3155, opened 2020-04-16
Erase the remaining _INSTALL_PREFIX and LIBSUFFIX in CMake files

Zalewa:
These seem to have survived from the migration to GNUInstallDirs in
9fb81c3be6dc30139ea32c92046c9794dca0e73e

_INSTALL_PREFIX was not set anywhere and resolved to empty value, which
in turn caused some paths in the installed lib config CMake files to
start from '/', effectively causing the failure of the "do the installed
files exist" checks.

After complete replacement of all uses of the _INSTALL_PREFIX with
GNUInstallDirs equivalents, the LIBSUFFIX variable has become unused
and can be removed too.

This is an amendment to PR #3126.

### DIFF
--- a/assimpTargets-debug.cmake.in
+++ b/assimpTargets-debug.cmake.in
@@ -9,12 +9,6 @@ set(ASSIMP_BUILD_SHARED_LIBS @BUILD_SHARED_LIBS@)
 
 get_property(LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS)
 
-if ("${LIB64}" STREQUAL "TRUE")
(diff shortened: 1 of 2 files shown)


vmatare:
I'm glad this problem has been noticed so quickly. It comes from a botched conflict resolution in https://github.com/assimp/assimp/pull/3035/commits/2d4eab6db087d67726da46d2bf2b4ef1cbf2b501, which was done to merge PR #3035, which should have been closed, not merged.

kimkulling:
Thanks for figuring this out.

kimkulling:

PR #3168, opened 2020-04-26
Use clang on Unix, msvc on Windows, Use Ninja everywhere

lukka:
CI build: 
  - use clang on Unix, msvc on Windows;
  - use Ninja as make program; this improves build time of 1 minute;

CMake project files:
 - Remove some explicit and specific to clang compiler flags in CMakeLists.txt;
- Drop the explicit -g flag;


kimkulling:
Thanks for your help! I tried this on my own as well but special the clang-build was broken all the time. I am not sure what went wrong there. I will close my try ( there is an open PR ). So we can concentrate on your work.

CMakeLists.txt
```diff
@@ -267,10 +267,10 @@ ELSEIF(MSVC)
   SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /D_DEBUG /Zi /Od")
 ELSEIF ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" )
   IF(NOT ASSIMP_HUNTER_ENABLED)
-    SET(CMAKE_CXX_FLAGS "-fPIC -std=c++11 ${CMAKE_CXX_FLAGS}")
-    SET(CMAKE_C_FLAGS "-fPIC ${CMAKE_C_FLAGS}")
+    SET(CMAKE_CXX_STANDARD 11)
+    SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
   ENDIF()
-  SET(CMAKE_CXX_FLAGS "-g -fvisibility=hidden -fno-strict-aliasing -Wall -Wno-long-long ${CMAKE_CXX_FLAGS}" )
```
kimkulling, line 273: Good catch, this was a temporary change I made some years ago.

kimkulling:
We need to another compiler into the matrix: vs-compiler. Only clang is not good enought because a lot of people are using Visual-Studio.

kimkulling:
Is it corrct that the cl..exe is based on the Visual-Studio-compiler?

kimkulling:

turol:
Now there appears to be no GCC build which will probably lead to breakage.

kimkulling:
Good point, we need to add this as well.

kimkulling:
And of course the sanitizers!

PR #3169, opened 2020-04-27
Add timescale for collada

inhosens:
I'm not sure that this approach is correct. But there is a mismatch between gltf2 and collada. gltf2 converts the timescale from sec to millisec but collada does not.
I want to make all the timescales the same.


kimkulling:
Thanks a lot for your contribution. To resolve your merge request I will open https://github.com/assimp/assimp/pull/3169

PR #3171, opened 2020-04-27
Add sanitizer support

kimkulling:


lukka:
you may be interested in integrating this (or similar) changes as well: #3173

.github/workflows/sanitizer.yml
```diff
@@ -0,0 +1,58 @@
...
+      with:
+        CXX: clang++
+        CC: clang
+    
+    - name: configure and build
+      uses: lukka/run-cmake@v2
+      with:
+        cmakeListsOrSettingsJson: CMakeListsTxtAdvanced
+        cmakeListsTxtPath: '${{ github.workspace }}/CMakeLists.txt'
+        cmakeAppendedArgs: '-GNinja -DCMAKE_BUILD_TYPE=Debug -DASSIMP_ASAN=ON'
```
turol, line 52: Wrong sanitizer here, you're using asan twice

.github/workflows/sanitizer.yml
```diff
@@ -0,0 +1,58 @@
...
+    branches: [ master ]
+
+jobs:
+  job1:
+    name: adress-sanitizer
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: lukka/get-cmake@latest    
+    - uses: ilammy/msvc-dev-cmd@v1
```
lukka, line 16: the `ilammy/msvc-dev-cmd` does not harm, but it is not necessary if you do not build with VS (i.e. msvc) afterwards. It could be removed in this line only.

.github/workflows/sanitizer.yml
kimkulling: Right, thanks for the hint. Fixed!

.github/workflows/sanitizer.yml
kimkulling: Thanks for the hint, fixed!

PR #3173, opened 2020-04-28
Enable gcc on Linux on CI, fix warnings

lukka:
- Enable building with `gcc` on Linux;
- Fix warning (mostly uninit'ed data);
- Remove some explicit flags specific for `gcc` with CMake "portable" ones;
- Use warning as error only on `assimp` target, not on 3rd party dependencies;


kimkulling:
Look good, merge is on its way!

kimkulling:

Issue #3177, opened 2020-04-28
Collada Exporter does not enforce unique geometry ids


RichardTea:
The Collada exporter uses the aiMesh.mName for both the geometry name and the geometry id.
However it does not check whether the geometry id is unique.

This means that exporting a Scene where several meshes have the same name results in an invalid Collada file where the multiple different geometries have the same id.

RichardTea:
Investigating further, it seems that the XMLIDEscaping for these isn't being applied/used everywhere so it is likely that animations don't export either.

PR #3188, opened 2020-04-29
Collada: Ensure export uses unique Mesh Ids

RichardTea:
Fixes issue #3177
This also changes the importer so that import/export round-trips make fewer changes.
The existing AI_CONFIG_IMPORT_COLLADA_USE_COLLADA_NAMES import option flag now also decides whether aiMesh names will be the mesh Id or the mesh name.

As with aiNodes, the default is to use the unique Id.

For completeness, I should also ensure unique ids for:

- aiNode
- aiMesh (done)
- aiAnimation
- aiTexture
- aiLight
- aiCamera

I also need to have another look at animations, I don't think the lookup is correct at the moment.


RichardTea:
This should now be ok to merge

kimkulling:

PR #3196, opened 2020-05-02
Kimkulling dev

kimkulling:
- Move importer / exporter to AssetLib


RichardTea:
Would have really appreciated some warning you were going to do this.

It's now made a lot of existing PRs extremely difficult to merge, including several very important fixes.

RichardTea:
I've also noticed that this PR does quite a lot more than simply rename/move files.

As a general rule it's extremely important to _only_ rename/move files in a single PR.
Don't change formatting or fix NULL/nullptr usage.

kimkulling:
You are right. Will will do the cleanup!

Issue #3206, opened 2020-05-05
GLTF2 Blendshape Import Bug: missing shapes


thomasbiang:
When I read in a standard gltf2 file from 
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphCube
It is supposed to be a cube with two blenshapes.

After read in, I find only one shape is there.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3207

PR #3207, opened 2020-05-05
Fix for issue# 3206: GLTF2 blendshape import missing shapes https://github.com/assimp/assimp/issues/3206

thomasbiang:
This is to fix gltf2 import bug: blendshapes (morph targets) are missing. 
It seems an index i (refers to primitive) is wrongly used as target index. Thus one primitive eventually only has one target.
This pr just change the target index to the right one.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Asset.inl
+++ b/code/AssetLib/glTF2/glTF2Asset.inl
@@ -1002,7 +1002,7 @@ inline void Mesh::Read(Value &pJSON_Object, Asset &pAsset_Root) {
                         // Valid attribute semantics include POSITION, NORMAL, TANGENT
                         int undPos = 0;
                         Mesh::AccessorList *vec = 0;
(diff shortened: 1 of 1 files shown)


code/AssetLib/glTF2/glTF2Asset.inl
```diff
@@ -1002,7 +1002,7 @@ inline void Mesh::Read(Value &pJSON_Object, Asset &pAsset_Root) {
                         // Valid attribute semantics include POSITION, NORMAL, TANGENT
                         int undPos = 0;
                         Mesh::AccessorList *vec = 0;
-                        if (GetAttribTargetVector(prim, i, attr, vec, undPos)) {
+                        if (GetAttribTargetVector(prim, j, attr, vec, undPos)) {
```
thomasbiang, line 1005: i is the mesh(primitive) index
j is the target index.
j should be used here, not i.

kimkulling:

PR #3214, opened 2020-05-07
[gltf2 Export] More robust handling for non-finites and 0-length normals

jercytryn:
Improves the tolerance for minor issues w/0-length normals and non-finite buffer data in the input models when exporting gltf2, as well as fixing a small bug that could cause writing out incomplete JSON w/o erroring. Addresses the following:

* If for whatever reason, the exporter generates incomplete JSON for gltf or gltf2, raise an exception so that the partial gltf does not get silently written out as a success case
* Addresses a cause of incomplete JSON for gltf2 that can happen when there are infs or NaNs in the buffer data when doing the accessor bounds recalc
* Addresses a cause of introduced NaNs in the normals buffer when the input model has any 0-length normals


Nimer-88:
Looks good to me :)

code/AssetLib/glTF2/glTF2Exporter.cpp
```diff
@@ -172,6 +172,13 @@ void SetAccessorRange(Ref<Accessor> acc, void* data, size_t count,
 		for (unsigned int j = 0 ; j < numCompsOut ; j++) {
 			double valueTmp = buffer_ptr[j];
 
+			// Gracefully tolerate rogue NaN's in buffer data
+			// Any NaNs/Infs introduced in accessor bounds will end up in
+			// document and prevent rapidjson from writing out valid JSON
+			if (!std::isfinite(valueTmp)) {
+				continue;
+			}
+
```
kimkulling, line 181: Good catch!

kimkulling:
Merged, thanks a lot for your contrubution.

Issue #2979, opened 2020-02-02
cmake assimp::assimp from "find_package" provides dll file, not lib


ElDesalmado:
I've installed v.5.0.0.rc2 and used ```find_package(assimp)``` to link against the library, but ```assimp::assimp``` yields the dll file, not lib. 

```${ASSIMP_LIBRARIES}``` yields the correct lib filename, however it does not provide the full or relative path to the lib file, so it can not be found when building

kimkulling:
Clsed by

kimkulling:
Thanks for the report by the way!

PR #3215, opened 2020-05-07
Add IMPORTED_CONFIGURATIONS property to cmake target.

kalyan-kumar:
When I tried linking assimp into my C++ project, cmake was unable to link the library even though the package was found. On investigating it, I found that this was due to `IMPORTED_LOCATION` not being set.

I looked at the `assimpTargets-release.cmake` and `assimpTargets-debug.cmake` files to find that an overridden version of `IMPORTED_LOCATION` was being set corresponding to each of these configurations, but the `CONFIGURATION` property itself was not added for non-windows OS. 

Suspecting that this was the issue, I set this property in the cmake file installed on my system and the library was found. Please let me know if this is the correct way to fix this, else I can update.

[This issue](https://github.com/assimp/assimp/issues/2979) is regarding this same problem.


kalyan-kumar:
Seems the checker is unable to download `ninja`.

kimkulling:

Issue #2359, opened 2019-03-08
glTF uses entire bufferView even if only small subset referenced


ardenpm:
It seems a common structure with glTF files is to store a large amount of data in a single buffer, then take a few bufferViews on this data, but share the same bufferView in multiple accessors then use indicies to look up the needed parts for a mesh.

The issue is that currently glTF2Importer::ImportMeshes just blindly copies the entire bufferView contents for the various vertex attribute data, even if only a sparse set of it is accessed. Since this is set on the mVertices, mNormals etc members of the aiMesh, even if the mesh has a handful of faces, it will reference all of the vertices in the buffer.

In reasonably large models when used with aiProcess_JoinIdenticalVertices and aiProcess_SortByPType flags, imports can take a huge amount of time and memory to complete. It is also blowing out memory requirements since the bufferView data is copied multiple times, once for each mesh that uses it.

If we turn off the vertex joining to speed up the process then we get meshes with huge vertex counts even with a handful of triangles. In one example I had a model that ended up with 483,550 triangles but 316,067.832 vertices (over about 1188 objects). This needed 10GB to load and process and a lot of time.

I don't see an obvious work around yet but am still looking at the code in glTF2Importer.cpp and specifically glTF2Importer::ImportMeshes. It looks like the face generation code could be moved in front of the vertex code since it only sets offsets, this could then be used to convert the sparse bufferView into a packed vertex array (though that will then require re-indexing all of the mesh data).

I'm still looking at it but if anyone else has come across the issue and has thoughts it would be interesting to hear.

kimkulling:
So if I get it right in short: by using lots of BufferViews we are doubling the used data, right?

ardenpm:
I should first clarify, this is only relevant to indexed meshes.

The _attributes_ of the _mesh_ reference _accessors_ which reference a very large _bufferView_. This is then subsetted by using the _indices_ property of the mesh.

Assimp takes the entire content of the bufferView referenced by the attributes accessor and memcpy's it into the vertex data for the mesh, even though only a subset of the data is actually used.

So this results in a large number of orphaned vertices if there is vertex data in the bufferView that the indices do not touch.

In the scene I found the issue with they basically have one big bufferView for all of the _POSISTION_ attribute data and this is used by all meshes. So each mesh gets the vertices of every other mesh, but then only uses a small number of them.

[...]

PR #3219, opened 2020-05-11
Gltf2 Support Importing sparse accessor 

thomasbiang:
This pr supports import gltf2 with sparse accessor.
It solves issue#2359: https://github.com/assimp/assimp/issues/2359

A Sparse struct is embedded in Accessor, following the gltf2 standard. 
Particularly bufferView of Accessor could be empty if there is a sparse section.

This pr is tested with gltf2 with two expected two blendshapes https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphCube
One gltf with original gltf2 of the morph cube, second gltf2 with sparse accessor for the morph cube. After importing both of files, blendshapes are exactly the same with no diffs.

NOTE: will make a separate pr that allows export gltf2 sparse accessor as well. Thus sparse accessor import/export can work together for gltf2 format.


kimkulling:

PR #3226, opened 2020-05-15
[GLTF2] Fix infinite recursion in skin/node parsing

M4T1A5:
Currently parsing a node causes the skin that is attached to it to be parsed, which causes the skins joint nodes to be parsed, which could cause the skin to be re-parsed leading to infinite or at the very least exponential recursion.

The fix is to just get a reference to a temporarily uninitialized skin as they were being parsed after the scene graph just to be safe anyway. This way we avoid the recursion problem and all the references
will be valid in the end.


kimkulling:

PR #3232, opened 2020-05-18
Verbose logging

MalcolmTyrrell:
I want to enable debug logging for assimp, to capture high-level internal messages about the importer and the processors. However, I don't want to be flooded by unneccessarily detailed debug messages, e.g. those concerning particular objects within a model. Unfortunately, there's currently no way to filter the messages at this level.

This PR does the following:
* adds the concept of verboseDebug
* switches many of the existing messages to use verboseDebug instead of debug
* Adds some additional standard debug messages to the FBX and GLFT2 importers.
* Fixes two typos in function names
* Makes the logging macros callable from outside the Assimp namespace.

I adopted the following logic when deciding whether a message is debug or verbosedDebug: If a message can be issued a data-dependent number of times, then it's probably a verbose debug message.

This is a mildly breaking change for anyone who has implemented a logger, since I've added a new pure virtual function. This can be avoided with a default implementation which makes verboseDebug call debug, which I'd be OK with.


kimkulling:

Issue #2187, opened 2018-10-19
play the bvh error


iamyaoting:
hello, i build the current version assimp project, found that it cannot play the bvh file correctly, so as to the assimp_viewer,  I use the visual stuido 2013 64 bit. 

the 4.10 can play the bvh file correctly, motionbuilder can play it correctly. 

the bvh file is below:
https://pan.baidu.com/s/1Bcf28qwmkztuZyRJQdKMdg

kshepherd2013:
I have encountered the same problem. It is caused by an error introduced when a channelMap was added to the file BVHLoader.cpp after the 4.10 release.  The problem is that the order of rotation specified in the file is lost, and becomes always XYZ.  This bug exists in v5.0.0 and 5.0.1 It is easily fixed.  Just edit the file code/BVH/BVHLoader.cpp and comment out the for loop at line 533, and replace it so that your code looks like that below:

    /*
    for (BVHLoader::ChannelType channel = Channel_RotationX; channel <= Channel_RotationZ; channel = (BVHLoader::ChannelType)(channel + 1))
    {
        //Find channel in node
        std::map<BVHLoader::ChannelType, int>::iterator mapIter = channelMap.find(channel);

        if (mapIter == channelMap.end())
            throw DeadlyImportError("Missing rotation channel in node " + nodeName);
        else {
            int channelIdx = mapIter->second;
            // translate ZXY euler angels into a quaternion
            const float angle = node.mChannelValues[fr * node.mChannels.size() + channelIdx] * float(AI_MATH_PI) / 180.0f;

            // Compute rotation transformations in the right order
            switch (channel)
            {
                case Channel_RotationX: 
                    aiMatrix4x4::RotationX(angle, temp); rotMatrix *= aiMatrix3x3(temp); 
                    break;
                case Channel_RotationY: 
                    aiMatrix4x4::RotationY(angle, temp); rotMatrix *= aiMatrix3x3(temp);  
                    break;
                case Channel_RotationZ: aiMatrix4x4::RotationZ(angle, temp); rotMatrix *= aiMatrix3x3(temp); 
                    break;
                default:
                    break;
            }
        }
    }
    */
    for (unsigned int channelIdx = 0; channelIdx < node.mChannels.size(); ++ channelIdx)
    {
        switch (node.mChannels[channelIdx])
        {
            case Channel_RotationX:
                {
[...]

PR #3233, opened 2020-05-18
Fixed rotation order bug in BVH Loader

kshepherd2013:
This bug has been present since after the 4.10 release.
It fixes the issue "play the bvh error" #2187
Almost all bvh files are affected, such as any of the CMU library.
The bug is caused by the introduction of channelMap in the BVHLoader.cpp - function void BVHLoader::CreateAnimation(aiScene *pScene).  The channelMap loses the rotation order present in the BVH file, and always applies the rotations in X,Y,Z order.


kimkulling:

PR #3235, opened 2020-05-19
Fix double free caused in FindInvalidDataProcess

rmstyrczula:
In debug builds, there's no issues, but in release, an invalid mesh that
is deleted in FindInvalidDataProcess will be double free'd when the
parent scene is destroyed.


turol:
Looks good to me

kimkulling:
For me, too!

kimkulling:

Issue #3176, opened 2020-04-28
export glb error.


copycd:
when assimp export to glb, I found error.
        =>  here. 
        int GLB_Chunk_count = 1;

        uint32_t binaryChunkLength = 0;
        if (bodyBuffer->byteLength > 0) {
            binaryChunkLength = (bodyBuffer->byteLength + 3) & ~3; // Round up to next multiple of 4

            =>  here. 
            auto curPaddingLength = binaryChunkLength - bodyBuffer->byteLength;
            GLB_Chunk_count ++;

            GLB_Chunk binaryChunk;
            binaryChunk.chunkLength = binaryChunkLength;
            binaryChunk.chunkType = ChunkType_BIN;
            AI_SWAP4(binaryChunk.chunkLength);

            size_t bodyOffset = sizeof(GLB_Header) + sizeof(GLB_Chunk) + jsonChunk.chunkLength;
            outfile->Seek(bodyOffset, aiOrigin_SET);
            if (outfile->Write(&binaryChunk, 1, sizeof(GLB_Chunk)) != sizeof(GLB_Chunk)) {
                throw DeadlyExportError("Failed to write body data header!");
            }
            if (outfile->Write(bodyBuffer->GetPointer(), 1, bodyBuffer->byteLength) != bodyBuffer->byteLength) {
                throw DeadlyExportError("Failed to write body data!");
            }

            => here.
            if (curPaddingLength && outfile->Write(&padding, 1, curPaddingLength) != curPaddingLength) {
                throw DeadlyExportError("Failed to write body data padding!");
            }
        }

        //
        // Header
        //

        GLB_Header header;
        memcpy(header.magic, AI_GLB_MAGIC_NUMBER, sizeof(header.magic));

        header.version = 2;
        AI_SWAP4(header.version);

         => here.
        header.length = uint32_t(sizeof(GLB_Header) + GLB_Chunk_count * sizeof(GLB_Chunk) + jsonChunkLength + binaryChunkLength);
        AI_SWAP4(header.length);

LoicFr:
Hello copycd,

Thank you for your code, it corrects an issue preventing several software to open GLB files generated by Assimp. I submitted a pull request with your code: #3246 

Regards,
Loïc

tellypresence:
Please add "Export" label to this issue

kimkulling:
Done, and it is already fixed.

PR #3246, opened 2020-05-28
glTF2: fix "file size doesn't match" issue when importing GLB in Blender (issue #3176 export glb error)

LoicFr:
Hello,

I tested the copycd's correction proposed in issue #3176 and it fixes the GLB export problem which begins to appear between early March and now. The problem was a wrong calculation of chunk size which causes several applications to not be able to import GLB files created by Assimp (Blender displays a "file size doesn't match" error and Paint3D also couldn't open such GLB files).

Many thanks to copycd !

Regards,
Loïc

### DIFF
--- a/code/glTF2/glTF2AssetWriter.inl
+++ b/code/glTF2/glTF2AssetWriter.inl
@@ -672,10 +672,13 @@ namespace glTF2 {
         // Binary chunk
         //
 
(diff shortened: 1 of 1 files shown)


kimkulling:
Will get merged here: https://github.com/assimp/assimp/pull/3323

kimkulling:
Merged!

PR #3249, opened 2020-05-28
Evaluated expressions and clean up some code in tests

hbina:


contrib/Open3DGC/o3dgcSC3DMCDecoder.inl
```diff
@@ -72,9 +72,12 @@ namespace o3dgc
...
         ifs.SetCCW             ((mask & 1) == 1);
-        ifs.SetSolid           ((mask & 2) == 1);
-        ifs.SetConvex          ((mask & 4) == 1);
-        ifs.SetIsTriangularMesh((mask & 8) == 1);
+        // (mask & 2) == 1
+        ifs.SetSolid           (false);
+        // (mask & 4) == 1
+        ifs.SetConvex          (false);
+        // (mask & 8) == 1
+        ifs.SetIsTriangularMesh(false);
```
hbina, line 80: I tried to find the implementation inside AMD's official repository and they seems to have changed.
https://github.com/amd/rest3d/blob/ecc287150791c1d5aef10f0606de46b2147c0944/server/o3dgc/src/o3dgc_decode_lib/src/o3dgcDynamicVectorDecoder.cpp#L118-L150

Either this part does not exist or this is changed to suit assimp.

test/unit/utMatrix3x3.cpp
```diff
@@ -73,9 +73,21 @@ TEST_F(utMatrix3x3Test, FromToMatrixTest) {
 
     aiVector3D from, to;
 
+    auto random_ratio = []() -> float {
+        return static_cast<float>(rand() / RAND_MAX);
+    };
+
     for (int i = 0; i < NUM_SAMPLES; ++i) {
-        from = aiVector3D(1.f * rand() / RAND_MAX, 1.f * rand() / RAND_MAX, 1.f * rand() / RAND_MAX).Normalize();
-        to = aiVector3D(1.f * rand() / RAND_MAX, 1.f * rand() / RAND_MAX, 1.f * rand() / RAND_MAX).Normalize();
```
hbina, line 78: This will trigger implicit int -> float conversion error around large integer values

```
/home/hbina/git/assimp/test/unit/utMatrix3x3.cpp:77:42: warning: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Wimplicit-int-float-conversion]
[build]         from = aiVector3D(1.f * rand() / RAND_MAX, 1.f * rand() / RAND_MAX, 1.f * rand() / RAND_MAX).Normalize();
[build]                                        ~ ^~~~~~~~
```

Can be silenced by explicitly casting them.

test/unit/utMatrix3x3.cpp
```diff
@@ -73,9 +73,21 @@ TEST_F(utMatrix3x3Test, FromToMatrixTest) {
 
     aiVector3D from, to;
 
+    auto random_ratio = []() -> float {
+        return static_cast<float>(rand() / RAND_MAX);
```
turol, line 77: This does an integer division and then casts to float which doesn't do the same thing

hbina: You are right, changing it now.

contrib/Open3DGC/o3dgcSC3DMCDecoder.inl
kimkulling: I guess we need to update it.

kimkulling:
Merged, thanks a lot for your cntribution.

PR #3250, opened 2020-05-28
Fixed dangling pointer and uninitialized variable.

hbina:
Also auto formatted some stuff :(

### DIFF
--- a/code/AssetLib/3DS/3DSConverter.cpp
+++ b/code/AssetLib/3DS/3DSConverter.cpp
@@ -231,10 +231,10 @@ void Discreet3DSImporter::ConvertMaterial(D3DS::Material &p_cMat,
     p_pcOut.AddProperty(&p_cMat.mEmissive, 1, AI_MATKEY_COLOR_EMISSIVE);
 
     // Phong shininess and shininess strength
-    if (D3DS::Discreet3DS::Phong == p_cMat.mShading ||
-            D3DS::Discreet3DS::Metal == p_cMat.mShading) {
+    if (D3DS::Discreet3DS::ShadeType3DS::Phong == p_cMat.mShading ||
+            D3DS::Discreet3DS::ShadeType3DS::Metal == p_cMat.mShading) {
         if (!p_cMat.mSpecularExponent || !p_cMat.mShininessStrength) {
-            p_cMat.mShading = D3DS::Discreet3DS::Gouraud;
+            p_cMat.mShading = D3DS::Discreet3DS::ShadeType3DS::Gouraud;
         } else {
             p_pcOut.AddProperty(&p_cMat.mSpecularExponent, 1, AI_MATKEY_SHININESS);
             p_pcOut.AddProperty(&p_cMat.mShininessStrength, 1, AI_MATKEY_SHININESS_STRENGTH);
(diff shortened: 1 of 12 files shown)


code/AssetLib/3DS/3DSLoader.cpp
```diff
@@ -156,9 +156,9 @@ void Discreet3DSImporter::InternReadFile(const std::string &pFile,
     mScene = &_scene;
 
     // Initialize members
-    D3DS::Node _rootNode("UNNAMED");
+    D3DS::Node *_rootNode = new D3DS::Node("UNNAMED");
     mLastNodeIndex = -1;
-    mCurrentNode = &_rootNode;
+    mCurrentNode = _rootNode;
     mRootNode = mCurrentNode;
```
hbina, line 162: I don't _quite_ understand this code, but it does not make sense regardless.
Taking a pointer to a local variable is _always_ bad.
Somewhere down below, these pointers (except `mCurrentNode`) gets invalidated to `nullptr` but only in debug...for some reason.

turol: No, there are valid reasons to take an address of a local variable. In this case it appears that `_rootNode` is used as a temporary container and in the end its contents are copied into the final scene. Your changes on the other hand caused a memory leak because you didn't delete the object you created.

hbina: After some rummaging, I find that `mRootNode` is indeed just a temporary variable. Although then its probably better to not make them a member variable?) On the other hand, `mCurrentNode` will get assigned a value here:

https://github.com/assimp/assimp/blob/1427e67b54906419e9f83cc8625e2207fbb0fcd5/code/AssetLib/3DS/3DSLoader.cpp#L636-L643

and slightly below that,

https://github.com/assimp/assimp/blob/1427e67b54906419e9f83cc8625e2207fbb0fcd5/code/AssetLib/3DS/3DSLoader.cpp#L648-L676

The second one seems correct enough, assuming it gets cleanup later as you said. But the first one is suspicious, because `FindNode` returns a pointer to the temporary variable `mRootNode` in this case.

https://github.com/assimp/assimp/blob/1427e67b54906419e9f83cc8625e2207fbb0fcd5/code/AssetLib/3DS/3DSLoader.cpp#L561-L574

As a result, I suggest adding an assert at the end of the function where `_rootNode` to verify that `mCurrentNode` have been assigned something else. This will not solve all the problems if it returns something in the child. I suppose one could introduce a more robust assertion that traverse through the entire graph.

code/AssetLib/3DS/3DSConverter.cpp
```diff
@@ -686,8 +686,8 @@ void Discreet3DSImporter::GenerateNodeGraph(aiScene *pcOut) {
     pcOut->mRootNode = new aiNode();
     if (0 == mRootNode->mChildren.size()) {
         //////////////////////////////////////////////////////////////////////////////
-        // It seems the file is so messed up that it has not even a hierarchy.
-        // generate a flat hiearachy which looks like this:
+        // It seems the file is so messed up that it does not even have a hierarchy.
+        // Generate a flat hiearachy which looks like this:
```
RichardTea, line 690: hiearachy -> hierarchy
(Should fix the spelling errors too)

hbina: I'll keep this in mind. Currently trying to work on some other stuff.

hbina:
Do you know how this works? @turol 

https://github.com/assimp/assimp/blob/1427e67b54906419e9f83cc8625e2207fbb0fcd5/code/AssetLib/3DS/3DSConverter.cpp#L424

One is `D3DS::Mesh` and the other is `aiMesh`. AFAIK they do not have the same layout.

turol:
It "works" because the original pointer which was stored in `mColors[0]` is actually a mesh. UBsan complains about it and it will probably break at some point. It would have already if we didin't disable strict aliasing optimization. Please fix it if you can.

RichardTea:
Thanks very much for digging into this importer, hbina. It is most appreciated!

PR #3263, opened 2020-06-06
Added rapidjson define to avoid warnings in c++17

AlecLafita:
Tested  by forcing to compile in c++17 and without changing CMakeLists


kimkulling:

PR #3268, opened 2020-06-08
Fbx Import: support channel name in blendshape name

thomasbiang:
Current blendshape name loses channel information, only keeps the shape info.
This pr supports importing channel name in blendshape name.  
The aiAnimMesh mName data will respect names in fbx file, in the following format:
**channelName.shapeName**
Both information is included.


kimkulling:

kimkulling:
Test is on its way as well!

PR #3270, opened 2020-06-09
Migenius fix texcoord

### LINKED ISSUE #3269: glTF 2.0 crash on non-contiguous TEXCOORD_x properties
glTF 2.0 states that TEXCOORD_x properties must have the index start at 0 and be contiguous. Currently if this is violated Assimp will crash since it still parses the file. So if only TEXCOORD_1 is given for example, it will still try to use the accessor for TEXCOORD_0 and crash.

This happens in glTF2Importer::ImportMeshes where the texture coordinates are processed. I think the import should fail if this is encountered since such a file does not meet the glTF 2.0 specification.

Will have a quick look and see if I can do a pull request for a fix.

Here is an intentionally corrupted glTF file which reproduces the issue.
[texcoord_crash.zip](https://github.com/assimp/assimp/files/4750145/texcoord_crash.zip)

ardenpm:
Fixes #3269 by adding check that texture coordinate accessor is valid before using. This may fail if the glTF 2.0 file is not compliant with the specification (e.g., non-contiguous texture sets). Making the error fatal since an invalid file should not be read.


kimkulling:
Thanks a lot for your contribution, I just updated it with the current master and will merge it here: https://github.com/assimp/assimp/pull/3277

PR #3271, opened 2020-06-09
Pyassimp - contextmanager for load function

DavidBerger98:
Thought it was more elegant to implement the load function with a contextmanager so that you can use it like "with open('/file.txt') as f: ..." and don't have to remember releasing it.

I also updated the examples in the README.md/README.rst.


kimkulling:
Looks good so far, I will merge it.

PR #3280, opened 2020-06-12
Customize Extras in Gltf2 Exporter with ExporterProperty Callback

thomasbiang:
This pr supports customizing extras section in gltf2 format. To make the flow convenient, a new extras pointer is added as a member to gltf2 class Asset , which points to a json object. This allows users specify its content flexibility,  through a callback function.

In addition to current exporter properties, we also support specifying callback function through Exporter Properties.  It allows users to set extra functions they need. In this pr, extras json object is returned from this callback function and handled properly during gltf2 export.

This pr is tested when users specify extras content. It has no influence if users don't include extras. It is purely optional and up to users.


rbsheth:
@kimkulling Also waiting for this one, if you can please take a look! 🙏

kimkulling:
On my way, sorry for the delay!

kimkulling:

PR #3288, opened 2020-06-20
Check invalid vertex id for bone weight

infosia:
assimp should not error out even when asset has invalid vertex id for bone weight.  Just ignoring it seems fine.

How to reproduce:

1. Download following glTF file [crash.glb](https://github.com/infosia/themachinery-issues/blob/master/0001/crash.glb)
2. Check if glTF file is valid (i.e. Open it with Windows 3D Viewer)
3. Load the glTF file in assimp
4. assimp errors out because of std::vector out of range.

Expected: assimp should not error out even when asset has invalid vertex id. Just ignoring it seems fine.


kimkulling:
Mrged, thanks a lot for your contribution.

PR #3290, opened 2020-06-22
changed "Sample_SimpleOpenGL.c" to use c-style cast

thisisleobro:
changed "Sample_SimpleOpenGL.c" to use c-style cast on void do_motion() since it was not compiling.


thisisleobro:
@kimkulling This is my first merge request and i don't quite understand how it works. How could i improve since i see it was closed?

PR #3292, opened 2020-06-22
FBXExporter: Use scene metadata for global settings

rmstyrczula:
I've been writing an internal tool to modify FBX data, and the models I'm working on have non-standard axes settings and scale.

This modifies WriteGlobalSettings to re-output the available non-animation related metadata keys instead of a static default.

I didn't want to modify the FBXExporter header so I created some internal helper functions.


rmstyrczula:
C-style casts have been changed!

kimkulling:

PR #3294, opened 2020-06-22
FindInvalidDataProcess: Mark removed meshes as "out"

rmstyrczula:
There's a code path where an input scene only has a single mesh.
In this case, if ProcessMesh returns 2, we delete the mesh and move on
with the postprocessing. UpdateMeshReferences is not called and so the
deleted mesh is left dangling in nodes.

In a later step (SplitMesh in my testing), it then tries to deference a
null pointer and we explode.

With `out = true`, we can now hit the DeadlyImportError instead of a
segfault.

---

I guess this can be considered a continuation of #3235 :)


kimkulling:

PR #3297, opened 2020-06-24
Allow users to customize the behavior of assert violations

MalcolmTyrrell:
Changes:
* ai_assert no longer uses the assert macro from assert.h.
* A new implementation is provided (by default it writes to stderr and calls abort).
* An API in the Assimp namespace is provided which allows the user to replace the function called when a violation occurs.
* A unit test is provided (which also tests the replacement functionality in non-debug builds).


kimkulling:
Looks good, I will merge it!

kimkulling:
MMerged, thanks a lot for your contribution.

Issue #3025, opened 2020-02-23
glTF access to extras and extensions properties


ardenpm:
glTF 2.0 allows for any node to have an _extras_ and/or _extensions_ property. These may expose data which is not directly of interest to Assimp itself but could be useful for applications using Assimp. Right now only extensions actually used by Assimp are parsed and extras properties are ignored.

It would be extremely useful if Assimp preserved this data somewhere (possibly as generic metadata). I'm not sure if there is a generic concept of metadata in Assimp, however just extending the glTF 2.0 Asset would be fine, particularly since these are fairly glTF specific properties.

RichardTea:
Assimp does indeed have arbitrary metadata.

I am working on this, and hope to propose a PR in a couple of weeksm

Evangel63:
#3298 implements this for Nodes via Assimp's metadata structure. I only implemented it for the ones I need, but adding others shouldn't be too difficult.

kimkulling:
Merged, so I will close this issue.

Futuremappermydud:
Is there any documentation on this feature?

singhraj82:
Hi, is there any example on how to access  the node information from extension? I an able to parse the extension as shown below but where is the actual node information for Node 2 and 3 stored in metadata. I would really appreciate any hints/ideas. Thanks
```
{
                "MSFT_lod": {
                    "ids": [
                        1,
                        2,
                        3
                    ]
                }
            }
```

PR #3298, opened 2020-06-26
Added arbitrary recursive metadata to allow for glTF2's extensions to…

Evangel63:
… be properly represented.

Primary changes are to include/assimp/metadata.h, adding in the aiMetadata GetAiType function,
adding the operator= to allow an aiMetadata type to be assigned, adding a check for the
AI_AIMETADATA type flag as it can't be trivially memcpy'd.

operator= is implemented with a by-value argument as then the copy is made by the copy
constructor and we can just swap everything out and let the destructor handle the mess.

Implemented parsing of the "extensions" flag on all glTF2 Nodes. Doesn't use the ReadValue
helper function on numbers as it did not seem to fill out the Nullable structure properly.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Asset.h
+++ b/code/AssetLib/glTF2/glTF2Asset.h
@@ -784,6 +784,48 @@ struct Mesh : public Object {
     void Read(Value &pJSON_Object, Asset &pAsset_Root);
 };
 
+struct CustomExtension : public Object {
+    //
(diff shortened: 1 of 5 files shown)


Evangel63:
#3025 refers to wanting this feature. Currently only added for glTF2 Node datatypes as that's all I needed, but adding for others shouldn't be too difficult.

kimkulling:

PR #3300, opened 2020-06-28
Use strrchr() when finding the '.' that begins the file extension.

mahiuchun:
Sometimes we encounter file paths like ../foo/bar.obj; searching from the end of
the string would yield the correct result.


kimkulling:

PR #3301, opened 2020-06-29
Fix build when ASSIMP_DOUBLE_PRECISION is on.

mahiuchun:
Tested:

```sh
$ cmake -DASSIMP_DOUBLE_PRECISION=1 -DASSIMP_BUILD_MDL_IMPORTER=0 ..; make
```

The MDL importer is disabled due to the fact that there is some float-only code in MDL. This also means bin/assimp and bin/unit would not link successfully with the above build flags.


kimkulling:

PR #3303, opened 2020-06-29
Fixed variable in loop - HL1MDLLoader.cpp

malortie:
Commit 7e5a0acc48efc54d7aa7900c36cd63db1fbeec9b made changes to HL1MDLLoader.cpp - Several variables have been renamed. A loop variable `i` was renamed to `faceIdx`, but one reference was left unchanged. This would break polygon winding order. Please see line 832 below.

https://github.com/assimp/assimp/blob/4ff5a06ef441e1df782906878d0dbf96251206d3/code/AssetLib/MDL/HalfLife/HL1MDLLoader.cpp#L831-L845


kimkulling:

PR #3304, opened 2020-06-30
Buildbreak

breiting:


kimkulling:
Hi, why closed? Does it makes no sence?

Kim

breiting:
Hey Kim, sorry, that was a wrong merge attempt :-)
Well, to my opinion, it would make sense to really integrate this export filter upstream. If you guys are open to get that code in, we would do a bit more code cleanup and prepare a PR. REXfile is the file format for our Augmented Reality platform called [REXos](https://www.rexos.org). The format is open sourced and can be found [here](https://github.com/roboticeyes/openrex/blob/master/doc/rex-spec-v1.md). This exporter uses the openrex library to dump out the Assimp structure.

Would be great to get your feedback if this makes sense for upstream.

PR #3307, opened 2020-07-02
Ensure asserts are defined where expected.

MalcolmTyrrell:
It seems that sometimes ai_assert.h was included before defs.h, which meant that asserts were sometimes not active when they should be. (I didn't try to disentangle the include tree to work out exactly why or where this was happening.) 

The problem is easily fixed this by including defs.h from ai_assert.h.

Making this change also required:
* a fix to ArmaturePopulate.cpp (since an assert there would not compile).
* the newly added aiAssertViolation needed to be marked with ASSIMP_API.


kimkulling:

PR #3309, opened 2020-07-04
Improve ToBinary() for double precision.

mahiuchun:
The constant 0x80000000 is specific to 32 bit types. Make the bit mask according to the size of types.


kimkulling:
Merged, thanks a lot for your contribution. Makes this creepy code much easier to read for me!

PR #3320, opened 2020-07-10
add triangle strip support to AC file loader

IOBYTE:
The AC loader advertises .acc file support which is an extension of the .ac file format used by TORCS and Speed Dreams but doesn't support triangle strips which is one extensions added in .acc.  Multiple textures and normals are the other extensions added by .acc but they are not needed for basic rendering.


kimkulling:
closes https://github.com/assimp/assimp/issues/3324

kimkulling:

PR #3322, opened 2020-07-12
Build viewer and publish artifacts on windows-msvc.

AndrewJDR:
This commit introduces Github Actions support for building the tools and viewer and making these available for download as a zip file in the artifacts area of the Github Actions page. This allows for continuous validation that the viewer and tools build successfully, and the download is useful for quick testing of the very latest assimp functionality without needing to download and build it from source.

This only applies to windows-msvc, since the assimp viewer is only supported on that platform. It downloads the June 2010 DirectX SDK from the Microsoft servers and installs it. It also uses a cache to prevent having to perform this DX SDK download and installation repeatedly for every commit.

Note, it's necessary install the older June 2010 DXSDK within the github actions runner, because assimp uses the now deprecated D3DX libraries, and these libraries are not included in the stock Windows Server image provided by Github Actions.

Also note, no DirectX libraries (dll files etc) are included in the resulting artifacts zip file, so there is not a licensing issue here. The end user still needs to have these installed on their system. Usually they are already present.


AndrewJDR:
As an example, artifacts are available for download here (must be logged into Github):

https://github.com/assimp/assimp/pull/3322/checks?check_run_id=862160711
image

kimkulling:

AndrewJDR:
@kimkulling thanks for merging!

By the way, what is the canonical viewer for assimp currently?

When I did this work, I thought it was the win32 api/directx based one (`tools/assimp_view`)... but I also saw some references in code to a qt-based viewer...

kimkulling:
I try to develop a New ohne. But time is a critical Ressource.

AndrewJDR:
@kimkulling Oh absolutely, just curious. Thanks!

PR #3326, opened 2020-07-14
DefaultIOStream: Remove assert on empty count

rmstyrczula:
fwrite() is valid to call with a 0 count, and will simply return 0.
See:
    https://en.cppreference.com/w/cpp/io/c/fwrite
    http://www.cplusplus.com/reference/cstdio/fwrite/

There are code paths where StreamWriter will call Tell(), which calls
Flush(), which calls Write(buffer.data(), 1, buffer.size()). This can
happen when nothing has yet been written to the buffer, so size is 0.

---

I ran into this yet again with some FBX exporting. Through testing, I found this assert first pop up in  6205af4efbe05bdfab239753fb503cfefdaf4312.

My stack trace:

```
Assimp::aiAssertViolation AssertHandler.cpp:71
Assimp::DefaultIOStream::Write DefaultIOStream.cpp:103
Assimp::StreamWriter<false, false>::Flush StreamWriter.h:121
Assimp::StreamWriter<false, false>::Tell StreamWriter.h:143
Assimp::FBX::Node::BeginBinary FBXExportNode.cpp:316
Assimp::FBX::Node::Begin FBXExportNode.cpp:182
Assimp::FBXExporter::WriteHeaderExtension FBXExporter.cpp:326
Assimp::FBXExporter::WriteAllNodes FBXExporter.cpp:292
Assimp::FBXExporter::ExportBinary FBXExporter.cpp:177
Assimp::ExportSceneFBX FBXExporter.cpp:114
Assimp::Exporter::Export Exporter.cpp:481
Assimp::Exporter::Export Exporter.hpp:503
```

The exporter was called like:
```
Assimp::Exporter exp;
exp.Export(scene, "fbx", "/tmp/out.fbx");
```

Now, I could be fixing the symptom and not the problem, but it looks as if multiple calls in a row to `StreamWriter::Tell()` would cause this assert to be raised due to the `Flush()`, which seems like a bug.


kimkulling:
Mercedes, thanks for YouTube contribution.

PR #3328, opened 2020-07-15
Fix MinGW builds (issues related to pragmas and format strings)

awr1:
For some context:

- Because CMake is set to consider warnings as errors, the pragmas to disable warnings on MSVC are wrong: they check `_WIN32` when they should check `_MSC_VER`
- Format strings on Windows are weird in regards to printing `size_t`s. I use `"%zu"` on compliant platforms, "%Iu" for MSVC, and on MinGW use the MinGW `sprintf` instead of the MSCVRT one.


kimkulling:

PR #3329, opened 2020-07-15
Handle Gltf2 files where a value in a mesh index buffer is out of range.

MalcolmTyrrell:
When index values were out of range, they hit the assert in CheckValidFacesIndices. It seems inappropriate for assimp to assert under these circumstances, and also it seems incorrect to ignore this issue when running without asserts.

This change does the following:
* When a mesh has some faces which have out-of-range indices, those faces are dropped.
    * A warning is issued.
    * Note: We don't bother resizing the allocation, and allow it to be longer than mNumFaces.
* When all the faces in a mesh are dropped, we presume there is a more serious problem, and issue a DeadlyImportError.
    * I made the meshes array "exception safe" to avoid leaking in this case.
* Example models and unit tests for both of these scenarios are included
    * Warning: it does not to cover every possible input case.

Things which possibly could be added to this change:
* CheckValidFacesIndices is now unnecessary and could be removed.
* I propose leaving it in for now, since it is debug only and validates the new code.


kimkulling:
Looks good.

kimkulling:

PR #3330, opened 2020-07-16
fix invalid pointer for bone animation

infosia:
Fix invalid pointer for bone animation. 

How to reproduce:

1. Download following glTF file [RiggedFigure.out.glb](https://github.com/infosia/themachinery-issues/blob/master/0002/RiggedFigure.out.glb)
2. Check if glTF file is valid (i.e. Open it with Windows 3D Viewer)
3. Load the glTF file in assimp
4. assimp crashes because of bad pointer

Expected: assimp should not crash.


kimkulling:
Mercedes, thanks a lot for YouTube contribution.

PR #3333, opened 2020-07-17
ACLoader: Use Surface type enums

RichardTea:
Swap the Surface type magic numbers for enums


kimkulling:

PR #3336, opened 2020-07-17
Use #ifdef _MSC_VER for pragma warnings (Issue 3332)

### LINKED ISSUE #3332: Default cmake config "warnings as errors" does not build in MSVC 2017 v15.9.25 x64
**Describe the bug**
Recent changes for mingw have broken the MSVC 2017 build.

MSVC 2017 v15.9.25 sets _MSC_VER to 1916, thus checks of this form do not disable the related warnings:
```
#if _MSC_VER > 1920
#pragma warning(disable : 4315)
#endif
```

**To Reproduce**
1. Sync to master (37056f0242a4c6e1d822d0cb7f6b886b0589de20)
2. Clear cmake cache, generate for MSVC2017 x64
3. Build
4. See a huge spool of C4244, C4315 and C4505 warnings.

By default, these are treated as errors thus breaking the build.

RichardTea:
Fix the build on MSVC 2017

fixes #3332 

I do not have a mingw environment to test against, however I believe this should not have made any changes there.


RichardTea:
PR #3328 looks to resolve the issue as well, that may be better than this one as they have access to mingw.

kimkulling:

PR #3337, opened 2020-07-20
clang with msvc backend

MeyerFabian:
Hey I spend some time to make ``clang-cl`` and ``clang with msvc abi`` compile, build and succeed on tests on windows.

1. Changes most closely to my heart to use assimp with clang would be the changes to ``include/assimp/Compiler/poppack1.h`` and ``include/assimp/Compiler/pushpack1.h``. As i still can compile assimp with MSVC and use it with clang with just these two lines.
2. Slight changes to CMakeLists to recognize ``clang-cl`` and ``clang with msvc abi``.
3. Rest is various warnings treated as errors.


kimkulling:

PR #3349, opened 2020-07-30
FBXExport: Fix crash if scene->mMetaData is null

rmstyrczula:
A [comment](https://github.com/assimp/assimp/commit/6619ec82531e76ebb926d0ba9e90bc31623b765e) popped up yesterday regarding #3292. This adds a check if `mMetaData` is null.


kimkulling:

PR #3350, opened 2020-07-30
FBX Version/Size Check

MalcolmTyrrell:
If you export a really big model as FBX from 3ds Max (and possibly other tools) and you have the export version set to FBX2013, then you will generate a corrupt FBX file, without the modeller giving you any warning. We seen a few people hit this. Assimp currently gives quite a technical error (FBX Tokenisation error) in these circumstances, but it is convenient to explain to users the cause, since they can fix it themselves by adjusting the export version.

This change augments the error message to read as follows:
* The FBX file is invalid. This may be because the content is too big for this older version (7300) of the FBX format. (FBX-Tokenize (offset 0xf20311a7) block offset is out of range)

Btw, the underlying issue happens is because the older, 32-bit format cannot describe large offsets.


kimkulling:

PR #3356, opened 2020-08-03
Added more undefined sanitizer flags

qarmin:
Now it should show more undefined situations, because `-fsanitize=undefined` doesn't show all this things by default


qarmin:
Strange but doesn't show any new errors, but still may be helpful

qarmin:
The only error which I have is
```
/home/rafal/Pulpit/assimp/include/assimp/metadata.h:386:22: runtime error: load of value 112, which is not a valid value for type 'bool'
```
which means that this variable is used before setting value

qarmin:
With Clang 10 I also have this error:
```
/home/rafal/Pulpit/assimp/code/../contrib/rapidjson/include/rapidjson/internal/stack.h:117:13: runtime error: applying non-zero offset 16 to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/rafal/Pulpit/assimp/code/../contrib/rapidjson/include/rapidjson/internal/stack.h:117:13 in 
```

kimkulling:

PR #3372, opened 2020-08-15
NFF importer double precision support

lsliegeo:
The NFF importer parses values to floats, which aren't correctly written to doubles inside an `aiVector3D` when compiling with **ASSIMP_DOUBLE_PRECISION=ON**.

There is no 'normal' float to double conversion possible since the double (`f` in the line below, which lives in an `aiVector3D`) is incorrectly referenced as a float. Because of this incorrect reference, the parsed float is written to the double _as if the double was a float_, causing corrupt values.
https://github.com/assimp/assimp/blob/9fc00cbf74abcdae4fb5fdd8b64b5c58f8fce368/code/AssetLib/NFF/NFFLoader.cpp#L99

The issue is fixed by changing the parsed value from float to ai_real, which matches the ai_real inside the `aiVector3D`.


kimkulling:

Issue #3376, opened 2020-08-18
gltf2Importer crashes on Linux


ywangsc:
**Describe the bug**
This bug only happens on linux machine. 
Use ReadFile function of Assimp::Importer to import glb file, it crashes at gltf2Importer, related to json.
MemberIterator FindMember(const Ch* name) {
GenericValue n(StringRef(name));
return FindMember(n);
 }

**To reproduce**
You need to be on a Linux machine. 
Use gltf from https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphCube/glTF
In code:

```
Assimp::Importer myImporter;
myImporter.ReadFile("AnimatedMorphCube.gltf");
```

It will crash at parsing gltf json and crash inside gltf2Importer.

tellypresence:
@ywangsc Is this still a problem?

tellypresence:
Closing due to no response to followup

PR #3382, opened 2020-08-27
Fix RapidJSON defines and add Hunter builds to CI

rbsheth:
Incorporating the newest version of assimp into our build process worked initially, but caused a crash. This is because of a forced define inside assimp for RapidJSON that we didn't have in code that used RapidJSON.

I have removed the forced defines from the offending files and added the definitions to the CMake build. In case of a Hunter build, the settings for these defines will be propagated automatically through CMake.

I also spent some time adding three new Github Actions pipelines for building and testing the assimp code with Hunter. The tests and samples build with Hunter too.

PTAL @kimkulling


ywangsc:
Great! This should fix https://github.com/assimp/assimp/issues/3376.

rahulshethsc:
@kimkulling I matched the Github Actions names with the required checks to fix the bug. Would you be willing to enable one of the Hunter pipelines (Ubuntu, since it is the fastest) as required? This PR should be ready to merge otherwise.

kimkulling (approved):

kimkulling:

PR #3383, opened 2020-08-28
collada: set aiAnimation->mTicksPerSecond to 1000.0

crocdialer:
In both glTF2- and Collada-importers a hardcoded-value for `aiAnimation->mTicksPerSecond` is used, but it's different. this causes .dae animations to have a 1000-times too long duration (when mTicksPerSecond is correctly applied).

to reproduce I used this model:
- gltf2 -> https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/BrainStem/glTF-Binary/BrainStem.glb
- collada -> https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/sourceModels/BrainStem/BrainStem.dae

- sets aiAnimation->mTicksPerSecond to 1000.0 in ColladaLoader.cpp
- this is matches the behaviour of gltf2-importer


kimkulling (approved):

kimkulling:

crocdialer:
likewise, thanks for a great library!

PR #3385, opened 2020-08-30
Fix Bad Ownership Acquisition

jnhyatt:
This fixes a crash when using a custom IOSystem/IOStream implementation. `IOSystem::Open` is supposed to return an `IOStream` implementation, but the memory management is ambiguous. In some places within the loader code, here for example:
```c++
    IOStream * fileIO = pimpl->mIOHandler->Open( pFile );
    uint32_t fileSize = 0;
    if (fileIO)
    {
        fileSize = static_cast<uint32_t>(fileIO->FileSize());
        pimpl->mIOHandler->Close( fileIO );
    }
```
The `IOStream` pointer is acquired, closed, and never touched again. We get the pointer by calling `IOSystem::Open`, and then release it by calling `Close`. Later, in the .fbx and .obj loaders, I found this:
```c++
    std::unique_ptr<IOStream> fileStream(pIOHandler->Open(file, mode));
```
When this goes out of scope, it calls `delete`. This works perfectly fine for a `DefaultIOStream` because `IOSystem::Close` simply calls `delete` on the stream it's given. Does that mean that `Close` is expected or required to call `delete`? Why is this a pure virtual for `IOSystem` if that's the case? The pattern needs to be consistent. Either `~IOStream()` should be expected to close the stream, or else we must close the stream exclusively through `IOSystem::Close`. Either way, a `unique_ptr` acquiring the stream within a method far downstream from the object's instantiation is undesirable because ownership is quickly confused.

This is also a potential memory leak -- if we don't require that `IOSystem::Close` delete the stream passed to it, the stream will never get deleted, for example in Importer.cpp (the first posted code example). The `IOSystem` is requested to `Close` it, but without a guarantee that this method will delete it, there is a memory leak with no safeguards. If `Close` is required to free the memory, `Open` must be required to allocate memory, a seemingly arbitrary requirement. Better would be to rely on `Close` and `Open` to allow the `IOSystem` to manage any memory it needs however it wants.

In my case, the crash was caused by a double delete -- the `unique_ptr` in the .fbx and .obj loader deletes my memory, and then my `IOSystem` closes and deletes any streams it still records as open when it destructs.

This fix simply replaces the `unique_ptr`s with raw pointers when the streams are acquired. This reflects the fact that the importer does not acquire the `IOStream` when the file is read. Corresponding calls to `Close` are added. This prevents my double-delete situation and clears up potential memory leaks. Let me know if you have suggestions!

### DIFF
--- a/code/AssetLib/FBX/FBXImporter.cpp
+++ b/code/AssetLib/FBX/FBXImporter.cpp
@@ -141,8 +141,11 @@ void FBXImporter::SetupProperties(const Importer *pImp) {
 // ------------------------------------------------------------------------------------------------
 // Imports the given file into the given scene structure.
 void FBXImporter::InternReadFile(const std::string &pFile, aiScene *pScene, IOSystem *pIOHandler) {
-	IOStream* pStream = pIOHandler->Open(pFile, "rb");
-	if (!pStream) {
+	auto streamCloser = [&](IOStream *pStream) {
+		pIOHandler->Close(pStream);
+	};
+	std::unique_ptr<IOStream, decltype(streamCloser)> stream(pIOHandler->Open(pFile, "rb"), streamCloser);
+	if (!stream) {
 		ThrowException("Could not open file for reading");
 	}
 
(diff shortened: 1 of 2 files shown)


jnhyatt:
CI threw some errors from the address sanitizer, and after some digging, I found that exceptions were caught outside the area where the memory for the IOStream is Opened and Closed, skipping the call to `Close` and leaking the memory. I assume this was the purpose of using the `unique_ptr`, to guarantee deletion of the stream. However, the problem still stands -- we can't simply call `delete` and expect implementations to work properly. To fix this, I added a custom deleter to the `unique_ptr` that properly closes the stream. For the default stream, this will indeed delete it. For custom streams, this will give them the opportunity to free memory themselves.

kimkulling:
Looks good. Let's see what CI tells us about you work. When CI is fine with that I will merge it!

kimkulling:

kimkulling:
Nice investigation, thanks for your detailed explanation!

PR #3394, opened 2020-09-10
GLTF2: Throw instead of assert when input file is invalid.

ms-maxvollmer:
What: Throw instead of assert when input file is invalid.

Why: Assimp shouldn't crash on invalid files. Asserts are disabled on Release builds.


ms-maxvollmer:
@kkulling Not sure if that's already planned, but I think we might at some point want to replace all asserts that check if a file is valid with throws, or with a simply if-statement, depending on context. Unfortunately as we are all constrained, for now I can only provide this local fix. I hope it's at least a step in the right direction :)

kimkulling (approved):

kimkulling:

PR #3412, opened 2020-10-02
Fbx report asset issues properly

MalcolmTyrrell:
We've started testing invalid assets to ensure we correctly reporting the error to users. We assume that bad input data should always lead to a DeadlyImportError being thrown.

This PR fixes two cases where an invalid FBX would not cause a DeadlyImportError:
1. An assert in BaseImporter::UpdateImporterScale states that fileScale should not be 0, but this was not checked during FBX load.
2. The float parsing routines in fast_atof.h always threw a std::invalid_argument exception. This has been generalized so they can throw a DeadlyImportError. The caller can decide what exception to throw but since almost every case was in an importer, DeadlyImportError was the default.


kimkulling (approved):

kimkulling:

PR #3446, opened 2020-10-09
fix for fbx files using stingray materials;

stromaster:


kimkulling:

PR #3452, opened 2020-10-12
Update Hunter for pugixml

rbsheth:
I saw https://github.com/assimp/assimp/commit/52826a03f8a8eb722221b595dc5eb993338eb334 but unfortunately pugixml uses a non-standard target name so that doesn't work. As described here https://github.com/cpp-pm/hunter/blob/master/examples/pugixml/CMakeLists.txt no namespace is needed. After changing the behavior for Hunter builds, things are working again properly. 

PTAL @kimkulling


kimkulling (approved):

kimkulling:

Issue #3445, opened 2020-10-08
Compilation failed


hyt589:
**Describe the bug**
Compilation failed with `CMake`. Bug seems to be from `assimp/code/AssetLib/XGL/XGLLoader.cpp`, line 602.

**To Reproduce**
Compile Assimp as a subdirectory using `CMake`:
```cmake
add_subdirectory(libs/assimp)
```

**Expected behavior**

**CMake Error message**
```
[build] /Users/hyt/workspaces/games101/big project/libs/assimp/code/AssetLib/XGL/XGLLoader.cpp:602:20: error: loop variable 'p' has type 'const pairt &' (aka 'const pair<unsigned int, Assimp::XGLImporter::TempMaterialMesh> &') but is initialized with type 'std::__1::__map_iterator<std::__1::__tree_iterator<std::__1::__value_type<unsigned int, Assimp::XGLImporter::TempMaterialMesh>, std::__1::__tree_node<std::__1::__value_type<unsigned int, Assimp::XGLImporter::TempMaterialMesh>, void *> *, long> >::value_type' (aka 'pair<const unsigned int, Assimp::XGLImporter::TempMaterialMesh>') resulting in a copy [-Werror,-Wrange-loop-analysis]
[build]         for (const pairt &p : bymat) {
[build]                           ^
[build] /Users/hyt/workspaces/games101/big project/libs/assimp/code/AssetLib/XGL/XGLLoader.cpp:602:7: note: use non-reference type 'pairt' (aka 'pair<unsigned int, Assimp::XGLImporter::TempMaterialMesh>') to keep the copy or type 'const std::__1::__map_iterator<std::__1::__tree_iterator<std::__1::__value_type<unsigned int, Assimp::XGLImporter::TempMaterialMesh>, std::__1::__tree_node<std::__1::__value_type<unsigned int, Assimp::XGLImporter::TempMaterialMesh>, void *> *, long> >::value_type &' (aka 'const pair<const unsigned int, Assimp::XGLImporter::TempMaterialMesh> &') to prevent copying
[build]         for (const pairt &p : bymat) {
[build]              ^~~~~~~~~~~~~~~~
```

**Desktop (please complete the following information):**
 - OS: macOS Catalina
 - Version 10.15.5

kimkulling:
Shall be closed!

PR #3454, opened 2020-10-13
Fix for issue #3445

jsmaatta:
Add const to avoid creating a temporary copy.

This should fix the cmake compilation error on XGLLoader.cpp shown in bug #3445

The fix is similar to commit bbe6f7f213b5e2ee689146df5c068dc1a55ea919


kimkulling:
Merged, thank a lot for your contribution.

PR #3460, opened 2020-10-15
AI_CONFIG_IMPORT_FBX_READ_WEIGHTS

MalcolmTyrrell:
It's useful to be able to configure the importer to ignore unneeded sections. Cameras, Lights and Animations were already configurable, but an option for weights was missing. This PR adds that option.

To allow the option to be configurable, the change in FBXImporter::SetupProperties is necessary. It isn't absolutely necessary to take the change in the config header, since customer code can just use a hard-coded copy of the property string. For consistency, I thought it was appropriate to add it. 

It might be good for these FBX specific settings to move out of the general config into a header which can be optional included by client code, but I haven't made that change in this PR.


kimkulling (approved):

kimkulling:

PR #3461, opened 2020-10-16
glTF1's orthgraphic camera & glTF2's skinning

inhosens:
1. glTF1's orthographic camera is not patched yet
2. glTF2's skinning ignored some bones that do not have weights since a moment.
    It is not correct approach because the empty bones can connect a bone with another.


kimkulling:

PR #3471, opened 2020-10-21
MSVC crash while importing fbx model workaround

MomoDeve:
I do not know why, but the following line causes crash if assimp was compiled using MSVC with RelWithDebInfo configuration:
`std::transform(shading.begin(), shading.end(), shading.begin(), Assimp::ToLower<char>);`
replacing `shading.begin()` with `shading.data()` fixes this issue.
изображение
you can download this model to reproduce the error: [teapot.zip](https://github.com/assimp/assimp/files/5418564/teapot.zip)


kimkulling (approved):

kimkulling:

RichardTea:
Which version of MSVC did you see this crash in?
I seem to recall that MS have fixed few code generation bugs around this sort of area.

MomoDeve:
MSVC version is 14.27.29110

PR #3484, opened 2020-11-02
Optimize FindDegenerates so it doesn't explode

MalcolmTyrrell:
The situation when meshes get completely removed because they only contain degenerate triangles is probably rare. However, it can happen that bad quality input data can have many small meshes whose triangles are all considered to be degenerate. The behaviour of the current processor is extremely poor in that situation.

In one case, this simple optimization brings the time of that processor down from ~3 hours to 14 seconds.


kimkulling (approved):

kimkulling:

PR #3487, opened 2020-11-03
Prevent to generate redundant morph targets for glTF2

inhosens:
For glTF2 morph animation,
assimp always makes morph targets' attributes based on the original mesh.

This patch prevents to make redundant morph targets.


kimkulling (approved):

kimkulling:

PR #3490, opened 2020-11-06
GLTF2: Null bufferview crash fix

MalcolmTyrrell:
A missing bufferview is not actually a bug (https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#sparse-accessors), but the code which checked bounds was dereferencing it and causing a crash. I've just put that code inside a conditional.

We were testing with an asset that is actually invalid, so we later got a DeadlyImportError "GLTF: data is NULL". That error message isn't particularly helpful, so I've also made an additional minor change to improve the message. It will now read: "GLTF2: data is null when extracting data from accessors[1]"

I've included another GLTF fix here: When element size is bigger than expected, GetValue can corrupt memory.


kimkulling:

PR #3502, opened 2020-11-16
Prevent crash with malformed texture reference

MalcolmTyrrell:
If you are loading in non-strict mode and the texture reference in the input FBX is malformed (for example, if the RelativeFilename is not a string) then the returned texture value is null. 

This change just prevents the value being dereferenced if the value is null.


kimkulling (approved):

kimkulling:

PR #3504, opened 2020-11-19
FBXParser.cpp - handle buffer over-read correctly

Neil-Clifford-FB:
This Pull-Request solves 2 issues w.r.t. FBXParser:
1. Handle potential integer overflow when calculating the amount of data to be read from buffer
   - i.e. count = INT_MAX
2. Raises ParseError in the case whereby trying to read more data than available in buffer
   - this issue can arise from corrupted FBX or maliciously modified files


kimkulling (approved):

kimkulling:

PR #3505, opened 2020-11-19
Sceneprecessor - potential memory leak

Neil-Clifford-FB:
Potential memory leak when allocating dummy Position, Rotation, Scaling tracks for the animation channel.
Ensures keys arrays are free'd before allocating the dummy arrays.


kimkulling (approved):

kimkulling:

PR #3506, opened 2020-11-19
Fbx exception safety

MalcolmTyrrell:
An exception in the FBXConvertor can lead to a crash when deleting the node chains. 

The crash occurs because assimp tries to destroy a vector of objects where there is already an ownership relationship between them, leading to double deletes.

I've adjusted the code so the node chains carry an object which may or may not own its node. The diff looks messy, but the changes are pretty minor.


kimkulling:

PR #3511, opened 2020-11-27
Simplification: textures_converted keys can just be pointers

MalcolmTyrrell:
In #2741, the keys of the textures_converted map were switched to Video objects rather than pointers. I don't think that's necessary:
* The video object are owned by the document
* In all the code I've looked at, the video objects are always const, so the keys always continue to match the objects they were copied from. 

All unit tests, including the one provided with that PR, still pass when pointers are used. (And the tests we use internally also pass.)

I thought I needed this PR for another change, but that is no longer true. However, I'll leave it here as an optional simplification.


kimkulling (approved):

kimkulling:
Looks good, merged. Thanks a lot for your contribution!

Issue #3489, opened 2020-11-06
SplitByBoneCountProcess looses morph targets along the way


boguscoder:
**Describe the bug**
when aiProcess_SplitByBoneCount is on and bone count is high enough so that SplitByBoneCountProcess runs all blendshapes are lost

**To Reproduce**
Have aiProcess_SplitByBoneCount set for import and import glb with large bone count (Im still checking if I can attach my .glb to the ticket since it could have IP problems)

**Expected behavior**
meshes are split but they still have blendshapes intact

**Desktop (please complete the following information):**
 All platforms

boguscoder:
I can propose a PR with smth like code below added to end of `SplitByBoneCountProcess::SplitMesh`
```
        // ... and copy all the morph targets for all the vertices which made it into the new submesh
        if (pMesh->mNumAnimMeshes > 0) {
            newMesh->mNumAnimMeshes = pMesh->mNumAnimMeshes;
            newMesh->mAnimMeshes = new aiAnimMesh*[newMesh->mNumAnimMeshes];
            
            for (unsigned int morphIdx = 0; morphIdx < newMesh->mNumAnimMeshes; ++morphIdx) {
                aiAnimMesh* origTarget = pMesh->mAnimMeshes[morphIdx];
                aiAnimMesh* newTarget = new aiAnimMesh;
                newTarget->mName = origTarget->mName;
                newTarget->mWeight = origTarget->mWeight;
                newTarget->mNumVertices = numSubMeshVertices;
                newTarget->mVertices = new aiVector3D[numSubMeshVertices];
                newMesh->mAnimMeshes[morphIdx] = newTarget;
                
                if (origTarget->HasNormals()) {
                    newTarget->mNormals = new aiVector3D[numSubMeshVertices];
                }
                
                if (origTarget->HasTangentsAndBitangents()) {
                    newTarget->mTangents = new aiVector3D[numSubMeshVertices];
                    newTarget->mBitangents = new aiVector3D[numSubMeshVertices];
                }
                
                for( unsigned int vi = 0; vi < numSubMeshVertices; ++vi) {
                    // find the source vertex for it in the source mesh
                    unsigned int previousIndex = previousVertexIndices[vi];
                    newTarget->mVertices[vi] = origTarget->mVertices[previousIndex];

                    if (newTarget->HasNormals()) {
                        newTarget->mNormals[vi] = origTarget->mNormals[previousIndex];
                    }
[...]

PR #3512, opened 2020-11-28
Fix for 3489 | Preserve morph targets when splitting by bone count

boguscoder:
Proposed fix for https://github.com/assimp/assimp/issues/3489


kimkulling (approved):

kimkulling:

PR #3518, opened 2020-12-01
Check _MSC_VER for MSVC specific pragma directives.

Biswa96:
This fixes build issues when compiled with compiles like Clang or GCC.


kimkulling (approved):

kimkulling:
Merged, thaks a lot for your contribution.

PR #3519, opened 2020-12-02
3ds Max 2021 PBR Materials in FBX

MalcolmTyrrell:
3ds Max 2021 has PBR material types (Metalness/Roughness and Specular/Gloss), which are stated to be scripted wrappers over the pre-existing Physical material. This change ensures that the textures and properties of the new Pbr Materials in 3ds Max 2021 are carried by the aiMaterial object.

Some of the new properties are duplicated by existing properties (possibly via a calculation). I have not attempted to translate the properties directly to common material properties, as the types are not always in agreement (RGBA vs RGB, for example). However, the changes I made to the property parsing means that the values are there as raw properties. A client that wants to create a PBR Material can check for the raw properties rather than trying to use the common material properties.

Note: In both of the new material types, there is a toggle in the UI selecting between glossiness and roughness. The property for the texture changes when this is toggled, but the property in the FBX file keeps its name (roughness or glossiness, depending on which of the two material types is chosen). There's a property of type "Integer" which communicates which of these is actually selected.


MalcolmTyrrell:
Temporarily closing while I investigate an issue.

MalcolmTyrrell:
Found a bug in our client code, so this PR should still be OK.

kimkulling (approved):

kimkulling:

PR #3524, opened 2020-12-07
FBXConverter : Fix timescales of FBX animations

inhosens:
FBX animations were recorded by framenumber, not by time.
This patch will change it as a base of milliseconds.

### DIFF
--- a/code/AssetLib/FBX/FBXConverter.cpp
+++ b/code/AssetLib/FBX/FBXConverter.cpp
@@ -79,7 +79,7 @@ using namespace Util;
 
 #define MAGIC_NODE_TAG "_$AssimpFbx$"
 
-#define CONVERT_FBX_TIME(time) static_cast<double>(time) / 46186158000LL
+#define CONVERT_FBX_TIME(time) (static_cast<double>(time) * 1000.0 / 46186158000LL)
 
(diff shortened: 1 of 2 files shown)


kimkulling (approved):

kimkulling:

code/AssetLib/FBX/FBXConverter.cpp
```diff
@@ -79,7 +79,7 @@ using namespace Util;
 
 #define MAGIC_NODE_TAG "_$AssimpFbx$"
 
-#define CONVERT_FBX_TIME(time) static_cast<double>(time) / 46186158000LL
+#define CONVERT_FBX_TIME(time) (static_cast<double>(time) * 1000.0 / 46186158000LL)
```
jeanlemotan, line 82: Shouldn't you also change `anim->mTicksPerSecond = anim_fps;` to `anim->mTicksPerSecond = 1000.0;` ?
The documentation mentions that animation times are expressed in ticks. To convert from ticks to seconds, a division by mTicksPerSecond is required.

corporateshark:
Animation duration loaded from https://developer.nvidia.com/orca/amazon-lumberyard-bistro is broken after this patch.

PR #3527, opened 2020-12-08
Disable rapidjson #define for android

tellypresence:
Originally added to fix warning on MSVC but breaks android build


tellypresence:
Superseded by https://github.com/assimp/assimp/pull/3528 which provides option for better control

Issue #3054, opened 2020-03-07
Gltf exporter converts multiple channels to multiple animations


gongminmin:
Repro steps:
1. Import a fbx with 1 animation, it contains 3 channels, each channel is touching 1 node's translation, rotation, and scale.
2. Export it to either gltf or glb
3. Import the gltf or glb file, then it has 3 animations, each one contains 1 channel of a node's translation, rotation, and scale. Also, animation names are not exported.

Issue #2234, opened 2018-11-20
FBX import to GLTF2 export does not handle multiple animations (nor one...)


fire:
1. Does not write the name of the animation
1. Splits the animation so each bone is a separate channel  and track.

![applicationframehost_2018-11-20_10-27-07](https://user-images.githubusercontent.com/32321/48794467-e8dc0c80-ecae-11e8-8237-2cea8c452cd0.png)

Image shows 17 animation takes, but the scene only has one animation in scene->mNumAnimations.

fire:
Model is from sketchfab https://sketchfab.com/models/7ece1ee1b602436dbcfb77e4f58ba8e5

fire:
The Animation name should be "Take 001". 

1. mNumChannels = 17
1. mNumMeshChannels = 0

fire:
[assimp_fbx_gltf2_debug.zip](https://github.com/assimp/assimp/files/2601116/assimp_fbx_gltf2_debug.zip)

Modified from https://sketchfab.com/models/7ece1ee1b602436dbcfb77e4f58ba8e5 has the gltf2 file and the fbx file.

fire:
https://gist.github.com/fire/8a894f3e76234e888d6b38ef7df45433 Has some analysis.

The animations has 17 animations instead of 1 animation with a name and 17 channels.

fire:
https://sketchfab.com/models/37c871cbd9174f67bda4c10d19a97ed0 has multiple animations and is importable. Can't figure out how to inject the animation name though.

kimkulling:
Thanks a lot for the report. I will try to take care.

fire:
There's an error near `name = mAsset->FindUniqueID(name, "animation");` If the animation is unnamed, it fails.

```c++
void glTF2Exporter::ExportAnimations()
{
    Ref<Buffer> bufferRef = mAsset->buffers.Get(unsigned (0));

    for (unsigned int i = 0; i < mScene->mNumAnimations; ++i) {
        const aiAnimation* anim = mScene->mAnimations[i];
        const float ticksPerSecond = static_cast<float>(anim->mTicksPerSecond);

        std::string name = "anim";
        Ref<Animation> animRef;
        Ref<Node> animNode;
        if (anim->mName.length > 0) {
            name = anim->mName.C_Str();	
        }

		name = mAsset->FindUniqueID(name, "animation");
		animRef = mAsset->animations.Create(name);		

[...]

PR #3531, opened 2020-12-10
Fixed a crash of the Gltf 2 exporter in the case of an animation without scale animation key.

JLouis-B:
In the GLTF 2 exporter, when we build the animation samplers, we check that the animation channel has at least one key, otherwise no need to add it.
The problem is that in this case, we don't initialize the sampler but we add them anyway which causes a crash at the time of writing.
The solution is simply not to add the animation channel in this case.


JLouis-B:
I added a commit that fixes another bug with this function: In an animation, the animation of each bone is exported in a different animation for each bone instead of a single common animation (https://github.com/assimp/assimp/issues/3054, https://github.com/assimp/assimp/issues/2234)

kimkulling (approved):

kimkulling:
Merged, thanks a lot!

Issue #3144, opened 2020-04-09
Compilation error in MinGW (ZipArchiveIOSystem.cpp)


daniilpetrov:
There is an error during the build:
> D:\NativeEngine\EngineSources\DevelopmentKits\GraphicFormats\Assimp 5.0.1\assimp-master\code\Common\ZipArchiveIOSystem.cpp: In static member function 'static zlib_filefunc_def Assimp::IOSystem2Unzip::get(Assimp::IOSystem*)':
> D:\NativeEngine\EngineSources\DevelopmentKits\GraphicFormats\Assimp 5.0.1\assimp-master\code\Common\ZipArchiveIOSystem.cpp:161:30: error: invalid conversion from 'uLong (*)(voidpf, voidpf, void*, uLong)' {aka 'long unsigned int (*)(void*, void*, void*, long unsigned int)'} to 'read_file_func' {aka 'unsigned int (*)(void*, void*, void*, unsigned int)'} [-fpermissive]
>   161 |         mapping.zread_file = read;
>       |                              ^~~~
>       |                              |
>       |                              uLong (*)(voidpf, voidpf, void*, uLong) {aka long unsigned int (*)(void*, void*, void*, long unsigned int)}
> D:\NativeEngine\EngineSources\DevelopmentKits\GraphicFormats\Assimp 5.0.1\assimp-master\code\Common\ZipArchiveIOSystem.cpp:162:31: error: invalid conversion from 'uLong (*)(voidpf, voidpf, const void*, uLong)' {aka 'long unsigned int (*)(void*, void*, const void*, long unsigned int)'} to 'write_file_func' {aka 'unsigned int (*)(void*, void*, const void*, unsigned int)'} [-fpermissive]
>   162 |         mapping.zwrite_file = write;
>       |                               ^~~~~
>       |                               |
>       |                               uLong (*)(voidpf, voidpf, const void*, uLong) {aka long unsigned int (*)(void*, void*, const void*, long unsigned int)}
> D:\NativeEngine\EngineSources\DevelopmentKits\GraphicFormats\Assimp 5.0.1\assimp-master\code\Common\ZipArchiveIOSystem.cpp:164:30: error: invalid conversion from 'long int (*)(voidpf, voidpf, uLong, int)' {aka 'long int (*)(void*, void*, long unsigned int, int)'} to 'seek_file_func' {aka 'long int (*)(void*, void*, unsigned int, int)'} [-fpermissive]
[...]

PR #3533, opened 2020-12-11
Common: Fix GCC error invalid conversion in MINGW.

Biswa96:
Fixes https://github.com/assimp/assimp/issues/3144

### DIFF
--- a/code/AssetLib/Collada/ColladaHelper.cpp
+++ b/code/AssetLib/Collada/ColladaHelper.cpp
@@ -53,7 +53,7 @@ const MetaKeyPairVector MakeColladaAssimpMetaKeys() {
     result.emplace_back("authoring_tool", AI_METADATA_SOURCE_GENERATOR);
     result.emplace_back("copyright", AI_METADATA_SOURCE_COPYRIGHT);
     return result;
-};
+}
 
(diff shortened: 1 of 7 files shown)


kimkulling (approved):

kimkulling:

JLouis-B:
Hello,
Assimp fail to compile with MinGW 7.3 since this PR :
`assimp-master\code\Common\ZipArchiveIOSystem.cpp|156|error: 'error_file_func' was not declared in this scope|`

Biswa96:
I have compiled assimp in MSYS2/MinGW-w64 toolchain without any issue now. Try to update your MinGW setup or use MSYS2 which provides latest toolchain.

Biswa96:
I have used MSYS2/MinGW-w64 toolchain to compile assimp. The `error_file_func` function pointer present in minizip v1.2 package here https://github.com/nmoinvaz/minizip/blob/1.2/ioapi.h

JLouis-B:
The issue is that ZipArchiveIOSystem.cpp use error_file_func which is defined in minizip 1.2 as you show, but assimp/contrib seems to be based on minizip 1.1 which does not define error_file_func but testerror_file_func

assimp/contrib/unzip+zip should probably be updated

PR #3540, opened 2020-12-14
 _dest may be destructed twice if _dest is not null in MergeScenes()

wasd845:


kimkulling (approved):

kimkulling:

PR #3541, opened 2020-12-14
GLTF2 fixes

ms-maxvollmer:
* Fix memory leak in GLTF2 Importer that occurs when an exception gets thrown while constructing a scene
* Fix crash in GLTF2 Importer when a scene with dangling/uninitialized pointers gets deleted after an exception was thrown
* Fix crash in GLTF2 Importer when channel sampler is not set (negative index)
* Throw instead of assert on invalid file input

I recommend viewing the changes in a third party tool, as github has problems with the indentation changes. Most of the big blocks are just the same code wrapped in a try-catch-throw to free memory that is otherwise lost after the exception is thrown.


kimkulling (approved):

kimkulling:

PR #3542, opened 2020-12-14
FBXBinaryTokenizer: Check length of property

ms-maxvollmer:
Throws an exception if property length is out of bounds. This wasn't checked before and could lead to memory being read from an invalid location.


kimkulling (approved):

kimkulling:
Looks good., I will merge it!

kimkulling:

PR #3543, opened 2020-12-15
Only consider continuation tokens at end of line

### LINKED ISSUE #3532: Problem with obj files and backslash (\)
UPDATE: The previous description was for how to reproduce this issue with [UnityMeshImporter](https://github.com/eastskykang/UnityMeshImporter), which uses [AssimpNet](https://github.com/assimp/assimp-net), but I have managed to recreate it with `assimp.exe` too, so I'm describing that workflow instead.

**Describe the bug**

In an obj file, having a backslash (`\`) in a comment before a vertex line seems to remove the vertex line from parsing.

**To Reproduce**

Try the following command with the obj file in [head.zip](https://github.com/assimp/assimp/files/5678264/head.zip): 
`assimp.exe info head.obj`.

**Expected behavior**

Information about the model is described.

**Actual behavior**

You get the following output:
```
ERROR: Failed to load file: OBJ: vertex index out of range
assimp info: Unable to load input file head.obj
```

**Screenshots**

Screenshot of what the model looks like.
image

**Desktop (please complete the following information):**

 - OS: Windows 10
 - Version: Update 2004

**Additional information**

I have done some research, and the problem goes away if I change line 6-7 from this

```
# head\Solid
v 1.32007420063019 0.0356011390686035 1.83193397521973
```

to either this:

```
# head-Solid
v 1.32007420063019 0.0356011390686035 1.83193397521973
```

or this

```
# head\Solid

v 1.32007420063019 0.0356011390686035 1.83193397521973
```

or by removing the comment altogether. It makes me think that the backslash somehow makes the parser believe that the row following it is part of the same row as the comment row (so that the vertex becomes part of the comment).

gris-martin:
Fixes #3532 

Before this PR, blocks parsed with `getNextDataLine(buffer, continuationToken)` will join the current line with the next if the `continuationToken` is encountered, no matter where in the line that is. It means that e.g. this:

```
# Path to model: Model\Mesh1
v -1.92181043063553E-16 -1.13401524746739E-16 \
9.81342513114214
```
will result in a buffer entry looking like this:
```
# Path to model: Model\Mesh1v -1.92181043063553E-16 -1.13401524746739E-16 9.81342513114214
```
given that `continuationToken` == `'\\'`

After this PR it will be parsed like this:
Call 1:
```
# Path to model: Model\Mesh1
```
Call 2:
```
v -1.92181043063553E-16 -1.13401524746739E-16 9.81342513114214
```


kimkulling (approved):

kimkulling:

PR #3554, opened 2020-12-23
Update minizip

JLouis-B:
Hi,

Compiling Assimp with MinGW (without ASSIMP_USE_HUNTER) doesn't work because of a recent change in ZipArchiveIOSystem.cpp :
```cpp
#if defined (ASSIMP_USE_HUNTER) || defined (__MINGW32__) // GH#3144
    mapping.zopen_file = (open_file_func)open;
    mapping.zread_file = (read_file_func)read;
    mapping.zwrite_file = (write_file_func)write;
    mapping.ztell_file = (tell_file_func)tell;
    mapping.zseek_file = (seek_file_func)seek;
    mapping.zclose_file = (close_file_func)close;
    mapping.zerror_file = (error_file_func)testerror;
#else
    mapping.zopen_file = open;
    mapping.zread_file = read;
    mapping.zwrite_file = write;
    mapping.ztell_file = tell;
    mapping.zseek_file = seek;
    mapping.zclose_file = close;
    mapping.zerror_file = testerror;
#endif
```

The issue is that the Minizip version provided by Hunter seems to be a Minizip 1.2 version which define `error_file_func`
And on the other hand, the version of minizip used in Assimp (in contrib/zip and contrib/unzip) is much less recent and doesn't define `error_file_func` but `testerror_file_func`
A quick and easy fix would be to remove the `|| defined (__MINGW32__)` and adding cast to the Minizip `testerror_file_func` in the else.

IMO a better solution would be to update the version of Minizip used in contrib (so we use similar headers in Hunter and contrib Minizip), which I propose via this PR.

The changed have been compiled and tested for MinGW and VS 2019, by exporting a 3MF package and reload it with Assimp Viewer

### DIFF
--- a/code/AssetLib/3MF/D3MFExporter.cpp
+++ b/code/AssetLib/3MF/D3MFExporter.cpp
@@ -395,7 +395,8 @@ void D3MFExporter::addFileInZip(const std::string& entry, const std::string& exp
     nullptr, 0, // comment
     6); // similar to the previous ZIP_DEFAULT_COMPRESSION_LEVEL value
 
-    zipWriteInFileInZip(m_zipArchive, exportTxt.c_str(), exportTxt.size());
+    uint32_t contentLength = static_cast<uint32_t>(exportTxt.size());
+    zipWriteInFileInZip(m_zipArchive, exportTxt.c_str(), contentLength);
 
     zipCloseFileInZip(m_zipArchive);
 }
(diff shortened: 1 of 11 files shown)


JLouis-B:
The version I upgraded to is not compatible with Hunter
I will start again from Hunter's package and I will reopen another PR

PR #3556, opened 2020-12-24
Update unzip contrib

JLouis-B:
Hi,
Compiling Assimp with MinGW doesn't work because of a recent change in ZipArchiveIOSystem.cpp :
```cpp
#if defined (ASSIMP_USE_HUNTER) || defined (__MINGW32__) // GH#3144
    mapping.zopen_file = (open_file_func)open;
    mapping.zread_file = (read_file_func)read;
    mapping.zwrite_file = (write_file_func)write;
    mapping.ztell_file = (tell_file_func)tell;
    mapping.zseek_file = (seek_file_func)seek;
    mapping.zclose_file = (close_file_func)close;
    mapping.zerror_file = (error_file_func)testerror;
#else
    mapping.zopen_file = open;
    mapping.zread_file = read;
    mapping.zwrite_file = write;
    mapping.ztell_file = tell;
    mapping.zseek_file = seek;
    mapping.zclose_file = close;
    mapping.zerror_file = testerror;
#endif
```

The issue is that the unzip provided by contrib and Hunter aren't based on the same Minizip version. The package provided by Hunter (based on a more recent minizip) define `error_file_func` and on the other hand, contrib use an older version of ioapi.h which doesn't define error_file_func but `testerror_file_func`
Although the addition of the ```|| defined (__MINGW32__) ``` does not seem to be an elegant solution to the initial issue since it generates others that I have just explained (in fact it only works in the case of an Hunter's unzip version.), a good solution seems to be to update contrib/unzip with a similar version of Hunter package, which I propose via this PR 

Compilation has been tested on MinGW GCC 8.1 and VS 2019 (with and without Hunter), and zip creation extraction have been tested by exporting a 3MF package and reload it with Assimp Viewer


JLouis-B:
Still needs some adaptations in ZipArchiveIOSystem to support large zip files due to the fact that the new versions of unzip support buffer sizes up to 65kb.

JLouis-B:
Tested with large files, it should be good now

kimkulling:

PR #3561, opened 2020-12-27
contrib/zlib/CMakeLists.txt: don't install zlib

ffontaine:
Don't install zlib static library to avoid potential conflicts with a system-wide zlib package

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


kimkulling:
This will not work for the windows systems in that way.

kimkulling:

makitsune:
When configuring CMake with `-DBUILD_SHARED_LIBS=OFF` and `-DASSIMP_BUILD_ZLIB=ON` on Linux (probably on macOS too), it fails with the error message:

`CMake Error: install(EXPORT "assimpTargets" ...) includes target "assimp" which requires target "zlibstatic" that is not in any export set.`

`libassimp.a` needs `libzlibstatic.a` which isn't available because of this pull request.

ffontaine:
Indeed, I think this PR should be reverted as this solution is unfortunately not working. I sent it because we had a build failure in buildroot with hiawatha and the zlib library installed by assimp. Instead of this PR, I'll disable the build of the embedded zlib library through `-DASSIMP_BUILD_ZLIB=OFF`. I'll also add a dependency on libzlib as it seems that assimp 5.0.1 is not compatible with zlib-ng.

ademets:
Yes, please revert. The IF(Win32) condition breaks all non-Windows builds, that want to build assimp with statically linked zlib.

PR #3582, opened 2021-01-12
Fix glTF vertex colors with types other than float

makitsune:
Whilst investigating why vertex colors weren't working, I figured out that older versions of Blender (before 2.91) exported vertex colors as **float (5126)** but has since changed to using **unsigned short (5123)**. I found this issue along the way: https://github.com/godotengine/godot/issues/44798

In the assimp code during `ExtractData`, it was copying 4 unsigned shorts (2 bytes, 8 total) to 4 floats or rather `aiColor4D` (4 bytes, 16 total). Because they aren't compatible, my fix was to create an array of `aiColor4t<unsigned short>` and export data to that.

This was only half of the problem because aiMesh colors expect aiColor4D (float). So the second part converts the `aiColor4t<unsigned short>` to `aiColor4D` by dividing each value by the maximum size of unsigned short.

This had to also be done for **unsigned byte (5121)** which is why there is macro code. I'm not sure if it's the best solution, but it's certainly readable. Look here for the glTF accessors spec, only float, unsigned byte and unsigned short is used:
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes

I'm glad I figured this out! :) It took me a while. I was able to test this with my own models and it looks perfect. Let me know if you want me to share them.


RichardTea:
Please don't use macros for this. ([C++ Core Guideline ES.31](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros2))

That macro is faking a templated function, so bets to use a templated free function.

glTFv2 can specify that an accessor must 'normalise', so this is in fact general functionality required for full glTFv2 compliance.
Have you checked whether this normalisation is currently supported in other glTF accessor contexts?

makitsune:
Sorry about the macro. I rewrote it as a template function.

kimkulling (approved):

kimkulling:

PR #3597, opened 2021-01-20
Increase float and double string export precision

### LINKED ISSUE #3573: Save/load roundtrip of single-precision mesh to ASCII STL does not preserve point positions
After an upgrade from 4.1.0 to 5.0.1, our tests noticed that a save/load roundtrip to ASCII-based STL no longer preserves point positions. After some investigation I ran into [this commit](https://github.com/assimp/assimp/commit/bf252c4452b76f9318b56d8c91f515bc13412a9c), reducing the text precision of floats to 8 digits (from a hardcoded 16). Based on the [docs for](https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10#:~:text=numeric_limits%3A%3Amax_digits10&text=The%20value%20of%20std%3A%3A,for%20all%20floating%2Dpoint%20types.) and values of `std::numeric_limits<>::max_digits10` I believe this should be 9 instead (and 17 for double precision) to guarantee a round-trip. 
While this precision reduction triggered the bug, it does not seem to be the main problem. I'm seeing results from the internally used `fast_atof.h` that differ from the STL implementation (specifically, Visual C++ 2015). Running the following snippet:
```
      float parsed_fast_atof;
      std::string testval_str{ "1.3049999 " };
      auto p_next_char = testval_str.data();
      Assimp::fast_atoreal_move(p_next_char, parsed_fast_atof);
      std::cout.precision(10);
      std::cout << "Parsing \"" << testval_str
                << "\" as float using:\n"
                   "  stof: "
                << std::stof(testval_str)
                << "\n"
                   "  fast_atoreal_move: "
                << parsed_fast_atof << "\n";

```
The output I get is:
```
Parsing "1.3049999 " as float using:
  stof: 1.304999948
  fast_atoreal_move: 1.304999828
```
It seems `fast_atoreal_move` does not yield the expected result here. This issue might have been masked by excess precision (16 digits for float) in the past?

RichardTea:
Should be 9 and 17 respectively.

Fixes #3573

Note: Unable to test ASSIMP_DOUBLE_PRECISION due to issue 3577


kimkulling (approved):

kimkulling:
I will test the double precision stuff when I have fixed the double precision errors. And thanks a lot for your contribution!

PR #3598, opened 2021-01-21
GLTF2: Allow Export Node in TRS format

thomasbiang:
In current assimp, node transformation is hard coded to be 

- matrix - if no animation
- TRS - if there is animation

There is no way to customize for users to decide which format they want to use. Particularly when matrix format is exported, there is no way to apply animations afterward. 

This pr does not change the default behavior of assimp node export, but just add an option so that users can specify explicitly if they want the node transformation being represented using TRS.


kimkulling (approved):

kimkulling:

Issue #3606, opened 2021-01-23
Collada Loader does not assign material indices to submeshes as expected


contriteobserver:
**Describe the bug**
The `ColladaLoader` always assigns the mesh material index to it's submeshes.
However, a common use case is for each submesh to have it's own material.

**To Reproduce**
`bin/assimp info ~/assets/tetrahedron.dae` (asset listed below)
currently, the output lists only one material
```
$ bin/assimp info ~/assets/tetrahedron.dae
Launching asset import ...           OK
Validating postprocessing flags ...  OK
Importing file ...                   OK
   import took approx. 0.00447 seconds

Memory consumption: 5460 B
Nodes:              2
Maximum depth       2
Meshes:             1
Animations:         0
Textures (embed.):  0
Materials:          1
Cameras:            0
Lights:             0
Vertices:           12
Faces:              4
Bones:              0
Animation Channels: 0
Primitive Types:    triangles
Average faces/mesh  4
Average verts/mesh  12
Minimum point      (-1.000000 -1.000000 -1.000000)
Maximum point      (1.000000 1.000000 1.000000)
Center point       (0.000000 0.000000 0.000000)

Meshes:  (name) [vertices / bones / faces | primitive_types]
    0 (Tetrahedron-mesh.Tetrahedron-mesh.Tetrahedron-mesh.Tetrahedron-mesh): [12 / 0 / 4 | triangle]

Named Materials:
    'Red'

Node hierarchy:
Scene
└╴Tetrahedron (mesh 0)
```

**Expected behavior**
`bin/assimp info ~/assets/tetrahedron.dae`
a listing of all four materials is expected
```
$ bin/assimp info ~/assets/tetrahedron.dae
Launching asset import ...           OK
Validating postprocessing flags ...  OK
Importing file ...                   OK 
   import took approx. 0.00149 seconds

Memory consumption: 10278 B
Nodes:              2
Maximum depth       2
Meshes:             4
Animations:         0
Textures (embed.):  0
Materials:          4
Cameras:            0
Lights:             0
Vertices:           12
Faces:              4
Bones:              0
Animation Channels: 0
Primitive Types:    triangles
Average faces/mesh  1
Average verts/mesh  3
Minimum point      (-1.000000 -1.000000 -1.000000)
Maximum point      (1.000000 1.000000 1.000000)
Center point       (0.000000 0.000000 0.000000)

Meshes:  (name) [vertices / bones / faces | primitive_types]
[...]

PR #3607, opened 2021-01-23
ColladaLoader now assigns individual material indices to submeshes as needed

contriteobserver:
Optional assignment of individual material indices to submeshes, addresses #3606


contriteobserver:
Please clue me in as to why the windows builds failed, I cannot seem to reproduce it.

kimkulling:

PR #3614, opened 2021-01-26
Implement import of Draco-encoded glTFv2 models

### LINKED ISSUE #2195: GLTF KHR_draco_mesh_compression
Adding support for the GLTF Extension, [KHR_draco_mesh_compression](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_draco_mesh_compression), I think will be a great addition to the GLTF exporter/importer. I mentioned this on twitter to @kimkulling, but I'm unaware of any current or pending work done by other contributors. I did see this extension mentioned in passing on some older threads that first added GLTF support to Assimp.

My company in particular would have interest in this, because our current workflow with GLTF is using it as a format that we push to a web server to store content, that can be fetched by clients to add to their project on-demand.

I might try and contribute this myself, since AssimpNet 4.1.0 has been released and I need something else to do :), but coming from .NET, working with cmake projects (other than building them) is a little intimidating. If not, at least we should have it on the radar as I'm sure other people would want this extension to be supported?

RichardTea:
Fixes #2195

Notes:
Draco support is enabled by default.

Adds a dependency on the Google draco library as a DLL/dylib.
Draco version 1.4.1 has been added to contrib

Note: Hunter almost certainly does _not_ work.
I'll need someone who understands Hunter to properly add the library for Hunter


RichardTea:
Looks like my cmake changes are definitely incorrect on Linux, macOS and Hunter.

I don't have access to a Linux environment at the moment, so would appreciate it if someone take a look at that platform.

contriteobserver:
Currently we require only CMake v3.0 on the rest of the project, Hunter requires only 3.2, if Draco is going to require CMake v3.12, we need to fail gracefully on older versions of CMake.

RichardTea:
@contriteobserver Thanks, I have added that check

I'm still having trouble with getting the draco library to link properly on macOS and Linux.
I do have access to macOS locally so will be able to work on that later today, but I don't have a Linux environment

rbsheth:
I can help add Hunter support, but we need to upgrade Draco to v1.4.1 (currently v1.3.4 I think).

RichardTea:
@rbsheth The changes to the github action appear to be calling out a wrong version.
I don't have that set up on my repo so couldn't check before merging.

Can you have a look?

rbsheth:
> @rbsheth The changes to the github action appear to be calling out a wrong version.
> I don't have that set up on my repo so couldn't check before merging.
> 
> Can you have a look?

Sorry, looks like you fixed it though? Thanks!

kimkulling (approved):

kimkulling:
Merged, thank you so much for this great contribution.

PR #3636, opened 2021-02-02
fix issue: 3482: invalid gltf2 properties

thomasbiang:
This pr fixes the reported issue 3482: gltf2 invalid properties are exported. 
Particularly it fixes: the unnecessary "jointName" under node, the unsupported "skeletons" under node and the unsupported "bindShapeMatrix" under skin.

export of "jointName" is removed, as node has "name" property already, jointName is just redundant.
export of "skeletons" and "bindShapeMatrix" is gated by GLTF2_CUSTOMIZE_PROPERTY. According to Gltf2 spec. these two properties are not valid, thus not exported by default. If users want to have the customized information, they can turn GLTF2_CUSTOMIZE_PROPERTY flag on.


code/AssetLib/glTF2/glTF2AssetWriter.inl
```diff
@@ -597,14 +596,13 @@ namespace glTF2 {
...
         }
 
-        AddRefsVector(obj, "skeletons", n.skeletons, w.mAl);
-
         if (n.skin) {
             obj.AddMember("skin", n.skin->index, w.mAl);
         }
-
-        if (!n.jointName.empty()) {
-          obj.AddMember("jointName", n.jointName, w.mAl);
```
thomasbiang, line 607: Redundant. Node has "name" already.

kimkulling:

PR #3653, opened 2021-02-15
Rust bindings

jkvargas:
Hi, I have created rust bindings to your library.
If you think that is ok, can I include it on your README?
Thanks.


kimkulling:
Great, thanks a lot for the conribution.

PR #3660, opened 2021-02-18
CMake fix for Android build with enabled JNI io system

ademets:
Cmake fails to finish configuration of build, when configuring project with ANDROID_JNIIOSYSTEM=ON. Used cmake version 3.19.5 to test with.
`CMake Error: install(EXPORT "assimpTargets" ...) includes target "assimp" which requires target "android_jniiosystem" that is not in any export set.`

Solution: Add android_jniiosystem to export set, so cmake file in /code folder can find and link with android_jniiosystem.


kimkulling (approved):

kimkulling:

PR #3667, opened 2021-02-23
Export Min/Max for Sparse Accessor

thomasbiang:
Currently Min/Max is only exported for regular accessor, but not Sparse ones. It causes alerts during gltf2 parsing.
This pr adds the Min/Max for Sparse Accessor as well.


kimkulling (approved):

kimkulling:

PR #3670, opened 2021-02-24
Check that normal count and tangent count match vertex count.

ms-maxvollmer:
This change adds checks to ensure that the normal count and tangent count match the vertex count. If they don't match we ignore the data and emit a warning. This prevents problems further down the line, as all code assumes these arrays are the same size.


kimkulling (approved):

kimkulling:

Issue #3690, opened 2021-03-06
thread-local storage is not supported for the current target


matthewswallace:
**the bug**

- On macOS I've built assimp from cmake --build
- cd into port/iOS/
- ./build.sh

I get the following error.

```
[ 92%] Building CXX object code/CMakeFiles/assimp.dir/AssetLib/Assjson/json_exporter.cpp.o

In file included from ~/assimp/code/Pbrt/PbrtExporter.cpp:87:
~/assimp/code/Pbrt/stb_image.h:931:1: error: thread-local storage is not supported for the
      current target
STBI_THREAD_LOCAL
^
~/assimp/code/Pbrt/stb_image.h:592:39: note: expanded from macro 'STBI_THREAD_LOCAL'
      #define STBI_THREAD_LOCAL       thread_local
                                      ^
~/assimp/code/Pbrt/stb_image.h:1070:8: error: thread-local storage is not supported for the
      current target
static STBI_THREAD_LOCAL int stbi__vertically_flip_on_load_local, stbi__vertically_flip_on_load_set;
       ^
~/assimp/code/Pbrt/stb_image.h:592:39: note: expanded from macro 'STBI_THREAD_LOCAL'
      #define STBI_THREAD_LOCAL       thread_local
                                      ^
~/assimp/code/Pbrt/stb_image.h:1070:8: error: thread-local storage is not supported for the
      current target
~/assimp/code/Pbrt/stb_image.h:592:39: note: expanded from macro 'STBI_THREAD_LOCAL'
      #define STBI_THREAD_LOCAL       thread_local
[ 92%] Building CXX object code/CMakeFiles/assimp.dir/AssetLib/Assjson/mesh_splitter.cpp.o
```

**Desktop **
 - OS: macOS 11.2.1
 - M1 mac mini 2021 computer
 - trying to compile for iOS
 
![Screen Shot 2021-03-06 at 3 24 23 PM](https://user-images.githubusercontent.com/123816/110221140-135a2f80-7e90-11eb-811f-1937a9ee4f10.png)

RichardTea:
I'm of the opinion that the STB image loaders shouldn't be in the assimp library at all.

Formats with embedded PNG images should pass the raw image file to the application. If nothing else, PNG contains a lot of metadata that stb_image doesn't support and applications may require.

matthewswallace:
I was able to solve the issue. The  `ports/iOS/build.sh` needed to be updated. 

[...]

PR #3692, opened 2021-03-08
updated MIN_IOS_VERSION to 10.0

matthewswallace:
I took some time to try and get assimp iOS port working the last couple of days I found that 1 variable in the ./build.sh file was the issue. 

After finding the issue and updating MIN_IOS_VERSION to 10.0 I was able to build the iOS binaries and I was able to do it on a new M1 Mac Mini. 

There were also some permission issue and some compile issues with compiling the example files however I did not update anything to fix the issue I just committed out that section in the main CMakeList.text so that I could run cmake. We should solve this issue for Mac users and I'm happy to continue to contribute if that would be helpful.


kimkulling:
You need to merge the latest master in.

RichardTea:
closes #3690 

@matthewswallace Can you update your branch to latest?

kimkulling:
Will be merged via https://github.com/assimp/assimp/pull/3737

PR #3695, opened 2021-03-10
[gltf2-exporter] Adding FB_ngon_encoding support

clems71:
Initial implementation of https://github.com/KhronosGroup/glTF/pull/1620 glTF2 extension proposal for handling ngons in a non destructive way.


devernay:
This implements only the conversion *to* glTF to preserve ngons, but not the conversion *from* glTF (with the extension) to a format that supports ngons.

One issue I see is that nothing in this implementation prevents two consecutive quads from sharing the same first vertex (which would cause them to be fused in a single ngon when converting back to an ngon-supporting format).
So you would have to implement the same `prev_first_indice` trick for quads and ngons.

Then an issue may appear if there are two consecutive ngons for which only one vertex that can be used for fanning, and it's the same vertex for both. What should be done in this case? Reorder the ngons?

clems71:
Indeed, you're right about the `prev_first_indice` trick, definitely something I overlooked!
Regarding the import process, this is not done yet. Will add on this PR asap.

kimkulling (approved):

kimkulling:
Do you want to get it merged. Or do you want to make the ngon conversion more stable first?

arodic:
> One issue I see is that nothing in this implementation prevents two consecutive quads from sharing the same first vertex

Good catch @devernay. I was able to confirm the asset you shared with me @clems71 has a bunch of quads fused from triangles that share the first vertex across quad boundaries.

kimkulling:
Could you please update the pull-request? I cannot merge it without the latest master-update,

devernay:
The problem can be solved for convex quads (aka arrowheads), because there are two choices for the fanning vertex (bad figure below), so even if they share the same arrow head. Only non-convex (even star-shaped) ngons with n>=5 pose an issue

image

clems71:
Hey all, sorry wasn't available to work on assimp the past week. I made a new commit to implement the right behavior for quads as discussed with @devernay . Another commit will come with the generic polygon case. It will likely be a best-effort solution for polygons though.

clems71:
@kimkulling Will do the rebase when the PR is complete (= generic polygons are properly handled)

clems71:
Just rebased on latest master branch. Ready to be reviewed now.

kimkulling (approved):

kimkulling:
Merged, thanks a lot for your conribution.

PR #3696, opened 2021-03-10
export with rotation

jaefunk:


tools/assimp_cmd/Main.cpp
```diff
@@ -325,6 +325,14 @@ bool ExportModel(const aiScene* pOut,
 		PrintHorBar();
 	}
 
+	aiMatrix4x4 rx, ry, rz;
+    aiMatrix4x4::RotationX(imp.rot.x, rx);
```
kimkulling, line 329: Is this correct? Y have always used the x rotation.

kimkulling (changes requested):

tools/assimp_cmd/Main.cpp
jaefunk: you right. my fault...
i fixed it

kimkulling (approved):

kimkulling:

PR #3707, opened 2021-03-15
Fix a set of glTF2 crashes on bad input

MalcolmTyrrell:
We hit null pointer dereferences in the gltf2Importer.cpp. The origin of the problem was bad types in the json, which the parsing code would simply ignore. 

This change makes glTF2 parsing more strict, and attempts to give a somewhat useful message.

Two other small changes were made.
1.  The "version" field is strictly a string, so I removed the option to parse it as a number. See here:
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/asset.schema.json
2. When a buffer is described as larger than the data it references, this is now a DeadlyImportError rather than an unhandled memory error.


kimkulling (approved):

kimkulling:

PR #3712, opened 2021-03-21
Fix for Xcode build.

lvalencia:
When you generate a static library framework build and you include it in a project 

`texture.h` and `material.h` result in the following errors

```
'./Compiler/pushpack1.h' file not found
'./Compiler/poppack1.h' file not found
```

Looks like it's the same issue as #2219, but for the two aforementioned files.

### DIFF
--- a/include/assimp/material.h
+++ b/include/assimp/material.h
@@ -457,8 +457,6 @@ enum aiBlendMode {
 #endif
 };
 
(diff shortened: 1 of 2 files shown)


lvalencia:
Oh I guess it's needed for the build, but not after? Okay, will cancel, don't worry about it.

PR #3718, opened 2021-03-24
Malcolm tyrrell/tangent check

MalcolmTyrrell:
If the target counts are inconsistent with the vertex count, then the importer can overrun buffers. 

Another closely related change ensures that the target data is skipped if the normal or tangent data was already excluded because of the similar checks that already exist.

I'm not including test data with this, as I don't have a small file which crashes. Adjust the gltf test file AnimatedMorphSphere so it's fifth accessor has size 16, it should crash without this change. The same change with the much smaller AnimatedMorphCube does not lead to a crash, but you can observe the buffer being read beyond its allocation in a debugger.


kimkulling (approved):

kimkulling:
Good catch.

kimkulling:

PR #3721, opened 2021-03-25
Added Blendshape Support to FBX Export

vfxgordon:
Kitbashed blendshape exports into the exporter, works well on the example glbs I have here (multi material, 30-40 blendshapes per mesh). Not 100% sure on the normals, however, but FBX normals are a bit iffy in general, anyway.


kimkulling (approved):

kimkulling:

Issue #3450, opened 2020-10-12
3ds export bug.


copycd:
**Describe the bug**

I found export bug in 3ds.
vertex is not transferred.
because, matrix exists.

That is 3DSExporter.cpp at void Discreet3DSExporter::WriteMeshes();

```
        // Vertices in world space
        {
            ChunkWriter curChunk(writer, Discreet3DS::CHUNK_VERTLIST);

            const uint16_t count = static_cast<uint16_t>(mesh.mNumVertices);
            writer.PutU2(count);
            for (unsigned int i = 0; i < mesh.mNumVertices; ++i) {
                //const aiVector3D &v = trafo * mesh.mVertices[i];
                => I fixed code.
                const aiVector3D &v = mesh.mVertices[i];
                writer.PutF4((float)v.x);
                writer.PutF4((float)v.y);
                writer.PutF4((float)v.z);
            }
        }
```

```
        // Transformation matrix by which the mesh vertices have been pre-transformed with.
        {
            ChunkWriter curChunk(writer, Discreet3DS::CHUNK_TRMATRIX);
            for (unsigned int r = 0; r < 4; ++r) {
                for (unsigned int c = 0; c < 3; ++c) {
                    writer.PutF4((float)trafo[r][c]);
                }
            }
        }
```

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3724

PR #3722, opened 2021-03-26
Fixing CHUNK_TRMATRIX translation sub chunk

bekraft:
and therefore omitting the transform aggregation in the CHUNK_VERTLIST. A similar bug has been posted already (see #3450) but unsure if the @copycd meant the same issue.

Found while comparing the 3DS exporter implementations of Blender see [export_3ds.py](https://github.com/blender/blender-addons-contrib/blob/f948f658ba33eb670a65e0bba058d43138abea7e/io_scene_3ds/export_3ds.py#L1034) to Assimp.


kimkulling:
Great catch, thanks a lot for your investigations. I really appreciate your efforts! Happy to see the progress.

Merged!

bekraft:
Hi Kim, to late for the PR: I saw that the import has the same issue. So I'll open a new PR.

PR #3723, opened 2021-03-28
Add zlibstatic to the list of exported targets

eliasdaler:
Hello.
This patch fixes zlibstatic being missing from export set when zlib is used as a static lib with assimp. Otherwise I get the error like this:
```
CMake Error: install(EXPORT "assimpTargets" ...) includes target "assimp" which requires target "zlibstatic" that is not in the export set.
```


kimkulling (approved):

kimkulling:
Merged, thanks for that.

PR #3743, opened 2021-04-05
Fix Q1 MDL group frame loading, e.g. Q1 progs/flame2.mdl

Garux:


Garux:
Idk what would be elegant solution for compiler warning here.

Garux:
So warning is fixed now, and code still produces good output.

kimkulling:

PR #3746, opened 2021-04-05
Renaming PI to aiPi.

### LINKED ISSUE #3310: header template error at line 100 pi error constant in mathfunctions.h
I'm using visualstudio 2019 16.3.3 maybe thats my problem, just kidding lol. I'm getting an error for line 100 in mathfunctions.h saying that the template statement is not correct for a constant, i'm not an expert with templates or c++ at the matter but i've read that static_cast is not a value to use in templates or headers and try integral_cast but same error. I know sometimes the compiler throws errors at line 100 when the problem is at line 95 or 99 as an example, but when i comment the template out it compiles. Sometimes i think c++ compatibility as c++11 or c++99 or something, the newer c++ standard is stricter in its checks maybe thats the build fails. I notice 5.0.0 doesn't have PI where 5.0.1 includes this line.
I'm using assimp in this modified project to load meshes into the 3d engine, not sure if that matters or what pi is used for.

https://github.com/GameInstitute/RuntimeMeshLoader

error from assimp compile
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2988	unrecognizable template declaration/definition	RUMTIME_MESH_LOADER	E:\Unreal Projects\RUMTIME_MESH_LOADER\Plugins\RuntimeMeshLoader\ThirdParty\assimp\include\assimp\MathFunctions.h	101	

Severity	Code	Description	Project	File	Line	Suppression State
Error	C2059	syntax error: 'constant'	RUMTIME_MESH_LOADER	E:\Unreal Projects\RUMTIME_MESH_LOADER\Plugins\RuntimeMeshLoader\ThirdParty\assimp\include\assimp\MathFunctions.h	101	

line causing issues
template<class T>
inline T PI() {
	
	return M_PI;
///    return static_cast<T>(3.14159265358979323846);

}

BaldricS:
Fixes #3310

This change renames the function PI to aiPi to avoid having it clash with the common macro convention of ALL_UPPERCASE_LETTERS. Because the function PI was included in a public header (that then got included everywhere), it would occasionally get stomped on if included in a translation unit that _also_ defined the macro PI (a fairly common definition I'd wager).

I chose `aiPi` over `ai_pi` as it seemed to better match some existing types I saw? If we want it to be `ai_pi` just let me know. :)

### DIFF
--- a/include/assimp/MathFunctions.h
+++ b/include/assimp/MathFunctions.h
@@ -87,7 +87,7 @@ inline IntegerType lcm( IntegerType a, IntegerType b ) {
     }
 	return a / t * b;
 }
-/// @brief  Will return the smallest epsilon-value for the requested type. 
+/// @brief  Will return the smallest epsilon-value for the requested type.
 /// @return The numercical limit epsilon depending on its type.
 template<class T>
 inline T getEpsilon() {
@@ -97,7 +97,7 @@ inline T getEpsilon() {
 /// @brief  Will return the constant PI for the requested type.
 /// @return Pi
 template<class T>
-inline T PI() {
+inline T aiPi() {
     return static_cast<T>(3.14159265358979323846);
 }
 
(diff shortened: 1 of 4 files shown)


kimkulling (approved):

kimkulling:
Thanks a lot for your help. Merged!

PR #3749, opened 2021-04-06
Fix incorrect indices in the MilkShape 3D loader

pcwalton:
These indices are quite bogus. I'm not sure how this was ever working before...


Garux:
Good catch here, no wonder that test models were crashing it consistently.
Minus one format from 'constant crash' list 📃

kimkulling (approved):

kimkulling:

PR #3757, opened 2021-04-13
Flip the check on _MSC_VER for using TR1 containers.

mahiuchun:
Non-TR1 version would be the default now, across all compilers. For MSVC the behavior stays the same.

Note that due to the typo `__GNUC___` above, GCC/Clang is not using unordered containers as of now.


kimkulling (approved):

kimkulling:

PR #3763, opened 2021-04-16
fixed glTF export stuff being pulled into the EXE even if building wi…

krishty:
…th ASSIMP_BUILD_NO_EXPORT

“LazyDictBase::WriteObjects()” in the two glTF implementations is only used for export. Since it’s a virtual method, and many compilers have trouble removing unreferenced virtual methods, glTF export stuff is pulled into the binary even if compiling without exports.

This commit removes said virtual function if only compiling for import.

This removes 75 KiB of useless code when compiled with Visual Studio for x64.


krishty:
Of course the Real Solution™ would be a logical separation of import and export functionality in the code design, but that’s way over my head.

kimkulling (approved):

kimkulling:

PR #3764, opened 2021-04-16
style fix - initializing and assigning empty std::string properly

krishty:
`std::string s(""); s = "";` calls the copy constructor, which in turn calls `strlen()`, … assigning a default-constructed string generates fewer instructions and is therefore preferred.

With C++11 uniform initialization, you’d simply write `s = { }` instead.


kimkulling (approved):

kimkulling:
Looks good, I will merge it.

kimkulling:

PR #3773, opened 2021-04-19
updated version string (if anyone cares)

krishty:
Also, shouldn’t it be “License*_d_* under …”


krishty:
Forgot to adjust the unit test; closing. Will re-open with a fix.

PR #3776, opened 2021-04-19
fixed export exceptions on import

krishty:
Ogre and 3MF imports threw `DeadlyExportError`s under some circumstances. Bad for people who assumed that they only needed to catch `DeadlyImportError`s. Changed them to `DeadlyImportError`s.


kimkulling:
Good catch, CI is on its way!

kimkulling:

PR #3787, opened 2021-04-21
Importer improvements

ms-maxvollmer:
These address a few issues where Assimp asserts or crashes when importing invalid files:
* Throw instead of assert on invalid file input
* Check JSON object type before accessing members
* Ensure samplers input and output references are set before accessing them


kimkulling (approved):

kimkulling:

PR #3788, opened 2021-04-22
Fix: Removing double delete of texture items after MergeScene.

BaldricS:
Textures were being double deleted after a merge scene because the
texture array wasn't being properly deleted at the end of merging.
Furthermore, the texture array was being sized to the number of
materials instead of the number of textures.


kimkulling (approved):

kimkulling:
Looks good.

kimkulling:
Merged, thanks a lot for your contribution,

Issue #3797, opened 2021-04-23
Blender importer has a newline in its name


JC3:
Library version: master @ cd42b99 (also present in 5.0.1 Bugfix release)
Platform: Windows 10 64-bit
Compiler: MSVC 2019 64-bit

The name reported for the Blender importer has a newline in it. It is the only importer whose name contains a newline. The newline should be removed as it causes problems in certain [G]UIs.

Example:
```
cout << '"' << aiGetImporterDesc("blend")->mName << '"' << endl;
```

Outputs
```
"Blender 3D Importer
http://www.blender3d.org"
```

krishty:
This is also annoying when you just want to log an exporter’s name and suddenly have a two-line log message.

JC3:
I've submitted a PR for this in #3822, as an immediate solution while I consider #3799.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3822

PR #3798, opened 2021-04-23
Remove newline from name of Blender importer.

JC3:
Addresses #3797.

Website should probably also be moved to the comments to be consistent with the rest of the importers, although I did not do it as the comment field is already occupied and I don't have enough coffee in me yet to make formatting decisions.

Alternate change could be to add a dedicated website field to importer descriptions.


JC3:
Regarding addition of a website field, see #3799 for discussion.

JC3:
Closed; see #3799 for explanation.

Issue #3791, opened 2021-04-23
STEP files cannot be imported - assimp seems to use randomly incorrect (?) importers


JC3:
Library version: master @ cd42b99
Platform: Windows 10 64-bit
Compiler: MSVC 2019 64-bit

STEP files cannot be imported, at least none that I've tried, because the STEP importer (which seems to be the IFC importer in this version) is not used. Instead, other importers are being used, and failing (so far I've observed the X3D, DXF and STL importers). I cannot identify the criteria that determines which importer is used.

A test program is [here](https://github.com/JC3/ModelSplit/tree/assimp-master-cd42b99/specific_tests/x3d_importer_fallback) (does not actually depend on Qt). This program just prints some info about the extension, and also tries to import a file and prints some info about the importer that was used. 

The output (slightly edited for brevity), when run against the model [here](https://github.com/JC3/ModelSplit/tree/assimp-master-cd42b99/specific_tests/x3d_importer_fallback/models) as well as a dummy zero-length file with a .stp extension, is:

```
assimp 5.0.1 (master @ cd42b995)
Extension: "stp"
  IsExtensionSupported:       true
  GetImporterIndex:           39
  GetImporter:                plugin=0x0000015974491640
  ->GetExtensionList:         ifc ifczip stp
  ->GetInfo->mName:           Industry Foundation Classes (IFC) Importer
  ->GetInfo->mFileExtensions: ifc ifczip stp

[created zero-length dummy file: dummy.stp]
  ReadFile (pFlags=0):
    ReadFile:                 success
    importerIndex:            46
    mName:                    Extensible 3D(X3D) Importer
    scene->mFlags:            0x00000000
[deleting dummy file]
---------------------------------------------------------------
[using specified file: models/scene.stp]
  ReadFile (pFlags=0):
    ReadFile:                 error: DXF: no data blocks loaded
    importerIndex:            16
    mName:                    Drawing Interchange Format (DXF) Importer
---------------------------------------------------------------
[using specified file: models/io1-cm-214.stp]
  ReadFile (pFlags=0):
    ReadFile:                 error: Failed to determine STL storage representation for models/io1-cm-214.stp.
    importerIndex:            14
    mName:                    Stereolithography (STL) Importer
---------------------------------------------------------------
[using specified file: models/io1-tu-203.stp]
  ReadFile (pFlags=0):
    ReadFile:                 error: DXF: no data blocks loaded
    importerIndex:            16
[...]

PR #3801, opened 2021-04-23
Simplified file detection and fixed a few bugs

krishty:
The search for a matching importer had a few issues, see #3791. There were two different mechanisms to determine whether an importer accepts a specific file extension:
1. `aiImporterDesc::mFileExtensions`, which was forwarded to the UI via `BaseImporter::GetExtensionList()`.
2. `BaseImporter::CanRead()` when called with `checkSig == false`, which determines whether to actually use that importer.

They were redundant and got out of sync repeatedly. I removed 2. completely and replaced it with 1., thereby syncing UI/import and shortening all `BaseImporter::CanRead()` implementations.

Further bugfixes:
- fixed glTF2 importer throwing exceptions when checking whether it can load a file
- removed `BaseImporter::SimpleExtensionCheck()` because it is no longer used and had a bug with case sensitivity

Since the `checkSig` parameter in `BaseImporter::CanRead()` is now useless, it can be removed completely. I’m not sure if this would break ABI compatiblity, so I’ll submit it with a later pull request.


kimkulling (approved):

kimkulling:
Cool approach, thank you very much for the great contribution. CI is on its way.

krishty:
Wow, sorry for the trouble!

Your first correction is right: Indeed, Microsoft’s STL seems to define `std::size()` in _any_ STL header, while libstd & Co. are more strict and only define it in a few selected headers. The additional `#include <iterator>` in `BaseImporter.h` is a good idea.

I don’t understand why it’s still failing. Did I rely on the wrong C++ standard? If this all falls apart, I’ll revert to hard-coded array sizes.

krishty:
Plan B:
`BaseImporter.h`
`+ #define AI_COUNT_OF(X) (sizeof(X) / sizeof((X)[0]))`
… replace all `std::size(` with `AI_COUNT_OF(`.

krishty:
@kimkulling here’s the fix I suggested above.

Unfortunately, two unit tests break and I’m sure it’s the test’s fault. Here’s what happens:
1. `glTF2Importer::CanRead()` did not catch any exceptions. This was an error for obvious reasons.
2. I fixed that; it now returns `false` a.k.a. “cannot read file” if an exception is thrown.
3. The test `utglTF2ImportExport.error_string_preserved` provides a file with a intentionally missing external reference. It is expected that the GLTF2 importer tries to load the file, notices the missing reference, and returns a valid error description (this is a good test).
4. When an importer for the test file is searched, `glTF2Importer::CanRead()` is called.
5. `CanRead()` loads the entire file. *I think this is trying too hard.*
6. While loading the entire file, it notices the missing reference and throws `GLTF: could not open referenced file "BoxTextured0.bin"`.
7. The exception is caught and translated to “GLTF2 importer cannot read this file”.
8. Since the GLTF2 importer cannot read this file, and no other importer can either, loading is aborted: No importer found.
9. The test fails.

Why did it work before? Since probing the file during `CanRead()` threw `GLTF: could not open referenced file "BoxTextured0.bin"` and `CanRead()` didn’t catch the exception, Assimp reported the correct error by accident, without even having determined the correct importer!

In order to pass the tests, it is required that GLTF2 only takes a quick look at the file during `CanRead()`, not loading it entirely and throwing detailed errors far in. I’m afraid I can’t provide that at this point because I’m not deep enough inside GLTF’s implementation.

So we must decide whether to
1. retire this pull request, or
2. disable the test;

… until `glTF2Importer::CanRead()` is less strict.

Garux:
> I don’t understand why it’s still failing. Did I rely on the wrong C++ standard? If this all falls apart, I’ll revert to hard-coded array sizes.

std::size() is C++17 https://en.cppreference.com/w/cpp/iterator/size

krishty:
@Garux Indeed, but that’s been worked around. The remaining blocking problem is the two broken unit tests.

Garux:
Mentioning this, as C++17 gonna be available soon, it seems.

krishty:
Very cool! Hit me up when it’s available and I’ll retrofit it by replacing `AI_COUNT_OF()` with `std::size`.

kimkulling:
Merge will be done here: https://github.com/assimp/assimp/pull/4342

PR #3806, opened 2021-04-24
removed dead code from 0d29203e24a8bc2c75278931a6bd25b2ae5848de

krishty:
Commit 0d29203e24a8bc2c75278931a6bd25b2ae5848de introduced a new scaling system into `BaseImporter`. But apart from one variable and one function, this scaling system was never used and has been dead for the past two years. The dead code pulled `std::map` and two virtual functions into every importer, wasting 4 KiB in the x64 binary. Apart from that, the implementation was very questionable in the first place.

Removing it may break ABI compatibility if `BaseImporter` is used across library bounds, hence the change should be used with caution.


kimkulling (approved):

kimkulling:

RichardTea:
Well, this commit broke my application. BaseImporter::GetFileScale() const has been removed. This breaks source and binary compatibility.

The FBX importer calls SetFileScale() to expose the internal scaling of the FBX file.
Removing GetFileScale() means I now have to find some other way of getting the FBX file scale.

krishty:
Yeah, I warned about ABI compatibility. Maybe I should have stated it like “This may break ABI compatibility and I have no idea what Assimp’s ABI policy is, someone please decide which release this commit should go into”.

Regarding FBX: I think the scale process is the only client of that information and you should use it instead.

I’d be happy to see a unified approach to this problem. Importers use at least four different methods of handling scale:
1. said `SetFileScale()` and `GetFileScale()`
2. a property value forwarded to the scale process (FBX and one other importer)
3. overwriting the root matrix of the imported scene (3DS, IIRC)
4. ignoring scale altogether (C4D)

Maybe `GetFileScale()` and `SetFileScale()` were meant to be the one thing to rule them all, but if so, it was implemented poorly and only realized in a very small number of importers, effectively leading to the [xkcd standards](https://xkcd.com/927/) situation.

kimkulling:
Hi,

at first the broken ABI-compatibility is my fault. I haven't checked the compatibility in a proper way that no other loader was using this code. I am sorry for that!

The policy in the past to deal with ABI was:
- Bugfix-Versions (5.01 to 5.0.2): Only fixes, no new features
- MInor versions ( 5.1 to 5.2 ): It is allowed to add new API-calls, removing them is forbitten because of broken ABI-compatibilities. We want to avoid that
- Major versions: It is allowed to break ABI-compatibility

My proposal to fix this would be to revert the broken ABI-compatibility.

What do you think?

krishty:
Reverting is absolutely fair.

How should I contribute such ABI-breaking commits? I have another one, but I can’t find a 6.0 branch. Should I just post pull requests anyway, tag them `6.0` so you can accept them just before the next major release?

PR #3808, opened 2021-04-24
SimpleTexturedDirectx11 sample: support embedded uncompressed textures

ericwa:
I wanted to try viewing a Half-Life .mdl in SimpleTexturedDirectx11 as an experiment (`bullsquid.mdl`).

This contains some fixes necessary to view the model:

- remove assumption that embedded texture names start with "*0", etc. The `bullsquid.mdl` has an embedded texture named "bottommap.bmp". Instead, just try to load the texture name with `aiScene::GetEmbeddedTexture()`, and if that fails, try to load it from disk.
- rename `ModelLoader::getTextureFromModel()` to `loadEmbeddedTexture()`
- support loading uncompressed embedded textures

Quake .mdl's now load with these changes, as well.

Note, although I can see the `bullsquid.mdl` now, triangle culling is inside-out. It looks correct on the Quake .mdl's and Quake 3 .md3's I tried though.


ericwa:
> Note, although I can see the bullsquid.mdl now, triangle culling is inside-out. 

Ah, sounds like this is addressed by https://github.com/assimp/assimp/pull/3842

kimkulling (approved):

kimkulling:

Issue #3802, opened 2021-04-24
3DS Regression: Transformation is broken


krishty:
https://github.com/assimp/assimp/pull/3724 changed the relative transformation during 3DS import. For me, this broke several files. I can only fix them by reverting the change.

Sample file:
[19127_Robot_Parallel_Manipulatar.zip](https://github.com/assimp/assimp/files/6368651/19127_Robot_Parallel_Manipulatar.zip)

Screenshot before:
![before](https://user-images.githubusercontent.com/51880096/115940375-a3d0eb80-a4a1-11eb-9d7c-15f1e846b174.png)
after:
![after](https://user-images.githubusercontent.com/51880096/115940379-a7647280-a4a1-11eb-9e36-3cf7522b651a.png)

I have more files that show this regression, though some are under NDA. No file improved, I only see regressions.

The changelog says this change happened “in alignment with 3DS exporter”, which I presume means [this change](https://github.com/assimp/assimp/pull/3722/).

Since this aligned import and export, I fear that both are wrong now.

Garux:
That change fixes the case i'v got
![unknown (11)](https://user-images.githubusercontent.com/1734117/116275097-f48b6180-a78b-11eb-863a-5fe192dca78c.png)
(smaller view is after the change)
But i haven't look what's actually going there, possibly wrong looking one is correct 🥨

Garux:
Ok, if i add aiProcess_PreTransformVertices:
![image](https://user-images.githubusercontent.com/1734117/116276856-9790ab00-a78d-11eb-9e49-78a2dedb73c6.png)
(and it's also laying on the side...)
And with the change undone:
![image](https://user-images.githubusercontent.com/1734117/116277891-98760c80-a78e-11eb-845f-fa33b81a6553.png)
(also on the side, but this seems to be *alright*)

krishty:
Thanks a lot for the testing! 👍 

Do I understand correctly:

1. change on, PreTransformVertices on: **wrong** (your 2nd pic)
2. change on, PreTransformVertices off: **OK** (1st pic, small)
3. change undone, PreTransformVertices on: **OK** (your 3rd pic)
4. change undone, PreTransformVertices off: **wrong** (1st pic, large)

Did you export the 3ds from Assimp to Blender? In this case keep in mind that the exporter may be broken, because “fixing” it is what broke the import int he first place.

[...]

PR #3826, opened 2021-04-27
reverted regression in 3DS transformation (issue #3802)

krishty:
The regression was introduced to align 3DS export and import, but in fact it broke the transformation matrices on import. This commit reverts the relevant lines. Furthermore, matrix layout was cross-checked with two other 3DS importers. Export was not considered.


kimkulling (approved):

kimkulling:
Review is done, fix is on its way. For bugs like this, we need a stronger test suite.

kimkulling:

PR #3833, opened 2021-04-28
Fix crash when reading 0 bytes

kimkulling:
- This is a valid option so crash shall not happen


krishty (commented):
What is this `if` for in the first place? `fread()` is specified to return zero if `pCount` equals zero, without any error whatsoever. Just delete the `if` and enjoy the ride.

PR #3837, opened 2021-04-29
added .step extension to IFC loader

krishty:
The extension .step is at least as common as .stp, so both should be supported.

Mind that this PR is only sufficient after #3801 is accepted, else the `CanRead()` function must be extended as well.


kimkulling (approved):

kimkulling:
Looks good, I will merge it.

kimkulling:

PR #3838, opened 2021-04-30
consider aiProcess_FlipWindingOrder in aiProcess_GenNormals & aiProcess_GenSmoothNormals

Garux:
Corrects generated normals direction, while aiProcess_FlipWindingOrder is used.
Assuming that aiProcess_FlipWindingOrder is used for change winding order to respect user's renderer, but not fixing wrong order after some importer.
Also user doesn't know which normals were generated, thus can't fix them on his own.


kimkulling:
Looks good, I will merge it.

kimkulling (approved):

kimkulling:

PR #3839, opened 2021-04-30
consider pScene->mRootNode->mTransformation set by some importers while using AI_CONFIG_PP_PTV_ROOT_TRANSFORMATION

Garux:
Some importers set `pScene->mRootNode->mTransformation` to orient a model properly.
Old behavior is to override it by `AI_CONFIG_PP_PTV_ROOT_TRANSFORMATION`, if used, which results in inconsistent orientation of various formats.


kimkulling (approved):

kimkulling:

PR #3840, opened 2021-04-30
fix mdl3457 normals direction

Garux:
Assuming here that test models have correct normals direction.
Importer loads them flipped:
image

after:
image

### DIFF
--- a/code/AssetLib/MDL/MDLLoader.cpp
+++ b/code/AssetLib/MDL/MDLLoader.cpp
@@ -708,6 +708,7 @@ void MDLImporter::InternReadFile_3DGS_MDL345() {
 
                 // read the normal vector from the precalculated normal table
                 MD2::LookupNormalIndex(pcVertices[iIndex].normalIndex, pcMesh->mNormals[iCurrent]);
+                pcMesh->mNormals[iCurrent] *= -1;
                 // pcMesh->mNormals[iCurrent].y *= -1.0f;
 
                 // read texture coordinates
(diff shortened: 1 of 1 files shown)


code/AssetLib/MDL/MDLLoader.cpp
```diff
@@ -708,6 +708,7 @@ void MDLImporter::InternReadFile_3DGS_MDL345() {
 
                 // read the normal vector from the precalculated normal table
                 MD2::LookupNormalIndex(pcVertices[iIndex].normalIndex, pcMesh->mNormals[iCurrent]);
+                pcMesh->mNormals[iCurrent] *= -1;
                 // pcMesh->mNormals[iCurrent].y *= -1.0f;
```
krishty, line 712: Now that you proved this outcommented, old code is wrong, you can delete it. Otherwise your commit looks a bit like another experiment

code/AssetLib/MDL/MDLLoader.cpp
```diff
@@ -763,6 +764,7 @@ void MDLImporter::InternReadFile_3DGS_MDL345() {
...
```
krishty, line 768: same here

code/AssetLib/MDL/MDLLoader.cpp
Garux: It was commented out here:
https://github.com/assimp/assimp/commit/c2d888154964d4720940a268e45d3dbb314cba5b#diff-2f88332cbd20db61ae0077d2a0e9238ed20e0fd0496692b7dbf03ce88946a8ebR733
image
Possibly whole model orientation is wrong, my commit just fixes normals in the current way, so it's really an experiment.

kimkulling:
OK, so let's close this one.

PR #3842, opened 2021-04-30
fix hl1 mdl orientation, tex coords, face windings order

Garux:
before:
image
after:
image
used half-life map editor for the reference:
image


kimkulling:

PR #3844, opened 2021-04-30
support missing closing brace in material list after Ascii Scene Exporter v2.51

Garux:
I'v got good amount of ase models, failing to load due to this exporter bug. Apparently some loaders do handle this.
Possibly more persistent way would be checking for global keywords at this point, if `MATERIAL_LIST` block may contain anything besides `MATERIAL_COUNT` & `MATERIAL` keywords.


kimkulling (approved):

Garux:
Example of such model; this is exporter of Blender afaik.
[tri_test_002_001.zip](https://github.com/assimp/assimp/files/6459208/tri_test_002_001.zip)

Issue #3777, opened 2021-04-19
m3da exporter should not be registered if M3D_ASCII not configured


JC3:
Library version: master @ cd42b995
Platform: Windows 10 64-bit
Compiler: MSVC 2019 64-bit

Currently, if M3D_ASCII support is not configured, the "m3da" exporter is still registered, but always fails with the error message "Assimp configured without M3D_ASCII support".

This is an issue because, since the exporter is registered, it will appear when enumerating supported export formats, and therefore will appear as an option in the UI of any client applications, but will always fail if chosen, possibly leading to users filing bug reports with the application developer, or just making the application look sloppy.

If the exporter cannot function without M3D_ASCII support configured, then unless M3D_ASCII support *is* configured, it should not be registered as an exporter, and should not appear in the list of exporters or be available to use.

Suggested fix: Do not register the "m3da" exporter if assimp was not compiled with the necessary configuration to support it.

kimkulling:
Fixed by #3848

PR #3848, opened 2021-05-01
build M3D ASCII support by default

contriteobserver:
seems the most straightforward way to address issue  #3777


RichardTea:
This format is disabled because it is unsafe. The reader suffers from a great many buffer overflows.
Sadly the original contributor of the format was not able to fix the issues, and so disabled it.

JC3:
> This format is disabled because it is unsafe. The reader suffers from a great many buffer overflows.
> Sadly the original contributor of the format was not able to fix the issues, and so disabled it.

Good to know. I'll include it my current round of bug fixes. I've got a solid test setup and the bug hunting has been very productive.

kimkulling (approved):

kimkulling:

bztsrc:
> This format is disabled because it is unsafe. The reader suffers from a great many buffer overflows. Sadly the original contributor of the format was not able to fix the issues, and so disabled it.

Liar. The ASCII support was disabled not because it's unsafe, but because YOU didn't liked setting LC_NUMERIC to C.

RichardTea:
I don't like code that invokes undefined behaviour.

Even if that wasn't explicitly specified as causing undefined behaviour in the C standard, libraries are guests in someone else's house. You do not change the carpets.

bztsrc:
> I don't like code that invokes undefined behaviour.

Which part involves allegedly UB? The ASCII importer is turned off, the other parts does not depend on printf nor on locales. I compile the code with -pedantic -Wall -Wextra flags, run through sanitizer and valgrind in many many scenarios. All the assimp's CI tests were green with all compilers on all platforms last time I commited, not even a single warning.

Cheers,
bzt

Cheers,
bzt

Issue #3849, opened 2021-05-01
M3D importer crash when file can't be opened


JC3:
The M3D loader crashes in `CanRead` when the file can't be opened (e.g. doesn't exist) and either has no extension or `checkSig` is set.

Issue is failure to check return value of `Open`, then subsequent access of null pointer:

https://github.com/assimp/assimp/blob/b516d940b67386e24f7a878450c16e2a7fda9cc5/code/AssetLib/M3D/M3DImporter.cpp#L138-L142

May be fixed in #3801, testing that is finally coming up on my todo list today.

JC3:
Addressed all in #3890:

![image](https://user-images.githubusercontent.com/5651184/117215210-e9f44c00-adcb-11eb-83b3-117f21bc42fe.png)

Error messages returned by importers are listed in [report.xml.gz](https://github.com/assimp/assimp/files/6430507/report.xml.gz). I'll leave figuring out the format of that as an exercise to the reader, lol.

JC3:
Oops wrong thread.

JC3:
Closed by #3850.

PR #3850, opened 2021-05-01
Fix crash in CanRead when file can not be opened.

JC3:
Addresses #3849


kimkulling (approved):

kimkulling:

Issue #3761, opened 2021-04-16
C4D: Supporting the latest SDK


krishty:
Assimp, if built accordingly, uses Maxon’s non-free Melange SDK to import C4D files.

This SDK has been renamed from _Melange_ to _Cineware_. Namespaces and types [have been renamed](https://developers.maxon.net/?p=3533) as well.

Supporting the latest SDK is pretty easy – just replace all occurences of _melange_ with _cineware_ in the source code, and _Melange_ with _Cineware_ in the build scripts. This may, however, break the build for people still using the older Melange SDK.

I think Assimp needs a decision on whether

1. not supporting new SDKs named Cineware,
2. supporting only new Cineware SDKs and dropping Melange SDK support,
3. supporting both.

kimkulling:
Right, for version 5.1 just supporting one C4D-SDK would be the easiest solution I guess. Does anyone depend on the Melange SDK?

krishty:
I’ll contribute the PR for the change to Cineware then. Should I use the master or is there a dedicated 5.1 branch?

kimkulling:
Master is the right one

kimkulling:
Fixed

PR #3851, opened 2021-05-01
updated C4D importer to use the Cineware SDK

krishty:
Maxon’s _Melange SDK_ has been renamed _Cineware SDK_ as of 21.004, and with it all namespaces and types (see #3761). This commit
- makes CMake use `contrib/Cineware` instead of `contrib/Melange`;
- renames Assimp’s `namespace melange` to `namespace cineware`;
- removes useless functions and formatter references from `class C4DImporter`;
- removes duplicate conversion of `cineware::String` to `aiString` in the importer;
- updates comments accordingly;
- updates copyright info.


kimkulling:

PR #3852, opened 2021-05-01
style fix: namespace instead of class with public static members

krishty:


kimkulling (approved):

kimkulling:

PR #3856, opened 2021-05-03
fixed bloat in SIB importer

krishty:
The SIB importer, upon needing an empty aiString, did not create a new one but rather copied a predefined global empty string.

Since aiStrings contain large buffers, Assimp copied 1028 B of zeros instead of setting five bytes (at least when compiled with Visual C++). Since aiString is a user-defined type without a constexpr constructor, Visual C++ had to generate a thread-safe run-time initializer as well.

Now it’s just two instructions.


kimkulling:

PR #3857, opened 2021-05-03
fix bug in vertex_color animation

spotaws:
First off, on behalf of the [Lumberyard](https://aws.amazon.com/lumberyard/) team at Amazon, thank you for assimp. It has been very helpful to us.

In our use of assimp, we have discovered a bug in vertex_color_animation, which we have fixed by skipping TransformComp_Rotation (since it is applied elsewhere in the code). This pull request is how we have fixed it, please consider it for merge.


spotrh:
On further testing, this causes other bugs. Please ignore, and apologies for the noise. :)

spotaws:
On further testing, this causes other bugs. Please ignore, and apologies for the noise.

Issue #3862, opened 2021-05-04
ogex importer prints garbage to stdout 


JC3:
The ogex importer prints garbage to stdout when parsing invalid input files containing binary data. This garbage is unsightly, sometimes messes up the terminal, and also the messages go straight to stdout and bypass assimp's logging system. Examples include:

    Error:Invalid token "–" expected "{"–qÌp!®À“e»¸ÄÉ.éL\ÉŸ~ï}²ò$ÿþñ3yu'stà¦Z­p„q•ê

    Error:Invalid token "Æ" expected "{"
    Æ×éLï@Ù[HÈRíNÛLWí¦u@$lÓ AÈIz£ž¡‹ÎôB½BBR¢ÄŠïl

    Error:Invalid token "d" expected "{"d+Q

Source of this message is the OpenDDLParser. 

Solution involves two changes:

1. Silence the OpenDDLParser.
2. Update ogex importer to set up its own logging callback for OpenDDLParser, then pass messages through assimp's logging facilities.

Optional: Also remove non-printable characters from messages at some point in the flow here. Otoh, maybe that should be left to the assimp loggers. 🤔

JC3:
> Silence the OpenDDLParser.

Addressed by PR #3863.

JC3:
> Update ogex importer to set up its own logging callback for OpenDDLParser, then pass messages through assimp's logging facilities.

Addressed by PR #3864.

This issue is now completely addressed by the above two PRs.

JC3:
I've closed both of the aforementioned PRs, all changes are in #3881 now.

JC3:
Closed by #3881.

PR #3863, opened 2021-05-04
Remove default log handler and unsolicited output.

JC3:
This addresses part of #3862.

- Remove default log handler.
- Log callback can now be set to nullptr, which just makes logging a no-op.
- Initial log callback is nullptr.

Assimp code that uses this may regain logging output by installing a callback and directing the output through appropriate logging facilities.


JC3:
Closing; will be part of a larger PR.

JC3:
Closing; will be part of a larger PR.

PR #3864, opened 2021-05-04
Direct OpenDDLParser log messages to assimp logger

JC3:
Addresses second part of #3862.


JC3:
Closing; will be part of a larger PR.

Issue #3867, opened 2021-05-04
bvh importer assertion failures when attempting to import invalid files


JC3:
BVH importer raises a large number of the following assertion failures when it attempts to load invalid files containing binary data:

    minkernel\crts\ucrt\src\appcrt\convert\isctype.cpp(36) : Assertion failed: c >= -1 && c <= 255

While I have not investigated the exact source, this particular assertion is almost certainly due to passing signed chars to one of the cctype functions, which are undefined for values outside of EOF & the range of unsigned chars.

Presuming that's the case, the fix is to convert to unsigned chars (I promise this is the correct fix; I just don't feel like citing C99 / C++ standards right now).

krishty:
Mind having a look at `ply_reader.cc`, `PlyReader::SplitWords()` as well? Smells bad to me …

krishty:
Okay, I just searched the codebase for `toupper()`, `tolower()`, `isalpha()`, `isdigit()` and it’s full of places where untrusted text like texture names is passed without first casting to `unsigned char`.

@JC3 if this is too much for you, I’ll put it on my TODO for another day.

JC3:
It's too much for both of us, so let's both put it on our TODO lists for
another day and we'll see who gets their first. 😂

I actually will have time eventually just maybe not in the next day or two,
as I'm about to travel out of town for work.

JC3:
Turns out, I've got time. PR forthcoming. Notes:

- I looked for [all 14 functions from cctype](https://en.cppreference.com/w/cpp/header/cctype).
- I went with the immediate fix: Converting params to unsigned chars as needed. I noticed, though, lots of random and duplicate string-related code spattered around, so the philosophically correct fix would be to kind of unify all that and provide safer implementations in a central place. Except, I don't care, and I suspect nobody else does, either. So, I'm not doing that.
- I stayed away from zlib; I figure whatever's going on in there is already under control.
- I did, however, fix a potential issue in gtest, since it was pretty clear that there was a minor oopsie, and what the fix was. I may or may not be motivated to submit it to official gtest, too. It probably hasn't caused any issues.

I'm not sure how to prevent potential issues from reappearing in the future, though. Maybe there's a code analysis tool that can spot ctype calls with char parameters that can be added to CI tests? No idea.

PR forthcoming, as soon as I figure out how tf GitHub forks work (I'm a dinosaur).

---

[...]

PR #3868, opened 2021-05-04
Fix cctype calls

JC3:
Cast to unsigned char as required by C++ (see C++ **[cctype.syn]** -> ISO C99 section 7.4, [see also](https://en.cppreference.com/w/cpp/string/byte/isspace)).

Addresses #3867.


JC3:
Closing. This will be part of a bigger set of changes to address problem library-wide.

PR #3871, opened 2021-05-04
- fix security issue

kimkulling:
use strln_s instead of strlen.


krishty:
Why? `strnlen_s()` is only available in C. If you don’t trust the source data, enforce a zero terminator after the maximum allowed length.

kimkulling:
Was reported from a security check.

kimkulling:
Good idea!

PR #3873, opened 2021-05-04
enable debug information in MSVC release build

krishty:
No effect on runtime speed/size. Slightly slower link time, but debugging experience improves by a million fold.

- `/Zi` – Store debug information in a .pdb file, not directly in the binary
- `/DEBUG:FULL` – generate debug information during link
- `/PDBALTPATH:%_PDB%` – do not store the file system path of the .pdb, just the filename and hash (no disclose paths on distribution)
- `/OPT:REF /OPT:ICF` – remove unreferenced functions and fold identical functions (this was enabled before, but is implictly disabled by `/DEBUG:FULL`)


RichardTea:
This is not required, and should be withdrawn.

The Cmake "ReleaseWithDebug" project configuration to create MSVC Release builds with debug information.

Messing with the defaults is almost always a really bad idea.

krishty:
> Cmake recommends using the default created "ReleaseWithDebug" project configuration for MSVC Release builds that have debug information.

Then I’ll open an issue for Assimp’s RelWithDebInfo being broken, because it embeds the debug information into the executable and has some optimizations disabled, therefore it’s very impractical for distribution.

Compare
RelWithDebInfo: 11 MiB DLL with 3 MiB PDB, optimized for Edit & Continue (incrementally linked)
Release: 5 MiB DLL, all optimizations, but no way to debug it
Release with this patch: 5 MiB DLL, all optimizations, 50 MiB PDB for offline debugging

RichardTea:
That's a recent regression then, I've been using RelWithDebug for a few years.
I am about three/four months behind, I think.

RichardTea:
Well, I'm wrong. It's a cmake issue. That's tedious.

See here: https://gitlab.kitware.com/cmake/cmake/-/issues/20812

Can you re-jig your patch to apply to the RelWithDebugInfo configuration, as that's supposedly the MSVC configuration that create where these options are _supposed_ to end up.

BTW, /DEBUG:FULL is implied by command-line builds, though best to set it anyway.

krishty:
> Note:
> Definitely also need: `/INCREMENTAL:NO`

No, this is implied by `/OPT:REF /OPT:ICF`. But specifying it explicitly could make matters more clear.

kimkulling (approved):

kimkulling:

PR #3878, opened 2021-05-04
Less string bloat

krishty:
**Fix 1:**
Consider
```static const std::string foo = "bar"; // at global scope```

If only `c_str()` is used on these strings, this commit converts them to C strings like so:
```static constexpr auto foo = "bar";```

Reason: optimizers have trouble understanding `std::string` is static, and Visual C++ in particular generates ~500 B of instructions and data per static `std::string` for nothing. See https://godbolt.org/z/7rx7e1P73 vs https://godbolt.org/z/oYxzra9Wq

**Fix 2:**
Converted
```
std::string left = …;
char * right = …;
if(0 == strncmp(left.c_str(), right, left.size()))
```
to
```
if(0 == strcmp(left.c_str(), right))
```
because `strncmp()` stops at the first zero it encounters, so passing the string length was absolutely useless. This allowed even more `std::string`s being coverted to string literals.

**Edit: @kimkulling was faster removing strings from 3MF, so I cancelled my 3MF changes. Numbers are outdated.**

Before:
```
assimp-vc142-mt.dll
Overall code:   4047642 B
Overall data:    605663 B
Overall BSS:       3471 B
```
After:
```
assimp-vc142-mt.dll
Overall code:   4024654 B (-23054 B, -0.6%)
Overall data:    599398 B (-6265 B, -1%)
Overall BSS:       3471 B
```


kimkulling (approved):

kimkulling:
Size reduced, horray! Thanks a lot for your contribution.

Issue #3855, opened 2021-05-03
stb_image.h is duplicated; both copies outdated


krishty:
There are two copies of `stb_image.h` in Assimp:
1. `contrib/stb_image/stb_image.h`
2. `code/AssetLib/M3D/m3d.h`

The `contrib` copy is outdated by almost 11 years (1.19 from 2010; current is [2.26](https://github.com/nothings/stb/blob/master/stb_image.h)). This may be bad regarding known vulnerabilities like [this one](https://www.cvedetails.com/cve/CVE-2019-15058/).

The `m3d.h` copy is outdated by 5 years (2.13) and modified in subtle ways. Apart from being vulnerable as well, it’s bad practice to include a library twice.

I’ll try to merge and/or update both copies.

krishty:
There’s a 3rd copy for the PBRT exporter: `code/Pbrt/stb_image.h` (but that one is at least up-to-date). I merged it as well.

If https://github.com/assimp/assimp/pull/3889 is merged and tested successfully, this issue can be closed.

kimkulling:
Already fixed.

PR #3889, opened 2021-05-05
Stb image updated

krishty:
The M3D importer and the PBRT exporter brought their own copies of `stb_image.h`. There was a third copy in `contrib`. Two of the copies were outdated (see #3855 ).

This commit merges all three copies into a central instance under `contrib`. Moreover, it updates the various `stb_image.h` versions to the current version 2.26.

This removes 8600 lines of duplicate code from the source, and 13 KiB of code+data from the x64 MSVC version (7.5 KiB if compiled for import only).


RichardTea:
I strongly believe stb_image should be removed entirely. The host application should be decoding PNG (and other common image formats) using its own maintained reader.

krishty:
I agree; with this patch it should at least become easier to remove it. PBRT export uses it for weird things like checking whether a texture has an alpha channel.

kimkulling (approved):

krishty:
Not sure what to do about the warnings (C/C++ CI / ubuntu-latest-g++-build-and-test). The old `stb_image.h` version compiled without warnings. I’ll diff Assimp’s old version and the public old version to see whether we just removed the corresponding lines to silence GCC.

kimkulling:
Normally gcc is a little bit more strickt. I will invest some time to see, what is this all about.

krishty:
> Normally gcc is a little bit more strickt
We compile with `/Wall` on GCC, but with `/W3` on VC. With `/W4`, we’d get similar warnings. GCC is not more strict per se, it’s just different.

The previous `stb_image.h` was modified to silence the warnings. Here’s the commit: https://github.com/assimp/assimp/commit/09c5564d5b1123125ba0a7f07f7f8c1dfcdf7a86

Guess I’ll just repeat that.

rbsheth:
> I strongly believe stb_image should be removed entirely. The host application should be decoding PNG (and other common image formats) using its own maintained reader.

I agree, very interested in seeing it consolidated and possibly removed. Also, header-only libraries like stb_image are a pain to deal with in fully static builds like the ones we do with Assimp and Hunter.

rbsheth:
> > Normally gcc is a little bit more strickt
> 
> GCC is not more strict per se, it’s just different and we use different warning levels. (We compile with `/Wall` on GCC, but with `/W3` on VC. With `/W4`, VC gives similar warnings.)
> 
> Assimp’s previous `stb_image.h` was modified to silence the warnings. Here’s the commit: [09c5564](https://github.com/assimp/assimp/commit/09c5564d5b1123125ba0a7f07f7f8c1dfcdf7a86)
> Guess I’ll just repeat that.
> 
> Regarding the hunter failures, I’m not sure how I should write the path correctly?

@krishty You need to use a system-like path for Hunter includes. The error in CI is:
```
Error: ../code/Common/Assimp.cpp:1280:13: fatal error: ../contrib/stb_image/stb_image.h: No such file or directory
   1280 | #   include "../contrib/stb_image/stb_image.h"
```

You'd have to add the library properly to the Hunter library declarations (see https://github.com/cpp-pm/hunter/blob/master/examples/stb/CMakeLists.txt) and then include the library this way: https://github.com/cpp-pm/hunter/blob/master/examples/stb/boo.cpp#L1. Let me know if that doesn't make sense.

rbsheth:
@kimkulling I see some updates - Hunter will need a block like this:

```cmake
hunter_add_package(stb)
find_package(stb CONFIG REQUIRED)
```

Issue #3888, opened 2021-05-05
Numerous loader misbehaviors when file can not be opened (crashes, assertions, bad returns) [amf, 3ds, sib, irrm, irrs, quick3d, ms3d, cob, nendo, xgl, assbin]


JC3:
Library version: latest master at the time of this post.
Platform: Win 10 64-bit, debug build, MSVC 2019 64-bit.

A fair number of loaders crash (usually due to assertion failures) or return empty scenes (instead of failing) when faced with files that can not be opened:

- 3DS, SIB, Quick3D, Milkshape, TrueSpace, Nendo: When the file can not be opened for any reason, these all throw null stream assertion failures in debug mode, and generally crash in release mode.
- AMF: When a zero-length file is opened, it appears to return and (correctly) fail, but an access violation (termination with 0xc0000005 on Windows) soon follows. Have not investigated yet.
- Assbin: When the file can not be opened for any reason, this returns an empty scene instead of failing.
- Irrlicht Mesh, Irrlicht Scene, XGL: When a zero-length file is opened, these return empty scenes instead of failing.

Will investigate and fix as much as possible. 

---

I've got a test program that spits this out pretty quickly:

![image](https://user-images.githubusercontent.com/5651184/117201994-4ef37600-adbb-11eb-98c8-1d19f846d403.png)

In that table:
- White = correctly failed with an error message
- Yellow = did not fail, returned an empty scene instead
- Red = terminated the program (ignore *noaccess_file*'s redness -- unrelated)
- (Dots aren't too important here but generally solid = something printed to stderr = an assertion failure in this context.)

Where the input files were:
- `empty_file`: a zero length file
- `empty_dir`: a directory
- `noaccess_file`: a file with no read permissions
- `nonexistent_file`: doesn't exist
- `invalid|path`: not a valid filename on Windows
- `glob*in*.path`: you get the idea
- `$%$%$%$%$%$%$%`

This again highlights the need for better unit test coverage, especially in failure cases.

JC3:
I failed to mention: This is in `ReadFile`. I think there's a slightly different set in `CanRead` (e.g. #3849), I guess I should include everything here. Will update in a bit.

JC3:
I failed to mention: I'm referring only to `ReadFile`. The only loader with a `CanRead` issue is M3D, already reported in #3849:

![image](https://user-images.githubusercontent.com/5651184/117208088-b7922100-adc2-11eb-8023-f98b322e66c4.png)

[...]

PR #3890, opened 2021-05-05
Fix importer ReadFile issues on file open error or when opening empty files

JC3:
Addresses #3888.

Affects importers: amf, 3ds, cob, ms3d, nendo, quick3d, sib, assbin, irr, irrmesh, xgl.

Does *not* address #3849 (CanRead). That is addressed by #3850.

This PR does not have any dependencies.


kimkulling:

PR #3894, opened 2021-05-07
Add export property to control blob names.

crud89:
The current export logic does not allow for certain scenarios when exporting blobs, that require auxiliary files. For example, when exporting an Wavefront OBJ file, that contains materials, the material library will be referenced within the master blob using the magic blob file keyword:

```
mtllib $blobfile.mtl
```

Currently, assimp has no way to know the name of the material file beforehand. In order to fix this issue and add support for other file formats that may require auxiliary files, this PR introduces a `AI_CONFIG_EXPORT_BLOB_NAME` export property. Calling `ExportToBlob` evaluates the property and selects a base name or defaults to `AI_BLOBIO_MAGIC`, if it is not provided. 

Furthermore, using `AI_BLOBIO_MAGIC` or not specifying an export base name does not change the current behavior. If a custom base name is specified, the blob names will be set from it. If no custom base name is specified, the blob names will either be empty (for the master blob), or only contain the file extension.

I've tested all behaviours using this slightly modified `ExportModel` function of the command line application:

```cxx
ExportProperties exportProps{};
exportProps.SetPropertyString(AI_CONFIG_EXPORT_BLOB_NAME, "test.obj");
//const aiReturn res = globalExporter->Export(pOut,pID,path);
auto res = globalExporter->ExportToBlob(pOut, pID, 0, &exportProps);
```

Not passing `exportProps` results in the same behavior as before, whilst passing them as shown above properly adds `mtllib test.mtl` to the output file.

### DIFF
--- a/include/assimp/BlobIOSystem.h
+++ b/include/assimp/BlobIOSystem.h
@@ -194,8 +194,14 @@ class BlobIOSystem : public IOSystem {
     friend class BlobIOStream;
     typedef std::pair<std::string, aiExportDataBlob *> BlobEntry;
 
+
 public:
-    BlobIOSystem() {
+    BlobIOSystem() :
+            baseName{} {
+    }
+
+    BlobIOSystem(const std::string &baseName) :
+            baseName(baseName) {
     }
 
     virtual ~BlobIOSystem() {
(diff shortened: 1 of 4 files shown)


include/assimp/BlobIOSystem.h
```diff
@@ -207,27 +213,32 @@ class BlobIOSystem : public IOSystem {
...
     // -------------------------------------------------------------------
     const char *GetMagicFileName() const {
-        return AI_BLOBIO_MAGIC;
+        return baseName.empty() ? AI_BLOBIO_MAGIC : baseName.c_str();
     }
 
     // -------------------------------------------------------------------
     aiExportDataBlob *GetBlobChain() {
+        const auto magicName = std::string(this->GetMagicFileName());
+        const bool hasBaseName = baseName.empty();
```
kimkulling, line 222: Is this correct?  

`const bool hasBaseName = !baseName.empty();`

I guess this version is more correct. What do you think?

kimkulling (changes requested):

include/assimp/BlobIOSystem.h
crud89: Indeed, I messed up this check, sorry for that. It is more appropriate to compare the base name to `AI_BLOBIO_MAGIC`, since `GetMagicFileName` already returns the base name, if it is set. I've fixed this and tested both cases using and obj/mtl file:

- Base name set results in blob names "test.obj"/"test.mtl"
- Base name not set results in ""/"mtl", as before.

Thanks for reviewing. 🙂

kimkulling (approved):

kimkulling:
Looks good, I will merge it.

kimkulling:

PR #3896, opened 2021-05-08
[Logger] Log a notification instead of silently dropping long log messages.

JC3:
Logs a notification instead of silently dropping long log messages, which can complicate debugging.

This way, if you don't see a message you expect to see, you'll immediately know why.

The *correct* approach would be to eliminate length filtering here entirely and use `snprintf` or whatever appropriately (also there's a tiny -- probably negligible -- performance hit here in calling `strlen` regardless of whether or not the verbosity level matches). Failing that, the second best option is to copy and truncate messages here. However, for now, this should be OK.


RichardTea:
I don't think this check is needed at all anymore, as `OnDebug()` et al now call `ai_snprintf` which truncates the passed message and ensures null termination.

I assume the check was from before the change to the "n" version of `sprintf`

kimkulling (approved):

kimkulling:

PR #3901, opened 2021-05-12
Add support for arm 64 bit

impala454:
The helper function which locates assimp for the python port fails on arm64 simply because it's only looking in the x86_64 directory.  This simple add makes pyassimp properly work on aarch64.


kimkulling (approved):

kimkulling:

PR #3903, opened 2021-05-15
Change the Assimp output directory vars to cached vars

ogjamesfranco:
Currently, the variables that control the library, runtime, and archive output directories are normal variables.
Changing them to cache variables would allow a parent project to set the output directories if desired.
I think this would be nicer than having to copy the files with a post-build step.


kimkulling (approved):

kimkulling:

PR #3375, opened 2020-08-18
Make internal errors accessible

MalcolmTyrrell:
There are two main parts to this PR:
1. DeadlyImportError and DeadlyExportError now have formatting built in.
    - I removed all instances of the formatter from callsites
    - I removed many uses of std::string::operator+ and several unnecessary std::string constructions.
    - I've simplified one or two code paths so the way they call DeadlyImportError is more straightforward (e.g. FBX tokenization errors).
2. The Importer API now has an additional method that provides access to the exception which was thrown via a std::exception_ptr. 
    - The existing GetErrorString API remains. 
    - GetException can be used to get full details of exceptions.

Note:  The two aspects of this PR are probably separable and, if requested, I'm willing to go to the trouble of separating them. I'd rather not though :D

A note on motivation: Both of these changes are extremely useful when you want to customize error handling. Part 1 centralizes the formatting, and allows a customized DeadlyImportError some control over the message they carry. Part 2 allows any additional details carried by a customized DeadlyImportError (e.g. a callstack) to be retrieved by the calling context. It additionally allows the calling context to try to make sense of the exceptions which would normally be discarded.

PR #3905, opened 2021-05-17
[Logger] Unify log formatting

MalcolmTyrrell:
Log messages were being formatted in several ways. For example, some used the _F version of the macros, while many others added std::strings together.

This change unifies the approach to logging, so that all log messages are now written with a sequence of comma separated arguments. The change is similar to the change I made to the exception classes in #3375.

I do not expect any performance impact for this change, since single argument log calls should not go through the formatting code path.


kimkulling (approved):

kimkulling:

Issue #2256, opened 2018-12-04
aiCamera: mHorizontalFOV documentation appears incorrect


twhittock:
It appears that all exporters I am aware of regard mHorizontalFOV as being the full width FOV of the Camera, not the half-width FOV, as the documentation states.

This appears to be the case with Collada and FBX, at least. Is the documentation or the code incorrect?

inhosens:
I agree, it is really confusing.

dlyr:
Confirm the issue on my side (tested with blender exported collada, the exported file have correct fov value, assimp loader give full fov, not half as confirmed by https://github.com/assimp/assimp/blob/92b8c02982980d6a324dbe3755a27c2d76e966ca/code/AssetLib/Collada/ColladaLoader.cpp#L443)
and here
https://github.com/assimp/assimp/blob/92b8c02982980d6a324dbe3755a27c2d76e966ca/code/AssetLib/Collada/ColladaParser.cpp#L999

easy fix: update the doc :)

kimkulling:
Closed by https://github.com/assimp/assimp/pull/3912

kvnnap:
This change is incomplete. The documentation [here](https://github.com/assimp/assimp/blob/aae93691b9923e13af089a2a092fdd75bbfab489/include/assimp/camera.h#L138C1-L144C26) still says:

```C++
    /** Horizontal field of view angle, in radians.
     *
     *  The field of view angle is the angle between the center
     *  line of the screen and the left or right border.
     *  The default value is 1/4PI.
     */
```

The angle between the center line of the screen and the left or right border is still half the FOV.

Additionally if the definition of `mHorizontalFOV` has now changed from half to full angle, what about other importers such as FBX which were correctly outputting half the FOV? see [here](https://github.com/assimp/assimp/blob/aae93691b9923e13af089a2a092fdd75bbfab489/code/AssetLib/FBX/FBXConverter.cpp#L449C34-L449C34).

PR #3912, opened 2021-05-21
Fix camera fov comment since full fov is stored

dlyr:
fix issue #2256 
only checked in collada loader source


kimkulling (approved):

kimkulling:

PR #3918, opened 2021-05-23
Adding basic support for lights in FBX exporter

Nodrev:
This probably doesn't handle all cases, but it does add a basic export to FBX for the assimp lights.
The generated exported files were tested using Blender 2.83 FBX importer.


kimkulling (approved):

kimkulling:

PR #3922, opened 2021-05-26
Fix bug in aiMetadata constructor that overwrites an array of one of …

Evangel63:
…aiString, aiVector3D, or aiMetadata with the first entry

aiMetadata copy constructor calls aiMetadata::Get on the copied from aiMetadata using the const aiString &key version. When
this is called on the metadata of an array type, this overwrites all entries with the first entry. This is due to the key
of all entries in an array being the name of the array. ie, in a glTF2 file with an extension:

```
"Extension" : [
	"Value1",
	"Value2",
	"Value3"
]

```
the aiMetadata struct for the "Extension" entry will have 3 entries with key/value pairs as:

```
"Extension"/"Value1"
"Extension"/"Value2"
"Extension"/"Value3"

```
So when the copy constructor calls the key based aiMetadata::Get, it will find "Value1" for all three entries.

This change simply replaces the key based aiMetadata::Get with the index based aiMetadata::Get

### DIFF
--- a/include/assimp/metadata.h
+++ b/include/assimp/metadata.h
@@ -202,17 +202,17 @@ struct aiMetadata {
             } break;
             case AI_AISTRING: {
                 aiString v;
-                rhs.Get<aiString>(mKeys[i], v);
+                rhs.Get<aiString>(i, v);
                 mValues[i].mData = new aiString(v);
             } break;
             case AI_AIVECTOR3D: {
                 aiVector3D v;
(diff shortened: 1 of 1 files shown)


Evangel63:
The Windows builds fail at what seems to be a task queuing stage rather than a compilation or testing stage. I'm unfamiliar with a Windows build environment and don't have one to test locally. Could someone else help me understand what's going on there?

kimkulling (approved):

kimkulling:

PR #3938, opened 2021-06-01
SceneCombiner memory issues when re-indexing textures.

BaldricS:
We've been doing a lot of work with pulling in multiple scenes and combining them into one scene that we later export. We've recently upgraded our scenes to include lots of textured models and noticed (after this change in sources), that we were tripping an assert in debug builds of assimp.

The assert in question can be found [here](https://github.com/assimp/assimp/blob/master/code/Material/MaterialSystem.cpp#L305).

After some head-scratching, I realized that when there are enough textures in the merged scenes to trigger a magnitude change on re-index (IE 0 -> 10 or 10 -> 100), then the `mDataLength` property of the `aiMaterialProperty` struct was off by the amount of the magnitude change.

This change addresses the change in magnitude by updating `mDataLength` by the delta of the character change between the old and new embedded texture index. In the case of no change in magnitude, the member will remain the same, but in the examples above, it'd increase by 1.

In my local tests, this changed stopped tripping the debug asserts in the MaterialSystem file.


BaldricS:
Looking at some errors I'm now receiving after testing further, I'm actually thinking this is leading to bad behavior in another section of my app, and I don't think this should be pulled just yet.

BaldricS:
Okay!

I believe I have solved the underlying issue effectively this time.

The issue: When updating texture index offsets, they sometimes changed in magnitude from 0-9 -> 10-100.
The problem: This tripped debug asserts because the length of the property wasn't being updated to account for this.
The deeper problem: Because we were re-interpreting prop->mData into an aiString, we were writing into memory we shouldn't have been writing to because aiString has a fixed buffer, but a prop->mData does _not_ and should not be modified through the aiString interface.

This fix addresses both problems by copying the mData into a new aiString which, after the mData buffer is regrown if the string changes in length, is copied back into the mData buffer. Furthermore, mDataLength is now updated when the size of the string changes (as it wasn't being changed before this PR).

kimkulling:

Issue #3887, opened 2021-05-05
PyAssimp cannot load PLY


Levi-Armstrong:
**Describe the bug**
This appears to be similar to this [issue](https://github.com/assimp/assimp/issues/2339) while loading ply. I have tried both ASCII and binary formats which produce the same error. I am currently using pyassimp version 4.1.4.

```
'Traceback (most recent call last):\n',
'  File "pyassimp/core.py", line 321, in load\n    scene = _init(model.contents)\n',
'  File "pyassimp/core.py", line 216, in _init\n    call_init(obj, target)\n', 
'  File "pyassimp/core.py", line 84, in call_init\n    _init(obj,parent=caller)\n', 
'  File "pyassimp/core.py", line 212, in _init\n    setattr(target, name, obj)\n', 
"AttributeError: 'bytes' object has no attribute 'aketrans'\n"
```

Levi-Armstrong:
Just tried version 4.1.3 and it works as expected.

kimkulling:
Hi,
thanks for your report. Indeed, there is an error in the iteration code for attributes of the loaded assimp structs. We have to check which kind of attribute we found and accidentally the wrong token was used to get the attributes. In the code, there was a check for the prefix m and for one case the starting m was used in the following code. 

Instead of this token we need to use the original token. 

I will try to fix the issue here: https://github.com/assimp/assimp/pull/3940 

I will test the new stuff with the 5.01. If this fix works we need to update the pyassimp version.

PR #3940, opened 2021-06-03
Use correct attribute name

kimkulling:
- closes https://github.com/assimp/assimp/issues/3887


mlopezantequera:
@kimkulling Just checking: will this also fix the same error on .fbx files? 
I got the same error when trying to load the .fbx file for this sample model: https://free3d.com/3d-model/-74958.html

PR #3945, opened 2021-06-08
Add GetEmbeddedTextureAndIndex() to aiScene.

mahiuchun:
It allows the caller to get the index of the embedded texture that is always computed anyway.


kimkulling:

Issue #3949, opened 2021-06-10
Generic PBR Materials Handling


RichardTea:
Most modern rendering engines prefer to use one or other variant of Physically-Based Rendering (PBR)

There are two main PBR workflows in current usage, which use the following texture maps
- Metallic/Roughness
  - Base Color (ARGB with color space and gamma correction)
  - Roughness (single value)
  - Metallic (single value)

- Specular/Glossiness
  - Diffuse/Albedo (ARGB with color space and gamma correction)
  - Glossiness (single value)
  - Specular (single value)

Additionally, the provided texture maps may combine two or more of the above single-value channels into a single map, and may invert the sense (eg Smoothness = 1.0 - Roughness)

**Current Situation**
Assimp does not have a defined consistent way to encode these PBR models.
The glTFv2 importer uses its own private set of material properties, while the FBX (Maya/3DS Max) importer uses another.

Eg glTFv2 places the Base Color into aiTextureType_DIFFUSE, and all the other PBR textures under `aiTextureType_UNKNOWN`, indexed as 0 thru 5.

Worse, the application has no way of knowing which color channels are in use, or what they mean. It has to know details about how a given format encodes the data. Formats that support multiple ways of encoding PBR data have no way of passing this information on to the application.

**Proposal**
Amalgamate and clean up the existing different sets of PBR material properties as follows:

***aiShadingMode***
Add
- `aiShadingMode_PBR`

Defines that the materials uses a PBR-type shading model.

Applications that support rendering of PBR materials then use the list of provided Properties (AI_MATKEY_XXX) and Textures (aiTextureType_XXX) to determine which PBR shading model is in use.

***aiTextureFlags***
Add
- `aiTextureFlags_UseRed`
- `aiTextureFlags_UseGreen`
- `aiTextureFlags_UseBlue`

Explicitly sets which channel(s) of a texture are actually in use for the specified texture function.

Combine with:
- `aiTextureFlags_InvertRed`
- `aiTextureFlags_InvertGreen`
- `aiTextureFlags_InvertBlue`
- `aiTextureFlags_InvertAlpha`

These allow a texture channel to be explicitly flipped if needed.
Eg a Normal Map might set `aiTextureFlags_UseRed | aiTextureFlags_UseGreen | aiTextureFlags_InvertRed | aiTextureFlags_InvertGreen` to indicate that it's using the Red and Green channels, and inverts them compared to Assimps' default left-handed setting.

***AI_MATKEY_XXX***
[...]

PR #3952, opened 2021-06-10
First pass at simplifying PBR

RichardTea:
Removed 'core' set of GLTF-specific properties, use the same generic ones used by FBX 3DS Max PBR importer

Added notes about use of PBR BRDF/BSDF properties
Added some additional values to material enums

Part of #3949


kimkulling (approved):

kimkulling:
Cool, thanks a lot for this first step. Merged!

lazka:
This breaks the qt5-quick3d build because it removes various macros. Was that on purpose?

RichardTea:
Sorry about that, I didn't think anyone was using them as they appeared to be a private prototype.
I'll put together a PR that puts them back into the gltf-pbr header and marks it as deprecated.

lazka:
I'm just a user/packager of qt5, so I don't know if qt is doing something wrong here :) I just updated to the new release.

I've downgraded to assimp 5.0 for now.

PR #3954, opened 2021-06-11
the expression does not throw an exception.

ihsinme:
maybe you just forgot this word.


kimkulling (approved):

kimkulling:
Good catch. Merged. Thanks a lot for your contribution.

PR #3964, opened 2021-06-22
Performance: Apply various performance fixes from clang-tidy

Skylion007:
* I used clang-tidy to fix a bunch of performance issues in the code base using the "*performance*" fixes. I only applied the automatic ones and made sure everything still compiled. There was one false positive in a Blender template which I reverted before committing. The fixes below mostly fall in the following categories:
1. Useless std::move (the type is trivially copyable or std::move would just do a copy anyway).
2. Missing std::move (a type that is movable is not actually moved but rather copied. Making it a move should fix performance).
3. Unnecessary pass-by-value param copying (many method args are not actually modified within the method and could be safely converted to const ref&. This preventsA an unnecessary copy each time the function is invoked.
4. Unnecesary copies in a loop. Removes useless implicit copies (ie. if each element could be const ref instead). This prevents each item in a container from being copied.
5. Calling find with a one character long string instead of a char. STL containers often have an overload for char which performs faster and should be used over the 1 character long string version.
Finally, I ran clang-format on the diff.


Skylion007:
Ping @kimkulling

kimkulling:
On my way!

kimkulling (approved):

kimkulling:
Looks good, I will merge it!

kimkulling:

PR #3966, opened 2021-06-23
include/material.h: Fixed broken C support

jerstlouis:
- The aiGetMaterialFloat() and aiGetMaterialInteger() C preprocessor definitions were broken because:
   - They had a space before the opening parenthesis
   - Using material key definitions expanding 1 to argument into 3 breaks the invocation of macros expecting 5 parameters

### DIFF
--- a/include/assimp/material.h
+++ b/include/assimp/material.h
@@ -1499,7 +1499,7 @@ ASSIMP_API C_ENUM aiReturn aiGetMaterialFloatArray(
         ai_real *pOut,
         unsigned int *pMax);
 
-#ifdef __cplusplus
+#if 1 //def __cplusplus
 
 // ---------------------------------------------------------------------------
 /** @brief Retrieve a single float property with a specific key from the material.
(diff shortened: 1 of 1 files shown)


include/assimp/material.h
```diff
@@ -1499,7 +1499,7 @@ ASSIMP_API C_ENUM aiReturn aiGetMaterialFloatArray(
         ai_real *pOut,
         unsigned int *pMax);
 
-#ifdef __cplusplus
```
kimkulling, line 1502: Is this correct?

include/assimp/material.h
```diff
@@ -1548,7 +1548,7 @@ ASSIMP_API C_ENUM aiReturn aiGetMaterialIntegerArray(const C_STRUCT aiMaterial *
         int *pOut,
         unsigned int *pMax);
 
-#ifdef __cplusplus
```
kimkulling, line 1551: Same here, the old preprocessor directive is not active. Maybe you have introduced this for debugging proposes?

kimkulling (changes requested):

include/assimp/material.h
jerstlouis: No that is as intended, it is the fix -- not to use those macros, because they are broken as I explained.
The inline function however work fine in C (with the fix for the one missing C_STRUCT).

I kept the `#ifdef` in case the macros can eventually be fixed, but they are problematic with the material keys which gets expanded from 1 to 3 args (but _after_ the higher level macro invocation).

kimkulling:
Ok, understood. I have removed the preprocessor else branch because I think this way the code will work and the idea in it is more visible. Thanks a lot for your investigation and the PR.

kimkulling:

PR #3969, opened 2021-06-27
Added support for custom properties ("extras") in glTF2 importer

Promit:
glTF2 allows nodes (as well as other objects) to have custom properties assigned to them, under the "extras" sub-object. These are used, for example, when exporting a Blender scene that uses its Custom Properties. This PR parses that data for nodes and exposes it under the assimp metadata. This is built on top of the work in #3298 with a few minor tweaks, and finishes the rest of the requested improvements from #3025.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Asset.inl
+++ b/code/AssetLib/glTF2/glTF2Asset.inl
@@ -372,6 +372,12 @@ inline void Object::ReadExtensions(Value &val) {
     }
 }
 
+inline void Object::ReadExtras(Value &val) {
+    if (Value *curExtras = FindObject(val, "extras")) {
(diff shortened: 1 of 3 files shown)


code/AssetLib/glTF2/glTF2Asset.inl
```diff
@@ -372,6 +372,12 @@ inline void Object::ReadExtensions(Value &val) {
     }
 }
 
+inline void Object::ReadExtras(Value &val) {
```
Skylion007, line 375: Shouldn't this be a const reference?

kimkulling (approved):

kimkulling:

Issue #3706, opened 2021-03-13
error 'GeneratorContextManager' object has no attribute 'meshes'


Mamartinezh:
**Describe the bug**
Hi! I'm new to assimp, I'm trying to load a model but when iterate over the meshes of the scene get this error : AttributeError: 'GeneratorContextManager' object has no attribute 'meshes'. I'm using python 2.7 and pyassimp, the assimp library compiled with CMake. Here is a snippe of the code that I'm trying to run using the load function of pyassimp, the object is a .OBJ format object and is just a plate (nothing fancy). Thank you in advance and regards!

**Screenshots**
![image](https://user-images.githubusercontent.com/64501300/111044844-f1752580-8418-11eb-9cef-e775357cd04d.png)

**Desktop (please complete the following information):**
 - OS: Windows 10

contriteobserver:
not able to reproduce this issue with this file, what is missing?
[plate.tar.gz](https://github.com/assimp/assimp/files/6154080/plate.tar.gz)
```
$ python3
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyassimp 
>>> source = "plate01.obj"
>>> scene = pyassimp.load(source, 0
... | pyassimp.postprocess.aiProcess_Triangulate
... | pyassimp.postprocess.aiProcess_SplitLargeMeshes
... | pyassimp.postprocess.aiProcess_GenNormals
... )
>>> for i, mesh in enumerate(scene.meshes):
...             print("num vertices(" + mesh.name + "): " + str(len(mesh.vertices)))
... 
num vertices(Plate): 9472
```

MateoeslaLey:
@contriteobserver Which version of assimp are you ussing? I see you are using python 3.6.9, could that be?  Also, I'm not 100% sure of my assimp compilling, I'm getting some warnings for change in integer precission but I dont think that would affect... Honestly I've tried so many thing that at this point dont know what could be the probabbly cause. Thank you so much for you respond!! A little hope for solving this issue!

contriteobserver:
just using python3-pyassimp 4.1.0

manuel-koch:
I have the same issue ( even with current master = 939dacd ) when loading my OBJ file(s).
assimp = 5.0.1
pyassimp = master

Same issue when using the `plate01.obj` provided by @contriteobserver:

```
[...]

PR #3979, opened 2021-07-01
Fix issues stopping PyAssimp's sample.py from working under Python 3

SaladDais:
### Revert https://github.com/assimp/assimp/commit/b4fc41bc094574400b0a688ec1381888e40ee99c

The issue it intended to fix was already fixed by https://github.com/assimp/assimp/commit/f305f105512237d7b44ec6c28492ac943e47fbcc, and setting `name` instead of `m` breaks access to `scene.rootnode` / `mRootNode` and any other non-iterable m-prefixed attr. Problem was just that the fix was only in `master` and they were testing against `4.1.4`.

This fixes failure with:

```
$ python scripts/sample.py ../../test/models/glTF2/simple_skin/simple_skin.gltf 
Traceback (most recent call last):
  File "scripts/sample.py", line 91, in <module>
    main(sys.argv[1])
  File "scripts/sample.py", line 25, in main
    scene = pyassimp.load(filename, processing=pyassimp.postprocess.aiProcess_Triangulate)
  File "/home/user/source/assimp/port/PyAssimp/pyassimp/core.py", line 330, in load
    recur_pythonize(scene.rootnode, scene)
AttributeError: 'Scene' object has no attribute 'rootnode'
```

### Split contextmanager portion of `pyassimp.load()` into `pyassimp.load_scoped()`

Added in efbabf3b0daccb976639a3734fa9018c4b3e13fa, without splitting this out there's no way to use a `scene` in a block other than the one it was allocated in. I added a separate `load_scoped()` that has the contextmanager semantics that were added to `load()`. It'd be nicer to add a `__exit__` to the `Scene` struct that could automatically clean things up, but doesn't seem like that'd be possible without pushing parts of `core.py` into `structs.py`

This fixes failure with:

```
$ python scripts/sample.py ../../test/models/glTF2/simple_skin/simple_skin.gltf 
MODEL:../../test/models/glTF2/simple_skin/simple_skin.gltf
SCENE:
Traceback (most recent call last):
  File "scripts/sample.py", line 89, in <module>
    main(sys.argv[1])
  File "scripts/sample.py", line 31, in main
    print("  meshes:" + str(len(scene.meshes)))
AttributeError: '_GeneratorContextManager' object has no attribute 'meshes'
```

### Make sample.py parse under Python 3

The bare `print` is a no-op under Python 3, so added a `future` import for Python 2 and changed it to `print("")`.

### Tests

I also added some simple smoke tests that can be run with `python setup.py test`.

### DIFF
--- a/port/PyAssimp/README.md
+++ b/port/PyAssimp/README.md
@@ -42,8 +42,8 @@ substituted by assertions ...):
 
 ```python
 
-from pyassimp import load
-with load('hello.3ds') as scene:
+from pyassimp import load_scopedz
+with load_scoped('hello.3ds') as scene:
 
   assert len(scene.meshes)
   mesh = scene.meshes[0]
@@ -58,8 +58,8 @@ scene:
 
 ```python
 
-from pyassimp import load
-with load('hello.3ds') as scene:
+from pyassimp import load_scoped
+with load_scoped('hello.3ds') as scene:
 
   for c in scene.rootnode.children:
       print(str(c))
--- a/port/PyAssimp/README.md
+++ b/port/PyAssimp/README.md
@@ -42,7 +42,7 @@ substituted by assertions ...):
 
 ```python
 
-from pyassimp import load_scopedz
+from pyassimp import load_scoped
 with load_scoped('hello.3ds') as scene:
 
   assert len(scene.meshes)
--- a/port/PyAssimp/README.rst
+++ b/port/PyAssimp/README.rst
@@ -49,8 +49,8 @@ substituted by assertions ...):
 .. code:: python
 
 
-    from pyassimp import load
-    with load('hello.3ds') as scene:
+    from pyassimp import load_scoped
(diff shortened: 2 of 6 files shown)


port/PyAssimp/README.md
```diff
@@ -42,8 +42,8 @@ substituted by assertions ...):
 
 ```python
 
-from pyassimp import load
-with load('hello.3ds') as scene:
+from pyassimp import load_scopedz
```
Skylion007, line 45: typo

SaladDais: fixed, thanks!

SaladDais:
Related to #3706

aleksandaratanasov:
Using pyassimp 5.2.5 (installed through PyPI) this issue is still here (Ferbruary 2024).

SaladDais:
For the record, pyassimp seems to no longer be maintained. I maintain a mostly-analogous API at https://github.com/SaladDais/Impasse.

PR #3981, opened 2021-07-01
Manage /R/N lines ends correctly on binary files, tested with solidworks PLY export

arkeon7:
Correction from issue #3711
The deduction was correct but the correction didn't work on the solidworks.ply test file from the issue.

This change force the skip the /n char in the beginning of the buffer when it's present.


kimkulling (approved):

kimkulling:
Looks good, I will merge it.

kimkulling:

PR #3985, opened 2021-07-07
Fix stb_image dependency

rbsheth:
A fixed version of #3889 that should work with Hunter.

PTAL @kimkulling


kimkulling (approved):

kimkulling:

PR #3988, opened 2021-07-13
Collada: Read <matrix> tags properly, assume <input set="0"/> when not present

### LINKED ISSUE #3984: [ColladaImporter] [Regression] [Bug] Fail to import .dae from makehuman
OS: Any (tested Windows x86_64 and Arch Linux x86_64)
Assimp Version: latest git; Release v5.0.1 is not affected
1. Create any character with MakeHuman
2. Export a resulting model to DAE Collada
3. ``aiImportFile("path/to/human.dae", anyPostrocessFlags)`` returns nullptr

I think this is bug of libassimp, since Blender and libassimp 5.0.1 are fine with importing the humans.

RichardTea:
"matrix" tags may have whitespace preceding the first value
"input" (shared) "set" attribute is optional, assume zero if not provided. (See Collada 1.4.1 5-39)

Closes #3984


kimkulling (approved):

kimkulling:

PR #3989, opened 2021-07-14
Find stb for Assimp

rbsheth:
@kimkulling Oops, I missed this in #3985 which causes a failure when building through Hunter: https://github.com/cpp-pm/hunter/pull/440/checks?check_run_id=3066470709


kimkulling (approved):

kimkulling:

PR #3992, opened 2021-07-16
Fix issues encountered during integration attempt

AdrianAtGoogle:
- when called on a file with no path, the directory was being computed as the file name.
- was throwing exceptions on gltf files with accessors with a count of zero (which is admittedly _silly_ but not exceptional)


kimkulling (approved):

kimkulling:
Looks good so far, let's merge it.

kimkulling:

PR #4001, opened 2021-07-28
StepExporter support polygon mesh 

xiaohunqupo:
StepExporter support polygon mesh


kimkulling:

PR #4016, opened 2021-08-02
Use strlen() rather than fixed length in fast_atof.h

mahiuchun:
This avoids reading past the length of the input string.


kimkulling (approved):

RichardTea:
This can overflow if the string is not null-terminated within a reasonable number of bytes.
Perhaps use strnlen_s?
(This requires C11)

PR #4017, opened 2021-08-02
Obj: make a predicate more robust.

mahiuchun:
Since we might encounter invalid input it is a good idea to check the actual
size of the array.


kimkulling (approved):

kimkulling:
Looks good, I will merge it.

kimkulling:

PR #4026, opened 2021-08-11
Fix: incorrect reading of PBR properties in FBX

Mykhailo418:
Fix issue of incorrect reading of PBR properties such as base and emissive color in FBX. In FBXConverter.cpp method GetColorPropertyFromMaterial adds unnecessary(for PBR properties) "Color" string to property name.


kimkulling:

PR #4036, opened 2021-08-19
fix sample build error

yzthr:
modify stb_image header path, fix opengl sample build error.


kimkulling:

PR #4044, opened 2021-08-24
Fix M3D import crash and memory leak.

kovacsv:
The same default material pointer was assigned to all the materials in the scene, so poor destructor tried to free the same pointer multiple times.

### DIFF
--- a/test/unit/utM3DImportExport.cpp
+++ b/test/unit/utM3DImportExport.cpp
@@ -50,35 +50,32 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 using namespace Assimp;
 
-class utM3DImportExport : public AbstractImportExportBase {
-public:
-	bool importerTest() override  {
-        Assimp::Importer importer;
-        const aiScene *scene = importer.ReadFile(ASSIMP_TEST_MODELS_DIR "/M3D/cube_normals.m3d", aiProcess_ValidateDataStructure);
+TEST(utM3DImportExport, import_cube_normals) {
+    Assimp::Importer importer;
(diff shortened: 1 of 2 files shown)


kovacsv:
Ok, the sanitizer detected some undefined behavior which I don't plan to fix. :)

Maybe I can remove the new test, so the sanitizer will remain silent...

test/unit/utM3DImportExport.cpp
```diff
@@ -50,35 +50,32 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #ifndef ASSIMP_BUILD_NO_M3D_IMPORTER
-		return nullptr != scene;
+    ASSERT_NE(nullptr, scene);
 #else
-        return nullptr == scene;
+    ASSERT_EQ(nullptr, scene);
 #endif // ASSIMP_BUILD_NO_M3D_IMPORTER
-    }
-
-#ifndef ASSIMP_BUILD_NO_EXPORT
```
kimkulling, line 65: What is the intention for rewriting the testcase, when the result is the same?

kimkulling:
I am fine with the solution, I am currently struggeling with the redesign of the tests.

kimkulling (changes requested):

test/unit/utM3DImportExport.cpp
kovacsv: I wanted to add another one, and I thought it would be better to follow the structure of another tests. But because the new test revealed another error, I'll revert the changes in this file.

kovacsv:
I've reverted back the test to the original one.

kimkulling (approved):

kimkulling:
Merged, thansk a lot for your contribution.

PR #4053, opened 2021-08-27
Add export property for assimp json exporter to write compressed json

kovacsv:
Added a new flag for `JSONWriter`, and an export property to handle it from the outside.


kimkulling (approved):

kimkulling:

PR #4057, opened 2021-08-31
Double Precision Issue

Madrich:


code/AssetLib/3DS/3DSLoader.cpp
```diff
@@ -449,7 +449,7 @@ void Discreet3DSImporter::ParseChunk(const char *name, unsigned int num) {
         // Read the lense angle
         camera->mHorizontalFOV = AI_DEG_TO_RAD(stream->GetF4());
         if (camera->mHorizontalFOV < 0.001f) {
-            camera->mHorizontalFOV = AI_DEG_TO_RAD(45.f);
+            camera->mHorizontalFOV = float(AI_DEG_TO_RAD(45.f));
```
kimkulling, line 452: Please use ai_real instead

code/PostProcessing/CalcTangentsProcess.cpp
```diff
@@ -56,7 +56,7 @@ using namespace Assimp;
 // ------------------------------------------------------------------------------------------------
 // Constructor to be privately used by Importer
 CalcTangentsProcess::CalcTangentsProcess() :
-        configMaxAngle(AI_DEG_TO_RAD(45.f)), configSourceUV(0) {
+        configMaxAngle(float(AI_DEG_TO_RAD(45.f))), configSourceUV(0) {
```
kimkulling, line 59: Please use ai_real instead

kimkulling (changes requested):

code/AssetLib/3DS/3DSLoader.cpp
Madrich: mHorizontalFOV and configMaxAngle are floats by definition and should not be changed to double. The AI_DEG_TO_RAD return ai_real, so we need to convert to float.

code/PostProcessing/CalcTangentsProcess.cpp
Madrich: same

kimkulling: Ok, I am fine with that.

code/AssetLib/3DS/3DSLoader.cpp
kimkulling: I am fine with that.

kimkulling (approved):

kimkulling:

PR #4071, opened 2021-09-08
Fix path for _IMPORT_CHECK_FILES_FOR_assimp

purushottamc:
As it is giving configuration error during validation of path
when used by 3rd party component to configure itself
using assimp::assimp target.
Compute the installation prefix relative to file.


purushottamc:
@kimkulling Could you please review this change?

kimkulling (approved):

kimkulling:

PR #4077, opened 2021-09-11
removed useless code

krishty:
In destructors, zeroing attributes or clearing containers is utterly useless.


turol:
Not really. When optimization is on these are dead. But in debug mode this will actually make sure that the reference is killed which helps tools like Valgrind and ASan. They could be replaced with a memset-like function that is guaranteed to not get optimized out, forget the name.

krishty:
Can you elaborate? If Valgrind/ASan did not actually recognize pointers in freed structures as inaccessible without zeroing them, it would be a pretty fundamental misunderstanding of C++ object lifetime, and hence draw them useless.

For a test, I just ran
```
struct AutoString_Zero {
    char * val;
    AutoString_Zero() : val(new char[16]) { }
    ~AutoString_Zero() { delete [] val; val = nullptr; }
};

struct AutoString_NoZero {
    char * val;
    AutoString_NoZero() : val(new char[16]) { }
    ~AutoString_NoZero() { delete [] val; }
};

int main()
{
    // Zero it
    {
        char * danglingPointer;
        {
            AutoString_Zero i;
            danglingPointer = i.val;
        }
        strcpy(danglingPointer, "blubblabblepboo");
    }
    // Don't zero it
    {
        char * danglingPointer;
        {
            AutoString_NoZero i;
            danglingPointer = i.val;
        }
        strcpy(danglingPointer, "blubblabblepboo");
    }
}
```
… and ASan reported the exact same error for both locations.

turol:
Number of years ago I ran into a strange issue where a pointer in a deleted block led to the tool thinking that the target was still reachable. I can't remember the details and it was probably a tool bug.

kimkulling:
For older compilers this was right. And that was the reason for me to add this code. But with newer compilers this hasn't occurred on my machine.
Any other opinions?

krishty:
https://stackoverflow.com/questions/3060006/is-it-worth-setting-pointers-to-null-in-a-destructor

kimkulling:
Merged, thanks a lot for  your contribution.

PR #4078, opened 2021-09-11
more const in format detection

krishty:
BaseImporter::SearchFileHeaderForToken() expected a pointer to a non-const token list. This was probably an oversight, as nobody would realistically expect the function to change the list. Furthermore, it prevented token lists from being compiled to read-only memory, in some cases (where also declared static) even causing the compiler to generate thread-safe initialization.

The list is now const and all callers declare their token lists static const, thus compiling them to read-only memory.


krishty:
In case BaseImporter::SearchFileHeaderForToken() is called from external programs/libraries (hard for me to tell), please place in a later release. It can be ABI-breaking because the mangled function name may change.

kimkulling (approved):

kimkulling:

PR #4083, opened 2021-09-15
including <exception>

markoffline:
When compiling on Linux, it was spitting out a error at include/assimp/Importer.hpp. Idk why "_#include <exception>_" was commented out, but I undid that comment.

I'm new to opensource btw, I don't really know what I'm doing.

### DIFF
--- a/include/assimp/Importer.hpp
+++ b/include/assimp/Importer.hpp
@@ -59,7 +59,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // Public ASSIMP data structures
 #include <assimp/types.h>
 
-//#include <exception>
+#include <exception>
 
 namespace Assimp {
 // =======================================================================


bartoszek:
_Besides blabbing like a noob and intangibly naming your PR, you're doing well dude_ 😏 

I'm bumping up the question "what's the rationale of commenting out `#include <exception>`" ?
It breaks build on the Linux host with the compiler throwing an error about:
```gcc
In file included from /build/assimp-git/src/assimp/include/assimp/GenericProperty.h:52,
from /build/assimp-git/src/assimp/code/Common/BaseProcess.h:46,
from /build/assimp-git/src/assimp/code/Common/BaseProcess.cpp:44:
/build/assimp-git/src/assimp/include/assimp/Importer.hpp:507:16: error: ‘exception_ptr’ in namespace ‘std’ does not name a type; did you mean ‘exception’?
507 |     const std::exception_ptr& GetException() const;
|                ^~~~~~~~~~~~~
|                exception
make[2]: *** [code/CMakeFiles/assimp.dir/build.make:118: code/CMakeFiles/assimp.dir/Common/BaseProcess.cpp.o] Error 1
```

krishty:
This PR is legit. The missing `exception_ptr` error is most likely a fault in the Linux configuration or in libc because `exception_ptr` is mandatory for C++11 and requires `#include <exception>`.

A few observations from [this discussion](https://trac.wxwidgets.org/ticket/16634):
- GCC versions before 4.7 do not support some atomic integer functionality which is required for `exception_ptr`. **Make sure to compile at least with GCC 4.7** or you won’t have full C++11 support.
- Platforms below Intel 486 do not provide that same atomic integer functionality. Hence, you cannot use C++11 if you don’t also specify a target architecture >=486. **Make sure to pass a valid target architecture**, e.g. `-march=core2` (should be hard finding an older CPU nowadays).

krishty:
I’m also totally for approving this pull request. The commented `#include <exception>` may be the side-effect of someone compiling with Visual C++, which is known for pulling in lots of unrelated headers, thus often making explicit `#include`s seem obsolete.

kimkulling (approved):

kimkulling:
Thanks for the PR. I will merge it.

kimkulling:

PR #4101, opened 2021-09-27
moved MD2/MDC tables from BSS to const data

krishty:
Visual C++ was unable to identify them as constant data during optimization. Now they definitely go to the read-only section.


krishty:
closed because duplicate

PR #4108, opened 2021-10-02
Fix aiString length not updated in the EmbedTextures postprocess task

davidepi:
`aiString` length was not updated when the texture path was replaced with the embedded path (i.e. `*0`) after performing the EmbedTextures postprocess task. 

This created all sort of problems when performing string manipulation using the `aiString::length` field instead of interpreting `aiString::data` as c-string (think of applications using UTF8 strings). 

I also wrote a small unit test for this change, but decided against pushing it because I don't know in which file to put it.
Let me know if I should add it to the pr.


kimkulling (approved):

kimkulling:

PR #4109, opened 2021-10-05
Spelling fixes

umlaeute:
Debian packaging found a few spelling mistakes, so I ran codespell on the sources and found a couple more.

There's two more that i haven't fixed, as they seem to be tokens for some model formats:

`transprency` as found in https://github.com/assimp/assimp/blob/babf3b8e3def5097e569c1d30b1b321e981ee07f/code/Pbrt/PbrtExporter.cpp#L617

`shpere` as found in https://github.com/assimp/assimp/blob/master/code/AssetLib/M3D/m3d.h#L615

if these two spellings are correct, i suggest to add a comment (as for `acuracy` in the FBXExporter).
if they are indeed bogus (and those break the import/export, please fix them)


umlaeute:
> That’s inside a comment ...

yea; I see the `#`, but as i don't know anything about pbrt, i couldn' t possibly assume that this is indeed a comment marker.

in any case, do you want me to fix this? or do you want to do it yourself?

krishty:
No problem, I had to look it up myself and thought I’d let you know. Personally, I’d prefer if you did it, because I prefer one large commit from one contributor over a few fragmented ones, but then again I’m not an official maintainer.

We‘re talking about a literal single-character comment fix here, so let’s speed it up: rate this post with thumbs up if you can quickly commit the fix, or with anything else if you have other things to do and I’ll fix it 😁

PR #4110, opened 2021-10-05
Fix a warning about deprecated array comparison

leapach:
I build my projects with "warnings as errors" (using assimp as a submodule in my own CMake based project).
The array comparison for the sheenColorFactor is deprecated in VS2019 and generates a warning which is promoted as an error.
This change simply uses a std::memcmp to compare the float arrays to have an explicit comparison.

Tiny change but would allow me to use vanilla upstream instead of using a patched version or disabling "warnings as errors" ;)


kimkulling:

PR #4111, opened 2021-10-06
Allow the gltf2 Importer to optionally use glTF 2.0 JSON schemas for initial validation

MalcolmTyrrell:
There are JSON schemas for gltf validation being worked on in the glTF repo. See here:
https://github.com/KhronosGroup/glTF/tree/main/specification/2.0/schema

This change allows the glTF2 importer to optionally make use of the those schemas for validation. (In fact, the change is implemented so that other importers which parse json based files could use schemas too.)

Changes:
* A pointer to a rapidjson::IRemoteSchemaDocumentProvider is provided as a property to the Importer
* Importers can query this for the name of specific schemas, and make use of them if they are available
* The logic in glTF2Importer::CanRead is adjusted
* Unit tests are provided which use a simple "fake" schema which checks a single property.

Note:  This change _does not_ add the glTF schemas to this repo, or cause them to be pulled from the source repo. User code is responsible for making the schemas available via the provider.

Here's some (untested) code of a provider which is similar to the provider we're using locally:
```
/// This class makes JSON schemas available to importers.
class SchemaDocumentProvider : public rapidjson::IRemoteSchemaDocumentProvider
{
public:
    SchemaDocumentProvider(std::filesystem::path schemaFolder)
        : m_schemaFolder(schemaFolder)
    {
    }

    /// The returned schemaDocument is owned by this object.
    const rapidjson::SchemaDocument* GetRemoteDocument(const char* uri, rapidjson::SizeType length) override {
        std::unique_ptr<const rapidjson::SchemaDocument>& ptr = m_cache[uri];
        if (ptr) {
            return ptr.get();
        }

        const std::filesystem::path filePath = m_schemaFolder / uri;
        if (!std::filesystem::exists(filePath))
        {
            return nullptr;
        }
        std::ifstream ifs(filePath);
        rapidjson::IStreamWrapper isw(ifs);
        rapidjson::Document schemaDoc;
        schemaDoc.ParseStream(isw);
        ai_assert(!schemaDoc.HasParseError());

        ptr = std::make_unique<const rapidjson::SchemaDocument>(schemaDoc, uri, length, this);
        return ptr.get();
    }

private:
    /// Cache the schemas we've returned.
    std::unordered_map<std::string, std::unique_ptr<const rapidjson::SchemaDocument>> m_cache;
    std::filesystem::path m_schemaFolder;
};
```


kimkulling:
Is this PR ready for a review?

MalcolmTyrrell:
Hi Kim. I believe it is already at quality, but we haven't integrated into the main branch of our product yet. I didn't want to publish it for review until it got it a bit more testing. I was intending this to only take a week or so, but it got delayed. I think I we will move forward with it in the next week.

MalcolmTyrrell:
I think this is ready for review now. Sorry for the draft in the PR list for so long :)

kimkulling (approved):

kimkulling:

PR #4114, opened 2021-10-09
Automatic updates of dependencies

krishty:
Assimp depends on 15 external libraries from the `contrib` directory. These external libraries are, in parts, outdated: `contrib/unzip`, for example, is some 2010 version that is not even available from [the official repository](https://chromium.googlesource.com/external/github.com/nmoinvaz/minizip) any more.

**I see outdated dependencies as a growing security/maintainability concern.**

With this pull request, I introduce automatic dependency updates via github’s Dependabot. (Just the ability to update them. The dependencies are still as close to Assimp’s state as possible.) It will periodically check our dependencies for updates and generate pull requests for them. Since Assimp uses CI, we can very quickly integrate new versions:
image

There are some catches, though, which I’d like to hear your opinion on.

Pro
====
- In the best case, **dependencies can be updated within 24 hours, with one click**. Dependabot will send separate alerts for **security-relevant updates**.
- Assimp now uses the official, trusted repositories to get its dependencies. No more customized versions with hand-crafted code from the Assimp team in them.
- What’s in our repo is now closer to what users get if they use package managers like Hunter.

Contra
====
- Since we no longer have a private copy of each dependency to mess with, **we have to upstream fixes to the respective original projects**. For example: Assimp’s poly2tri library was modified to fix compiler warnings. This fix was never pushed back to poly2tri’s original repository, so we now have compiler warnings again and I had to disable the treat-warnings-as-errors compiler flag. I’ll do my best to upstream as many of our modifications as possible, but the fixes will only be available after we updated to the latest version and such things will keep happening.
- We now have significantly **more files in our repository**. Git submodules do not support mapping just one directory from a repo, so we have to pull it all in. On the other hand, we also get the test suites of our dependencies, which is positive.
- For some dependencies, **I had to change paths**. E.g. our `contrib` directory contained only a small subdirectory of the Google Test reposity. Since git submodules only support pulling in the entire repo (see above), the path changed from `contrib/gtest` to `contrib/googletest/googletest/include/gtest`.
- After cloning the repository, you need to run `git submodule update --init --recursive` to get the dependencies.

Open Questions
====

### zlib
I couldn’t replace our zlib with the official version due to CMake problems. Our version has a heavily modified CMake configuration, and OTOH the [official zlib repo](https://github.com/madler/zlib) has countless open CMake issues. I’m no CMake expert and I’m having a hard time figuring out what to do.

### clipper
The [official repository](https://sourceforge.net/p/polyclipping/code/HEAD/tree/) is not git but SVN. Git submodules do not support SVN repositories. We must either abandon the idea of automatic updates, or mirror the SVN in a git repository to include it.

### Open3DGC
Not urgent because the official repo hasn’t been updated in eight years, but: Our current version looks like someone copied arbitrary files from all over the repo and stuffed them into one directory. This requires more effort, but I’m scared of breaking something.

### Feedback from Users who use package managers
Assimp is already compatible with [Hunter](https://hunter.readthedocs.io/en/latest/). Sadly, Dependabot does not support Hunter, so I had to go to submodules. I’d like to hear about your experience of using Assimp with package management; especially about what problems to avoid! Where does Assimp play well with current versions of dependencies and where not? Etc.


krishty:
Yeah, git submodules do have their caveats (see failing checks).

@turol Any suggestions for such a tool? It should be available as a github action for automation. (Dependabot was just my first choice because it is recommended by github and integrated into the web interface.)

@lerppana Dependabot [does not support vcpkg](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem) and [the feature request is not being updated](https://github.com/dependabot/dependabot-core/issues/2027). Hence, the same question: Any suggestions for an automated github action that supports vcpkg?

krishty:
Rest assured, I’ll definitely keep support for Hunter and all other package managers! I’m not trying to replace other package managers, I just want an automated solution to provide up-to-date packages to users who don’t use package managers.

I also realize that vcpkg is easier to use than submodules, and I’d happily switch if I found any way to automate it in github Actions. There is still hope since both vcpkg and github are Microsoft-owned.

Regarding breaking updates: I have the same fear that we pull in untested code. Dependabot will currently always update to the latest version, not to the latest _tagged_ version. This is indeed unfortunate ([issue](https://github.com/dependabot/dependabot-core/issues/1639)).

Three things come to mind here:
1. A submodule is assigned a specific branch. Two dependencies have dedicated branches for stable versions:
    - [android-cmake](https://github.com/taka-no-me/android-cmake): I picked _master_, not _experimental_
    - [minizip-ng](https://chromium.googlesource.com/external/github.com/nmoinvaz/minizip): _master_ is latest stable version; _dev_ is development (for now, I had to pick *1.1* for compatibility anyway)
2. We could route Dependabot updates towards an internal branch and merge it only to _master_ after sighting by a maintainer. However, this touches the bigger topic of Assimp branch management (X3D support in Assimp’s master is severely broken, shouldn’t we have a development branch anyway?!)
3. Our CI runs on every pull request by Dependabot. I realize this is nothing we should rely on, but it’s better than nothing.

The more I think about it, the more I wish for Assimp to have separate branches for development/release(s).

nightlark:
For zlib, it might be worth taking a look at [zlib-ng](https://github.com/zlib-ng/zlib-ng) as a replacement.

It's supposed to be a modern fork and appears to have:
- [better performance](https://github.com/zlib-ng/zlib-ng/discussions/871)
- better CMake support
- more recent releases

Though that likely comes at a cost of dropping support for legacy compilers/platforms.

krishty:
I agree that zlib is an alternative we should try – just not with this pull request. I don’t know if dropping legacy platform support may become a problem, and I try to break as little as possible here.

jdumas:
Hi there. Rather than submodules, have you considered using CMake [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) for dependency management? Conan/vcpkg/hunter are all very nice, but there require dependency on another package manager. Here are [some examples](https://github.com/adobe/lagrange/tree/main/cmake/recipes/external) of how I am managing external dependencies via FetchContent. Combining this with Hunter is actually pretty straightforward, as long as you use the same target names.

krishty:
Does FetchContent integrate with Dependabot or a similar Github Action? I couldn’t find anything with a quick google search …

The aim of this PR is automated dependency updates here in the repository. Submodules just happen to be the tool that seemed suitable for this because github recommends using Dependabot, and Dependabot’s choice for C++ projects is submodules.

RichardTea:
> I oppose. git submodules are _horrible_.

I'm with Turol here. git submodules have really limited support, and several GUIs are _totally broken_.
For example, the official GitHub Desktop application _will destroy_ your repo as it will randomly commit submodule changes when you change branches.

AMZN-alexpete:
> Does FetchContent integrate with Dependabot or a similar Github Action? I couldn’t find anything with a quick google search …

Though we do not use Dependabot or GitHub Actions, [O3DE uses CMake to download 3rdparty packages](https://github.com/o3de/o3de/blob/development/cmake/3rdPartyPackages.cmake#L122)  given a [URL and a checksum](https://github.com/o3de/o3de/blob/development/cmake/3rdParty/Platform/Windows/BuiltInPackages_windows.cmake).    If you do end up looking into using CMake to download dependencies as ZIP files from a URL, that might be a useful resource.

jcfr:
To follow up on @AMZN-alexpete comment regarding used of the `file(DOWNLOAD ...)` command, using the [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) introduced in CMake 3.11 would allow to integrate third-party projects specifying a git repository and git sha (or a download URL &  checksum) without having to maintain additional custom CMake code.

The file `contrib/CMakeLists.txt` would be similar to this:

```cmake
include(FetchContent)

FetchContent_Populate(poly2tri
  GIT_REPOSITORY https://github.com/jhasse/poly2tri.git
  GIT_TAG        f72dce22bd25d10f8575ec54cf51bf5a5fdcb5ba
  )
add_subdirectory(${poly2tri_SOURCE_DIR} ${poly2tri_BINARY_DIR})

...
```

jcfr:
Last, if some of the "vendorized" dependencies are not build statically, the corresponding symbols would have to be mangled to avoid symbol clashes when integrating `Assimp` in client projects.

jcfr:
> > Does FetchContent integrate with Dependabot or a similar Github Action?

> Very good question, I look into this and follow up.

Short answer would be `No`. And currently `dependabot-core` does not accept contribution to support new ecosystems. See https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md#contributing-new-ecosystems

PR #4146, opened 2021-10-29
Fix fuzzer crashes

alpire:
This PR addresses [all but one of the OSS-Fuzz issues](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=proj&q=Type%3DBug-Security%20label%3ADeadline-Exceeded%20-label%3AStability-UndefinedBehaviorSanitizer%20label%3Areproducible%20proj%3Aassimp&can=2) marked as security relevant.

Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33238
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24553 (Fix #3425)
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24465 (Fix #3348)
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31795
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24463 (Fix #3429)
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36594
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25401 (Fix #4096)
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24502 (Fix #4098)
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25341 (Fix #4097)
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34416

PR #4147, opened 2021-10-30
Fix out-of-bounds read in RemoveLineComments

alpire:
Follow up to #4146, which was not sufficient to fix one of the bugs.

Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24553


kimkulling (approved):

kimkulling:

PR #4163, opened 2021-11-09
Better aiMesh ABI compatibility with 5.0.1, make smaller

### LINKED ISSUE #4162: Change aiMesh.mTextureCoordsNames field to a pointer and move its declaration after mMethod field (proposal)
First, I am a C# developer so my terminology may not be the most accurate for C / C++.

**Is your feature request related to a problem? Please describe.**

1. 
On May 3 2021 the Mesh.h file was changed by adding mTextureCoordsNames (https://github.com/assimp/assimp/commit/f91b439f79cc106f08678a265562191c67b184cf). Because the field was added in the middle of the Mesh struct, the offset addresses of all the fields after this field were changed. Because this includes some of the core mesh fields (mFaces, mBones, mMaterialIndex, etc.) the C# wrappers (and probably other wrappers too) crashed when they wanted to read the content of the moved fields.

Therefore I would ask you to move the mTextureCoordsNames field after the mMethod or mAABB fields.

This would preserve the old field offsets and will make all the old wrappers work as before.
This should not break any C / C++ library because they get struct definitions from header files and therefore they will be immediately able to use the new layout.

2.
The aiMesh.mTextureCoordsNames field is defined as `C_STRUCT aiString mTextureCoordsNames[AI_MAX_NUMBER_OF_TEXTURECOORDS];` - a fixed array of 8 aiString structs. Each aiString struct is defined by one int (string's Length) and fixed 1024 bytes for the string content. This means that this field takes 8 kb of memory for each Mesh. The content of this array is set only by the FBX importer and only when the mesh defines the texture coordinates. For all other file formats, that 8 kb per Mesh are taken but not used. Also, most of the meshes that define texture coordinates define them in only one channel so only the first aiString is usually used. 

Therefore I propose that the mTextureCoordsNames is changed into a pointer to an array of aiStrings, or even better to a pointer to pointers of aiString. This way the memory will be taken only when the content of mTextureCoordsNames is actually defined.

I understand that this point is a breaking change for all the users that read the mTextureCoordsNames and that I am late with this proposal.

Additional notes:
The mTextureCoordsNames is defined in Mesh.h. 
It is used only in FBXConverter.cpp line 1131 and in AssxmlFileWriter.cpp line 604.

RichardTea:
Move new mTextureCoordsNames member to end of struct
Convert to pointer-to-array, saving ~8KB per aiMesh in almost all cases
Add C++ accessor functions for simpler usage

Fixes #4162


abenedik:
Thank you very much for this PR! I have tested it and it works correctly with an old Assimp C# wrapper (this wrapper crashes when using the Assimp 5.1.rc1 version without this PR).

kimkulling (approved):

kimkulling:
Merged, thanks a lot for yon contribution.

PR #4164, opened 2021-11-09
Remove optimization fence preventing automatic move

Skylion007:
I ran clang-tidy on the most recent RC, and found a performance bug. There was a const qualifier on a return value which prevented the compiler from doing an automatic move of the shared_ptr return value.


kimkulling (approved):

kimkulling:
Cool, nice finding. I will merge it.

kimkulling:

PR #4165, opened 2021-11-09
Added UE4 plugin

irajsb:
This is most complete UE assimp plugin that currently wraps  90% of the assimp code to be used in unreal engine blueprints and c++ . I'll be honored if you guys mention this plugin  in readme so we can get more contributions .


kimkulling (approved):

kimkulling:

PR #4167, opened 2021-11-10
Update anim.h

### LINKED ISSUE #4158: Bug: null check error
**Describe the bug**

The variable mArmature, mNode is not initialized, so that it occur null check error. In this regard, I think it seems that other processing is needed that is copy and delete

struct aiBone in mesh.h

kimkulling:
- Add missing whitespaces
- closes #4158

### DIFF
--- a/include/assimp/anim.h
+++ b/include/assimp/anim.h
@@ -98,6 +98,7 @@ struct aiVectorKey {
     bool operator<(const aiVectorKey &rhs) const {
         return mTime < rhs.mTime;
     }
+    
     bool operator>(const aiVectorKey &rhs) const {
         return mTime > rhs.mTime;
     }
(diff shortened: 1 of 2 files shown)


RichardTea:
I don't think that copy constructor is correct. Surely it should be copying the pointers to the armature and node?

kimkulling:
I do not know honestly speaking. When copying a node also all its data will be copied and this is a bad situation for our scenegraph concept. All the nodes will get invalid because they are not unique any more.
So I have decided to avoid the copy-step for armature. When I remove the copy constructor completly I guess the users will kill me.

Other opinions?

RichardTea:
So I guess this change does at least mean you know that SceneCombiner broke the copy's armatures.
That's a lot better than the pointers-to-nowhere it did before.

SceneCombiner's copy function needs fixing up, but that's a much bigger task than is sane at this point.

I guess that it is better for it to actively break the links than to leave them pointing at the original aiScene.

kimkulling:
Now I know it. I will create an issue for that problem. Thanks for the hint!

PR #4186, opened 2021-11-17
Fix CanRead() to return true on success path

weitjong:
One of our user has submitted this bug fix to our project. We think the fix should be sent to upstream as well.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Importer.cpp
+++ b/code/AssetLib/glTF2/glTF2Importer.cpp
@@ -122,7 +122,7 @@ bool glTF2Importer::CanRead(const std::string &pFile, IOSystem *pIOHandler, bool
 		return asset.CanRead(pFile, extension == "glb");
 	}
 
-    return false;
+    return true;
 }
 
 static aiTextureMapMode ConvertWrappingMode(SamplerWrap gltfWrapMode) {


RichardTea:
That's a failure path.

If there is no pIOHandler then it definitely cannot read as there is nowhere to read the file content from.

weitjong:
Hey, I think I know why we have come to different conclusion on the final code path.

This was from v4.1.0 + the patch. 
```
bool glTF2Importer::CanRead(const std::string& pFile, IOSystem* pIOHandler, bool checkSig) const
{
    const std::string &extension = GetExtension(pFile);

    if (extension != "gltf" && extension != "glb")
        return false;

    if (checkSig && pIOHandler) {
        glTF2::Asset asset(pIOHandler);
        try {
            asset.Load(pFile, extension == "glb");
            std::string version = asset.asset.version;
            return !version.empty() && version[0] == '2';
        } catch (...) {
            return false;
        }
    }

    return true;
}
```

I have used the git diff and git apply today to apply our fix to the current master. I only saw one line has changed as the result and thought the patch has been applied cleanly and all is good before submitting the PR. I didn't realize the code in master has in fact already changed to:
```
bool glTF2Importer::CanRead(const std::string &pFile, IOSystem *pIOHandler, bool checkSig ) const {
	const std::string &extension = GetExtension(pFile);
	if (!checkSig && (extension != "gltf") && (extension != "glb"))
		return false;
	if (pIOHandler) {
		glTF2::Asset asset(pIOHandler);
		return asset.CanRead(pFile, extension == "glb");
	}

    return false;
}
```

So, in this case the last return may be the failure path as you suggested. At this point, I am not sure which one is right anymore.

RichardTea:
I believe this is already correct in current master.
There has been a lot of work on the glTF importers over the last 18 months or so.

It would be great if you could submit a PR for a test case against the issue your user found, just to be sure.

weitjong:
In any case I think I should close this PR as the patch should not be applied as it is on the latest version of code in the master branch. The change in the conditional check since the v4.1.0 (that we have) is crucially changing the meaning of the function parameters.

PR #4187, opened 2021-11-17
Collada: Read value, not attribute

### LINKED ISSUE #4179: Bug: Collada parsing partly broken since v5.1.0
On v5.1.0 (and probably earlier) some Collada parts aren't parsed correctly, mainly lights, samplers, and some material properties.

There's a lot of code that does something like this:

```c++
if (currentName == "constant_attenuation") {
    XmlParser::getRealAttribute(currentNode, "constant_attenuation", pLight.mAttConstant);
}
```

which tries to read a *child attribute*, not the current attribute. This fails and the value is never written.

This happens in several places (non-exhaustive list):

https://github.com/assimp/assimp/blob/05115b07b4d7ca12a52b9fd0e0534029ac8f2cb6/code/AssetLib/Collada/ColladaParser.cpp#L951
https://github.com/assimp/assimp/blob/05115b07b4d7ca12a52b9fd0e0534029ac8f2cb6/code/AssetLib/Collada/ColladaParser.cpp#L1111
https://github.com/assimp/assimp/blob/05115b07b4d7ca12a52b9fd0e0534029ac8f2cb6/code/AssetLib/Collada/ColladaParser.cpp#L1144

Seems to have been changed in [this commit](https://github.com/assimp/assimp/commit/d393f677ce8af2343260fed3231222170b511300)

**To Reproduce**
1. Import https://github.com/assimp/assimp/blob/master/test/models/Collada/lights.dae
2. Attenuation values for "Point-light" are 1, 0, 0 instead of 0.1, 0.7, 0.9

**Expected behavior**
Same behaviour as 5.0.1

RichardTea:
Some less-common Collada properties were misconverted when changing from IrrXML to pugixml.
These are values, not attributes.

Fixes #4179

### DIFF
--- a/code/AssetLib/Collada/ColladaParser.cpp
+++ b/code/AssetLib/Collada/ColladaParser.cpp
@@ -924,6 +924,8 @@ void ColladaParser::ReadMaterial(XmlNode &node, Collada::Material &pMaterial) {
 void ColladaParser::ReadLight(XmlNode &node, Collada::Light &pLight) {
     XmlNodeIterator xmlIt(node, XmlNodeIterator::PreOrderMode);
     XmlNode currentNode;
+    // TODO: Check the current technique and skip over unsupported extra techniques
+
     while (xmlIt.getNext(currentNode)) {
         const std::string &currentName = currentNode.name();
         if (currentName == "spot") {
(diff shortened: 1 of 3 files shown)


kimkulling:

PR #4193, opened 2021-11-20
Update blender importer to work with Blender 2.8+ files

### LINKED ISSUE #3316: Expected at least one object with no parent - blend importer
When I trying to import my model with 2 animations I get
`Assimp error: BLEND: Expected at least one object with no parent`

If I convert `blend` to `fbx` via Blender, then Assimp imports `fbx` model without any errors

**Steps to reproduce the behaviour:**
1. Download the attached `blend` model
2. Try to import it
3. See error

**Expected behaviour**
Import model without any error

**Additional info**
Blender 2.83 (both Linux and Windows)
Assimp 5.0.1

Models: https://drive.google.com/drive/folders/1YvkMsG9NdMnRNyLhXpeifyndIs5DZSVS?usp=sharing

PencilAmazing:
Updates the blender importer to be aware of newly added collections and try to keep behavior as close to previous as possible. Closes #3316.
The way this works is that it reads the top level collection as root, and interprets each collection as an extra node.
I'm not exactly proud of this, but I think it does the job and does pass all tests on my machine.


kimkulling (approved):

PencilAmazing:
Took me a while before I realized that LeakSanitizer doesn't work on win32 and had to learn to use WSL. Also I think on windows master branch has memory leaks? Running `_CrtDumpMemoryLeaks()` always returns a lot but I've never got symbolizing to work with that.
Anyway, here's me hoping this fixes the tests.

kimkulling:

ruby0x1:
Thanks for this, been waiting eagerly for this. I'll be sure to report any issues!

PR #4204, opened 2021-11-23
Allow dlclose of so library by avoiding unique symbols.

TThulesen:
The libassimp.so library contains one symbol of type STB_GNU_UNIQUE, meaning that the library can not be unloaded with dlclose after it has been loaded with dlopen. If a plugin links with the assimp library, this plugin will not be able to unload either.

The following symbol is the problem:
nm -DC /lib/x86_64-linux-gnu/libassimp.so.5.0.0 | grep " u "
0000000000c40f20 u Assimp::IOSystem::CurrentDirectory[abi:cxx11]() const::Dummy

A simple fix will be to remove the inlining of this function.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -174,6 +174,7 @@ SET( Common_SRCS
   Common/ImporterRegistry.cpp
   Common/DefaultProgressHandler.h
   Common/DefaultIOStream.cpp
+  Common/IOSystem.cpp
   Common/DefaultIOSystem.cpp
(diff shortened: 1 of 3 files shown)


RichardTea:
Note: This breaks binary compatibility on all platforms. May also affect language wrappers, unsure what they do for inlined functions.

kimkulling:
So we need to merge this later on.

kimkulling:
OK, lets increase the minor release number and try to get lot of feature into it.

PR #4211, opened 2021-11-26
Added checks for out of bounds data access/writing

ms-maxvollmer:
Three fixes:
* Check that positions exist in a mesh before accessing them
* Reject files with an invalid byteLength value
* Ensure we don't access the vector in a Ref object with an out of bounds index


kimkulling:

Issue #4177, opened 2021-11-13
Bug: X3D importer CanRead returns true for virtually all files


pezcode:
`X3DImporter::CanRead` uses `pFile.find_last_of(".x3d")` which returns an offset for if *any* of those characters is in the string. This means virtually all files with a dot in the name and all memory files (`$$$___magic___$$$.<hint>`) are recognized as X3D.

This was changed in https://github.com/assimp/assimp/pull/2966. The previous behaviour looks to be correct.

This by itself wouldn't be quite that tragic, but invalid X3D files lead to an assert in `X3DImporter::InternReadFile` because `.front()` is called on an empty `NodeElement_List`.

umlaeute:
as mentioned in #4201, the `CanRead` method if thee X3DImporter does a bogus file extension check.
instead it should look like:

~~~C++
bool X3DImporter::CanRead(const std::string &pFile, IOSystem * /*pIOHandler*/, bool checkSig) const {
    if (checkSig) {
        std::string::size_type pos = pFile.rfind(".");
        if (pos != std::string::npos) {
            std::string ext=pFile.substr(pos+1);
            if ("x3d" == ext)
                return true;
        }
    }

    return false;
}
~~~

while this makes the X3DImporter at least reject files that don't have the `.x3d` extension, it still crashes on the X3D-models in the test/ folder

PR #4217, opened 2021-12-01
Fix file-extension check for X3D-files

umlaeute:
using the pre-existing and well-tested GetExtension() (which happens to also normalize the extension), rather than attempting a new buggy implementation of the same thing...

Closes: https://github.com/assimp/assimp/issues/4177

### DIFF
--- a/code/AssetLib/X3D/X3DImporter.cpp
+++ b/code/AssetLib/X3D/X3DImporter.cpp
@@ -235,10 +235,8 @@ void X3DImporter::ParseFile(const std::string &file, IOSystem *pIOHandler) {
 
 bool X3DImporter::CanRead(const std::string &pFile, IOSystem * /*pIOHandler*/, bool checkSig) const {
     if (checkSig) {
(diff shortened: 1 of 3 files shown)


umlaeute:
i've also added a bit of code that prevents crashes when reading X3D files like the `ComputerKeyboard.x3d`.
i don't know whether the fix is correct though.

in any case, the model read by assimp differs significantly from what e.g. `view3dscene` would display (when displaying the `EscapeButton` group)

kimkulling (approved):

kimkulling:
Thanks for that. Special for the keyboard model I wasn't able to find a fix.

umlaeute:
> Special for the keyboard model I wasn't able to find a fix.

well, as i said: it's not really a fix, it just stopped crashing.
afaict the current code only extracts 3 keyboard keys instead of 98, and one of them is totally off...


with assimp-5.0:
assimp-5.0


with assimp-5.1:
assimp5.1


but i think not crashing is a net win :-)

Issue #4224, opened 2021-12-07
 Bug: Importer crashes on one of the MDL models


umlaeute:
With assimp-5.1.2, i noticed that assimp also crashes on one of the MDL example models:

~~~
$ gdb --args ./bin/assimp info ../test/models/MDL/MDL\ \(HL1\)/manT.mdl
(gdb) run
Starting program: @PATH@/build/bin/assimp info ../test/models/MDL/MDL\ \(HL1\)/manT.mdl
Launching asset import ...           OK
Validating postprocessing flags ...  OK
ai_assert failure in @PATH@/code/PostProcessing/RemoveRedundantMaterials.cpp(175): iNewNum > 0

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
49      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007ffff6c05536 in __GI_abort () at abort.c:79
#2  0x00007ffff74ccd31 in Assimp::defaultAiAssertHandler (failedExpression=0x7ffff7b2e804 "iNewNum > 0", file=0x7ffff7b2e7b0 "@PATH@/code/PostProcessing/RemoveRedundantMaterials.cpp", line=175)
    at @PATH@/code/Common/AssertHandler.cpp:56
#3  0x00007ffff74ccd75 in Assimp::aiAssertViolation (failedExpression=0x7ffff7b2e804 "iNewNum > 0", file=0x7ffff7b2e7b0 "@PATH@/code/PostProcessing/RemoveRedundantMaterials.cpp", line=175)
    at @PATH@/code/Common/AssertHandler.cpp:71
#4  0x00007ffff74f2937 in Assimp::RemoveRedundantMatsProcess::Execute (this=0x555555598d10, pScene=0x55555559a2a0) at @PATH@/code/PostProcessing/RemoveRedundantMaterials.cpp:175
#5  0x00007ffff7487f2e in Assimp::BaseProcess::ExecuteOnScene (this=0x555555598d10, pImp=0x7fffffffdec0) at @PATH@/code/Common/BaseProcess.cpp:78
[...]

Issue #4209, opened 2021-11-25
Bug: LWOImporter crashes on one of the test-files


umlaeute:
after the discussion in #4201 and #4177, i disabled the `X3DImporter`, only to discover that assimp also crashes on one of the LWO examples:

~~~
$ gdb --args ./assimp info ../../test/models/LWO/LWO2/UglyVertexColors.lwo
(gdb) run
Starting program: @PATH@/assimp/obj-x86_64-linux-gnu/bin/assimp info ../../test/models/LWO/LWO2/UglyVertexColors.lwo
Launching asset import ...           OK
Validating postprocessing flags ...  OK

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff74f9d97 in Assimp::LWOImporter::InternReadFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, aiScene*, Assimp::IOSystem*) ()
   from @PATH@/assimp/obj-x86_64-linux-gnu/bin/libassimp.so.5
(gdb) bt
#0  0x00007ffff74f9d97 in Assimp::LWOImporter::InternReadFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, aiScene*, Assimp::IOSystem*) ()
   from @PATH@/assimp/obj-x86_64-linux-gnu/bin/libassimp.so.5
#1  0x00007ffff73207e0 in Assimp::BaseImporter::ReadFile(Assimp::Importer*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Assimp::IOSystem*) ()
   from @PATH@/assimp/obj-x86_64-linux-gnu/bin/libassimp.so.5
#2  0x00007ffff7336e8c in Assimp::Importer::ReadFile(char const*, unsigned int) () from @PATH@/assimp/obj-x86_64-linux-gnu/bin/libassimp.so.5
#3  0x0000555555574be6 in Assimp::Importer::ReadFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int) ()
#4  0x0000555555573656 in ImportModel(ImportData const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
#5  0x0000555555577f10 in Assimp_Info(char const* const*, unsigned int) ()
#6  0x00005555555733a1 in main ()
(gdb) 
~~~

other LWO-files seem to work fine:
~~~sh
[...]

PR #4226, opened 2021-12-07
Bugfix/import crashes

umlaeute:
this fixes a few crashes when running `assimp info` on the test-models.

notably:
- `LWOImporter`: it seems there was a typo, swapping `j` for `i` (loop-vars rock!), which would create an out-of-bounds access on a vector
  - Closes: #4209 
- `RemoveRedundantMatsProcess`: this crashed on an assertion (that there are actually materials left after removing all the redundant ones). i replaced the assertion by a throwing a `DeadlyImportError`
  - the documentation of BaseProcess says that the *process()* method should raise an `ImportErrorException`: https://github.com/assimp/assimp/blob/70f5cca9c3ebd84b89c5775852ac9aa2449f6c5b/code/Common/BaseProcess.h#L217-L224
    however, there is no such exception and other postprocessors used the `DeadlyImportError`, so i used that one as well (probably the docs should be updated)
    - Closes: #4224 
    - Closes: #4225


kimkulling (approved):

kimkulling:
Great work, thanks a lot for your contribution.

PR #4258, opened 2021-12-15
Fix duplicate call to Export function

sherpya:
In the exporter the `exp.mExportFunction` is called twice, the indentation it's also broken


RichardTea:
This was fixed in 2019 and is in the 5.1.0 release.
The 5.0 release branch is outdated now.

kimkulling:
So thanks for the PR, but it is outdated. So I will close it.

sherpya:
@kimkulling I've seen some sort of activity on that branch so I tried :)

PR #4259, opened 2021-12-16
fix fbx import metalness

VyacheslavVanin:
Can't import metalness textures in fbx importer.
$raw.ReflectionFactor material property was not available via aiGetMaterialTexture(...) with aiTextureType_METALNESS


kimkulling (approved):

kimkulling:

PR #4264, opened 2021-12-18
fix test/models/3DS/IMAGE1.bmp: is jpg

Garux:


kimkulling:

PR #4265, opened 2021-12-18
add Inter-Quake Model (IQM) Importer

Garux:
Basic import support here. Possible imporvements:
* iqe ASCII flavor support (intended for export only however)
* import animations (srsly no clue on this topic)
* figure out what are 4-component tangents in this format (tangents w/o bitangents are not enough)
* support version 1 format
* support all possible custom vertex array data types (only trivial compatible ones atm)


Garux:
Am i supposed to fix all these MSVC warnings?
While latter are question of style, the first `'block': unreferenced formal parameter` is questionable.
It's really unused in LE code, good solution would be https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.
And after all these are fine diagnostic warnings, but not pointing to actual errors.

RichardTea:
Yes, assimp needs to build without warnings.

There's very few big-endian targets left in common usage - x86/amd64 and ARM are all LE - so the whole function should be elided for almost all build environments.

I'd suggest removing the swap_block() function entirely and just doing the endian conversion in the four(?) places the endian conversion would be needed.
It's not immediately clear what swap_block() is doing, so fairly likely someone will make a mistake in future maintenance.

Garux:
Fixed these warnings.
However the fact, that only MSVC emits these warnings, looks like inconsistency. May be they must be disabled for MSVC or enabled for others?

kimkulling:
Looks good, CI is on its way. Sorry for the delay!

kimkulling:
Merged, thanks a lot for this great contribution.

RichardTea:
@Garux 
Excellent, thank you!
> Fixed these warnings. However the fact, that only MSVC emits these warnings, looks like inconsistency. May be they must be disabled for MSVC or enabled for others?

In my experience the three major compilers emit slightly different sets of warnings, usually all valid.
I assume a lot of it is due to the order of parsing and code generation in the toolchains - in this case you won't get a warning if the unused variable is passed onto the function, but you will if the function was already inlined by a previous step.

Issue #963, opened 2016-08-05
Feature-request: Command line export progress


deargonaut:
_What a wonderful program you've created!!_

What I am missing though (except the lack of SolidWorks import possibility (#830)), is the possibility to view/read the progress of converting a model in command line.

As I am using the command-line for convert models to STL through a website this would be an awesome feature.

Thanks in advance!

kimkulling:
Good idea, let's put this on trello!

deargonaut <notifications@github.com> schrieb am Fr., 5. Aug. 2016, 10:13:

PR #4293, opened 2021-12-26
Add console progresshandler

kimkulling:
- Commandline-tool assimp will show a progress in the console.
- closes https://github.com/assimp/assimp/issues/963


tools/assimp_cmd/Main.cpp
```diff
@@ -286,6 +303,9 @@ const aiScene* ImportModel(
 
 	// do the actual import, measure time
 	const clock_t first = clock();
+    ConsoleProgressHandler *ph = new ConsoleProgressHandler;
```
SirLynix, line 306: Why use new instead of putting it on the stack?

PR #4311, opened 2021-12-28
CMake: Replace CMAKE_COMPILER_IS_MINGW by MINGW

SirLynix:
Assimp can't detect MinGW with CMake >=3.21 because [CMAKE_COMPILER_IS_MINGW got removed](https://discourse.cmake.org/t/cmake-compiler-is-mingw-removed/4539), thus won't set some flags such as `-Wa,-mbig-obj` which are required for compilation.

This commit replaces CMAKE_COMPILER_IS_MINGW with [MINGW](https://cmake.org/cmake/help/latest/variable/MINGW.html)  variable (which appeared in CMake 3.2, so this should be fine).

Please note I haven't tested if [that android hack](https://github.com/assimp/assimp/blob/2356d9cf024e74de511e74e894203217b8a2b3fb/contrib/android-cmake/android.toolchain.cmake#L1617) for MinGW still works.


SirLynix:
Has been tested here: https://github.com/xmake-io/xmake-repo/runs/4650952688

kimkulling:
Look good. I will merge it when the CI-build is ready.

kimkulling:

PR #4318, opened 2022-01-03
Fix bone fitted check in gltf2 exporter

vpzomtrrfrt:
I'm not entirely sure what's going on here but this seems to fix a crash I encountered with some models


inhosens:
Looks good to me.

kimkulling (approved):

kimkulling:
The fix looks useful to me. In case of an error, we will return -1 so checking for this makes much more sense.

Thanks a lot for your contribution.

PR #4323, opened 2022-01-05
Locale independent meter scale

tanolino:
`XmlParser::getRealAttribute(...)` will call `strtod` (or `wcstod`) which are both locale dependent. So on a German locale system a scale of 0.01 meter will be parsed to 0. In order to avoid that I use the `fast_atoreal_move<ai_real>()` method.


kimkulling (approved):

kimkulling:

Issue #4329, opened 2022-01-09
5.1.5 Android Compile Targets fail with 32bit - x86 / armv7 due to zip typedef redefinition 


danoli3:
**Describe the technical debt**
Failing to compile in armv7 / x86 
Working in x86_64 / arm64-v8

Similar issue of the past:
https://github.com/assimp/assimp/issues/2954

**Desktop (please complete the following information):**
 - OS Linux / macOS - compiling Android
 - Version 5.1.5
 - Android NDK 23.1

**Additional context**
**contrib/unzip/crypt.h:20:23: error: typedef redefinition with different types ('unsigned long' vs 'unsigned int')**

```
Building C object code/CMakeFiles/assimp.dir/__/contrib/unzip/crypt.c.o
591
[...]

PR #4330, opened 2022-01-09
Assimp Patch Android LTS NDK 23 Fix

danoli3:
**Fixes**: https://github.com/assimp/assimp/issues/4329
**Error**: Compile targets armv7, x86 fail in NDK 23+ 

**Issue**: redefinition in Assimp zlib unzip for Android NDK causing typedef redefinition on 32bit compile targets for Assimp due to a conflicting in NDK which has a more updated version of zlib 

**Solution**: Using the PreProcessor define setup in newer Zlib which can be used to prevent issue

**Code change**: Minimal:
```
#if ZLIB_VERNUM < 0x1270
#if !defined(Z_U4)
typedef unsigned long z_crc_t;
#endif
#endif
```


danoli3:
@kimkulling can we get the tests running

kimkulling (approved):

kimkulling:
Thanks for the fix, makes sense to me. CI is on its way!

PR #4336, opened 2022-01-10
Move Base64 encode/decode functionality to the common folder

kovacsv:
Later I'd like to use base64 encoding in json export, because at the moment it's really slow to write textures byte-by-byte. So I decided to move Base64 encode/decode from the glTF-related code to the common folder.


kimkulling (approved):

kimkulling:
I just have changed some minor formattings and the copyrights (we are in 2022 right now).

Thanks for the PR. I will merge it!

PR #4337, opened 2022-01-10
Fix imported target include directory

lopsided98:
Fix the imported target include directory when the headers are not installed in `${CMAKE_INSTALL_PREFIX}/include`. Without this patch, the CMake config script ends up referring to non-existent paths.


kimkulling (approved):

kimkulling:

PR #3363, opened 2020-08-07
Fixed runtime output directory overridden.

malortie:
See issues https://github.com/assimp/assimp/issues/3239 and https://github.com/assimp/assimp/issues/3006. See also this PR https://github.com/assimp/assimp/pull/3013.

I introduced 3 new variables:

```
ASSIMP_RUNTIME_OUTPUT_DIRECTORY => assigned to the respective 'bin' folder.
ASSIMP_LIBRARY_OUTPUT_DIRECTORY => assigned to the respective 'bin' folder.
ASSIMP_ARCHIVE_OUTPUT_DIRECTORY => assigned to the respective 'lib' folder.
```

A macro `TARGET_USE_COMMON_OUTPUT_DIRECTORY` has been made to automatically set target output directory to use the 3 new variables as output directory.

I also removed the conditional `IF ( CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR )`, as the 3 new variables will be automatically assigned to a proper bin/lib directory.

The changes have been tested on x86 and x64 in Debug and Release configurations.

The following projects were tested:

- assimp
- assimp_cmd
- assimp_simpleogl
- assimp_simpletextureddirectx11
- assimp_simpletexturedogl
- assimp_viewer
- unit

All tests were run.

PR #4338, opened 2022-01-11
CMAKE: Respect top-level CMAKE_*_OUTPUT_DIRECTORY variables

### LINKED ISSUE #3239: CMakeListst.txt: Runtime output directory overridden
Hello,

I am building assimp over my own project. Today after updating the sources (master branch) I had a problem after compiling the shared library. Compilation finished successfully but the compiled library was missing in the output folder. After some research I found the problem in the main CMakeLists.txt:

\# If this is an in-source build (CMAKE_SOURCE_DIR == CMAKE_BINARY_DIR),
\# write the library/executable files to the respective directories in the
\# source tree. During an out-of-source build, however, do not litter this
\# directory, since that is probably what the user wanted to avoid.
IF ( CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR )
  SET( CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_HOME_DIRECTORY}/bin" )
  SET( CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_HOME_DIRECTORY}/lib" )
  SET( CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_HOME_DIRECTORY}/bin" )
ELSE()
  SET(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/lib")
  SET(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/bin")
  SET(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/bin")
ENDIF ()

The 'ELSE()'-block overriddes my global project configuration and the library is put under '"${CMAKE_CURRENT_BINARY_DIR}/bin'.

As workaround I tried disabling the 'ELSE()'-block ... result: -> It works as expected.

I don't know the background why this changes where made, neither I don't know if they are needed by critical parts. Would it be possible to remove the 'ELSE()'-block?

leonvictor:
This is a follow up to #3363 which only partially fixed #3239. In cases where assimp is included as a module in another project and included via CMake's `add_subdirectory()`, the top level build should be able to specify where assimp targets files are built.

This PR changes the initialization of assimp's `RUNTIME_OUTPUT_DIRECTORY`, `LIBRARY_OUTPUT_DIRECTORY` and `ARCHIVE_OUTPUT_DIRECTORY` properties, and set them to the value of  `CMAKE_RUNTIME_OUTPUT_DIRECTORY` and other counterparts variables, if they are set.

This is coherent with CMake's own [documentation](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html)

This only happens if theses variable are set in a upper-level CMakeFiles.txt so any other include methods should stay as they were

I've also noticed the option `ASSIMP_INSTALL` is defined as "Disable this if you want to use assimp as a submodule.". I'm not sure this is exactly related, as right now it looks like it's only associated to packaging behavior, but it could be used to condition the initialization. Any insight on this is welcome !


kimkulling (approved):

kimkulling:

PR #4339, opened 2022-01-12
Collada: Read all instance_material child nodes

jsigrist:
The Collada spec allows multiple `instance_material` child nodes under the `instance_geometry` tag. There's an example of this in the collada [spec](https://www.khronos.org/files/collada_spec_1_5.pdf) on page 8-69.

Right now the Collada parser only reads the first `instance_material` node, which can cause some materials to be skipped when reading a collada file. This PR checks all the child nodes for the `instance_material` label and adds them to the map of parsed materials.


kimkulling (approved):

kimkulling:
You need to merge master into this branch, please. Looks fine, I will merge it after that!

kimkulling:

PR #4342, opened 2022-01-16
Krishty new file detection

kimkulling:
- Make file-detection much easier.
- Based on @krishty branch
- Check https://github.com/assimp/assimp/pull/3801 for more detailed information


kimkulling:
You are right, I missed an important part of your work. My understanding was that you have switched off some importers which worked before.

I guess I was wrong, right?

krishty:
Yeah, please remove it so we have just one mechanism to maintain!

But in order to do so, you have to fix the two glTF unit tests as described here: https://github.com/assimp/assimp/pull/3801#issuecomment-832302128

The fix would be making glTF’s `CanRead()` much simpler. Instead of loading and validating the entire file(!), it should just check quickly whether the file uses the glTF schema. Then it can succeed the two unit tests saying *“this is a glTF file … but I cannot load it completely because a texture is missing“* instead of denying to read the file in the first place because of `CanRead()` throwing an exception.

kimkulling:
(Not to worry, I have the same issues with my job!)

I will try to find the time for this! Thanks for the hint!

Issue #4358, opened 2022-01-26
5.2.0 build fails on big endian archs


krop:
The assimp 5.2.0 build fails on both ppc64 (be) and s390x:

```
[  794s] [ 24%] Building CXX object code/CMakeFiles/assimp.dir/AssetLib/IQM/IQMImporter.cpp.o
[...]

PR #4360, opened 2022-01-26
Update IQMImporter.cpp

kimkulling:
- Add missing namespace
- closes https://github.com/assimp/assimp/issues/4358


umlaeute:
an alternative fix for the problem can be found in my PR #4367

PR #4367, opened 2022-01-31
use fully qualified namespace in byteswap macros

umlaeute:
the `AI_SWAP4` macros and friends assume that they are called in a context that resolves the `Assimp` namespace.

rather than adding an explicit namespace to the macro-invocation whereever it is used (as suggested in #4360), this PR modifies the macros themselves so they are expanded with the full namespace.

Closes: https://github.com/assimp/assimp/issues/4358


kimkulling:
Thanks a lot for your fix. I was not able to find the right way to fix that issue.

kimkulling (approved):

kimkulling:

PR #4381, opened 2022-02-08
Kimkulling/introduce compression

kimkulling:
- Introduce block-wise decompression in Compression
- Adapt more importer
- Fix X-Importer decompress

### DIFF
--- a/code/AssetLib/X/XFileParser.cpp
+++ b/code/AssetLib/X/XFileParser.cpp
@@ -5,8 +5,6 @@ Open Asset Import Library (assimp)
 
 Copyright (c) 2006-2022, assimp team
 
-
-
 All rights reserved.
 
 Redistribution and use of this software in source and binary forms,
@@ -60,26 +58,11 @@ using namespace Assimp::Formatter;
 
 #ifndef ASSIMP_BUILD_NO_COMPRESSED_X
 
-/* #ifdef ASSIMP_BUILD_NO_OWN_ZLIB
-#include <zlib.h>
-#else
-#include "../contrib/zlib/zlib.h"
-#endif*/
 #include "Common/Compression.h"
 
 // Magic identifier for MSZIP compressed data
-#define MSZIP_MAGIC 0x4B43
-#define MSZIP_BLOCK 32786
-
-// ------------------------------------------------------------------------------------------------
-// Dummy memory wrappers for use with zlib
-/* static void *dummy_alloc(void * opaque, unsigned int items, unsigned int size) {
-    return ::operator new(items *size);
-}
-
-static void dummy_free(void * opaque, void *address) {
-    return ::operator delete(address);
-}*/
+constexpr unsigned int MSZIP_MAGIC = 0x4B43;
+constexpr size_t MSZIP_BLOCK = 32786l;
 
 #endif // !! ASSIMP_BUILD_NO_COMPRESSED_X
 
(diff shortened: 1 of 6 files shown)


code/Common/Compression.h
```diff
@@ -56,7 +63,7 @@ class Compression {
 
     /// @brief  Will open the access to the compression.
     /// @return true if close was successful, false if not.
-    bool open();
+    bool open(Format format);
```
AMZN-alexpete, line 66:
```suggestion
    bool open(Format format = Format::Binary);
```
A default of Format::Binary might be convenient for the most common use case.

code/Common/Compression.h
```diff
@@ -48,6 +48,13 @@ namespace Assimp {
 /// @brief This class provides the decompression of zlib-compressed data.
 class Compression {
 public:
+    enum class Format {
+        Binary = 0,
+        ASCII,
+
+        NumFormats,
+        InvalidFormat
+    };
```
AMZN-alexpete, line 57:
```suggestion
    enum class Format {
        InvalidFormat -1,
        Binary = 0,
        ASCII,

        NumFormats
    };
```
Keep InvalidFormat out of the range of any future valid formats

code/AssetLib/X/XFileParser.cpp
```diff
@@ -209,41 +183,26 @@ XFileParser::XFileParser(const std::vector<char> &pBuffer) :
             P1 += ofs;
             est_out += MSZIP_BLOCK; // one decompressed block is 32786 in size
```
AMZN-alexpete, line 184: Odd that the spec seems to indicate it is 32KB + 12 bytes https://docs.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-mci/3048c471-f739-4d30-bcb3-a962cfe84a69

code/AssetLib/X/XFileParser.cpp
```diff
@@ -60,25 +58,11 @@ using namespace Assimp::Formatter;
...
-// Dummy memory wrappers for use with zlib
-static void *dummy_alloc(void * /*opaque*/, unsigned int items, unsigned int size) {
-    return ::operator new(items *size);
-}
-
-static void dummy_free(void * /*opaque*/, void *address) {
-    return ::operator delete(address);
-}
+constexpr unsigned int MSZIP_MAGIC = 0x4B43;
+constexpr size_t MSZIP_BLOCK = 32786l;
```
AMZN-alexpete, line 65: Just happend to notice this value `32786l` does not match the comment below 
```
// one decompressed block is 32786 in size
```
or the spec https://docs.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-mci/3048c471-f739-4d30-bcb3-a962cfe84a69

But I'm not familiar with MSZIP, just happened to notice

code/Common/Compression.h
kimkulling: I prefer to support such options as an explicit parameter. It makes clear what the intention is behind that.

code/AssetLib/X/XFileParser.cpp
kimkulling: Let's try it.

code/AssetLib/X/XFileParser.cpp
kimkulling: Right, but this is the original code which has worked for years.

PR #4391, opened 2022-02-17
cmake: Fix assimp target install rule fully specifying component

jcfr:
This commit ensures generated install rules for the `assimp` target are
all consistently associated with components named after `LIBASSIMP_COMPONENT`
or `LIBASSIMP-DEV_COMPONENT` CMake variables.

It streamlines the packaging of specific components of assimp in client
projects leveraging the `CPACK_INSTALL_CMAKE_PROJECTS` variable.

Failing to do so leads to `/path/to/assimp-build/code/cmake_install.cmake`
file having install rules associated with a component called `Unspecified`.

See https://cmake.org/cmake/help/v3.10/command/install.html#installing-targets
and https://cmake.org/cmake/help/v3.10/module/CPack.html#variable:CPACK_INSTALL_CMAKE_PROJECTS


kimkulling:

PR #4393, opened 2022-02-18
Added support for "map_Bump -bm"

TerenceRussell:
Hi,

I've added Obj model support for the bump map -bm #### multiplier. I've attempted to follow the gltf use of "$tex.scale" as a guide for making it available.

An example Obj line with the parameter:
`map_Bump -bm 0.398056 textures\main_normal.png`

These changes will load the 0.398056 factor into a material property for either Normal or Height maps (ie: Obj map_Bump). The property is only set if the scalar is not 1.0.

The property can be retrieved like so:
`aiMaterial* mtl = scene->mMaterials[m];`
`ai_real bm = 1.0;`
`aiReturn rc = aiReturn_SUCCESS;`

`if( mtl->GetTexture(aiTextureType_NORMALS, texIndex, &path) == AI_SUCCESS) { `
`rc = aiGetMaterialFloat(mtl, AI_MATKEY_OBJ_BUMPMULT_NORMALS(0), &bm);`
`}`
`else if(mtl->GetTexture(aiTextureType_HEIGHT, texIndex, &path) == AI_SUCCESS) {`
`    rc = aiGetMaterialFloat(mtl, AI_MATKEY_OBJ_BUMPMULT_HEIGHT(0), &bm);`
`}`

### DIFF
--- a/code/AssetLib/Obj/ObjFileData.h
+++ b/code/AssetLib/Obj/ObjFileData.h
@@ -195,6 +195,9 @@ struct Material {
     //! PBR Anisotropy
     ai_real anisotropy;
 
+    //! bump map multipler (normal map scalar)(-bm)
+    ai_real bump_multiplier;
+
     //! Constructor
     Material() :
             diffuse(ai_real(0.6), ai_real(0.6), ai_real(0.6)),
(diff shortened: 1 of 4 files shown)


kimkulling (approved):

kimkulling:

PR #4397, opened 2022-02-20
update the calculation and orthogonalization for bitangent

youkeyao:
I think the calculation of bitangent needs to negate if the tangent space is TBN.


kimkulling:

Issue #4390, opened 2022-02-16
Bug: DefaultIOStream::FileSize() broken on 32bit systems


flixr:
**Describe the bug**
`stat` can fail on 32bit systems if not compiled with `-D_FILE_OFFSET_BITS=64`.
That could be because inode number is too large (happens on many file systems) or the file itself is too large.

This was 
* originally reported in #975
* then fixed with https://github.com/assimp/assimp/commit/ec67a1f2eae7d343e1c56a8a47bf80fdce93474c
* then commented out again with https://github.com/assimp/assimp/commit/a8e65a1e8aff4a254453f26719df95f8cacdaadf
* and finally completely removed in https://github.com/assimp/assimp/commit/417c4fcb457ab8282e8698596cba2de99c604e92

I don't know why it was apparently removed because of Android builds (cmake check for android not working)?
But for 32bit armhf this is needed for sure.

**To Reproduce**
Try reading a file with assimp on armhf (or probably any other 32bit arch) which has a inode number not representable in 32bit.
E.g. from a NFS mount or XFS file system.

**Expected behavior**
`DefaultIOStream::FileSize()` returns correct size instead of 0.

**Platform (please complete the following information):**
 - OS: Linux/armhf

**Additional context**
Might be the (or at least one) cause for autopkgtests failing for 32bit archs:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=995859

kimkulling:
FuXXX, sorry for that. This bug was introduced by myself. The fix is on its way.

kimkulling:
Shall be fixed now.

flixr:
actually adding the definition is still commented out...

PR #4398, opened 2022-02-20
Fix stat for 32-biut linuxes

kimkulling:
- Reenable fix to ensure that stat works correctly on 32-bit linuxes again
  - stat will return 32-bit inodes when checking a file. So when this call will be used on a 64-bit linux this will cause errors like:
Error writing to foo: Value too large for defined data type
File I/O error: foo
- closes https://github.com/assimp/assimp/issues/4390

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,6 +243,13 @@ SET(ASSIMP_LIBRARY_SUFFIX "" CACHE STRING "Suffix to append to library names")
 IF( UNIX )
   # Use GNUInstallDirs for Unix predefined directories
   INCLUDE(GNUInstallDirs)
+  # Ensure that we do not run into issues like http://www.tcm.phy.cam.ac.uk/sw/inodes64.html on 32 bit linux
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -243,6 +243,13 @@ SET(ASSIMP_LIBRARY_SUFFIX "" CACHE STRING "Suffix to append to library names")
 IF( UNIX )
   # Use GNUInstallDirs for Unix predefined directories
   INCLUDE(GNUInstallDirs)
+  # Ensure that we do not run into issues like http://www.tcm.phy.cam.ac.uk/sw/inodes64.html on 32 bit linux
+  IF( ${OPERATING_SYSTEM} MATCHES "Android")
+  ELSE()
+    IF ( CMAKE_SIZEOF_VOID_P EQUAL 4) # only necessary for 32-bit linux
+      #ADD_DEFINITIONS(-D_FILE_OFFSET_BITS=64 )
```
flixr, line 250: still commented out....

PR #4409, opened 2022-02-23
Expose the original OBJ "illum" value

TerenceRussell:
This commit exposes the original Obj model  `illum` value (illumination_model) as a material property. 

The `illum` value is defined in a .mtl material file like so:
`illum 3`

Background:
`ObjFileImporter::createMaterials` in `ObjFileImporter.cpp` partially interprets the illum value as an `aiShadingMode` enum and stores it in the `AI_MATKEY_SHADING_MODEL` material property. Illumination model values 3 thru 10 are lost in the interpretation. Exposing the original illum value allows user interpretation of the intended model.

Example accessor usage:
`    aiScene* scene = ...`
`    int illum = 1;`
`    aiMaterial* mtl = scene->mMaterials[m];`
`    aiGetMaterialInteger(mtl, AI_MATKEY_OBJ_ILLUM, &illum);`

### DIFF
--- a/code/AssetLib/Obj/ObjFileImporter.cpp
+++ b/code/AssetLib/Obj/ObjFileImporter.cpp
@@ -605,6 +605,9 @@ void ObjFileImporter::createMaterials(const ObjFile::Model *pModel, aiScene *pSc
 
         mat->AddProperty<int>(&sm, 1, AI_MATKEY_SHADING_MODEL);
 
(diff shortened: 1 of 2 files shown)


kimkulling (commented):
Looks good.

kimkulling:
Mered, thanks a lot for your contribution.

PR #4416, opened 2022-02-25
Optimize the problem of excessive memory allocation in FBX import

SolaToucher:
This problem may cause memory explosion.

In FBXConvert.cpp::ConvertMeshMultiMaterial, the number of mesh vertices after splitting has been preprocessed in advance, but the number of original mesh vertices was incorrectly used when allocating memory later.


kimkulling (approved):
Good catch, I will merge it!

kimkulling:

PR #4425, opened 2022-03-03
glTF2: Metallic roughness split

sacereda:
Now that we have the PBR properties inside material.h, I guess we should use them instead of AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE?

So we don't have to check both options inside the app.

The pull request is splitted in two commits (clang-format + patch).


kimkulling (approved):

kimkulling:

Issue #4424, opened 2022-03-03
How to detect if PBR is enabled


sacereda:
After updating to v5.2.2, I'm struggling on how to detect if a material is PBR or regular. Based on this commit from Obj importer:

https://github.com/assimp/assimp/blob/bbcefbd0348c5407e57f9f66a16a69b61159f537/code/AssetLib/Obj/ObjFileImporter.cpp#L620

https://github.com/assimp/assimp/commit/19371af6e65608ffc968df646bf20278e6d99414

It seems PBR properties like AI_MATKEY_METALLIC_FACTOR or AI_MATKEY_ROUGHNESS_FACTOR are added always, so I cannot check if the material has any of those in order to apply or not PBR properties on my application.

I was thinking maybe material properties like AI_MATKEY_USE_METALLIC_MAP or AI_MATKEY_USE_ROUGHNESS_MAP, but they are only added in FBXConverter.

Maybe there should be an AI_MATKEY_USE_PBR property or similar?

TerenceRussell:
I'm not sure about the general case, but for OBJ models a PBR hint flag could be added to the material.

If any of the newer PRB fields are encountered on load, the flag would be set.

The hint could be retrieved from the material like so:

`int isMaterialPBR = 0; // 0 = no; 1 = possibly.`
`aiGetMaterialInteger(mtl, AI_MATKEY_OBJ_PBR_HINT, &isMaterialPBR);`


This would be simple to add.

sacereda:
Hi Terence,

I think this will hide the root problem, as the obj is still exporting PBR properties on non-pbr materials, and we will end up adding an importer-specific PBR hints on each importer.

Please see the output from 'assimp info' (after applying https://github.com/assimp/assimp/pull/4426):

> $ bin\debug\assimpd.exe info test\models\OBJ\box.obj
...
Named Materials:
    'Default' (prop) [index / bytes | texture semantic]
        0 (?mat.name): [0 / 12 | n/a]
        1 ($mat.shadingm): [0 / 4 | n/a]
        2 ($mat.illum): [0 / 4 | n/a]
        3 ($clr.ambient): [0 / 12 | n/a]
        4 ($clr.diffuse): [0 / 12 | n/a]
        5 ($clr.specular): [0 / 12 | n/a]
        6 ($clr.emissive): [0 / 12 | n/a]
        7 ($mat.shininess): [0 / 4 | n/a]
        8 ($mat.opacity): [0 / 4 | n/a]
[...]

PR #4426, opened 2022-03-04
Add properties information on assimp info command line

sacereda:
This comes from debugging the issue/question from: 

https://github.com/assimp/assimp/issues/4424

Adding property listing can come in handy when displaying a quick overview from an asset. 

I also wanted to add some information from each property (color, float, integer, etc), but maybe it will be overkill. I gave it a try, but found myself re-implementing the AssJson exporter, at least on the material property point of view.

Same as before, I added a clang-format commit before comitting the real patch. If this is not the point please let me know, but my editor is configured to apply clang-format on save if a .clang-format is found and DisableFormat directive is not set.


kimkulling:

PR #4434, opened 2022-03-09
Fix 'i >= 0' always true bug

karjonas:
If 'disk_filename' does not contain a dot (.) then 'i' would overflow.
Making 'i' an int makes sure the for loop works as intended.


kimkulling (approved):

kimkulling:
Good catch, thanks a lot for your help. I will merge that!

PR #4440, opened 2022-03-11
Avoid setting PBR properties when they are not found on mtl file

sacereda:
This comes from https://github.com/assimp/assimp/issues/4424.

Metallic/roughness/sheen/clearcoat properties should only be added to the material if they appear on .mtl file.

Pinging @errissa here, feel free to suggest/change if you find anything odd.


errissa:
I’ll leave it to the core developers to comment, if necessary, on the implementation details. However, the concept expressed with this PR to optionally add the PBR properties based on their presence in the MTL file makes sense to me and shouldn’t impact anyone using a PBR workflow with OBJ.

code/Common/Maybe.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
```
kimkulling, line 1: Copyright is missing

code/Common/Maybe.h
```diff
@@ -0,0 +1,29 @@
...
+#include <assimp/ai_assert.h>
+
+template <typename T>
```
kimkulling, line 4: Documentation is missing

kimkulling (changes requested):
Just 2 small findings. Would you be able to solve these things? When done I will merge it. And maybe you can consider using std::optional? 

Just a question from my side.

sacereda:
Sure, I can check. I'm not really versed with std::optional, never used it before, and just tried to use something similar to STEPFile.h::Maybe, in case you were all used to this.

But yes, I can change it if you prefer to use std.

turol:
`std::optional` is C++17. What do we want our minimum standard requirement to be?

kimkulling:

PR #4451, opened 2022-03-16
GLTF2 attribute name/parse bug fix

Promit:
Fixing mistakes in attribute names for GLTF. This causes attribute index to not be parsed out correctly (e.g. JOINTS_1) and fails the scene load. See https://github.com/assimp/assimp/blob/a916a0eda5dfa5c9d1a6f0c4657af6d40886f0a8/code/AssetLib/glTF2/glTF2Asset.inl#L1380 for where the parsing breaks.


kimkulling (approved):

kimkulling:
Great catch! Thanks a lot for your contribution. Merged!

PR #4463, opened 2022-03-23
Prevent nullptr access to normals-array in bitangent computation

jakrams:
This came up in a fuzzing test. The input mesh ended up having no normals, but tangents, and assimp attempted to compute the bitangents, which does require normals and ended up crashing.

The fix checks whether normals are missing, and skips the tangent/bitangent computation and prints a warning instead.

Also executed clang-format on the file, which is why there are several whitespace changes.


Skylion007 (approved):

kimkulling:

PR #4465, opened 2022-03-25
Fix double precision build.

mahiuchun:


mahiuchun:
Fixed by #4473 already, closing.

PR #4481, opened 2022-04-07
Some Windows/MSYS2-specific fixes 

hgdagon:
After struggling with Github Desktop, I lost the fight. Se here's a new PR.


code/Common/DefaultIOStream.cpp
```diff
@@ -63,7 +63,7 @@ inline int select_fseek(FILE *file, int64_t offset, int origin) {
 
 
 
-#if defined _WIN32 && (!defined __GNUC__ || __MSVCRT_VERSION__ >= 0x0601)
+#if defined _WIN64 && (!defined __GNUC__ || __MSVCRT_VERSION__ >= 0x0601)
```
kimkulling, line 66: What about win32?

code/Common/DefaultIOStream.cpp
```diff
@@ -149,7 +149,7 @@ size_t DefaultIOStream::FileSize() const {
         //
         // See here for details:
         // https://www.securecoding.cert.org/confluence/display/seccode/FIO19-C.+Do+not+use+fseek()+and+ftell()+to+compute+the+size+of+a+regular+file
-#if defined _WIN32 && (!defined __GNUC__ || __MSVCRT_VERSION__ >= 0x0601)
+#if defined _WIN64 && (!defined __GNUC__ || __MSVCRT_VERSION__ >= 0x0601)
```
kimkulling, line 152: Same here?

kimkulling (changes requested):
Just check the comments.

code/Common/DefaultIOStream.cpp
hgdagon: In Msys2 clang32 environment this resolves to true, resulting in a build failure. `__MSVCRT_VERSION__` is `0xE00` in clang32.

code/Common/DefaultIOStream.cpp
hgdagon: Ditto

kimkulling (approved):
Ok, thanks a lot for your explanation!

code/Common/DefaultIOStream.cpp
Osyotr: This change breaks x86-windows builds...

PR #4484, opened 2022-04-13
MMD (pmx) fixes

RedSkittleFox:
Changes:

# Reuploaded MMD test model
I reuploaded mmd (pmx) model (test/models-nonbsd/MMD/Alicia_blade.pmx) from the original source:
https://3d.nicovideo.jp/alicia/
https://3d.nicovideo.jp/works/td14712
I have no idea what the issue was with the previous one, but it appeared to have ~10000 textures. (Corrupted Binary?)

# Enabled unit test for MMD
I re-enabled unit test for MMD again. It works as intended after reuploading the model.

# Fixed bug in MMDImporter
MMDImporter fails to read `pmx` files from the memory (`importer.ReadFileFromMemory`). The problem appears to be not using `IOSystem pIOHandler` but rather standard file stream. I've reimplemented it in the similar fashion as the [`FBXImporter::InternReadFile`](https://github.com/assimp/assimp/blob/dc8ba610951cd784ab97986e5094a7c83fde176a/code/AssetLib/FBX/FBXImporter.cpp#L136).

The fix could be probably improved by changing the signature of [`void PmxModel::Read(std::istream *stream)`](https://github.com/assimp/assimp/blob/dc8ba610951cd784ab97986e5094a7c83fde176a/code/AssetLib/MMD/MMDPmxParser.cpp#L513) to take a different stream type.


RedSkittleFox:
Changes in [5ab1be8](https://github.com/assimp/assimp/pull/4484/commits/5ab1be8d04da9ec3729e617fc002c634242c4c01):
1. Removed C++17 features: `std::data(vec)` -> `vec.data()` and `std::size(vec)` -> `vec.size()`
2. Replaced deprecated `std::strstream` with `std::stringstream`. That being said, right now it's initialized with a copy of the data loaded to the vector. That cannot be resolved using `std::stringstream` even when using `std::string` as the original storage container. Until C++20 STL does not provide `std::string` move constructor for `std::stringstream`. 
This could be potentially resolved by providing custom stream class wrapper for vector or something.

kimkulling (approved):
Looks good, let's see if the test gets green.

kimkulling:
It works! Merged, thanks a lot for your contribution.

PR #4490, opened 2022-04-18
FbxConverter: update bone mOffsetMatrix 

Mykhailo418:
For some FBX files, there is an incorrect calculation of _mOffsetMatrix_ of bones. According to those 2 FBX files  "**box_bone_animation.FBX**" and  "**box_bone_animation_moved.FBX**"(you can find them in [primitives.zip](https://github.com/assimp/assimp/files/8506104/primitives.zip) archive), they both have the same box with skeleton animation, but in the file "**box_bone_animation_moved.FBX**"  box is moved by 50 by the x-axis. After converting them into the Assimp Scene, in the first example and the second example Assimp Scenes have different _mOffsetMatrix_ of their bones, but they should have the same.

I have found that there is another parameter "_Transform_" in "_SubDeformer::Cluster_" in FBX file that contains the exact matrix that corresponds to _mOffsetMatrix_ that should be. Additionally, in both cases, this property has some value. Also, I have checked it in other 3D Models. In **FBXDocument.h** I have found that the class "_Cluster_" has the property "_transform_" contains a value of the same parameter "_Transform_" of "_SubDeformer::Cluster_" in the FBX file.

So, this pull request is my proposal to change the way of assigning _mOffsetMatrix_ value of bones.

### DIFF
--- a/code/AssetLib/FBX/FBXConverter.cpp
+++ b/code/AssetLib/FBX/FBXConverter.cpp
@@ -232,7 +232,6 @@ void FBXConverter::ConvertNodes(uint64_t id, aiNode *parent, aiNode *root_node) 
             nodes_chain.clear();
             post_nodes_chain.clear();
 
-            aiMatrix4x4 new_abs_transform = parent->mTransformation;
             std::string node_name = FixNodeName(model->Name());
             // even though there is only a single input node, the design of
             // assimp (or rather: the complicated transformation chain that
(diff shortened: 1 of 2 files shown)


kimkulling (approved):
Makes sense to me, but we will need some more testing. I will merge it to master.

kimkulling:
Merged, thanks a lot for your contribuion.

PR #4494, opened 2022-04-20
Small allocation fix in fbx loader

FlorianBorn71:
This PR addresses a problem with loading very large binary fbx files. We had a case where the parsing code performed one billion(!) single small allocation which eventually crashed on any system.
This PR introduces class StackAllocator which can be used to allocate many small objects in a deque pattern. After parsing, the single objects are not de-allocated individually but instead the whole allocator is discarded.
This fix not only allows for loading much larger fbx files (in our case ~16GB with 3 million individual parts) but also improves parsing speed significantly on all other medium to large fbx files.


turol:
First, `StackAllocator` should be made noncopyable to avoid mishaps. Second, is this actually safe? It never runs destructors and the allocated objects are not trivial. `LazyObject` contains a `std::unique_ptr` and `Connection` contains a `std::string`. This sounds like a memory leak to me.

FlorianBorn71:
Thanks Turo, indeed the current version leaks memory - I will have to re-iterate

FlorianBorn71:
proper deallocation handled in new commit

code/CMakeLists.txt
```diff
@@ -193,6 +193,8 @@ SET( Common_SRCS
   Common/ScenePreprocessor.cpp
   Common/ScenePreprocessor.h
   Common/SkeletonMeshBuilder.cpp
+  Common/StackAllocator.h
```
kimkulling, line 196: PLease inline the implementation in an inl-file.

kimkulling (changes requested):
The stack-allocator looks great. I would recommend not to include a cpp-file. If you need this for linkage just define this in an inline file and include this in the header.

code/CMakeLists.txt
FlorianBorn71: done

FlorianBorn71:
Hi @kimkulling, the compilation issues above do not make too much sense, it looks like it's compiling with a cached old .h file.
Is there a way to trigger a clean build?

code/Common/StackAllocator.h
```diff
@@ -0,0 +1,96 @@
...
+    // Releases all the memory owned by this allocator.
+    // Memory provided through function Allocate is not valid anymore after this function has been called.
+    inline void FreeAll();
+
+private:
+    constexpr const static size_t g_maxBytesPerBlock = 64 * 1024 * 1024; // The maximum size (in bytes) of a block
+    constexpr const static size_t g_startBytesPerBlock = 16 * 1024;  // Size of the first block. Next blocks will double in size until maximum size of g_maxBytesPerBlock
+    size_t m_blockAllocationSize = g_startBytesPerBlock; // Block size of the current block
+    size_t m_subIndex = g_maxBytesPerBlock; // The current byte offset in the current block
+    std::deque<uint8_t *> m_storageBlocks;  // A list of blocks
```
RedSkittleFox, line 88: The `std::vector<uint8_t*>` might be a better choice here.

code/AssetLib/FBX/FBXDocument.h
```diff
@@ -80,6 +80,10 @@ class BlendShape;
 class Skin;
 class Cluster;
 
+#define new_LazyObject new (allocator.Allocate(sizeof(LazyObject))) LazyObject
+#define new_Connection new (allocator.Allocate(sizeof(Connection))) Connection
+#define delete_LazyObject(_p) (_p)->~LazyObject()
+#define delete_Connection(_p) (_p)->~Connection()
```
RedSkittleFox, line 86: Object construction and destruction should be delegated to the function related to the allocator. Generally it would be better not to use macross like that overall. 
```cpp
// #include <new>

public:
    template<class T, class... Args>
    T* ConstructObject(Args&&... args)
    {
        return ::new(this->Allocate(sizeof(T))) T(std::forward<Args>(args)...);
    }

    template<class T>
    void DestroyObject(T* obj) noexcept
    {
        obj->~T();
    }
```

code/Common/StackAllocator.h
RedSkittleFox, line 88: `std::unique_ptr<uint8_t[]>` can be used here to improve safety.

code/Common/StackAllocator.inl
```diff
@@ -0,0 +1,84 @@
...
+
+
+inline StackAllocator::StackAllocator() {
+}
+
+inline StackAllocator::~StackAllocator() {
+    FreeAll();
+}
+
+inline void *StackAllocator::Allocate(size_t byteSize) {
```
RedSkittleFox, line 56: This function returns unaligned memory. It should take in additional argument called alignment and align memory internaly.

```cpp
StackAllocator sa;

char* c = new (sa.Allocate(sizeof(char))) char(1);
std::string* s = new (sa.Allocate(sizeof(std::string))) std::string("str"); // This is UB?
std::cout << reinterpret_cast<size_t>(s) % alignof(std::string) << '\n';
```

kimkulling:
Will get fixed by https://github.com/assimp/assimp/pull/5096

Issue #4450, opened 2022-03-16
Bug: Intel oneAPI 2022.0.2 compilation fails with tautological-constant-compare in AssetLibs


wdconinc:
**Describe the bug**
With the Intel oneAPI 2022.0.2 C and C++ compiler on Ubuntu 21.10, compilation of assimp master branch fails due to `-Werror,-Wtautological-constant-compare` in various AssetLib files.

**To Reproduce**
1. Install the Intel oneAPI 2022.0.2 C and C++ compilers on Ubuntu 21.10 (instructions at https://www.intel.com/content/www/us/en/develop/documentation/installation-guide-for-intel-oneapi-toolkits-linux/top/installation/install-using-package-managers/apt.html)
2. `source /opt/intel/oneapi/setvars.sh`
3. `CC=icx CXX=icpx cmake -Bbuild -S.`
4. `make -Cbuild -k`

**Expected behavior**
Compilation should succeed.

**Actual behavior**
Various warnings-turned-errors such as
```console
/home/wdconinc/git/assimp/code/AssetLib/IFC/IFCCurve.cpp:549:30: error: comparison with infinity always evaluates to false in fast floating point modes [-Werror,-Wtautological-constant-compare]
    ai_assert( min_diff[ 0 ] != inf );
               ~~~~~~~~~~~~~ ^  ~~~
/home/wdconinc/git/assimp/include/assimp/ai_assert.h:67:45: note: expanded from macro 'ai_assert'
#   define ai_assert(expression) (void)((!!(expression)) || (Assimp::aiAssertViolation(#expression, __FILE__, __LINE__), 0))
                                            ^~~~~~~~~~
```
and
```console
/home/wdconinc/git/assimp/code/AssetLib/IFC/IFCOpenings.cpp:644:27: error: comparison with infinity always evaluates to false in fast floating point modes [-Werror,-Wtautological-constant-compare]
        if (std::fabs(s0) == inf && std::fabs(n0_to_m0.x) < smalle) {
            ~~~~~~~~~~~~~ ^  ~~~
```
All warnings are attributed to tautological-constant-compare. Full error log attached as [assimp-tautological-constant-compare.log](https://github.com/assimp/assimp/files/8273519/assimp-tautological-constant-compare.log)

**Platform (please complete the following information):**
 - OS: Ubuntu 21.10
 - Version: master a916a0eda5dfa5c9d1a6f0c4657af6d40886f0a8

RichardTea:
[...]

PR #4507, opened 2022-04-28
Disable assertion tests floating point against inf for Intel oneAPI

kimkulling:
- closes https://github.com/assimp/assimp/issues/4450


turol:
It might be better to rewrite those tests to use `std::fpclassify`. Then comparing the representations becomes the implementation's problem.

PR #4509, opened 2022-04-30
Add an option to treat warnings as errors

hgdagon:
Add an option to treat warnings as errors for finer control.


CMakeLists.txt
```diff
@@ -108,7 +108,11 @@ OPTION( ASSIMP_INSTALL
   "Disable this if you want to use assimp as a submodule."
   ON
 )
-OPTION ( ASSIMP_ERROR_MAX
+OPTION ( ASSIMP_WARNINGS_AS_ERRORS
+  "Treat all warnings as errors."
+  OFF
```
kimkulling, line 113: The current policy is to have this feature enabled. Could you please change this to be enabled as the default option? I would really apprechiate this!

kimkulling (changes requested):
One finding abouth the default value.

CMakeLists.txt
hgdagon: @kimkulling Done. Also merged the 2 together, so as not to mess with the logic of enabling one for the other, and updated the list of options in Build.md.

kimkulling (approved):
Looks nice, I will merge it.

kimkulling:

CMakeLists.txt
rdb: Why should it be on by default, though? -Werror is fairly universally considered harmful to ship by default:
https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/
http://blog.schmorp.de/2016-02-27-tidbits-for-the-love-of-god-dont-use-werror.html
https://flameeyes.blog/2009/02/25/future-proof-your-code-dont-use-werror/

It seems like an option only assimp developers might want to use, or something to enable in CI, but not to ship by default for people who just want to build and use assimp.

PR #4510, opened 2022-04-30
[Feedback needed]Resource script updates

hgdagon:
I have a very limited knowledge of Windows resource scripts and all of it is based on the MS documentation, which isn't exactly rich in detail. 

If I understand correctly, everything I removed is not needed for a library or a command line application. Please correct me if I'm wrong.

I was not able to test these in MSVC, since the generate command fails, and I have no idea how to circumvent that.


kimkulling (approved):
Looks fine for me, can I merge it?

kimkulling:

PR #4525, opened 2022-05-11
Clang-Tidy performance fixes (make values const-ref where needed).

Skylion007:
* Apply fixes from clang-tidy (mostly just ensuring that values that should use const ref end up doing so).


kimkulling (approved):
Look good, I will merge it!

kimkulling:

PR #4534, opened 2022-05-18
Inifinite loop on bad import files

tanolino:
I had an import file that caused an infinite loop. I don't exactly know how this algorithm works here but maybe we should change it more.
Probably calculate the amount of steps and chop linear.
Also I am not sure if I am allowed to share the file with you so I will add it here when I get the go.


tanolino:
[Hyperspace_Madness_Killamari_Minion.zip](https://github.com/assimp/assimp/files/8727329/Hyperspace_Madness_Killamari_Minion.zip)

I don't know where it is from.

kimkulling (approved):
Looks good, I will merge it.

PR #4541, opened 2022-05-24
I ran into an error while processing coloured binary stl. Just a type but better be fixed

blackhorse-reddog:
I ran into an error while processing coloured binary stl. Just a type but better be fixed


turol:
This is wrong. Those bitops are for masking off specific bits. If you want them to be hex they should be `0x1f`.

turol:
It might be better to fix them but leave them as hex like `0x1f`.

blackhorse-reddog:
I thought about that too, and chose the decimal format to indicate it is the same 31 as the 31.0 used to construct InvVal. But you are right, there is a risk that 0x will be put in front of the number again, if it stays that way, so I rewrote it into hex.

kimkulling (approved):
Looks fine!

kimkulling:

PR #4542, opened 2022-05-25
Fix GNUC check on Windows

### LINKED ISSUE #4540: Bug: x86-windows build failure
**Describe the bug**
Build fails when targeting x86 on windows.

**To Reproduce**
Try to build latest release with 

**Additional context**
The breakage was introduced in this PR: https://github.com/assimp/assimp/pull/4481
Easiest solution is to replace revert _WIN32 to _WIN64 change and add another check for msys2 clang32 environment.

cc @hgdagon

hgdagon:
@Osyotr reported an issue on MSVC. @Osyotr please check if this fixes the issue and report back.

resolves #4540


hgdagon:
OK. This seems to be safest route, but I don't know if this will work if assimp is built with clang-cl.

Osyotr:
I'm happy if CI is happy :)


To make github happy: Fixes #4540

Osyotr:
I'm happy if CI is happy :)


To make github happy: Fixes #4540

hgdagon:
@Osyotr That's the thing. CI was happy with [my previous PR](https://github.com/assimp/assimp/runs/5868443534?check_suite_focus=true) too, which I also thought should prove it's correct, but evidently that wasn't the case. I cannot test MSVC build on my machine.

Osyotr:
MSVC builds are successful with this patch, thanks!

kimkulling (approved):
Sorry for the delay, I was not sure if this fix is correct. Now I found the time to check it and I will merge it.

kimkulling:

PR #4549, opened 2022-05-28
fix msvc warnings-as-errors

Gargaj:


code/Common/FileSystemFilter.h
```diff
@@ -300,7 +300,7 @@ class FileSystemFilter : public IOSystem
 
         const char separator = getOsSeparator();
         for (it = in.begin(); it != in.end(); ++it) {
-            int remaining = std::distance(in.end(), it);
+            int remaining = (int)std::distance(in.end(), it);
```
turol, line 303: A better fix would be to change the type of `remaining`.

Issue #4516, opened 2022-05-07
Static Linking - OPENDDL_STATIC_LIBARY and ASSIMP_BUILD_DLL_EXPORT


RobertMtx:
**Describe the bug**
DLLs are being exported in code because `OPENDDL_STATIC_LIBARY `is not defined when attempting to link statically, which causes the final project to compile differently, which results in the generation of lib and exp files for exe builds. I recommend defining `OPENDDL_STATIC_LIBARY `when `ASSIMP_BUILD_DLL_EXPORT `is undefined. This should fix the issue.

**To Reproduce**
Attempt to link assimp statically and build exe. Compiler will generate exp and lib files automatically.

**Expected behavior**
Normal exe build without exp and lib files.

**Platform (please complete the following information):**
 - OS: Windows 10

Thanks for a great library. Love this tool!

Underdisc:
After some more digging, I realized the export tags were added in `include/assimp/defs.h`, even though I thought I already addressed it. 🤷 This is because `ASSIMP_BUILD_DLL_EXPORT` is always defined no matter the build type as can be seen [here](https://github.com/assimp/assimp/blob/b7a70ac707cfabe73b18ec8219cb80fd3d201805/code/CMakeLists.txt#L1103).

PR #4555, opened 2022-06-01
Remove dllexport tags from static library builds.

Underdisc:
This is a fix to issue #4516. A definition needs to be added depending on the library build type.


kimkulling (approved):
Looks fine, I will merge it"

PR #4564, opened 2022-06-06
Pragma warnings cause build fail with MinGW

ethaninfinity:
Using pragma warning caused builds using MinGW bundled with the latest version of CLion (2022.1.2). Removing it seems to build correctly and fix the problem.

### DIFF
--- a/contrib/unzip/crypt.c
+++ b/contrib/unzip/crypt.c
@@ -43,10 +43,10 @@
 
 #include "crypt.h"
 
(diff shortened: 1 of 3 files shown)


turol:
No, the guard should be changed to MSVC-specific macro (`_MSC_VER` I think).

kimkulling (approved):
Looks cool, CI is on its way!

kimkulling:

PR #4577, opened 2022-06-16
Add support for _WIN32

PercentBoat4164:
My machine failed to compile this project with the MSVC, but does succeed with MinGW. I edited this pragma to add a check for _WIN32 to the 32-bit variant of the code. My computer then compiled the project successfully under both the MSVC and MinGW. MinGW seemed to use the 64-bit variant while the MSVC used the 32-bit variant. Perhaps I am fixing this in a way that makes it bad, but this fix does make it functional, at least on my machine.

### DIFF
--- a/code/Common/DefaultIOStream.cpp
+++ b/code/Common/DefaultIOStream.cpp
@@ -156,7 +156,7 @@ size_t DefaultIOStream::FileSize() const {
         if (0 != err)
             return 0;
         mCachedSize = (size_t)(fileStat.st_size);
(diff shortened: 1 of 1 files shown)


PercentBoat4164:
I also noticed that changing _WIN64 in the pragma above to _WIN32 also allows the program to compile. Perhaps this is a better solution?

PercentBoat4164:
I have also noticed #4542. This seems to be addressing the same problem. I am going to close this as it has already passed all checks and seems to have approval.

PR #4599, opened 2022-06-30
Add mingw

kimkulling:


kimkulling:
I guess this approach is too complicated.

PR #4630, opened 2022-07-08
Fix stl for over 4 gb

tanolino:
I am not sure how often we have STL files of over 4 GB in size. Anyway, let's make them work.


kimkulling (approved):

kimkulling:
Great finding, thanks a lot!

Issue #4334, opened 2022-01-10
Building v5.1.6 gives error: 'zlib_filefunc_def' {aka 'struct zlib_filefunc_def_s'} has no member named 'zopendisk_file'


brechtsanders:
Building v5.1.6 gives error: 'zlib_filefunc_def' {aka 'struct zlib_filefunc_def_s'} has no member named 'zopendisk_file'.

I'm on Windows 11 using MinGW-w64 GCC 11.2.0 and I have all dependancies installed from latest versions.

Whe building assimp 5.1.6 I get:
```
R:/winlibs64-11.2.0ucrt/assimp-5.1.6/code/Common/ZipArchiveIOSystem.cpp: In static member function 'static zlib_filefunc_def Assimp::IOSystem2Unzip::get(Assimp::IOSystem*)':
R:/winlibs64-11.2.0ucrt/assimp-5.1.6/code/Common/ZipArchiveIOSystem.cpp:200:13:
error: 'zlib_filefunc_def' {aka 'struct zlib_filefunc_def_s'} has no member named 'zopendisk_file'; did you mean 'zopen_file'?
  200 |     mapping.zopendisk_file = (opendisk_file_func)opendisk;
      |             ^~~~~~~~~~~~~~
      |             zopen_file
R:/winlibs64-11.2.0ucrt/assimp-5.1.6/code/Common/ZipArchiveIOSystem.cpp:200:31:
error: 'opendisk_file_func' was not declared in this scope; did you mean 'open_file_func'?
  200 |     mapping.zopendisk_file = (opendisk_file_func)opendisk;
      |                               ^~~~~~~~~~~~~~~~~~
      |                               open_file_func
ninja: build stopped: subcommand failed.
```

When checking I noticed there are different `ioapi.h` versions around with and without `zopendisk_file` defined as member of `zlib_filefunc_def_s`. From what I see I think `ZOPENDISK64` seems like a good macro to check to determine if `zopendisk_file` is to be used or not.

So I was able to build assimp 5.1.6 after the following patch:
```sh
patch -ulbf code/Common/ZipArchiveIOSystem.cpp << EOF
@@ -198,3 +198,5 @@
     mapping.zopen_file = (open_file_func)open;
+#ifdef ZOPENDISK64
     mapping.zopendisk_file = (opendisk_file_func)opendisk;
+#endif
     mapping.zread_file = (read_file_func)read;
EOF
```

But I'm not sure if that breaks anything else...

[...]

PR #4631, opened 2022-07-08
Build fix for compiling against minizip.

robertosfield:
From v5.1.6 onwards Assimp fails to build on my Ubuntu 21.10 system's default minizip install.  Looking at Issues tracker, #4334, others have the same build problem on a range of Linux distos.   I've added a #ifdef check around the problematic line to keep things compiling.

Unfortunately this #ifdef approach looks like it'll disable the functionality that caused this break in the build - I don't know how intrusive this would be as I'm not familiar with the wider code, there may well be more changes that would need disabling or handling to more elegantly handle these features that can't be implemented against minizip.

I provide this build fix so that others who are hitting up against the same problem can at least get Assimp built.


kimkulling (approved):
Let's give it a try!

PR #4641, opened 2022-07-15
Fixed warning & improved algorithm

Koekto-code:
Build was failing because of check if (char) more than 127.
Now it checks if there are NUL, SOH, STX, ETX symbols and symbols that are out of STL syntax (like Ђ € ‰).
AssimpView (using library with this fix) loads successfully all files from /test/models/STL except triangle_with_empty_solid.stl.


kimkulling (approved):
Looks feasible for me.

PR #4643, opened 2022-07-16
Fixes Heap-buffer-overflow in SuperFastHash

sashashura:
Fixes Heap-buffer-overflow in SuperFastHash revealed by fuzzing:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44964

The root cause is that a call to `ai_snprintf` in `LWSImporter::SetupNodeName` returns the number of characters that would have been written if the buffer had been sufficiently large.
It leads to creation of `aiString` object with incorrect length that causes out of bounds buffer read in `SuperFastHash` later.


sashashura:
Do you have any questions about this fix for out of bounds memory read?

kimkulling (approved):
Looks fine.

kimkulling:
Sorry, I was in vacation.

kimkulling:

sashashura:
> Sorry, I was in vacation.

No worries, I saw there is no activity in the repo. It is much worse when there is the activity, but my pull requests are being ignored :(

PR #4644, opened 2022-07-16
Fixes Heap-use-after-free in Assimp::DXFImporter::ExpandBlockReferences

sashashura:
Fixes Heap-use-after-free in Assimp::DXFImporter::ExpandBlockReferences revealed by fuzzing:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=42899

The root cause is that `bl_src.lines` and `bl.lines` in `DXFImporter::ExpandBlockReferences` may be the same vector. Iterator may become invalid when new objects are pushed into the same vector.


kimkulling (approved):
Looks fine!

kimkulling:

PR #4646, opened 2022-07-16
Fixes Heap-buffer-overflow in Assimp::ObjFileParser::getFace

sashashura:
Fixes Heap-buffer-overflow in Assimp::ObjFileParser::getFace revealed by fuzzing:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40589

The root cause is that `const int iVal(::atoi(&(*m_DataIt)));` reads beyond of `m_DataItEnd` if the string is not null terminated.


kimkulling (approved):
Looks fine, I will merge it.

PR #4647, opened 2022-07-16
Fixes Crash in Assimp::ObjFileMtlImporter::getFloatValue

sashashura:
Fixes Crash in Assimp::ObjFileMtlImporter::getFloatValue revealed by fuzzing:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25240

The root cause is that in the `getFloatValue(m_pModel->m_pCurrentMaterial->alpha);` call the `m_pModel->m_pCurrentMaterial` is null and `getFloatValue` receives a memory reference to 0 + offsetof(alpha). This leads to write memory violation.


kimkulling (approved):
Looks fine, let's merge it!

PR #4666, opened 2022-07-30
Improvements and optimizations for the obj-parsers.

kimkulling:
- Avoid creation of std::string to parse integer number from face declaration
- Replace one std::string constant with a character array
- Use std::vector to store pod face types instead of storing pointers to each face


kimkulling:
Unfortunately, the face is not a pure POS type. I need to rework this again.

PR #4678, opened 2022-08-08
Use unqualified uint32_t everywhere in FBXBinaryTokenizer

villevoutilainen:
The use of qualified std::uint32_t requires including <cstdint> instead
of <stdint.h> on some implementations, and that breaks the build of Qt 6
on GCC 13. Just use the unqualified name everywhere.


kimkulling:

PR #4679, opened 2022-08-09
(Mostly) Blender fixes

turol:
Fix loading .blend files and some other miscellaneous fixes I found along the way.

Most shipped .blend files now load. The only one that doesn't is "BlenderDefault_262.blend" but I'm not sure what's wrong with it.


turol:
@kimkulling ping. Is this ok to merge?

turol:
@kimkulling If you don't object by tomorrow I will merge this.

PR #4699, opened 2022-08-23
SMD fixes

turol:
A number of cleanups and a fix for possible out of bounds write. I'm not entirely sure this is correct since the use of indices is extremely complicated. The unit tests pass and the included .smd files still load.


kimkulling (approved):

kimkulling:
Not too bad I would say ;-). Thanks for the contribution.

PR #4700, opened 2022-08-23
Apply modernize-use-emplace clang-tidy rule

Skylion007:
This PR applies the modernize-use-emplace clang-tidy rule to the code base to improve performance/readability. The clang-tidy check prefers "emplace" to "push" on STL containers when its could improve performance. 

Using emplace prevents a temporary from being constructed as the arg to push method call and instead calls the constructor of the object at the appropiate memory location within the STL container. This particularly useful for constructing objects from c_string pointers or adding std::tuple, std::pairs to containers etc since no move operation is required. This can dramatically speed up some operations involving push items onto a STL container (like a std::vector).

### DIFF
--- a/code/AssetLib/3DS/3DSLoader.cpp
+++ b/code/AssetLib/3DS/3DSLoader.cpp
@@ -319,7 +319,7 @@ void Discreet3DSImporter::ParseObjectChunk() {
     case Discreet3DS::CHUNK_MAT_MATERIAL:
 
         // Add a new material to the list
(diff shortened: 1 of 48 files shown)


JG-Adams:
This look like a good practice.
I wonder why this isn't taught to be used everywhere?

turol:
You should make one commit per importer or even one per file. If this breaks something it will be very hard to bisect or revert.

Skylion007:
> You should make one commit per importer or even one per file. If this breaks something it will be very hard to bisect or revert.

My previous static analysis PRs have been monolithic. The automated fixes are the conservative side so I really doubt they will break anything.

kimkulling (approved):
Looks fine.

kimkulling:

PR #4701, opened 2022-08-24
Fix out of bounds reads in X3D importer

turol:
And add FIXME comment to questionable code. The included test model does not exercise this code. It would be useful to find one that does.


kimkulling:
You are right!

kimkulling:

PR #4703, opened 2022-08-25
clang-tidy: explicitly default all empty ctors and dtors

Skylion007:
Applying yet another static analysis clang-tidy rule: [modernize-use-equals-default](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-default.html). 

This check explicitly defaults empty ctors destructors. This tells the compiler that the destructor is trivial and may enable more aggressive compiler optimizations. Specifically, it make some class and structs trivial if all its members are trivial. Without this, compilers may not treat the class as trivial even if the custom defined ctor / dtor is empty. It also increases readability and consistency with the rest of the codebase.

### DIFF
--- a/code/AssetLib/3DS/3DSExporter.cpp
+++ b/code/AssetLib/3DS/3DSExporter.cpp
@@ -209,9 +209,7 @@ Discreet3DSExporter::Discreet3DSExporter(std::shared_ptr<IOStream> &outfile, con
 }
 
 // ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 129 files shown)


kimkulling (approved):
Thanks a lot for your work! Looks fine.

Skylion007:
I added a couple of more automated fixes that should be safe. Mostly just forgetting to apply some checks to the headers and catching a few more push_backs that should be emplace.

kimkulling:

PR #4704, opened 2022-08-26
Wrong object is created here!

JG-Adams:
If I'm not mistaken. Because this is originally push_back(aniMeshVertex) instead of push_back(v).
And the fact that aniMeshVertex is just getting created and destroyed for nothing.
It need to be replaced with this then? A Clang-tidy syntax to create the same aniMeshVertex.

I found this in pull request #4527 I think it may need to be reviewed too.
Hope it help!


kimkulling (approved):
Looks fine, thanks!

kimkulling:

JG-Adams:
> Merged, thanks a lot for your contribution.

Happy to have helped! :)

PR #4707, opened 2022-08-31
fix vertices being joined duplicating weights

Gargaj:
This bug causes weights to duplicate, i.e. one bone will have several (equal) weights assigned to the same vertex - which then has the potential to cause even bigger problems when combined with `LimitBoneWeightsProcess`.

The fix is not the prettiest but I'm a week before deadline here.


kimkulling (approved):

kimkulling:

PR #4715, opened 2022-09-07
Kimkulling/cleanup after review

kimkulling:
- Add doc + unittest for small class Maybe.

### DIFF
--- a/code/Common/Maybe.h
+++ b/code/Common/Maybe.h
@@ -49,7 +49,7 @@ namespace Assimp {
 template <typename T>
 struct Maybe {
     /// @brief 
-    Maybe() : _valid(false) {}
+    Maybe() = default;
 
     /// @brief 
     /// @param val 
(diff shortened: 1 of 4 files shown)


code/Common/Maybe.h
```diff
@@ -1,29 +1,89 @@
...
         return _val;
     }
 
-private:
     Maybe &operator&() = delete;
+    Maybe(const Maybe &) = delete;
+
+private:
+    T _val;
+    bool _valid;
```
Skylion007, line 86:
```suggestion
    bool _valid = false
```
Then you can simplify the default ctor to 
```
Maybe() = default;
``` 
which is might be a bit more readable.

kimkulling: Done!

Skylion007 (approved):

code/Common/Maybe.h
```diff
@@ -82,7 +82,7 @@ struct Maybe {
     Maybe(const Maybe &) = delete;
 
 private:
-    T _val;
+    T _val = false;
```
Skylion007: @kimkulling Wrong line. the thing below it should be default initialized.

kimkulling: Saw it, already fixed. Thanks for the hint!

PR #4716, opened 2022-09-07
Minor obj export bugfix

HiMemX:
Fixed a small typo, which caused verts to be exported with one space too much after their declaration character


kimkulling (approved):
Looks fine, let's merge it.

PR #4717, opened 2022-09-09
Perfect forward val to utMaybe.cpp

Skylion007:
Something I should have caught in my review of https://github.com/assimp/assimp/pull/4715 . The val arg should be perfectly forwarded with a universal reference to utMaybe. This allows the value T to be move assigned which allows both can make it more efficient and also allows utMaybe to work with values that are only move assignable. By using perfect forwarding, it will also just pass the reference if T is a reference.

Sorry for not catching this yesterday @kimkulling


kimkulling:
Not to worry! Thanks you so much for the great review. I will merge it.

kimkulling (approved):
Looks fine.

kimkulling:

PR #4733, opened 2022-09-19
WIP - skeleton

inhosens:
Hi,
This is not a pull request but just a suggestion about an usage of a new skeleton implementation.
I understand there are some differences between different assets for skinning and assimp tries to solve the problem with it.
I think if assimp makes an additional data in the node having meshes in order to indicate corresponding skeletons, it will be very helpful to support gltf's skin implementation. As you know, current assimp cannot support an example, RecursiveSkeletons in glTF-Sample-Models. (https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/RecursiveSkeletons)
Thanks to all the developers and contributers here.

### DIFF
--- a/code/AssetLib/Blender/BlenderScene.cpp
+++ b/code/AssetLib/Blender/BlenderScene.cpp
@@ -569,7 +569,7 @@ void Structure ::Convert<MVert>(
         const FileDatabase &db) const {
 
     ReadFieldArray<ErrorPolicy_Fail>(dest.co, "co", db);
-    ReadFieldArray<ErrorPolicy_Fail>(dest.no, "no", db);
+    ReadFieldArray<ErrorPolicy_Warn>(dest.no, "no", db);
     ReadField<ErrorPolicy_Igno>(dest.flag, "flag", db);
     //ReadField<ErrorPolicy_Warn>(dest.mat_nr,"mat_nr",db);
     ReadField<ErrorPolicy_Igno>(dest.bweight, "bweight", db);


inhosens:
@kimkulling Can you review this patch and comment on it? I'm not sure if this is what you expected for aiSkeleton.

kimkulling:
Just started to review it. Thanks for your work!

kimkulling:
So each mesh will store their own skeleton data, right? And yes: we need to have a real example implementation. I am currently working on a IMGUI-based viewer. Maybe this could be the right place for it?

code/AssetLib/FBX/FBXConverter.cpp
```diff
@@ -1451,34 +1425,38 @@ unsigned int FBXConverter::ConvertMeshMultiMaterial(const MeshGeometry &mesh, co
...
-        if (bone == nullptr) {
+        const unsigned int skinId = mesh2skin_map[i];
+        if (skinId == AI_NO_SKELETONS || skins_set[skinId])
             continue;
+
+        aiSkeleton *skeleton = mSkeletons[skinId];
+        skeleton->mNumBones = mesh->mNumBones;
+        skeleton->mBones = new aiSkeletonBone*[skeleton->mNumBones]();
+        for (size_t j = 0; j < mesh->mNumBones; ++j) {
+            aiBone *bone = mesh->mBones[j];
```
jackalbe, line 1450: to avoid junk data in the pointer, should the skeleton's bone pointer be set to null?
such as: 
skeleton->mBones[j] = nullptr

PR #4737, opened 2022-09-21
Prefix symbols from stb_image.h

p12tic:
This makes it possible to link assimp statically into an executable with another static library that includes stb_image.h without hiding its symbols (be it prefixing names like here, using internal linkage or not exporting in case of shared libraries).

In this particular case prefixing is the only option. Using internal linkage is not efficient because the library is used from multiple translation units. Not exporting is not possible in cases assimp is compiled as a static library.


kimkulling (approved):
Great idea.

kimkulling:

PR #4738, opened 2022-09-21
Make FBX parser resilient to missing data streams

FlorianBorn71:
A customer of ours hit this problem in the "ByPolygonVertex" code path. The same model loads fine in other tools (e.g. Maya viewer) so I think it's worthwhile to make this parsing more robust, also for some other code paths.

Interestingly, a similar fix had been added in this commit:
https://github.com/assimp/assimp/commit/de8adacb54570ca58aa233ed59e5210cefb6f497

...but not for the other code paths.


kimkulling:
You are completly right: we missed the other code paths when fixing this older bug. Thanks a lot for your work.

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #4744, opened 2022-09-30
Fixed issue with clang complaining about sprintf and vsprintf being depreciated

slinky55:
Clang was complaining about a security issue with sprintf and vsprintf, and that they were depreciated and produced a compiler error.


turol:
You should use either sizeof or a define so if someone changes the buffer definitions the function calls automatically stay in sync.

slinky55:
Good point, done.

kimkulling:
Merged, thansk for your contribution.

PR #4755, opened 2022-10-14
* fix bug reading ply file in case of presence of "end_header\n<BINARY_DATA>..." with <BINARY_DATA> starting by "\n"

emvivre:
* fix bug reading ply file in case of presence of "end_header\n<BINARY_DATA>..." with <BINARY_DATA> starting by "\n"


kimkulling:

PR #4757, opened 2022-10-16
Use correct flags for openddl for static libs

kimkulling:
- Update CMakeLists.txt for openddl lib
- Use correct configuration for openddl lib when building static lib
- closes  https://github.com/assimp/assimp/issues/4516


kimkulling:
Shall worl now.

PR #4783, opened 2022-11-02
Updated check for is_verbose_format on scene export

NaftaliD:
I believe the check for verbose was set up unlike the comment above it hints.

If a scene had AI_SCENE_FLAGS_NON_VERBOSE_FORMAT set then the code would ignore it and go on to check with MakeVerboseFormatProcess::IsVerboseFormat(pScene) and if indeed the scene was verbose it will result in a fatal exception when trying to run "GenVertexNormalsProcess".

The thing is if a scene didn't have the flag set (as the comment above says users creating scenes from scratch might not know of it) the code will take the lack of the flag as true and not check if the scene is indeed verbose, resulting in a later exception.

This change reverses the situation and makes it so that:
1. Flag is set - take that as true
2. Flag is not set - check if it was a user error (and act accordingly)
3. If the flag is set and the data is verbose it means the code will run the MakeVerboseProcess but for me, it didn't result in any harm, and its a much less likely scenario to happen rather than a user not defining their flag on scene creation.


turol:
Typo in comment message

code/Common/Exporter.cpp
```diff
@@ -370,7 +370,7 @@ aiReturn Exporter::Export( const aiScene* pScene, const char* pFormatId, const c
     // format. They will likely not be aware that there is a flag in the scene to indicate
     // this, however. To avoid surprises and bug reports, we check for duplicates in
     // meshes upfront.
-    const bool is_verbose_format = !(pScene->mFlags & AI_SCENE_FLAGS_NON_VERBOSE_FORMAT) || MakeVerboseFormatProcess::IsVerboseFormat(pScene);
+    const bool is_verbose_format = !((pScene->mFlags & AI_SCENE_FLAGS_NON_VERBOSE_FORMAT) || !MakeVerboseFormatProcess::IsVerboseFormat(pScene));
```
turol, line 373: Double negation like this does not look right. Are you sure it should not be 
```
(pScene->mFlags & AI_SCENE_FLAGS_NON_VERBOSE_FORMAT) || MakeVerboseFormatProcess::IsVerboseFormat(pScene)
```

NaftaliD: no, because then if the scene has the flag (the scene is not verbose) it will set is_verbose_format to true, the idea was to make the line better fool-proof and remove the possibility of a fatal exception.
I just read the comment above and was not sure that was the intent in the line.

maybe instead of double negative:
```
MakeVerboseFormatProcess::IsVerboseFormat(pScene) || !(pScene->mFlags & AI_SCENE_FLAGS_NON_VERBOSE_FORMAT)
```

This will result in the Verbose check always being run, and the flag as an extra step if the check results in negative.

kimkulling:
No progress, so I will close this one.

PR #4785, opened 2022-11-03
chore: add missing std moves and perfect forwards

Skylion007:
Optimize the code by adding a lot of missing std::move and a few std::forwards for perfect forwarding. This PR should reduce the number unnecessary copies and overall increase efficiency.


kimkulling (approved):
Looks fine.

PR #4786, opened 2022-11-03
Removed KHR_materials_pbrSpecularGlossiness, added KHR_materials_specular

Beilinson:
pbrSpecularGlossiness - https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Archived/KHR_materials_pbrSpecularGlossiness/README.md 
Has been superseded by:
KHR_material_specular - https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_specular

I replaced the outdated pbrSpecularGlossiness with the up-to-spec KHR_material_specular extension, which doesn't overwrite the base pbrMetallicRoughness material and can be used with any other KHR extensions.


code/AssetLib/glTF2/glTF2AssetWriter.inl
```diff
@@ -390,34 +390,31 @@ namespace glTF2 {
...
           unlit.SetObject();
           exts.AddMember("KHR_materials_unlit", unlit, w.mAl);
         }
 
+        if (m.materialSpecular.isPresent) {
+            Value materialSpecular(rapidjson::Type::kObjectType);
+            materialSpecular.SetObject();
+
+            MaterialSpecular &specular = m.materialSpecular.value;
+			if (specular.specularFactor != 0.f) {
```
turol, line 404: Broken indentation

Beilinson:
the failed tests are because it is attempting to run tests on the pbrSpecularGlossiness that I replaced, not due to legitimate issues

RichardTea:
Please don't remove _import_ of the deprecated extension, as this will leave assimp unable to properly import existing assets.

There will be a lot of existing assets using `KHR_materials_pbrSpecularGlossiness`, and they should continue to import as they did before.

Instead, Assimp should prefer the new `KHR_materials_specular` extension in both import and export, but support should not be removed in either direction.

- Any models using the deprecated extension should import as they did before.
- Assimp should export using `KHR_materials_specular` instead of `KHR_materials_pbrSpecularGlossiness`
- Add an exporter property so applications can choose to continue exporting using the deprecated `KHR_materials_pbrSpecularGlossiness`, should they need to.

Remember that the goal of the assimp project is to import as many formats as practicable, regardless of whether they are ancient history or the newest hotness.

Beilinson:
ok, will do so

Beilinson:
Quick q: 
According to this: https://kcoley.github.io/glTF/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/#best-practices
An asset with only the pbrSpecularGlossiness and `pbrSpecularGlossiness` in `extensionsUsed` will render as if no material. from my experience it is hit or miss, if the pbrSpecularGlossiness diffuse color has an opacity of 1.0 then it may be rendered, while with an opacity below 1.0 and alpha mode blend then it is just not rendered at all.
I propose to use pbrSpecularGlossiness then in addition with the regular pbr metallic-roughness, so on unsupported platforms the whole model will still load properly, as that was why I made the change in the first place, all models exported with pbrSpecularGlossiness were simply unsupported on any platform I attempted to open them on.

Beilinson:
> Models using both (should not happen) should import using the new `KHR_materials_specular` extension

> Models using both (should not happen) should import using the new KHR_materials_specular extension

Are you sure thats the right choice? I currently left it to import both although I agree that is a strange behavior

include/assimp/material.h
```diff
@@ -1000,6 +1000,7 @@ extern "C" {
 // Specular Color.
 // Note: Metallic/Roughness may also have a Specular Color
 // AI_MATKEY_COLOR_SPECULAR
+#define AI_MATKEY_USE_GLTF_PBR_SPECULAR_GLOSSINESS "$mat.useGltfPbrSpecularGlossiness", 0, 0
```
RichardTea, line 1003: Sorry, I didn't explain properly.
This should not be an AI_MATKEY_, it should be an exporter property.

Exporter properties are set by the application to customise the behaviour of the exporter, they're not part of the model.

`glTF2Exporter(const char *filename, IOSystem *pIOSystem, const aiScene *pScene,
            const ExportProperties *pProperties, bool binary);`

See:
https://github.com/assimp/assimp/blob/cfed74516b46a7c2bdf19c1643c448363bd90ad7/include/assimp/Exporter.hpp#L372

Beilinson: > Sorry, I didn't explain properly. This should not be an AI_MATKEY_, it should be an exporter property.
> 
> Exporter properties are set by the application to customise the behaviour of the exporter, they're not part of the model.

This creates an issue, if an asset is imported that uses the pbrSpecGlos extension there will be no way to guarantee that the exporter will use it since the importer can't use the SetPropertyBool of the exporter. 
With my implementation the imported material will havethe defined flag so that when the same model is exported that material will be unchanged and will also have the pbrSpecGlos extension.
Thoughts?

Beilinson:
> Sorry, I didn't explain properly. This should not be an AI_MATKEY_, it should be an exporter property.
> 
> Exporter properties are set by the application to customise the behaviour of the exporter, they're not part of the model.

This creates an issue, if an asset is imported that uses the pbrSpecGlos extension there will be no way to guarantee that the exporter will use it since the importer can't use the SetPropertyBool of the exporter. 
With my implementation the imported material will havethe defined flag so that when the same model is exported that material will be unchanged and will also have the pbrSpecGlos extension.
Thoughts?

code/AssetLib/glTF2/glTF2AssetWriter.inl
```diff
@@ -550,7 +570,7 @@ namespace glTF2 {
 
     inline void Write(Value& obj, Mesh& m, AssetWriter& w)
     {
-		/****************** Primitives *******************/
+        /****************** Primitives *******************/
```
turol, line 573: Whitespace

code/AssetLib/glTF2/glTF2AssetWriter.inl
```diff
@@ -980,7 +1004,7 @@ namespace glTF2 {
         if (d.mObjs.empty()) return;
 
         Value* container = &mDoc;
-		const char* context = "Document";
```
turol, line 983: Whitespace

test/unit/utglTF2ImportExport.cpp
```diff
@@ -220,7 +220,9 @@ TEST_F(utglTF2ImportExport, importglTF2AndExport_KHR_materials_pbrSpecularGlossi
             aiProcess_ValidateDataStructure);
     EXPECT_NE(nullptr, scene);
     // Export
-    EXPECT_EQ(aiReturn_SUCCESS, exporter.Export(scene, "glb2", ASSIMP_TEST_MODELS_DIR "/glTF2/BoxTextured-glTF-pbrSpecularGlossiness/BoxTextured_out.glb"));
+    ExportProperties props;
+    props.SetPropertyBool(AI_CONFIG_USE_GLTF_PBR_SPECULAR_GLOSSINESS, true);
+    EXPECT_EQ(aiReturn_SUCCESS, exporter.Export(scene, "glb2", ASSIMP_TEST_MODELS_DIR "/glTF2/BoxTextured-glTF-pbrSpecularGlossiness/BoxTextured_out.glb", 0, &props));
```
turol, line 225: This should test both old and new ways.

turol (changes requested):
Broken whitespace and the unit test needs to test both code paths.

Also the history contains broken commits and way too many merges. This will make bisecting painful. Squash, rebase and force-push please.

code/AssetLib/glTF2/glTF2AssetWriter.inl
Beilinson: I actually fixed it, it was using tabs before while the rest of the file is white space

turol: That should go in a separate commit then. Mixing functional and whitespace changes makes reviewing harder and can create unnecessary conflicts.

code/AssetLib/glTF2/glTF2Asset.inl
```diff
@@ -1250,7 +1250,7 @@ inline void Material::Read(Value &material, Asset &r) {
     ReadMember(material, "alphaCutoff", this->alphaCutoff);
 
     if (Value *extensions = FindObject(material, "extensions")) {
-        if (r.extensionsUsed.KHR_materials_pbrSpecularGlossiness) {
+        if (r.extensionsUsed.KHR_materials_pbrSpecularGlossiness) {
```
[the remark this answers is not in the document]
kimkulling: Fixed.

kimkulling (approved):
Looks fine now.

kimkulling:

PR #4792, opened 2022-11-08
Modernize smartptrs and use C++11 literals

Skylion007:
I added some more clang-tidy checks:
* I applied several checks that should make the codebase more readable:
* Use make_shared when possible which is more efficient than calling the ctor or reset.
* Remove redundant calls to `.get()` for smartprs
* Use nullptr and bool literals whenever possible.


kimkulling (approved):
Thanks for the great work!

kimkulling:
Merged, thanks a lot for your great work.

PR #4804, opened 2022-11-21
Kimkulling/use zip at submodule

kimkulling:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "./contrib/zip"]
+	path = ./contrib/zip
+	url = https://github.com/kuba--/zip.git
```
Skylion007, line 3:
```suggestion
[submodule "contrib/zip"]
	path = contrib/zip
	url = https://github.com/kuba--/zip.git
```
Seems like the '.' might be throwing it off? I tried to regenerate the submodule locally, and it worked with `git submodule status`, where the cloned version of this branch did not (at least on OSX and gave the same error as the CI).

Skylion007: Ping @kimkulling if you want to revive this PR.

kimkulling:
Will not work, we need to deal with all the warnings.

PR #4816, opened 2022-11-29
[BlenderDNA.h] Declare explicit specializations

tkoeppe:
It is an ODR violation to use a template specialization for which an explicit specialization is defined somewhere whose declaration is not visible at the point of use.

Found via clang's `-Wundefined-func-template` warning.


kimkulling (approved):

kimkulling:

tkoeppe:
My great pleasure!

PR #4824, opened 2022-12-02
Fixed bug when exporting binary FBX

umesh-huawei:
Fixed vector subscript out of range bug when NULL_RECORD is passed to PutString and is actually null (i.e a string "" with size 0). Replaced it with a std::string initialized by '\0' time NumNullRecords. The cause of this issue might be that the std::string constructor gets the correct const char* but stops after seeing that the first element is '\0'.


kimkulling (approved):
Looks fine to me.

kimkulling:

Issue #4831, opened 2022-12-08
Bug: Issue 49797: assimp:assimp_fuzzer: Heap-buffer-overflow in Assimp::ScenePreprocessor::ProcessMesh


kimkulling:
Status: New
Owner: ----
CC: kim.k...@[googlemail.com](http://googlemail.com/) 
Labels: Restrict-View-Commit ClusterFuzz Stability-Memory-AddressSanitizer Reproducible Engine-libfuzzer OS-Linux Security_Severity-Medium Proj-assimp Reported-2022-08-05
Type: Bug-Security

New issue 49797 by ClusterFuzz-External: assimp:assimp_fuzzer: Heap-buffer-overflow in Assimp::ScenePreprocessor::ProcessMesh
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=49797

Detailed Report: https://oss-fuzz.com/testcase?key=4578023118077952

Project: assimp
Fuzzing Engine: libFuzzer
Fuzz Target: assimp_fuzzer
Job Type: libfuzzer_asan_assimp
Platform Id: linux

Crash Type: Heap-buffer-overflow READ 4
Crash Address: 0x60e000000298
Crash State:
  Assimp::ScenePreprocessor::ProcessMesh
  Assimp::ScenePreprocessor::ProcessScene
  Assimp::Importer::ReadFile
  
Sanitizer: address (ASAN)

Recommended Security Severity: Medium

Crash Revision: https://oss-fuzz.com/revisions?job=libfuzzer_asan_assimp&revision=202208040605

Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=4578023118077952

Issue filed automatically.

See https://google.github.io/oss-fuzz/advanced-topics/reproducing for instructions to reproduce this bug locally.
When you fix this bug, please
  * mention the fix revision(s).
  * state whether the bug was a short-lived regression or an old bug in any stable releases.
  * add any other useful information.
This information can help downstream consumers.

kimkulling:
Still open!

kimkulling:
Shall be closed by https://github.com/assimp/assimp/pull/4836.

PR #4836, opened 2022-12-10
Fixes Heap-buffer-overflow READ 4 in Assimp::ScenePreprocessor::ProcssMesh

sashashura:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=49797

The root cause of the crash is that [InternReadFile](https://github.com/assimp/assimp/blob/0fdae2879d78864693ee730610dcf8ee10707875/code/AssetLib/OFF/OFFLoader.cpp#L108) allocates 9 faces.
```cpp
    mesh->mNumFaces = numFaces;
    aiFace* faces = new aiFace[mesh->mNumFaces];
```

Later in the same function in [`for (unsigned int i = 0; i < numFaces; )`](https://github.com/assimp/assimp/blob/0fdae2879d78864693ee730610dcf8ee10707875/code/AssetLib/OFF/OFFLoader.cpp#L284) loop in goes into
```cpp
        if(!idx || idx > 9) {
            ASSIMP_LOG_ERROR("OFF: Faces with zero indices aren't allowed");
            --mesh->mNumFaces;
            continue;
        }
```
which decrements the number of faces an continues the loop. Note that the loop doesn't have `i++`, so it repeats the loop with the same `i` value.
This allows to decrement the number of faces more than its value 9, which leads to integer underflow and it becomes very big (4294967293).
Later it leads to heap read out of bounds since only 9 faces were allocated.

### DIFF
--- a/code/AssetLib/OFF/OFFLoader.cpp
+++ b/code/AssetLib/OFF/OFFLoader.cpp
@@ -290,11 +290,12 @@ void OFFImporter::InternReadFile( const std::string& pFile, aiScene* pScene, IOS
         sz = line; SkipSpaces(&sz);
         idx = strtoul10(sz,&sz);
         if(!idx || idx > 9) {
-	    ASSIMP_LOG_ERROR("OFF: Faces with zero indices aren't allowed");
+	        ASSIMP_LOG_ERROR("OFF: Faces with zero indices aren't allowed");
             --mesh->mNumFaces;
+            ++i;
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

kimkulling:
- closes https://github.com/assimp/assimp/issues/4831

kimkulling (approved):
Looks fine.

PR #4839, opened 2022-12-12
Actions: curse artifacts to be generated again

kazu0617:
This PR fixes an issue where repositories no longer publish artifacts because they currently have a configuration but no matching criteria.
based pr, first time to add publish artifact, but not worked now: https://github.com/assimp/assimp/pull/3322


kazu0617:
I tried to work it in "SharedProject:fix-artifact".

For visitors:
This PR also works in #4938 plz double checks.

PR #4846, opened 2022-12-18
illegal token on right-side-of ::Windows

rohit-kumar-j:
Error without:
```
assimp\material.inl(101,47): message : error recovery skipped: ') ?'
```

Reference : https://stackoverflow.com/questions/2561368/illegal-token-on-right-side-of


kimkulling (approved):
I removed the undef max. Seems that only min is causing the problem.

kimkulling:

PR #4858, opened 2023-01-03
{cmake} Remove dead code

asmaloney:
`BUILD_LIB_TYPE` isn't used anywhere.

I think the intent with _add_definitions_ here was to affect cmake (?), but it's actually just adding `#define DDL_STATIC_LIBRARY OFF` which isn't used in the code anywhere.


kimkulling (approved):
That is correct. This usage of this option was removed by an older pR and this dead code is a deprecated artifact. Thanks a lot for figuring this out.

kimkulling:
Merged, thank a lot for your contribution.

PR #4872, opened 2023-01-12
Generalize JoinVerticesProcess for multiple UV and color channels

drbct:
On the current master branch, `JoinVerticesProcess` silently ignores all UV channels except the first as well as all color channels. This causes vertices to be merged unexpectedly if they differ only in one of those attributes. There is a parameter in the local comparison function `areVerticesEqual` which enables the associated checks but it is always set to `false`.

The proposed change replaces the stateless comparator used in `JoinVerticesProcess::ProcessMesh` with one that caches the number of UV and color channels in the current mesh. This allows `areVerticesEqual` to check all relevant vertex attributes without performing unnecessary comparisons.

### DIFF
--- a/code/PostProcessing/JoinVerticesProcess.cpp
+++ b/code/PostProcessing/JoinVerticesProcess.cpp
@@ -105,7 +105,11 @@ void JoinVerticesProcess::Execute( aiScene* pScene) {
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine.

kimkulling:

PR #4878, opened 2023-01-15
Fix #4877

### LINKED ISSUE #4877: Bug: Collada: Whitespace between <v> tag and first number is not removed
**Describe the bug**
Colladas XML parser does not remove whitespace between a <v> tag and the first number. The first call to strtoul10() then returns 0 and the index pairs are not properly parsed. This leads to memory corruption and segfault.

**To Reproduce**
run "assimp info Lokfuerer2.dae.txt" with attached dae file, crash occurs. Without the space in line 666 after the <v> tag the file is parsed fine. 

**Expected behavior**
assimp should parse the file and show the file information for well-formed dae files and handle ill-formed dae files without segfaulting.

 - OS: Windows, Linux
 - Current master, 5.x and earlier

**Additional context**
I am proposing a fix for the issue with a PR.
[Lokfuerer2.dae.txt](https://github.com/assimp/assimp/files/10421387/Lokfuerer2.dae.txt)

MMory:
Remove whitespace between a <v> tag and the first number, otherwise first call to strtoul10() returns 0 and the indices are broken, leading to possible out-of-bound access and memory corruption/crash.

I am an outsider and not familiar with the code base, hence I kindly ask whether this issue is present in more instances in the code base. There are many places where SkipSpacesAndLineEnd() is called.

### DIFF
--- a/code/AssetLib/Collada/ColladaParser.cpp
+++ b/code/AssetLib/Collada/ColladaParser.cpp
@@ -762,6 +762,7 @@ void ColladaParser::ReadControllerWeights(XmlNode &node, Collada::Controller &pC
                 if (text == nullptr) {
                     throw DeadlyImportError("Out of data while reading <vertex_weights>");
                 }
+                SkipSpacesAndLineEnd(&text);
                 it->first = strtoul10(text, &text);
                 SkipSpacesAndLineEnd(&text);
                 if (*text == 0) {


MMory:
Fixes #4877

kimkulling (approved):
Looks fine!

PR #4879, opened 2023-01-16
Fix #4262 Build With M3D Import Only

### LINKED ISSUE #4262: Bug: likely m3d build error with ASSIMP_BUILD_NO_EXPORT
M3DImporter uses M3DWrapper:
https://github.com/assimp/assimp/blob/master/code/AssetLib/M3D/M3DImporter.cpp#L190
which is disabled with ASSIMP_BUILD_NO_EXPORT:
https://github.com/assimp/assimp/blob/master/code/AssetLib/M3D/M3DWrapper.h#L51
as well as required `m3d.h` inclusion.

krishty:
`M3DWrapper.h` is designed to omit the definition of `class M3DWrapper` if neither M3D import nor M3D export are compiled. 608bccd9cf3a28c55163c10e84aec025293b7ab0 touched the corresponding preprocessor checks and introduced a bug:
```
#ifndef ASSIMP_BUILD_NO_M3D_IMPORTER
#if !(ASSIMP_BUILD_NO_EXPORT || ASSIMP_BUILD_NO_M3D_EXPORTER)
class M3DWrapper {
```
When compiling
- with M3D import enabled,
- but with either export generally disabled or M3D export disabled specifically, These checks evaluate to the wrong result and skip the definition, leading to a build failure in dependent code.
```
#if 1 // import enabled
#if !(1 || 1) // export disabled and M3D export disabled
```
This commit fixes the check to compile the definition if neither import is disabled.

### DIFF
--- a/code/AssetLib/M3D/M3DWrapper.cpp
+++ b/code/AssetLib/M3D/M3DWrapper.cpp
@@ -39,8 +39,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 ----------------------------------------------------------------------
 */
-#ifndef ASSIMP_BUILD_NO_M3D_IMPORTER
-#if !(ASSIMP_BUILD_NO_EXPORT || ASSIMP_BUILD_NO_M3D_EXPORTER)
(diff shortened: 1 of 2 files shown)


kimkulling (approved):
Looks fine, thanks for the fix.

kimkulling:

Issue #4435, opened 2022-03-09
glTF2 Importer uses incorrect code to calculate horizontal FOV


jamesn031199:
The [importer code](https://github.com/assimp/assimp/blob/master/code/AssetLib/glTF2/glTF2Importer.cpp#L846) for glTF2.0 assumes multiplying the vertical FOV by the aspect ratio is how horizontal FOV is calculated.
However, this is not correct and the correct way to convert between the two is by [this method](https://i.imgur.com/mz5NpQu.png) referenced in [this Reddit post](https://www.reddit.com/r/Planetside/comments/1xl1z5/brief_table_for_calculating_fieldofview_vertical/).

kvnnap:
This [code ](https://github.com/assimp/assimp/blob/aae93691b9923e13af089a2a092fdd75bbfab489/code/AssetLib/glTF2/glTF2Importer.cpp#L911C20-L911C35)does not respect the aiCamera's interface. It computes the correct horizontal FOV, however, the camera struct `aiCamera` [mHorizontalFOV](https://github.com/assimp/assimp/blob/aae93691b9923e13af089a2a092fdd75bbfab489/include/assimp/camera.h#L144C6-L144C6) is defined as being **half** the horizontal field of view according to the documentation. I have noticed this issue since when I import FBX files the cameras load properly, however with gltf2 the FOV is always wrong. I tested a scene with cameras in this [online gltf viewer](https://gltf-viewer.donmccurdy.com/), and indeed they load fine here. Additionally, I checked the FBX importer code [here](https://github.com/assimp/assimp/blob/aae93691b9923e13af089a2a092fdd75bbfab489/code/AssetLib/FBX/FBXConverter.cpp#L449C34-L449C34). This code computes half the horizontal FOV as seen below:

```c++
// FBX fov is full-view degrees. We want half-view radians.
out_camera->mHorizontalFOV = AI_DEG_TO_RAD(fov_deg) * 0.5f;
```

The code [here](https://github.com/assimp/assimp/blob/aae93691b9923e13af089a2a092fdd75bbfab489/code/AssetLib/glTF2/glTF2Importer.cpp#L911C17-L911C17) should therefore change from:

```c++
aicam->mHorizontalFOV = 2.0f * std::atan(std::tan(cam.cameraProperties.perspective.yfov * 0.5f) * ((aicam->mAspect == 0.f) ? 1.f : aicam->mAspect));
```

to 

```c++
[...]

PR #4886, opened 2023-01-16
Fix: fix incorrect math for calculating the horizontal FOV of a perspective camera in gltf2 import #4435

shimaowo:
Uses the correct math, as detailed in #4435.

This also preserves the behavior that if mAspect == 0, the computed horizontal fov will just pass through the vertical fov unchanged, so that it can be recovered and computed properly according to the gltf2 spec later.

Further math reference [here](https://stackoverflow.com/questions/5504635/computing-fovx-opengl), with a nice visual demo of why this matters [here](https://twgljs.org/examples/fov-checker.html).


kimkulling:

PR #4889, opened 2023-01-17
Fix MSVC Warnings With “emplace_back()”

krishty:
Several places in the code call `std::vector<aiVector3D>.emplace_back(0, 0, 0)`. The constructor of `aiVector3D` actually expects arguments of the type `ai_real`, (alias of `float` if compiling without `ASSIMP_DOUBLE_PRECISION`) but the literal `0` is of type `int`.

`emplace_back()` does support promotion, but `int` to `float` is a potentially lossy conversion. tl;dr: On warning level 4, MSVC spits out a very deeply nested `warning C4244: 'argument': conversion from '_Ty' to 'TReal', possible loss of data with _Ty=int and TReal=ai_real`.

I used `static_cast<ai_real>()` in the affected places because both `ai_real(0)` and `(ai_real)0` are regarded as C-style casts and frowned upon by various linters, e.g. clang-tidy.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #4866, opened 2023-01-05
Bug: .PLY comments support


jonathanpokrass:
**Describe the bug**
.ply file parser fails when there is a comment after the scene definition,
 example file:
```
ply
format binary_little_endian 1.0
element vertex 1
comment Points
property float x
property float y
property float z
end_header
000000000000
```
error message:

> “ERROR: Failed to load file: Invalid .ply file: Unable to extract mesh data” 

without the comment (line 4), all is working.

PencilAmazing:
Should assimp discard comments or attempt to preserve them? Does the internal assimp format support comments?

PR #4899, opened 2023-01-23
Fix issue #4866 by continuing to resetting read loop after hitting a comment

PencilAmazing:
Simple one line fix.


test/models/PLY/issue623.ply
```diff
@@ -2,6 +2,7 @@ ply
 format ascii 1.0
 comment Created by Blender 2.77 (sub 0) - www.blender.org, source file: ''
 element vertex 24
+comment Manually added comment line to fix issue #4866
```
kimkulling, line 5: Do we need this?

kimkulling (commented):
One finding.

test/models/PLY/issue623.ply
PencilAmazing: It's for test coverage I guess. Comments can be anywhere in the header, and I added one to 'simulate' the issue.

kimkulling:

PR #4900, opened 2023-01-23
Add S3O Loader (more to come).

jochumdev:
I'm very rusty in C++, please have a closer look.


turol:
Don't add your own vscode or other local project files. This will break things for anyone else using it.

kimkulling:
Thanks for the PR. Could you provide some information about the used data-format? I am not aware of S30. Sorry for that.

jochumdev:
Sadly there's not much information about the S3O format available on the net.

This code is based on [spring](https://github.com/springrts/spring/blob/develop/rts/Rendering/Models/S3OParser.cpp) which is GPLv2 licensed, sorry for not telling that in advance.

s3o -> assimp -> obj seems to work fine here (obj imported into blender 3.x)

turol:
It appears you have indeed directly copied GPL code into this PR. Since assimp is BSD-licensed this is not acceptable.

jochumdev:
@turol while this is heavily based on the spring code, it's not a copy.

turol:
Compare these two:
https://github.com/spring/spring/blob/6777eb1ddb3ebebba3764f46a5421f04d48b5c10/rts/Rendering/Models/s3o.h#L69
https://github.com/assimp/assimp/pull/4900/files#diff-095f13e63fcdf4f05e0aa91db5c725eb367eb0b7e0db5288fbb80d38be386e54R97

They're identical except for formatting. You copied GPL-licensed code. This is not allowed.

PR #4901, opened 2023-01-23
Optimized usedVertexIndices in JoinVerticesProcess by using bitmask instead of unordered_set

AdamCichocki:


turol:
Do you have benchmarks?

AdamCichocki:
Cuts the JoinVertices time in half, while Join vertices take almost half the time of the whole import of models with simple vertices (i.e. no tangents etc).

Skylion007:
Wouldn't the `std::vector<bool>` specialization be a bit more readable / helpful here? https://en.cppreference.com/w/cpp/container/vector_bool

AdamCichocki:
@Skylion007 Fair enough, updated the PR.

Skylion007 (approved):

kimkulling (approved):
Looks fine, much better!

kimkulling:

PR #4932, opened 2023-02-02
Unit test warning fixes

turol:
GCC was warning about possibly uninitialized variables. Initialize them and use values which are nonzero and distinct from each other and any real value passed to any set method. This should prevent any false positives from zero initialization.


kimkulling (approved):
Looks fine, many thanks!

kimkulling:

PR #4933, opened 2023-02-02
Fix: Close leak.

kimkulling:
- closes https://github.com/assimp/assimp/issues/3416

### DIFF
--- a/include/assimp/MemoryIOWrapper.h
+++ b/include/assimp/MemoryIOWrapper.h
@@ -2,8 +2,7 @@
 Open Asset Import Library (assimp)
 ----------------------------------------------------------------------
 
-Copyright (c) 2006-2022, assimp team
-
+Copyright (c) 2006-2023, assimp team
 
 All rights reserved.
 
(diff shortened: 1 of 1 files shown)


include/assimp/MemoryIOWrapper.h
```diff
@@ -162,7 +157,11 @@ class MemoryIOSystem : public IOSystem {
     }
 
     /** Destructor. */
-    ~MemoryIOSystem() = default;
+    ~MemoryIOSystem() {
+        for (auto &it : created_streams) {
+            MemoryIOSystem::Close(it);
```
turol, line 162: `MemoryIOSystem::Close` modifies `created_streams` while the destructor is iterating over it.

turol:
Try something like this:
```
while (!created_streams.empty()) {
  MemoryIOSystem::Close(created_streams.front());
}
```
Not very nice but probably the best we can do since the destructor has to remove itself from the map

kimkulling:
It seems that there is a mismatch in the way other importers aree closing the files. I will check it this weekend.

kimkulling:
Seems not to fix the issue.

Issue #3093, opened 2020-03-28
Terragen importer is broken on x64 Platform


kimkulling:
When trying to import the terragen-examples the loading breaks caused by the following exception:

`End of file or read limit was reached.`

turol:
Bisected.
e8d2b840178b2b22336eb02b2535f32f3d5cc37b is first bad commit.

@kimkulling Please 
1. Don't make monster commits like this
2. Make sure all of your commits build on all platforms. I had to make a lot of local fixes during bisect because many commits were just completely broken.

kimkulling:
That is the reason for having our CI. I am wondering why there was not red build. Which platforms have I missed?

And you are right: I will avoid these in the future, sorry for that.

turol:
There was a test but it was commented out. You might want to check all the other tests and then that all formats have a test.

PR #4934, opened 2023-02-03
Fix Terragen loader

turol:
Fixes https://github.com/assimp/assimp/issues/3093


kimkulling (approved):
Looks fine. But why has this change fix the problem?

kimkulling:

turol:
Because in the original the value assigned to dtt and therefore used in addition was between 0-3. After your change the check was done against a different value than the addition.

PR #4935, opened 2023-02-04
ACLoader: add support for reading more than one texture per object

IOBYTE:
The .ac format supports 0 or 1 textures per object.  The .acc extended format supports 0 to 4 textures per object.   Currently the last texture is the one selected when multiple textures are read.  This loader doesn't support multiple textures so the last layer above the base texture is used.  This PR selects the first (base) texture to use.


kimkulling:

Issue #4871, opened 2023-01-11
Bug: Assimp 5.2.5 Reader Crash on ply (Double Precision) 


Nosenzor:
**Describe the bug**
Hi Here,
I compiled assimp 5.2.5  with ASSIMP_DOUBLE_PRECISION.
Assimp crash on ReadFile on the attached file.
`Assimp::Importer importer;
    
    const aiScene* scene = importer.ReadFile(iInputFile.string(),
                                             aiProcess_OptimizeGraph |
                                               aiProcess_Triangulate |
                                               aiProcess_JoinIdenticalVertices); `

I will try to reproduce with another platform and add a comment later here.
**To Reproduce**
Compile Assim

**Expected behavior**
No Crash ;-)

**Screenshots**

**Platform (please complete the following information):**
 - OS: Mac OS 13.1
 Compiler :
 - Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin22.2.0

**Additional context**
[73612.ply.zip](https://github.com/assimp/assimp/files/10391212/73612.ply.zip)

Nosenzor:
Reproduced on MacOs 10.15.7 (intel) with Apple clang version 12.0.0 (clang-1200.0.32.29)
[libassimpd.5.dylib] Assimp::TriangulateProcess::TriangulateMesh(aiMesh*) 0x0000000106076570
[libassimpd.5.dylib] Assimp::TriangulateProcess::Execute(aiScene*) 0x0000000106075d99
[libassimpd.5.dylib] Assimp::BaseProcess::ExecuteOnScene(Assimp::Importer*) 0x0000000105fbd13e
[libassimpd.5.dylib] Assimp::Importer::ApplyPostProcessing(unsigned int) 0x0000000105fd03e3
[libassimpd.5.dylib] Assimp::Importer::ReadFile(char const*, unsigned int) 0x0000000105fcee85
it looks like an out of bound access. I don't have yet  the exact line that leads to the crash.

Nosenzor:
Reproduced on Windows too and it gives me the line of crash : 
            // Collect all vertices of of the polygon.
            for (tmp = 0; tmp < max; ++tmp) {
                **temp_verts3d[tmp] = verts[idx[tmp]];**
            }

tmp=3
max=61
idx[tmp]=9175808

kimkulling:
Thanks for your investigations. Seems to be not a crash anymore when using the latest master. But the importer for double precision is still not ok.

[...]

PR #4936, opened 2023-02-04
Fixes PLY reader when the header ends with \r\n

TinyTinni:
When "getNextBlock" was called after "getNextLine", the pointer could still on the newline.

The pointer to a newline could not advance enough, when the line ended with \r\n. The resulting buffer was correct, as the buffer range went from <start> until \r, but that the pointer increased by just 1 could lead to the problem that the next pointer points at \n, which is still part of the newline and therefore, "getNextBlock" got 1 byte too much.

Refs Issue #4871

I added a test with a copy of the PLY file inside the test directory. I couldn't figure out a good name. I hope this is ok. A name idea would be appreciated.

The comment removed is an outdated comment from me.


kimkulling (approved):
Looks fine.

kimkulling:

PR #4944, opened 2023-02-09
Fix double free when the mesh contains duplicate bones.

dmarcinkovic:
The aiMesh class destructor iterates the array of bones and deletes them. 
In the case of duplicate bones, the same bone is attempted to be deleted twice which causes the 'double free or corruption error. 
We can store bones in the set and delete each bone only once.

### DIFF
--- a/include/assimp/mesh.h
+++ b/include/assimp/mesh.h
@@ -59,6 +59,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <assimp/types.h>
 
 #ifdef __cplusplus
+#include <unordered_set>
+
 extern "C" {
 #endif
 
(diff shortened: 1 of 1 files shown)


cstdbool:
Noticed this bug with FBX files which have bones with duplicate names.

kimkulling:
I am not able to update your branch. Could you do this, please? That would help a lot.

kimkulling:
I see that the problem exists and that we need a solution for that. UNfortunately, this solution will only work for c++. The stuff gets inlined only when we are using a c++ compiler.

turol:
Why are there duplicate bones in the first place? Can we just define that bones must be unique and fix any loaders that create duplicates?

dmarcinkovic:
I will update the branch and I'll try to provide a solution that will work with the C compiler as well.

dmarcinkovic:
@turol, I got a double free or corruption error when I tried to import an fbx created with the Autodesk Maya tool. 
I do agree that the Assimp library should refuse to load an fbx file when the mesh contains duplicate bones, and it is exactly what Assimp does. However, the problem is in the cleanup process because the same memory address is attempted to be deleted twice.

turol:
Can you share a failing model?

dmarcinkovic:
Unfortunately, the model I used is proprietary.

kimkulling:
Will get merged by https://github.com/assimp/assimp/pull/5291

PR #4946, opened 2023-02-09
Add skeleton doc

kimkulling:
- Please comment

### DIFF
--- a/include/assimp/mesh.h
+++ b/include/assimp/mesh.h
@@ -952,7 +952,7 @@ struct aiMesh {
 #endif // __cplusplus
 };
 
-*
+/**
  * @brief  A skeleton bone represents a single bone is a skeleton structure.
  *
  * Skeleton-Animations can be represented via a skeleton struct, which describes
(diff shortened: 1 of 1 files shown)


include/assimp/mesh.h
```diff
@@ -1007,41 +1025,53 @@ struct aiSkeletonBone {
...
         mWeights = nullptr;
     }
 #endif // __cplusplus
 };
 /**
- *  @brief
+ * @brief A skeleton represents the bone hierarchy of an animation.
+ *
+ * Skeleton animations can be described as a tree of bones:
+ *                  root
```
jackalbe, line 1039: It might be good to be explicit which mBones bone is the root in this doc. Since the aiSkeletonBone constructor sets the mParent for all new aiSkeletonBone's to -1, it is hard to know how to determine the root node. Is it the first node in the mBones?

kimkulling: Good question. Initially the root node will not be set when you have to use the default constructor. I will introduce another constructor where you can explicitly specify which root node shall be used.

PR #4963, opened 2023-02-16
Fix a leak in FBXDocument when duplicate object IDs are found

avaneyev:
When a duplicate ID is encountered, existing LazyObject is overwritten. Previously allocated instance leaks.

This change deletes the previously allocated instance before overwriting the pointer.


kimkulling (approved):
Looks fine.

kimkulling:

PR #4965, opened 2023-02-16
Fix implicit conversion errors on macOS

aaronmjacobs:
These changes fix 'Implicit conversion loses integer precision' errors by explicitly casting to the desired type. I am currently running Apple clang version 14.0.0 (clang-1400.0.29.202) on macOS 13.2.1 (22D68).

### DIFF
--- a/code/AssetLib/MDL/MDLLoader.cpp
+++ b/code/AssetLib/MDL/MDLLoader.cpp
@@ -975,7 +975,7 @@ void MDLImporter::CalcAbsBoneMatrices_3DGS_MDL7(MDL::IntBone_MDL7 **apcOutBones)
                     }
 
                     // store the name of the bone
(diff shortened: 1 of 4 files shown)


turol:
This might cause warnings from MSVC about potential loss of data since you're casting to a smaller type.

You should use C++ style casts instead of C -style casts.

aaronmjacobs:
I've updated the casts to be C++ style (except for miniz, since that is a C library).

I'm not sure I understand the concerns about warnings in MSVC. I would expect this to remove warnings (if there are any), as it is converting the casts from being implicit to explicit.

kimkulling (approved):
Looks fine. I am wondering why other compilers haven't found these things before.

kimkulling:

PR #4985, opened 2023-02-28
Two bug fixes in Python port.

MakerOfWyverns:
Bug fix: Correctly set the name of the attributes. PR #3940 did not seem to update in Master?
Bug fix: We should not be accessing `mPrivate` according to structs.Scene and fixes a crash when `mPrivate` points to invalid data.


MakerOfWyverns:
I misread the commit #3940. The rest of the code assumes it uses the `name` as the attribute, not `m`. So, if we want to change that attribute name, we need to fix other areas of the code to match that change. My change reverts it back so it is usable again. Sorry for any misunderstanding.

kimkulling (approved):
Good catch, tanks for your great work.

kimkulling:

Issue #4998, opened 2023-03-06
Out-of-memory for some gltf exported with "Sketchup glTF Exporter by Centaur"


FlorianBorn71:
**Describe the bug**
Assimp seems to run into exactly the same issue that is described here:
[https://github.com/KhronosGroup/UnityGLTF/issues/197](https://github.com/KhronosGroup/UnityGLTF/issues/197)

**More details after my investigation:** Gltf allows to share the attribute reference between meshes. However, function glTF2Importer::ImportMeshes does not consider this and allocates the same buffer n times. Our sample model, exported by "Sketchup glTF Exporter by Centaur", has only 2M vertices, but it's shared across ~20k mesh parts **which results in allocating 20k * 2M vertices**.

I tried to hack in the sharing, but it will not work further down in the processing pipeline. I'm leaning towards shrinking the vertex array to the actual vertices that are used by the index buffer. I think similar stuff is one by the FBX importer?

Any suggestions?

FlorianBorn71:
I have a suggested PR open here that reduces memory footprint for the customer model **by factor 1000x** (!)

https://github.com/assimp/assimp/pull/5003

GLTF Models without shared attributes won't profit from this fix.

Let me know what you think

FlorianBorn71:
[SharedAttributesOOM.zip](https://github.com/assimp/assimp/files/11039567/SharedAttributesOOM.zip)

The attached zip contains a repro case gltf file. Without the fix, loading through Assimp passes out after allocating a few hundred GB of memory. With the fix, loading peaks at ~170MB, so more than 1000x improvement.

This is an artificially generated model to expose the problem, but our customer had a similar real model, where the fix saves > 1000x memory.

kimkulling:
Fixed

PR #5003, opened 2023-03-09
GLTF Importer: Build a list of the actual vertices so it works well with shared attribute lists

FlorianBorn71:
This PR addresses this bug:
[Out-of-memory for some gltf exported with "Sketchup glTF Exporter by Centaur"](https://github.com/assimp/assimp/issues/4998)

It shows that the memory consumption during import for said customer model now goes down f**rom >800GM (->out-of-memory) to <700MB, so more than 1000x !**

In my opinion this change makes sense, but I wanted to ask if this looks like a good approach to the problem for you as well.
Of course this needs more testing with other assets,


code/AssetLib/glTF2/glTF2Importer.cpp
```diff
@@ -641,20 +678,15 @@ void glTF2Importer::ImportMeshes(glTF2::Asset &r) {
             aiFace *facePtr = nullptr;
             size_t nFaces = 0;
 
-            if (prim.indices) {
-                size_t count = prim.indices->count;
-
-                Accessor::Indexer data = prim.indices->GetIndexer();
-                if (!data.IsValid()) {
-                    throw DeadlyImportError("GLTF: Invalid accessor without data in mesh ", getContextForErrorMessages(mesh.id, mesh.name));
```
FlorianBorn71, line 649: This error is now thrown above

code/AssetLib/glTF2/glTF2Importer.cpp
```diff
@@ -453,6 +453,12 @@ void glTF2Importer::ImportMeshes(glTF2::Asset &r) {
     unsigned int k = 0;
     meshOffsets.clear();
 
+
+    const unsigned int unusedIndex = ~0;
+    std::vector<unsigned int> usedVertexIndices;
+    std::vector<unsigned int> reverseMappingIndices;
```
jankrassnigg, line 459: For containers sizes that are unknown upfront, a deque usually performs better, since it never needs to reallocate existing data. Both reverseMappingIndices and indexBuffer don't seem to be needed to be contiguous, you only access individual elements by index, so they could be deque's, unless you have a good approximate size to reserve them with up front.

code/AssetLib/glTF2/glTF2Asset.inl
```diff
@@ -980,18 +981,20 @@ void Accessor::ExtractData(T *&outData) {
...
-    if (stride == elemSize && targetElemSize == elemSize) {
+    outData = new T[usedCount];
+    if (remappingIndices == nullptr && stride == elemSize && targetElemSize == elemSize) {
         memcpy(outData, data, totalSize);
     } else {
-        for (size_t i = 0; i < count; ++i) {
-            memcpy(outData + i, data + i * stride, elemSize);
+        for (size_t i = 0; i < usedCount; ++i) {
+            size_t srcIdx = remappingIndices != nullptr ? static_cast<size_t>((*remappingIndices)[i]) : i;
+            memcpy(outData + i, data + srcIdx * stride, elemSize);
```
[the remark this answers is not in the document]
FlorianBorn71: I considered breaking this up into two loops, however on a PC the CPU's branch prediction works so well that you will probably not notice any difference in timings. I thought it would be more worthwhile to keep the code compact.

code/AssetLib/glTF2/glTF2Importer.cpp
FlorianBorn71: Good point, I'll consider this

code/AssetLib/glTF2/glTF2Asset.inl
```diff
@@ -980,18 +981,20 @@ void Accessor::ExtractData(T *&outData) {
     }
 
     const size_t maxSize = GetMaxByteSize();
-    if (count * stride > maxSize) {
-        throw DeadlyImportError("GLTF: count*stride ", (count * stride), " > maxSize ", maxSize, " in ", getContextForErrorMessages(id, name));
+    if (usedCount * stride > maxSize) {
```
FlorianBorn71, line 984: I just noticed that this error check does not make sense if we pass an index list

code/AssetLib/glTF2/glTF2Asset.inl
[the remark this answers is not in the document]
FlorianBorn71: Edit: Now that I realized that I have to do separate index range checking in the index case, I will break it up into two loops

code/AssetLib/glTF2/glTF2Importer.cpp
FlorianBorn71: Ah, now I remember: While I do like the ideas of deque's, the std version of a deque is completely unusable imho, because it has a ridiculously small buffer size. So leaving it like this. BTW, the reallocation concern is the reason why I declared arrays outside both loops, so it reuses pre-warmed arrays

jankrassnigg: The deque implementation of some major compiler vendor(s) are debatable, indeed. The no-realloc argument is valid anyway. Obviously there is a trade-off point where either std::vector or std::deque is better, and without concrete testing it is really hard to say which one is better. Since you already measured a major speedup with the current solution, I guess it's good enough.

code/AssetLib/glTF2/glTF2Importer.cpp
```diff
@@ -583,23 +620,23 @@ void glTF2Importer::ImportMeshes(glTF2::Asset &r) {
...
                             aiVector3D *positionDiff = nullptr;
-                            target.position[0]->ExtractData(positionDiff);
+                            target.position[0]->ExtractData(positionDiff, vertexRemappingTable);
                             for (unsigned int vertexId = 0; vertexId < aim->mNumVertices; vertexId++) {
                                 aiAnimMesh.mVertices[vertexId] += positionDiff[vertexId];
                             }
                             delete[] positionDiff;
                         }
                     }
                     if (needNormals) {
```
FlorianBorn71, line 634: I have to revisit whether the following block still uses the correct indexing of normals/tangets for animated parts

test/unit/utglTF2ImportExport.cpp
```diff
@@ -380,7 +380,7 @@ TEST_F(utglTF2ImportExport, importglTF2PrimitiveModeLines) {
     const aiScene *scene = importer.ReadFile(ASSIMP_TEST_MODELS_DIR "/glTF2/glTF-Asset-Generator/Mesh_PrimitiveMode/Mesh_PrimitiveMode_08.gltf", aiProcess_ValidateDataStructure);
     EXPECT_NE(nullptr, scene);
     EXPECT_EQ(scene->mMeshes[0]->mNumVertices, 4u);
-    std::array<unsigned int, 5> l1 = { { 0u, 3u, 2u, 1u, 0u } };
+    std::array<unsigned int, 5> l1 = { { 0u, 1u, 2u, 3u, 0u } };
     EXPECT_EQ(scene->mMeshes[0]->mFaces[0].mNumIndices, 2u);
```
FlorianBorn71, line 384: extracting the used vertices changes the order of vertices and thus the index buffer

FlorianBorn71:
@kimkulling , what do you think about this improvements?
Summarized, I changed the GLTF loading code to build a list of vertices that are actually needed per primitive. This gives us huge memory saving in case the full vertex attribute stream is shared for many primitives. In case of the customer model, the memory saving was 1000x

kimkulling:
I have started to take a look on it. Stay tuned :-).

kimkulling (approved):
So when identitying the indices are the same this fix will just reuse them to save memory allocation. Great idea! Thanks a lot for your investigations and the PR as well! I really appreciate this!

kimkulling:

PR #5025, opened 2023-03-22
FBX import: Fix camera rotation

inhosens:
Assimp always used FbxCamera's initial values, but there are the transformed values and they should be used.


inhosens:
This patch was tested in blender with following patch.
https://projects.blender.org/blender/blender-addons/pulls/104500

kimkulling (approved):
Looks good!

kimkulling:

PR #5026, opened 2023-03-22
Add more AMF unit tests

turol:
`3_bananas.amf.7z` is compressed and we don't currently have an in-memory decompressor for it so it's not tested.

Two test models (`test8.amf` and `test9.amf`) leak memory and are disabled. 

Also adds a comment to `AMFImporter_Postprocess.cpp` where the leak happens.


turol:
Bug report for the leak: #5027

PR #5037, opened 2023-03-30
Fix link issue in UWP builds without functional replacement

liedtkeInTUM:
After trying a version update from 5.2.5 I discovered some build issues if included in UWP projects.
I come up with a simple solution, skipping just the missing function in windows libraries. There seems to be some system function in UWP to get a similar behavior. 

For me, it doesn't matter if this function is executed that way. If there is a way to just not include unzip code or replace with UWP support, I didn't find it the last days.

### DIFF
--- a/contrib/unzip/crypt.c
+++ b/contrib/unzip/crypt.c
@@ -103,7 +103,7 @@ int cryptrand(unsigned char *buf, unsigned int len)
     unsigned __int64 pentium_tsc[1];
     int result = 0;
 
-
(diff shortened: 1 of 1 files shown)


kimkulling:
The used API is marked as deprecated. I guess I need to update the used dependency here. Thanks for figuring this out. The used function is not used in our software, so I am fine with the fix.

kimkulling (approved):
Looks fine!

Issue #3937, opened 2021-06-01
UWP builds appear to be broken with latest master (c04759c)


NoTuxNoBux:
**Describe the bug**
Latest master (commit c04759c) seems to be broken when building for ARM64 UWP (Universal Windows Platform) in Visual Studio 2019, several compiler errors are shown:

```
assimp\code\Common\ImporterRegistry.cpp(213,31): error C2039: 'getenv': is not a member of 'std'
contrib\unzip\crypt.c(107,28): error C2220: the following warning is treated as an error
contrib\unzip\crypt.c(107,28): warning C4013: 'CryptAcquireContext' undefined; assuming extern returning int
contrib\unzip\crypt.c(109,32): warning C4013: 'CryptGenRandom' undefined; assuming extern returning int
XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(486,5): error MSB4181: The "CompileXaml" task returned false but did not log an error.
```

**To Reproduce**

1. Make sure your repository is clean.
2. Run `cmake.exe  -G "Visual Studio 16 2019" -A ARM64 -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0 -DASSIMP_BUILD_TESTS=OFF`
3. Open `Assimp.sln` in Visual Studio 2019.
4. Set `assimp (Universal Windows)` as active startup project.
5. Set the `RelWithDebugInfo` or `Release` configuration as active, and ARM64 as architecture.
6. Build.

**Expected behavior**
The build succeeds, - and the resulting DLL works in an UWP context.

**Platform (please complete the following information):**
 - OS: Windows 10 64-bit
 - Browser: N.A.
 - Version: commit c04759c
 - CMake version: 3.16.3

**Additional context**
Both building and running stable Assimp 5.0.1 works fine. I only wanted to test latest master because I have several issues with 5.0.1 regarding FBX reading, which appear to be fixed already.

It is likely that this problem also exists with plain x86-64 UWP Assimp, and is not limited to ARM64.

[...]

PR #5039, opened 2023-03-31
std::getenv is not supported using uwp

liedtkeInTUM:
Found another line that is not supported in uwp environments. Already discussed in #3937. Without std:: is offered by spdlogger but also just return nothing as uwp app.


code/Common/ImporterRegistry.cpp
```diff
@@ -378,7 +383,7 @@ void GetImporterInstanceList(std::vector<BaseImporter *> &out) {
     out.push_back(new IQMImporter());
 #endif
     //#ifndef ASSIMP_BUILD_NO_STEP_IMPORTER
-    //    out.push_back(new StepFile::StepFileImporter());
+    //     out.push_back(new StepFile::StepFileImporter());
```
kimkulling: Dead code.

kimkulling: Fixed.

kimkulling (commented):
Lokks fine for me.

kimkulling (approved):

kimkulling:

PR #5050, opened 2023-04-01
Fix warning related to nonportable-system-include-path.

Jackie9527:
Fix the following warnings when `-Wno-nonportable-system-include-path` is removed.
```
D:\workspace\CODE\assimp\code\Common\DefaultIOSystem.cpp(57,14): error : non-portable path to file '<Windows.h>'; speci
fied path differs in case from file name on disk [-Werror,-Wnonportable-system-include-path] [D:\workspace\CODE\assimp\
build-msvc-clang\code\assimp.vcxproj]
```


turol (changes requested):
This is wrong. On Linux mingw this file is called `windows.h` with all lowercase letters and Linux is case sensitive. These changes break this platform.

Jackie9527:
> This is wrong. On Linux mingw this file is called `windows.h` with all lowercase letters and Linux is case sensitive. These changes break this platform.

ok.
Let's keep `-Wno-nonportable-system-include-path` to make the compiler happy for all the platforms.

PR #5056, opened 2023-04-03
Fix pretransform vertices with cameras

sutajo:
mLookAt is a position vector inside the reference frame of its node, not a direction vector, so translation should be applied to it.


kimkulling (approved):
Good catch, thanks! Looks good.

kimkulling:

Issue #2115, opened 2018-09-02
Suggested rollback to FBXConverter camera changes


schuttejoe:
I just recently updated from 4.1.0 to latest and found that my cameras were being incorrectly exported from FBX files. I tracked the problem down to this commit: https://github.com/assimp/assimp/commit/0b15d5cd46caaa7fcb2d82e025a914f9e21cbeef

which is undoing the changes from a prior commit with a thorough comment explaining why it had done things this way: https://github.com/assimp/assimp/commit/5be2351dc6aa06502a561a5e2e8905d6c60e39ca

inhosens:
Sigh, I found it was reverted today. But I think @schuttejoe used some tools like blender for exporting his FBX files and those tools are not correctly exporting the camera's properties. I already tested it with FBX sdk.

PR #5057, opened 2023-04-04
ConvertToLHProcess now inverts viewing direction

sutajo:
The viewing direction of cameras should be inverted during conversion to a left handed coordinate system.

### DIFF
--- a/code/AssetLib/FBX/FBXConverter.cpp
+++ b/code/AssetLib/FBX/FBXConverter.cpp
@@ -421,6 +421,8 @@ void FBXConverter::ConvertCamera(const Camera &cam, const std::string &orig_name
 
     out_camera->mAspect = cam.AspectWidth() / cam.AspectHeight();
 
+    // NOTE: Camera mPosition, mLookAt and mUp must be set to default here.
+    // All transformations to the camera will be handled by its node in the scenegraph.
     out_camera->mPosition = aiVector3D(0.0f);
     out_camera->mLookAt = aiVector3D(1.0f, 0.0f, 0.0f);
     out_camera->mUp = aiVector3D(0.0f, 1.0f, 0.0f);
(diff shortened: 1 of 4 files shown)


sutajo:
Reverted commit https://github.com/assimp/assimp/commit/cdfdd75a66e2323bd834f923798d3043ad3deaf8 because it makes the same mistake as the gltf 2 importer. Cameras must be initialized to their default values and should only be pretransformed if requested with PretransformVertices.

This has been an issue before:
https://github.com/assimp/assimp/issues/2115

kimkulling (approved):
Looks fine for me.

kimkulling:

beholders-eye:
While cross building it from Exherbo Linux, I'm getting the error bellow:

```
/var/tmp/paludis/build/media-libs-assimp-scm/work/assimp-scm/code/PostProcessing/ConvertToLHProcess.cpp:244:26: error: no match for 'operator*' (operand types are 'float' and 'aiVector3D' {aka 'aiVector3t<double>'})
  244 |     pCam->mLookAt = 2.0f * pCam->mPosition - pCam->mLookAt;
      |                     ~~~~ ^ ~~~~~~~~~~~~~~~
      |                     |            |
      |                     float        aiVector3D {aka aiVector3t<double>}
I
```

Using gcc 12:

```
gcc --version
gcc (GCC) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Also tried with gcc 9

Maybe I'm doing something wrong while setting CMake build options?

sutajo:
This should be fixed by #5073

beholders-eye:
Yeah, it did, thanks a lot @sutajo !

monsieurgustav:
A bit late obviously but I'd like to discuss this change.
To me it doesn't make sense: the camera is looking at a location, already expressed in a left or right handed space.
Z being forward or backward must be handled when building the view matrix, not in the look-at vector.

@sutajo @kimkulling Could you explain what it made sense for you?

Issue #4918, opened 2023-02-01
Bug: Issue 54900 in oss-fuzz: assimp:assimp_fuzzer: Abrt in aiMesh::~aiMesh


kimkulling:
Status: New
Owner: ----
CC: kim.k...@[googlemail.com](http://googlemail.com/) 
Labels: Restrict-View-Commit ClusterFuzz Reproducible Stability-UndefinedBehaviorSanitizer Engine-libfuzzer OS-Linux Proj-assimp Reported-2023-01-07
Type: Bug

New issue 54900 by ClusterFuzz-External: assimp:assimp_fuzzer: Abrt in aiMesh::~aiMesh
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54900

Detailed Report: https://oss-fuzz.com/testcase?key=5966142718214144

Project: assimp
Fuzzing Engine: libFuzzer
Fuzz Target: assimp_fuzzer
Job Type: libfuzzer_ubsan_assimp
Platform Id: linux

Crash Type: Abrt
Crash Address: 0x053900082fb5
Crash State:
  aiMesh::~aiMesh
  std::__1::unique_ptr<aiMesh, std::__1::default_delete<aiMesh> >::reset
  std::__1::unique_ptr<aiMesh, std::__1::default_delete<aiMesh> >::~unique_ptr
  
Sanitizer: undefined (UBSAN)

Regressed: https://oss-fuzz.com/revisions?job=libfuzzer_ubsan_assimp&range=202206020602:202206030600

Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=5966142718214144

Issue filed automatically.

See https://google.github.io/oss-fuzz/advanced-topics/reproducing for instructions to reproduce this bug locally.
When you fix this bug, please
  * mention the fix revision(s).
  * state whether the bug was a short-lived regression or an old bug in any stable releases.
  * add any other useful information.
This information can help downstream consumers.

turol:
Did not fix it.

turol:
Valgrind trace:

```
Invalid write of size 8
   at 0x3B536C: Assimp::ObjFileImporter::createVertexArray(Assimp::ObjFile::Model const*, Assimp::ObjFile::Object const*, unsigned int, aiMesh*, unsigned int) (ObjFileImporter.cpp:501)
[...]

PR #5066, opened 2023-04-17
Fix possible dereferencing of invalid pointer.

kimkulling:
- closes https://github.com/assimp/assimp/issues/4918


turol:
This doesn't look like it should fix anything. `unique_ptr` is generally safer than explicit `new`/`delete` Have you tested that it fixes the test case? Can you put the testcase somewhere where it can be freely downloaded?

turol:
Looks like google has fixed their crap and it's now possible to download testcases without login.

turol:
This did not fix the bug and might have introduced memory leaks in error paths by removing `unique_ptr`. Please revert.

PR #5070, opened 2023-04-19
Fix warning related to missing-variable-declarations.

Jackie9527:
Fix the following warnings when `-Wno-missing-variable-declarations` is removed.
```
D:\workspace\CODE\assimp\code\AssetLib\Blender\BlenderCustomData.cpp(99,1): message : declare 'static' if the variable
is not intended to be used outside of this translation unit [D:\workspace\CODE\assimp\build-msvc-clang\code\assimp.vcxp
roj]
```


Jackie9527:
There is a conflict, I will fix it ASAP.

PR #4658, opened 2022-07-25
Added node metadata export for gltf2

bengtr:
glTF2Importer reads node metadata but glTF2Exporter did not write node metadata. Added node metadata export of all aiMetadataType types except AI_AIVECTOR3D.

PR #5092, opened 2023-05-11
Fix: Copy aiMetadata in SceneCombiner

luho383:
Metadata with sub-objects failed to copy due to default assertions.

Pushed this, but have not seen that #4658 has the same fix, so this could also be merged.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5101, opened 2023-05-22
Added more Maya materials

Sanchikuuus:
Added Maya materials to list and add setting texture properties for getting lost textures to mesh


include/assimp/material.h
```diff
@@ -292,6 +292,11 @@ enum aiTextureType {
     aiTextureType_DIFFUSE_ROUGHNESS = 16,
     aiTextureType_AMBIENT_OCCLUSION = 17,
 
+    aiTextureType_MAYA_BASE = 22,
```
turol, line 295: Why not use the existing types?

Sanchikuuus: Existing types is not including support of Specular/Gloss textures.
Screenshot 2023-05-22 at 14 02 01

kimkulling (approved):
Looks fine.

kimkulling:
You need to adapt the unittests as well.

kimkulling (changes requested):
Unittests are broken.

kimkulling:

PR #5117, opened 2023-06-02
Fix Heap-buffer-overflow READ in Assimp::FileSystemFilter::Cleanup

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53488

`it` was getting pass the `in.end()` because it does `it += 3;` and `it += 2;` in the loop and the `const size_t remaining = std::distance(in.end(), it);` becomes wrong.


kimkulling:

PR #5125, opened 2023-06-03
Update pugiXML library

mosfet80:
Release highlights:

xml_attribute::set_value, xml_node::set_value and xml_text::set now have overloads that accept pointer to non-null-terminated string and size Fix error handling in xml_document::save_file that could result in the function succeeding while running out of disk space Fix memory leak during error handling of some out-of-memory conditions during xml_document::load


kimkulling (approved):
Looks fine.

kimkulling:

PR #5133, opened 2023-06-06
Fixed missing config property lookup for removal of empty bones.

Biohazard90:
The post processing step LimitBoneWeightsProcess currently always removes unused bones because the config property AI_CONFIG_IMPORT_REMOVE_EMPTY_BONES has not been used to toggle this behavior on and off.

This change adds a property look up so LimitBoneWeightsProcess won't remove unused bones anymore when the client doesn't want this behavior.


kimkulling (approved):
A much better to do it this way. Thanks a lot! I will merge it.

kimkulling:

PR #5129, opened 2023-06-04
Fix UNKNOWN READ in aiTexture::~aiTexture

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25108
Closes #3440

PR #5120, opened 2023-06-02
Fix Stack-buffer-overflow READ in aiMaterial::AddBinaryProperty

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45079
Closes #4447 

The stack [buffer read overflow](https://github.com/assimp/assimp/blob/d9a8837a5b0e0d6b34b35a5bf1e1288ea23721aa/code/Material/MaterialSystem.cpp#L485) happens because the real `pInput` size is one byte less than `pSizeInBytes` in `aiMaterial::AddBinaryProperty`. The `pSizeInBytes` is [calculated](https://github.com/assimp/assimp/blob/d9a8837a5b0e0d6b34b35a5bf1e1288ea23721aa/code/Material/MaterialSystem.cpp#L528) as `static_cast<unsigned int>(pInput->length + 1 + 4)` which assumes that `pInput->length` doesn't include `\0` terminating character. This is in line with the [comment](https://github.com/assimp/assimp/blob/d9a8837a5b0e0d6b34b35a5bf1e1288ea23721aa/include/assimp/types.h#L393) in `aiString` for the member: `Binary length of the string excluding the terminal 0`. However in the case of `MDLImporter::ParseSkinLump_3DGS_MDL7` the aiString [is initialized as](https://github.com/assimp/assimp/blob/d9a8837a5b0e0d6b34b35a5bf1e1288ea23721aa/code/AssetLib/MDL/MDLMaterialLoader.cpp#L494-L499):
```cpp
        aiString szFile;
        const size_t iLen = strlen((const char *)szCurrent);
        size_t iLen2 = iLen + 1;
        iLen2 = iLen2 > MAXLEN ? MAXLEN : iLen2;
        memcpy(szFile.data, (const char *)szCurrent, iLen2);
        szFile.length = static_cast<ai_uint32>(iLen2);
```
When `iLen` is bigger or equal to `MAXLEN` the `szFile` length is set to `MAXLEN` and `data` contains non zero terminated characters. It is not entirely clear who breaks the contract, but I assume it is fair to expect aiString to have `\0` terminated data and fixed the implementation of `MDLImporter::ParseSkinLump_3DGS_MDL7`.

PR #5134, opened 2023-06-07
Fix memory leak

sashashura:
Related https://github.com/assimp/assimp/pull/5129
The only hypothesis I have why https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25108 wasn't closed is that the same reproducer file triggers memory leaks with a different sanitizer. The memory leak fix was submitted as a part of [another pull request](https://github.com/assimp/assimp/pull/5120). Extracting it here.


kimkulling (approved):
Looks fine, thanks!

kimkulling:

Issue #5139, opened 2023-06-15
Bug: ObjImporter generates a wrong vertex color mapping if vertex colors are partially specified


emaame:
**Describe the bug**
ObjImporter generates a wrong vertex color mapping if vertex colors partially specified (i.e., specified / omitted are mixed), such as:

```only_a_part_of_vertexcolors.obj
v  0.0  0.0  0.0
v  0.0  0.0  1.0  0.0  0.0  1.0
v  0.0  1.0  0.0
v  1.0  0.0  0.0  1.0  0.6  0.3
```

I think it is not an officially spec, however it is necessary in my case.

**Actual Vertex Colors**

```
(0.0, 0.0, 1.0, 1.0)
(1.0, 0.6, 0.3, 1.0)
uninitialized
uninitialized
```

because ObjFileParser::getTwoVectors3 assumes vertices and vertex colors have the same size.

**To Reproduce**
Remove a subset of vertex color declaration from an OBJ file then load it.

**Expected behavior**
Handle omitted vertex colors as `(0.0, 0.0, 0.0, 1.0)` or `(0.0, 0.0, 0.0, 0.0)`.
I guess `(0.0, 0.0, 0.0, 1.0)` is more similar to an explicitly declared file.

```
(0.0, 0.0, 0.0, 1.0)
(0.0, 0.0, 1.0, 1.0)
(0.0, 0.0, 0.0, 1.0)
(1.0, 0.6, 0.3, 1.0)
```

or

```
(0.0, 0.0, 0.0, 0.0)
(0.0, 0.0, 1.0, 1.0)
(0.0, 0.0, 0.0, 0.0)
(1.0, 0.6, 0.3, 1.0)
```

**Additional context**
Currently I'm going to post a PR.

turol:
What software produces files like this?

emaame:
I heard that exported from Cinema4D.
I'm so sorry I have no more information about it (version et. al).

(I also post here again)

emaame:
fix PR merged

PR #5140, opened 2023-06-15
Fix Objimported supports OBJ partially specified vertex colors

emaame:
The omitted vertex colors are treated as (0, 0, 0, 1).
for issue #5139 

e.g

```only_a_part_of_vertexcolors.obj
v  0.0  0.0  0.0
v  0.0  0.0  1.0  0.0  0.0  0.0
v  0.0  1.0  0.0
v  1.0  0.0  0.0  1.0  0.6  0.3
```

I'm not sure that which is better (0,0,0,1) or (0,0,0,0) for omitted one.
In this patch, it is treated as (0, 0, 0, 1), the same as the next file.

```.obj
v  0.0  0.0  0.0  0.0  0.0  0,0
v  0.0  0.0  1.0  0.0  0.0  0.0
v  0.0  1.0  0.0  0.0  0.0  0,0
v  1.0  0.0  0.0  1.0  0.6  0.3
```

I'm not good at C++ now.
thanks


turol (changes requested):
I think this is entirely the wrong approach. You should do the defaulting in the code which calls `getTwoVectors3`. It should check how many components there are and act appropriately.

emaame:
Thank you. You are right.
The resizing before calling `getTwoVectors3` is what I needed for.

turol (changes requested):

turol:
The code is much better and simpler now. But what software actually creates files like this? Why have we not seen such before?

Need to squash these commits to make a nicer history.

emaame:
I heard that exported from Cinema4D. but I apology I have no more information about it (version et. al).
I guess the reason is simple that vertex-color is a flavor and the most software is well mannered.

I noticed the code need to handle vertices without color at the end.
I will add a test case, and check empty and resize after getting a vector3.

I agree that `m_hasVertexColors` is not necessary, so I will remove it as well
(this member was also present in the original version, should I keep it?).

Then I will resubmit with a clean history.

kimkulling (approved):
Cool, great work! Thanks a lot for your help!

PR #5142, opened 2023-06-16
Update LWSLoader.cpp

kimkulling:
- closes https://github.com/assimp/assimp/issues/5141

### DIFF
--- a/code/AssetLib/LWS/LWSLoader.cpp
+++ b/code/AssetLib/LWS/LWSLoader.cpp
@@ -921,6 +921,7 @@ void LWSImporter::InternReadFile(const std::string &pFile, aiScene *pScene, IOSy
             SkeletonMeshBuilder builder(pScene);
         }
     }
+    delete master;
 }
 
 #endif // !! ASSIMP_BUILD_NO_LWS_IMPORTER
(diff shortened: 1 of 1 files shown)


code/AssetLib/LWS/LWSLoader.cpp
```diff
@@ -888,6 +891,7 @@ void LWSImporter::InternReadFile(const std::string &pFile, aiScene *pScene, IOSy
     if (anims.size()) {
         master->mAnimations = new aiAnimation *[master->mNumAnimations = 1];
         aiAnimation *anim = master->mAnimations[0] = new aiAnimation();
+        master->mAnimations[0] = nullptr;
```
turol, line 894: This one is definitely wrong, it forgets the pointer that was just create above. Prefer one assignment per line to avoid such confusing code.

code/AssetLib/LWS/LWSLoader.cpp
```diff
@@ -921,6 +925,7 @@ void LWSImporter::InternReadFile(const std::string &pFile, aiScene *pScene, IOSy
             SkeletonMeshBuilder builder(pScene);
         }
     }
+    delete master;
```
turol, line 928: Prefer `unique_ptr` over explicit `new/delete`.

code/AssetLib/LWS/LWSLoader.cpp
kimkulling: Right.

Issue #2036, opened 2018-06-22
ERROR, DXF IMPORT


Ave23:
The file DXFLoader.cpp is contained the line  
unsigned int indices[4]; 
This array is used to store face indices.
But, it is known, the DXF file can contain negative face indices to describe dummy edges. Therefore, the import result is incorrect if the file contains faces with dummy edges.

kimkulling:
Do you have an example for this case?

PR #5146, opened 2023-06-19
Add handling for negative indices.

kimkulling:
- closes https://github.com/assimp/assimp/issues/2036


kenichiice:
@kimkulling 
This PR currently skips faces with negative indices. However, it is the 'edges' that become invisible due to negative indices. Shouldn't the faces still need to be constructed?

PR #5148, opened 2023-06-20
update utf8 from 2.3.4 to 3.2.3

mosfet80:
New convenience API for C++ 11 and later compilers. The library still works with C++ 98/03 compliant compilers, just without the new functions. advance() function works in both directions.
The following deprecated functions were removed:

previous() - deprecated since version 1.02.
is_bom() - deprecated since version 2.3.
Fix of the project version number at CMakeLists.txt Continuous Integration with Google Tests and CircleCI A minor release that contains fix for Issue #31 Program fails to link when including utf8.h in multiple files. This release adds one new API call: unchecked::replace_invalid(). Optional support for C++ 17 std::string_view.
The release contains the fix the inclusion of both cpp11 and cpp17 headers on C++17 compilation. Also some additional tests for using string literals and string objects with modern compilers.


kimkulling:

PR #5153, opened 2023-06-22
Unify way to check readable blender files.

feuerste:
Previously the `InternReadFile` logic was much more restrictive than the `CanRead` logic, but able to also parse compressed files. With this change the behavior is unified.

### DIFF
--- a/code/AssetLib/Blender/BlenderLoader.cpp
+++ b/code/AssetLib/Blender/BlenderLoader.cpp
@@ -115,15 +115,12 @@ BlenderImporter::~BlenderImporter() {
     delete modifier_cache;
 }
 
-static const char * const Tokens[] = { "BLENDER" };
+static const char Token[] = "BLENDER";
 
 // ------------------------------------------------------------------------------------------------
 // Returns whether the class can handle the format of the given file.
 bool BlenderImporter::CanRead(const std::string &pFile, IOSystem *pIOHandler, bool /*checkSig*/) const {
-    // note: this won't catch compressed files
-    static const char *tokens[] = { "<BLENDER", "blender" };
-
-    return SearchFileHeaderForToken(pIOHandler, pFile, tokens, AI_COUNT_OF(tokens));
+    return ParseMagicToken(pFile, pIOHandler).error.empty();
 }
 
 // ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


code/AssetLib/Blender/BlenderLoader.cpp
```diff
@@ -1338,4 +1295,55 @@ aiNode *BlenderImporter::ConvertNode(const Scene &in, const Object *obj, Convers
     return node.release();
 }
 
+std::string BlenderImporter::ParseMagicToken(std::shared_ptr<IOStream>& stream) const {
```
turol, line 1297: You should not pass `shared_ptr` via reference. Either pass by value or pass the raw pointer.

feuerste: In this specific case the stream is an input/output parameter, as it can internally be reset, if the file is compressed, and will be reused outside of this function. So passing by value wouldn't work. Passing by raw pointer would require us to clean up the old memory, so the whole point of using smart pointers is lost.

Of course we could rewrite more of the code here, but I didn't want to touch too much of the original logic (besides converting exceptions to strings).

feuerste:
@turol Would you mind to restart the workflows and take another look? Thanks a lot!

turol:
I still don't like the use of reference parameters. I think the correct solution would be to return a `std::variant` with either an error or the new buffer/stream. This might require C++ 17 however.

kimkulling (approved):
Looks fine!

code/AssetLib/Blender/BlenderLoader.h
```diff
@@ -180,10 +180,17 @@ class BlenderImporter : public BaseImporter, public LogFunctions<BlenderImporter
             const Blender::MTex *tex,
             Blender::ConversionData &conv_data);
 
-    // Returns an error if it can't parse the magic token.
-    std::string ParseMagicToken(
-            std::shared_ptr<IOStream>& stream,
-            std::vector<char>& uncompressed) const;
+    struct StreamOrError {
```
turol, line 183: Add a comment that this should use a `std::variant` when we move to C++17.

code/AssetLib/Blender/BlenderLoader.h
```diff
@@ -180,10 +180,17 @@ class BlenderImporter : public BaseImporter, public LogFunctions<BlenderImporter
...
+        std::shared_ptr<IOStream> stream;
+        std::shared_ptr<std::vector<char>> input;
```
turol, line 185: Why is `input` a shared_ptr? Why not just have vector that might be empty?

code/AssetLib/Blender/BlenderLoader.cpp
```diff
@@ -142,25 +140,23 @@ void BlenderImporter::SetupProperties(const Importer * /*pImp*/) {
...
-        ThrowException(error);
+    StreamOrError streamOrError = ParseMagicToken(pFile, pIOHandler);
+    if (!streamOrError.error.empty()) {
+        ThrowException(streamOrError.error);
     }
 
     char version[4] = { 0 };
-    file.i64bit = (stream->Read(version, 1, 1), version[0] == '-');
-    file.little = (stream->Read(version, 1, 1), version[0] == 'v');
+    file.i64bit = (streamOrError.stream->Read(version, 1, 1), version[0] == '-');
```
turol, line 149: You should move stream into a local variable so there's less changes to existing code.

turol:
If you're going to rebase it you might as well squash the whole thing into one commit.

code/AssetLib/Blender/BlenderLoader.h
feuerste: Done.

code/AssetLib/Blender/BlenderLoader.cpp
feuerste: Done.

feuerste:
> If you're going to rebase it you might as well squash the whole thing into one commit.

Done.

kimkulling:

PR #5154, opened 2023-06-22
Improve binary check for gltf and gltf2.

feuerste:
By checking the magic token we don't depend on the extension any more and follow the official way to detect a gltf file as binary, see also https://github.com/KhronosGroup/glTF/blob/main/extensions/1.0/Khronos/KHR_binary_glTF/README.md#header and https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#binary-header.

Before this fix a binary glb file called e.g. "blob" (without any extension) couldn't successfully be loaded, as it was assumed to be gltf.


kimkulling (approved):
Looks fine!

feuerste:
@kimkulling Thanks for reviewing this! I just realized that we should include the version as well.
Please don't merge yet, I'll be working on a fix.

kimkulling:
OK, I will wait until you have added the version check.

feuerste:
@kimkulling Thanks for waiting, I double-checked everything, this PR is fine, we don't need the versioning here, as it is really only checking for the binary format. Please go ahead with the merge, and thanks again for all your work!

kimkulling:
Ok, merged. Thanks a lot for your double-check and your contribution.

PR #5155, opened 2023-06-22
Extend token search flag from alpha to graph.

feuerste:
This way we can fix Blender exported files being wrongly detected as obj files, if the filename didn't contain any extension information, e.g. when loading a binary glb file named `blob`.


feuerste:
This will fix #1605 for Blender exported glb files.

kimkulling:
Cool finding!

kimkulling (approved):
Looks fine.

kimkulling:

PR #5156, opened 2023-06-22
Handle gcs cloud storage file extensions with versioning.

feuerste:
If a user uses its own `IOSystem`/`IOStream` for handling google cloud storage uris, with this PR the extension will still be correct. Together with #5157 the following will be supported without going over all possible importers and trying to detect the format:
```
Assimp::Importer importer;
importer.SetIOHandler(new GsIOSystem(...));
const aiScene* scene = importer.ReadFile("gs://bucket/model.glb#1234", ...);
...
```

### DIFF
--- a/code/Common/BaseImporter.cpp
+++ b/code/Common/BaseImporter.cpp
@@ -59,6 +59,31 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <memory>
 #include <sstream>
 
+namespace {
+// Checks whether the passed string is a gcs version.
+bool IsGcsVersion(const std::string &s) {
+    if (s.empty()) return false;
+    return std::all_of(s.cbegin(), s.cend(), [](const char c) {
+        // gcs only permits numeric characters.
+        return std::isdigit(static_cast<int>(c));
+    });
+}
+
+// Removes a possible version hash from a filename, as found for example in
+// gcs uris (e.g. `gs://bucket/model.glb#1234`), see also
+// https://github.com/GoogleCloudPlatform/gsutil/blob/c80f329bc3c4011236c78ce8910988773b2606cb/gslib/storage_url.py#L39.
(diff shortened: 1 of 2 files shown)


turol:
I'm not sure this is a good idea. `#` is technically a valid character in file names so this might break something for someone.

feuerste:
> I'm not sure this is a good idea. `#` is technically a valid character in file names so this might break something for someone.

Thanks @turol I agree that `#` is a valid file name character.
However, in this case we are explicitly looking at file extensions, and looking through https://en.wikipedia.org/wiki/List_of_file_formats reveals no single extension having a `#` inside. Furthermore, s3 and gcs regularly add their file versions behind the extension (and marking them with `#`), so it would be really useful for everyone extending `IOSystem`/`IOStream` with s3/gcs support to already detect the corresponding files based on their extension (see also #5157 in combination with this PR) and not having to wait for the loop going over all files, opening them, reading magic bytes or other ways to unify files.

feuerste:
@kimkulling This PR depends on #5157. Once the other one is merged, I will update this one and let you know.

turol:
I haven't used C++ regex but have heard many people (even on the committee) call it bad. We probably don't want to use it.

feuerste:
> I haven't used C++ regex but have heard many people (even on the committee) call it bad. We probably don't want to use it.

Fine with me, too.

feuerste:
@kimkulling This one is ready for review now as well :)

kimkulling:
Just working on it.

kimkulling:
If I get it right hashes will only get filtered out when they re part of the extension. So for me this looks fine. Some unittests would be great to have.

feuerste:
> If I get it right hashes will only get filtered out when they re part of the extension. So for me this looks fine. Some unittests would be great to have.

Yes, this is correct. I added a few unit tests as well. PTAL!

kimkulling (approved):
Looks fine to me.

kimkulling:

PR #5161, opened 2023-06-27
fix incorrect default for material::get with aiColor3D

malytomas:
Accessing material properties with Get method should not modify the provided value if the property is not present in the material (this is by specification). However, the method for aiColor3D resets the value to black instead.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5164, opened 2023-06-29
[pyassimp] Fix passing of file extension string.

feuerste:
Previously only the first character of the string was passed to c, now it passes the whole extension.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5168, opened 2023-07-03
Fix UNKNOWN READ in Assimp::SMDImporter::ParseNodeInfo

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57218
Closes #5107

PR #5169, opened 2023-07-03
Fix memory leak

sashashura:
This is related to #5168 but I'm not sure if I got it right. Thus making as a separate pull request.

### DIFF
--- a/code/Common/SceneCombiner.cpp
+++ b/code/Common/SceneCombiner.cpp
@@ -63,6 +63,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <assimp/scene.h>
 #include <stdio.h>
 #include <assimp/DefaultLogger.hpp>
+#include <memory>
 
 namespace Assimp {
 


include/assimp/SceneCombiner.h
```diff
@@ -223,7 +223,7 @@ class ASSIMP_API SceneCombiner {
      *    deletes the input scenes afterwards. There may be duplicate scenes.
      *  @param flags Combination of the AI_INT_MERGE_SCENE flags defined above
      */
-    static void MergeScenes(aiScene **dest, aiScene *master,
+    static void MergeScenes(aiScene **dest, std::unique_ptr<aiScene>& master,
```
turol, line 226: This is part of the public API, you can't change it (without major version bump). It's already documented as taking ownership of the pointer so just use `.release()` at the call site.

turol (changes requested):

include/assimp/SceneCombiner.h
turol: Store it in a `unique_ptr` first thing inside `MergeScenes`?

code/Common/SceneCombiner.cpp
```diff
@@ -250,17 +250,18 @@ void SceneCombiner::AttachToGraph(aiScene *master, std::vector<NodeAttachmentInf
 }
 
 // ------------------------------------------------------------------------------------------------
-void SceneCombiner::MergeScenes(aiScene **_dest, std::unique_ptr<aiScene>& master, std::vector<AttachmentInfo> &srcList, unsigned int flags) {
+void SceneCombiner::MergeScenes(aiScene **_dest, aiScene *master, std::vector<AttachmentInfo> &srcList, unsigned int flags) {
+    std::unique_ptr<aiScene> masterAutoPtr(master);
```
turol, line 254: I would call this `master` and rename the parameter to `master_` or `master_raw`

PR #5170, opened 2023-07-03
Add DIFFUSE_ROUGHNESS_TEXTURE for gltf2 exporter

vulcanozz:
I am using the assimp exporter and I have noticed that I can't add aiTextureType_DIFFUSE_ROUGHNESS when exporting to  glb2 format. Therefore, I would like to add it.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Exporter.cpp
+++ b/code/AssetLib/glTF2/glTF2Exporter.cpp
@@ -838,6 +838,16 @@ void glTF2Exporter::ExportMaterials() {
 
         GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, aiTextureType_DIFFUSE_ROUGHNESS);
 
+        if (!m->pbrMetallicRoughness.metallicRoughnessTexture.texture) {
+            //if there wasn't a aiTextureType_DIFFUSE_ROUGHNESS defined in the source, fallback to aiTextureType_METALNESS
+            GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, aiTextureType_METALNESS);
+        }
+
+        if (!m->pbrMetallicRoughness.metallicRoughnessTexture.texture) {
+            //if there still wasn't a aiTextureType_METALNESS defined in the source, fallback to unknown texture
+            GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE);
+        }
+
         if (GetMatColor(mat, m->pbrMetallicRoughness.baseColorFactor, AI_MATKEY_BASE_COLOR) != AI_SUCCESS) {
             // if baseColorFactor wasn't defined, then the source is likely not a metallic roughness material.
             //a fallback to any diffuse color should be used instead
--- a/code/AssetLib/glTF2/glTF2Exporter.cpp
+++ b/code/AssetLib/glTF2/glTF2Exporter.cpp
@@ -844,7 +844,7 @@ void glTF2Exporter::ExportMaterials() {
         }
 
         if (!m->pbrMetallicRoughness.metallicRoughnessTexture.texture) {
-            // if there still wasn't a aiTextureType_METALNESS defined in the source, fallback to unknown texture
+            // if there still wasn't a aiTextureType_METALNESS defined in the source, fallback to AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE
             GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE);
         }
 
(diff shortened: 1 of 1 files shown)


turol:
There's no need to continually merge the base branch. The maintainers will do that when it's getting merged. You're just adding noise to the history. If you want to keep the branch up to date use rebase.

kimkulling (approved):
Looks fine!!

code/AssetLib/glTF2/glTF2Exporter.cpp
```diff
@@ -870,6 +870,7 @@ void glTF2Exporter::ExportMaterials() {
             }
         }
 
+        GetMatTex(mat, m->normalTexture, aiTextureType_DIFFUSE_ROUGHNESS);
```
feuerste, line 873: Are you sure you want to assign diffuse roughness to the normal texture? This doesn't make sense to me...

feuerste (changes requested):

code/AssetLib/glTF2/glTF2Exporter.cpp
vulcanozz: Sorry，it was a mistake and has been fixed.

code/AssetLib/glTF2/glTF2Exporter.cpp
```diff
@@ -836,7 +836,7 @@ void glTF2Exporter::ExportMaterials() {
             GetMatTex(mat, m->pbrMetallicRoughness.baseColorTexture, aiTextureType_DIFFUSE);
         }
 
-        GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE);
+        GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, aiTextureType_DIFFUSE_ROUGHNESS);
```
feuerste, line 839: In the importer `AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE`, `aiTextureType_METALNESS`, and `aiTextureType_DIFFUSE_ROUGHNESS` are all set to `pbrMetallicRoughness.metallicRoughnessTexture`. Would it make sense to also try all three here for backwards compatibility, having a similar fallback strategy as above for the `pbrMetallicRoughness.baseColorTexture`?

feuerste (changes requested):
Thanks for your adjustments!

code/AssetLib/glTF2/glTF2Exporter.cpp
vulcanozz: It looks like what you said makes sense, and I have followed your suggestion. Additionally, I took the opportunity to format the code.

code/AssetLib/glTF2/glTF2Exporter.cpp
```diff
@@ -832,20 +832,30 @@ void glTF2Exporter::ExportMaterials() {
...
+
+        if (!m->pbrMetallicRoughness.metallicRoughnessTexture.texture) {
+            // if there wasn't a aiTextureType_DIFFUSE_ROUGHNESS defined in the source, fallback to aiTextureType_METALNESS
+            GetMatTex(mat, m->pbrMetallicRoughness.metallicRoughnessTexture, aiTextureType_METALNESS);
+        }
+
+        if (!m->pbrMetallicRoughness.metallicRoughnessTexture.texture) {
+            // if there still wasn't a aiTextureType_METALNESS defined in the source, fallback to unknown texture
```
feuerste, line 847: `AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLICROUGHNESS_TEXTURE` instead of `unknown`?

code/AssetLib/glTF2/glTF2Exporter.cpp
feuerste: Thanks for the adjustments. Using `!= AI_SUCCESS` checks may be more consistent with the rest of the file though. Can you please also verify, whether it is better to first try `aiTextureType_METALNESS` or `aiTextureType_DIFFUSE_ROUGHNESS`?

feuerste (commented):
Thanks for all your work!

code/AssetLib/glTF2/glTF2Exporter.cpp
vulcanozz: OK

code/AssetLib/glTF2/glTF2Exporter.cpp
vulcanozz: I'm sorry, but the GetMatTex function seems to not have a return value, and I have not found any interface with an aiReturn return value that has the same functionality as GetMatTex.
About the order: I have tried exporting according to aiTextureType_METALNESS, aiTextureType_DIFFUSE_ROUGHNESS, and exporting with both together. In the exported glb file, there is only 'pbrMetallicRoughness', which should be consistent with the GLB file specification. I also attempted to re-import with Assimp, and regardless of whether I exported with aiTextureType_METALNESS, aiTextureType_DIFFUSE_ROUGHNESS, or both, I was always able to read the output using either aiTextureType_DIFFUSE_ROUGHNESS or aiTextureType_METALNESS. I further tested by importing the models exported with these methods into Blender to check. In Blender, the material effects of each model were consistent. Thus, I believe the order of aiTextureType_METALNESS and aiTextureType_DIFFUSE_ROUGHNESS should not be crucial.

feuerste: Okay, thanks a lot for digging in!

feuerste (approved):

kimkulling:

PR #5175, opened 2023-07-10
Bug fix and improvement to FBX camera field-of-view during import.

sfjohnston:
For FBX files, mHorizontalFOV was being computed from the full field-of-view, not from the center to the edge. Scaled by 0.5.

Also, some software (Maya) does not put the field-of-view in FBX, but it can be computed from focal length and image width. This patch detects if FOV hasn't been set in the file and computes it.

This is not an exhaustive fix to FBX cameras, but is an improvement for common uses.


code/AssetLib/FBX/FBXDocument.h
```diff
@@ -55,9 +55,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define _AI_CONCAT(a,b)  a ## b
 #define  AI_CONCAT(a,b)  _AI_CONCAT(a,b)
 
+/* Use an 'illegal' default FOV value to detect if the FBX camera has set the FOV. */
+#define FBX_FOV_UNKNOWN -1.0f
```
kimkulling, line 59: Please use a constexpr float instead, more typesafe.

code/AssetLib/FBX/FBXDocument.h
```diff
@@ -55,9 +55,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define _AI_CONCAT(a,b)  a ## b
 #define  AI_CONCAT(a,b)  _AI_CONCAT(a,b)
 
+/* Use an 'illegal' default FOV value to detect if the FBX camera has set the FOV. */
```
kimkulling, line 58: Please use // style comments. This will make it much easier to put bigger blocks into comments.

kimkulling (changes requested):
See comments, thanks a lot for the PR!

kimkulling (approved):
Looks fine, thanks!

kimkulling:

PR #5176, opened 2023-07-11
Fix detection of `KHR_materials_specular` on glTF2 export.

feuerste:
Currently when exporting as glTF2 the detection of a specular material requires that both `AI_MATKEY_COLOR_SPECULAR` and `AI_MATKEY_SPECULAR_FACTOR` are set. One of them should actually be enough, which I think was also the original intention, according to the comment.

### DIFF
--- a/test/unit/utglTF2ImportExport.cpp
+++ b/test/unit/utglTF2ImportExport.cpp
@@ -60,7 +60,7 @@ using namespace Assimp;
 
 class utglTF2ImportExport : public AbstractImportExportBase {
 public:
-    virtual bool importerMatTest(const char *file, bool spec_gloss, std::array<aiTextureMapMode, 2> exp_modes = { aiTextureMapMode_Wrap, aiTextureMapMode_Wrap }) {
+    virtual bool importerMatTest(const char *file, bool spec, bool gloss, std::array<aiTextureMapMode, 2> exp_modes = { aiTextureMapMode_Wrap, aiTextureMapMode_Wrap }) {
         Assimp::Importer importer;
         const aiScene *scene = importer.ReadFile(file, aiProcess_ValidateDataStructure);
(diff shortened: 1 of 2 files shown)


feuerste:
@Beilinson Can you please have a look?

Beilinson:
Yes this is correct

feuerste:
@Beilinson Thanks for having a look!
@kimkulling Would you mind merging the fix? Thank you very much!

kimkulling (approved):
Looks fine.

kimkulling:
Looks fine so far. Do we have any kind of test file to validate the fix?

feuerste:
> Looks fine so far. Do we have any kind of test file to validate the fix?

I adjusted the corresponding unit test to catch the bug.

test/unit/utglTF2ImportExport.cpp
```diff
@@ -60,7 +60,7 @@ using namespace Assimp;
 
 class utglTF2ImportExport : public AbstractImportExportBase {
 public:
-    virtual bool importerMatTest(const char *file, bool spec_gloss, std::array<aiTextureMapMode, 2> exp_modes = { aiTextureMapMode_Wrap, aiTextureMapMode_Wrap }) {
+    virtual bool importerMatTest(const char *file, bool spec, bool gloss, std::array<aiTextureMapMode, 2> exp_modes = { aiTextureMapMode_Wrap, aiTextureMapMode_Wrap }) {
```
kimkulling, line 63: Override missing?

kimkulling (approved):
Looks fine.

wedesoft:
Ah, ok. I [reported it on LWJGL](https://github.com/LWJGL/lwjgl3/issues/909). Hopefully it gets fixed soon. Thanks for looking into it :)

PR #5183, opened 2023-07-15
Remove /WX from CMakeLists for MSVC

SirLynix:
Assimp CMakeLists forcefully adds /W4 /WX flags with MSVC, even when `ASSIMP_WARNINGS_AS_ERRORS` is disabled. This fixes it


kimkulling:
Looks fine, good catch!

kimkulling:

PR #5187, opened 2023-07-21
Clipper update

mosfet80:
update clipper from 4.8.8 to 6.4.2


kimkulling:
See https://github.com/assimp/assimp/pull/5220

PR #5192, opened 2023-08-02
IRR - Fix UTF-16 file parsing (and possibly more?)

PencilAmazing:
Simple one line change. https://pugixml.org/docs/manual.html#xml_document::load_string states that load_string does not do any auto encode detection, so switch to load_buffer instead.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5193, opened 2023-08-04
Collada: added import property to disable unit size scaling

martinweber:
I needed the option to ignore the unit size scaling for Collada imports.

This PR adds a property `AI_CONFIG_IMPORT_COLLADA_IGNORE_UNIT_SIZE` which controls if the unit size scaling is applied to the root node's transform or not. It works similar to the existing `AI_CONFIG_IMPORT_COLLADA_IGNORE_UP_DIRECTION` property.

The default value is false, so the behavior for Collada imports will not change unless explicitly requested.

### DIFF
--- a/code/AssetLib/Collada/ColladaLoader.cpp
+++ b/code/AssetLib/Collada/ColladaLoader.cpp
@@ -95,6 +95,7 @@ ColladaLoader::ColladaLoader() :
         noSkeletonMesh(false),
         removeEmptyBones(false),
         ignoreUpDirection(false),
+        ignoreUnitSize(false),
         useColladaName(false),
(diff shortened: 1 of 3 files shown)


turol:
There's no need to merge master branch, we will do that when we're ready to merge. If you want to update your branch prefer rebase and force push to avoid excessive amount of merge commits.

martinweber:
Ah sorry. I thought I hit the rebase option through the Github web interface. I will refrain from further updates.

kimkulling:
So, I have merged the current master. Now I will start doing the code review.

kimkulling (approved):
Looks fine!

kimkulling:

PR #5195, opened 2023-08-06
Fix Heap-buffer-overflow READ in ODDLParser::OpenDDLParser::parseStructure

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58650
`in` can be equal to `end` and `*in` causes out of bounds read.


turol:
This should go upstream to https://github.com/kimkulling/openddl-parser/ and then get pulled in from there.

PR #5198, opened 2023-08-08
Fix Heap-buffer-overflow READ in ODDLParser::OpenDDLParser::parseFloatingLiteral

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58766

Requires https://github.com/kimkulling/openddl-parser/pull/83.

If left as is, later it crashes because `pScene->mRootNode` is nullptr. I think it is better to throw if parsing has failed.


sashashura:
Looks like the change is necessary.

PR #5199, opened 2023-08-08
Fix UNKNOWN READ in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<ch

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58770

In this case `temp` is max unsigned int. `temp + 76` was overflowing the integer and `reader.IncPtr` was reading 75 bytes. However later `std::string(head, temp)` was reading the maximum number of characters, leading to out of bounds read.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5211, opened 2023-08-23
Bug: Lose when Import Gltf2 and then export Gltf2


guguTang:
**Describe the bug**
lose materials/occlusionTexture/strength property in gltf description file when import gltf2 and then export gltf2

**Expected behavior**
retain texture strength property in gltf description file

**Screenshots**
![image](https://github.com/assimp/assimp/assets/17616476/a3f588db-5ea7-48fc-9417-3968d463d0cd)

**Platform (please complete the following information):**
 - master

PR #5212, opened 2023-08-23
Ensure that the strength attribute is the same when importing or exporting gltf2

guguTang:
Fixed: https://github.com/assimp/assimp/issues/5211


kimkulling (approved):
Good catch!

Issue #4269, opened 2021-12-19
Bug: Missing Pdb Files On CMake Install


aWilson41:
Using 5.1.3, & now also tested with 5.1.4. Cmake install fails when building shared libs on Visual Studio 16 2019. I use the following cmake variables:
 - DASSIMP_BUILD_ASSIMP_TOOLS:BOOL=OFF
 - DASSIMP_BUILD_TESTS:BOOL=OFF
 - DASSIMP_BUILD_ZLIB:BOOL=ON
 - CMAKE_INSTALL_PREFIX:PATH="<my superbuilds install directory>"

Install fails with:
```
1>  CMake Error at code/cmake_install.cmake:163 (file):
1>    file INSTALL cannot find
1>    "<My Build Directory Here>/code/RelWithDebInfo/assimp-vc142-mt.pdb":
1>    No such file or directory.
1>  Call Stack (most recent call first):
1>    cmake_install.cmake:77 (include)
```

Where <My Build Directory Here>/code/RelWithDebInfo directory is never created/used. The pdb file only exists in <My Build Directory Here>/bin.

There  appears to be a bit of logic here: https://github.com/assimp/assimp/blob/master/code/CMakeLists.txt#L1328 . I'm unsure why this is on a conditional for static builds only. Removing it appears to solve the issue.

kimkulling:
Thanks for the report. I guess we need some refactoring in the CMake-Files. Some stuff there is already outdated or nt in an optimal state I guess.

PR #5213, opened 2023-08-23
Use correct PDB paths

AnyOldName3:
The previously-specified paths were only correct for a static library build.
I therefore fenced off that code to be specific to static library builds.

For shared library builds, I added a generator-expression-based alternative.
Annoyingly, this won't work with static library builds, so we don't get the concise, reliable version in all circumstances.

I've avoided modifying any of the paths anything ends up at to avoid breaking changes.

Resolves https://github.com/assimp/assimp/issues/4269


kimkulling (approved):
Looks fine from my point of view.

kimkulling:

PR #5215, opened 2023-08-30
Fix compilation with minizip-ng

modbw:
minizip-ng doensn't import the typedefs voidpf and uLong from zlib


kimkulling:
This one is empty. I will close it. Please add a new PR to fix your issue.

PR #5252, opened 2023-09-26
Reduce memory consumption in JoinVerticesProcess::ProcessMesh() signi…

ockeymm:
When execute JoinVerticesProcess::ProcessMesh()， 'uniqueVertices' and 'uniqueAnimatedVertices' Array will consumption huge memory, some device will oom when execute this funcation.
For example： 
         The size of each array's element - Vertex structure is 272 bytes. If there is a mesh of 100000 Vertex and 52 animationMeshes, these two arrays will consume 100000 * 272 * 53=1441600000 memory during function operation.
         I changed the elements of the array from the Vertex structure to the index of the vertex in the original data, and then assigned the vertex to the new target by searching for the original data in the updateXMeshVertices function. In the above example, using this method, the memory usage can be reduced to 100000 * 4 * 53=21200000


code/PostProcessing/JoinVerticesProcess.cpp
```diff
@@ -157,52 +157,52 @@ void updateXMeshVertices(XMesh *pMesh, std::vector<Vertex> &uniqueVertices) {
 
     // Position, if present (check made for aiAnimMesh)
     if (pMesh->mVertices) {
-        delete [] pMesh->mVertices;
+        aiVector3D *oldVertices = pMesh->mVertices;
```
turol, line 160: You should use `unique_ptr` instead of manual delete.

kimkulling:
A good idea. Let's see what the CI says to it.

kimkulling:
This will decrease the memory consumption for the process by 68. Whowow!

kimkulling (approved):
Looks fine for me!

kimkulling:

PR #5265, opened 2023-10-05
Fix output floating-point values to fbx

Riv1s-sSsA01:
Failed to get floating values (e.g. opacity) from scene material when ASSIMP_DOUBLE_PRECISION is defined, so they are not reflected to output fbx file.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5271, opened 2023-10-11
Fix: Implicit Conversion Error

Ipomoea:
In latest versions of XCode when building Assimp for ARM processor `Implicit conversion loses integer precision` warning is treated as error.
It happens when we try to assign `unsigned long` value (which is 8 bytes sized) to `int` variable (which is 4 bytes). `static_cast` covers this problem here and it is safe because length check before conversion already exists.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #4864, opened 2023-01-05
Bug: ld: library not found for -lminizi on Mac OS


rongcuid:
**Describe the bug**
I tried to compile assimp 5.2.5 on Mac OS 13.1, but failed with library not found.

**To Reproduce**
1. `cmake -DBUILD_SHARED_LIBS=OFF -DASSIMP_WARNINGS_AS_ERRORS=OFF ..`
2.  `make`
3. See error

**Expected behavior**
assimp compiles

**Screenshots**
```
make -j16 V=1
[ 64%] Built target assimp
[ 65%] Linking CXX executable ../bin/unit
ld: library not found for -lminizip
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/unit] Error 1
make[1]: *** [test/CMakeFiles/unit.dir/all] Error 2
make: *** [all] Error 2
```

**Platform (please complete the following information):**
 - OS: Mac OS 13.1
 - Version 5.2.5

MiniZ is available:

```
ls /opt/local/lib/libminizip*
/opt/local/lib/libminizip.1.dylib /opt/local/lib/libminizip.a       /opt/local/lib/libminizip.dylib
```

kimkulling:
Do you have a complete log, please?

kimkulling:
Weight!

kimkulling:
I guess this issue is caused by the installation from minizip. Have you tried to enable building our own zlibs?

tellypresence:
Closing due to no response to followup

Issue #3933, opened 2021-05-30
macOS build fails "ld: library not found for -lminizip" even though CMake has found it


MarkCallow:
**Describe the bug**
After CMake configure and generate the build fails with the error `ld: library not found for -lminizip`. As can be seen in the screen shot below CMake has found libminizip.dylib. And as can be seen in the link command given in Additional Context there is no option to add /opt/local/lib as a place to search for libaries. So it is not surprising the link fails.

**To Reproduce**
1. Install minizip via macports - `sudo port install minizip`
2. Run CMake GUI. Configure - selecting default build tools (Xcode), and selecting only importers (no tools, no exporters).
3. Generate project
4. `cd <directory with generated project>; cmake --build . --config Release `
5. See error

**Expected behavior**
I expect the build to succeed.

**Screenshots**
Screenshot of CMake GUI showing it found libminizip.dylib in /opt/local/lib:

<img width="825" alt="Screen Shot 2021-05-30 at 9 59 23 PM" src="https://user-images.githubusercontent.com/2244683/120105034-32253680-c10c-11eb-8582-496705e0f389.png">

**Platform (please complete the following information):**
 - OS: macOS
 - Browser: n/a
 - Version 11.4 (Big Sur)

**Additional context**
Here is the link command and output:

```
Ld /Users/mark/Projects/github/assimp/macOS/bin/Release/libassimp.5.0.1.dylib normal (in target 'assimp' from project 'Assimp')
    cd /Users/mark/Projects/github/assimp/assimp
[...]

PR #5278, opened 2023-10-14
link to external minizip with full path

aumuell:
This let's cmake create config files that allow linking to minizip even when in an unrelated prefix. This can happen with package managers that install every package into their own prefix, such as [Spack](https://spack.io).

Unfortunately, the variable containing the full path to the libraries discovered via pkg-config (`XXX_LINK_LIBRARIES` instead of `XXX_LIBRARIES`) is only documented for CMake >= 3.12.

Perhaps this also helps solving #4864 and #3933.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5279, opened 2023-10-15
utf8 header not found

TarcioV:
utf8 header not found while compiling assimp with samples.


samples/SimpleTexturedDirectx11/SimpleTexturedDirectx11/main.cpp
```diff
@@ -23,6 +23,7 @@
 #include <dxgi1_2.h>
 #include <DirectXMath.h>
 #include <d3dcompiler.h>
+#include <utf8.h>
```
umlaeute, line 26: why do you need to include `<utf8.h>` here?

samples/SimpleTexturedOpenGL/SimpleTexturedOpenGL/src/model_loading.cpp
```diff
@@ -17,6 +17,7 @@
 #include <stdio.h>
 #include <GL/gl.h>
 #include <GL/glu.h>
+#include <utf8.h>
```
umlaeute, line 20: and why do you need to include `<utf8.h>` here?

code/CMakeLists.txt
```diff
@@ -923,7 +923,7 @@ IF(ASSIMP_HUNTER_ENABLED)
   hunter_add_package(utf8)
   find_package(utf8cpp CONFIG REQUIRED)
 ELSE()
-  # utf8 is header-only, so Assimp doesn't need to do anything.
+  INCLUDE_DIRECTORIES("../contrib/utf8cpp/source")
```
umlaeute, line 926: i agree that it is much better to just fix the include path, rather than including different files. based on context as is done now: https://github.com/assimp/assimp/blob/020554e213365e009f5455ae3572bb0ae10026eb/code/Common/BaseImporter.cpp#L360-L364

umlaeute (commented):
i welcome the move from fixing the include-dirs and having a unified `#include <utf8.h>` wherever it is needed. (the latter part is missing from this patch).

also, i find it somewhat strange that you are adding includes without mentioning that in the commit (it probably should be a separate commit, adding the utf8.h header wherever it has been forgotten)

code/CMakeLists.txt
umlaeute: obviously this PR does not (yet) change the actual includes (e.g. in `BaseImporter.cpp`.

after adding the include_directories, it could be as simple as:

```C
#include <utf8.h>
```

kimkulling (approved):
Looks better for me doing it this way.

kimkulling:

PR #5281, opened 2023-10-16
Rm unnecessary deg->radian conversion in FBX exporter

### LINKED ISSUE #5137: Bug: Exported FBX with skeleton is corrupted
When importing a .glb file (CesiumMan.glb from the glTF 2.0 samples repo) , which contains skeleton animation and exporting it to .fbx file, the output file is messed up as can be seen in the following screenshot:
image

This is the result with the original .glb file:
image

Exporting to .glb works fine.

 I use Assimp version 5.2.5 and for the import/export the pre/post processing flags were set to 0.
I also tried other Assimp versions as suggested in other bug reports related to this issue, but I had no success with them.

JulianKnodt:
It seems that rotation matrices later expect radians. This conversion breaks it, and was validated on the conversion of `cesium_man.glb` --> `cesium_man.fbx`

cc @ronenv-ten @gongminmin
closes #5137

EDIT: actually, I found that for animated meshes, this works. For meshes without animations, this seems to break things. Will continue to investigate further.

EDIT 2: This actually can be fixed entirely within `FBXConverter.cpp` by converting deg to rad on 3219, 3242, and 3250, will get to it tonight.


JulianKnodt:
I'll need to double check tmrw if this is correct, but should be good to review now.

code/AssetLib/FBX/FBXConverter.cpp
```diff
@@ -3243,6 +3243,7 @@ aiNodeAnim* FBXConverter::GenerateSimpleNodeAnim(const std::string& name,
     if (keyframeLists[TransformationComp_Rotation].size() > 0) {
         InterpolateKeys(outRotations, keytimes, keyframeLists[TransformationComp_Rotation], defRotation, maxTime, minTime, rotOrder);
```
JulianKnodt, line 3244: todo: convert defRotation to rad

code/AssetLib/FBX/FBXExporter.cpp
```diff
@@ -2434,7 +2432,7 @@ void FBXExporter::WriteObjects ()
                 aiMatrix4x4 m(k.mValue.GetMatrix());
                 aiVector3D qs, qr, qt;
                 m.Decompose(qs, qr, qt);
-                qr *= DEG;
+                qr = AI_DEG_TO_RAD(qr);
```
JulianKnodt, line 2435: this should be rad_to_deg

code/AssetLib/FBX/FBXExporter.cpp
```diff
@@ -2515,9 +2513,10 @@ void FBXExporter::WriteModelNode(
             );
         }
         if (r != zero) {
+            r = AI_DEG_TO_RAD(r);
```
JulianKnodt, line 2516: this should be rad_to_deg

code/AssetLib/FBX/FBXExporter.cpp
```diff
@@ -2601,8 +2600,7 @@ void FBXExporter::WriteModelNodes(
             transform_chain.emplace_back(elem->first, t);
             break;
         case 'r': // rotation
-            r *= float(DEG);
-            transform_chain.emplace_back(elem->first, r);
+            transform_chain.emplace_back(elem->first, AI_DEG_TO_RAD(r));
```
JulianKnodt, line 2603: this should be rad_to_deg

kimkulling (approved):
Looks fine.

kimkulling:
-merged, thanks a lot for your contribution.

PR #5282, opened 2023-10-17
Fix invalid read of `uint` from `uvwsrc`

JulianKnodt:
Previously was reading a uint, which always failed. Since the output was never checked, this seemed to work, and works fine for most models since they only use UV channel 0.


kimkulling (approved):
Looks fine, is there a model to test this?

JulianKnodt:
Let me find one, I think I cannot release the one I've been testing with

JulianKnodt:
image

[sphere_two_uv.zip](https://github.com/assimp/assimp/files/13300350/sphere_two_uv.zip)

I had to zip it because github apparently doesn't support just uploading raw glb files?

kimkulling:
Merged, thanks a lot dor your contribution.

Issue #4648, opened 2022-07-18
Bug: Unable to load basisu compressed images in GLTF2 model


X-skyer:
**Describe the bug**
Recently I want to load the gltf2 model with basis universal image compression. So I download the sample model from here and pack the model to a single glb file.

[BoxTextured_basisu.zip](https://github.com/assimp/assimp/files/9131493/BoxTextured_basisu.zip)

But when i use assimp to import this model in assimp view, no texture displayed. Then I step into the gltf2importer code, I found the embedded images array to be length of Zero.

I've checked that last year, assimp has support basisu images.
https://github.com/assimp/assimp/pull/3893

tellypresence:
Closing due to no response

PR #5308, opened 2023-11-03
Added support for basisu extension to GLTF2 importer

Promit:
A GLTF2 file contains a textures array, which references items from the images array and the samplers array:
`{"sampler":0,"source":0}`
Assimp follows these references to lazy load the corresponding image and sampler. Simple enough. But when you're dealing with a GLTF2 that uses Basisu textures, the texture item comes out like this instead:
```
{
            "sampler": 0,
            "extensions": {
                "KHR_texture_basisu": {
                    "source": 0
                }
            }
        }
```
So the regular code flow doesn't work and the extension needs to be checked. I've put in a relatively straightforward addition here that recognizes the alternate source and causes Assimp to follow the references correctly.

Fixes issue #4648 .


Promit:
There is a similar change needed to support the Microsoft extension for DDS textures in GLTF2 (MSFT_texture_dds) but that extension hasn't been recognized in Assimp at all and I thought it better to start with this relatively simple fix. I think it would be worth adding the MSFT extension as well.

kimkulling (approved):
Looks fine.

kimkulling:
You need to update your branch.

alexrp:
@kimkulling I don't know if you're aware of this, but FWIW, GitHub introduced a feature somewhat recently that allows you to easily update a PR that's out of date with respect to the base branch. You just need to go to the repo Settings tab and enable this:

image

Then PRs will show this:

image

(That said, if that doesn't work for some reason, I can rebase and resubmit this PR.)

kimkulling:
I am not allowed to update this branch. In other pull requests, this button is visible. Maybe there is something wrong with the access rights?

Promit:
I kinda forgot this was here, I have to admit. I clicked the update button, if all the tests pass then maybe we can get this all buttoned up.

kimkulling:
Already merged.

PR #5316, opened 2023-11-06
Drop explicit inclusion of contrib/ headers

umlaeute:
this PR removes the following pattern:

```C
#ifdef ASSIMP_USE_HUNTER
#  include <${HEADER}>
#else
#  include "../contrib/utf8cpp/source/${HEADER}"
#endif
```

since the build-system already sets the appropriate include paths (e.g. `-I.../contrib/utf8cpp/source/`) we can simply use
```C
#include "${HEADER}"
```
instead, thus simplifying the code-base and making it easier to selectively use system-installed headers, without having to use `-DASSIMP_USE_HUNTER=1` to get the proper includes (when not using hunter at all).

this PR builds on #5279


umlaeute:
@kimkulling i'm not exactly sure how to understand https://github.com/assimp/assimp/pull/5279#pullrequestreview-1715443220

do you mean that **this** PR is the wrong way to do it?

kimkulling (approved):
Looks fine.

kimkulling:

PR #5322, opened 2023-11-13
Fix buffer overflow in FBX::Util::DecodeBase64()

ttxine:
`FBX::Util::DecodeBase64()` has `maxOutLength` parameter, but doesn't use it to check bounds when writing to `out`. This may result in a buffer overflow.

Add bounds check in `FBX::Util::DecodeBase64()`.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5323, opened 2023-11-13
Fix double free in Video::~Video()

ttxine:
If `contentLength` is zero, then `content` was deallocated in `Video::Video()` or was not allocated at all. Thus, deleting `content` in destructor without a check resulted in double free.

Add a check in `Video::~Video()`.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #1325, opened 2017-06-28
Crash in `BlenderModifier_Subdivision :: DoIt` when importing a blender mesh 


lamogui:
line 314 
```c++
    here ---> aiMesh** const meshes = &conv_data.meshes[conv_data.meshes->size() - out.mNumMeshes];
    std::unique_ptr<aiMesh*[]> tempmeshes(new aiMesh*[out.mNumMeshes]());
```

[conv_data.meshes->size() is 1 and  out.mNumMeshes is 0 maybe it miss a -1 here ?

lamogui:
[avion_mig.zip](https://github.com/assimp/assimp/files/1108879/avion_mig.zip)

kimkulling:
Finally fixxed.

PR #5338, opened 2023-11-22
Fix: Check if index for mesh access is out of range

kimkulling:
- closes https://github.com/assimp/assimp/issues/1325


kimkulling:
Done!

PR #5345, opened 2023-11-25
Add unit test for subdivision modifier on Blender importer

aavenel:
Catmul-Clark subdivision modifier is implemented in Assimp but was lacking unit-test.

This PR add an unit-test of a Blend file composed of default cube with subdivision modifier applied and same cube with subdivision modifier.
image

Vertices should match if Assimp is running the subdivision code correctly.


test/unit/utBlenderImportExport.cpp
```diff
@@ -221,3 +222,29 @@ TEST(utBlenderImporter, importFleurOptonl) {
...
+/*
+This test contains a default cube with subdivision surface modifier
+and a default cube with subdivision surface applied.
+Vertices should be identical.
+*/
+TEST_F(utBlenderImporterExporter, importBlendWithSubdivisionSurface) {
+    Assimp::Importer importer;
+    const aiScene *scene = importer.ReadFile(ASSIMP_TEST_MODELS_DIR "/BLEND/subdivision_test_277.blend", aiProcess_ValidateDataStructure);
+    EXPECT_NE(nullptr, scene);
+    EXPECT_EQ(scene->mNumMeshes, 2);
```
turol, line 236: This must be `ASSERT_EQ` to not crash on the next line in case of failure.

kimkulling:

PR #5347, opened 2023-11-26
KHR_materials_specular fixes

rudybear:
I found 2 issues while exporting the official assets ( https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/SpecularTest)
1. It doesn't provide proper properties for an extension, default values are different in the spec
2. It overrides the Specular texture type (specular and specularColor by the spec are separate)


kimkulling:

Issue #5132, opened 2023-06-06
Bug: FBX importer, doubt about current mOffsetMatrix


Biohazard90:
It looks like the mOffsetMatrix does currently not contain any bone bind pose transformations due to this change: https://github.com/assimp/assimp/commit/863e3e0ccbd9266fd6f87282a6832cdde578641e

As far as I understand, the GetTransform matrix describes the mesh transform and not the bone transform. See here: https://stackoverflow.com/questions/71956620/in-fbx-file-i-cannot-understand-difference-between-transform-and-transformlink So I feel like this change has introduced a bug.

---

My own animation system/format works a bit differently than how it is designed in assimp, but I wasn't able for the life of me to construct the bone bind pose in model space with the mOffsetMatrix as it was. I personally changed this code to simply provide me with GetTransformLink which allowed me to construct the bind pose.

Additionally I noticed that bone weights weren't normalized which also messed up my results, but I haven't dug deeper into that and simply normalized them in my client code for now.

While trying to solve my own issues, I noticed that there are a number of posts about issues related to this and I think there is a good chance that the aforementioned commit has introduced an issue.

kimkulling:
The old status is there. I will add a new bug-report for the issue which has caused the broken fix.

PR #5349, opened 2023-11-27
Update FBXConverter.cpp

kimkulling:
- Temporary fix bind pose storage in offset matrix
- Will close https://github.com/assimp/assimp/issues/5132


mwestphal:
This is causing https://github.com/assimp/assimp/issues/5449

PR #5351, opened 2023-11-27
Readme.md:  correct 2 errors in section headers

stephengold:
`prebuild` denotes something used or occurring before a build.
`doc's` denotes something associated with a doc(tor).


kimkulling (commented):
Good point, looks nice. Thanks for your support!

kimkulling (approved):

Issue #5354, opened 2023-11-28
Bug: Some gLTF files do not have target names(aiAnimMesh name) properly loaded.


Futuremappermydud:
**Describe the bug**

When loading a binary glTF2 file and iterating over each animMesh in each mesh, some models have empty names for the animMesh. 

Edit: Cause in comment below

**To Reproduce**
```
#include <assimp/Importer.hpp>
#include <assimp/postprocess.h>
#include <assimp/scene.h>
#include <iostream>

int main()
{
    Assimp::Importer importer;

    const aiScene* scene = importer.ReadFile("C:/Users/User/OneDrive/Desktop/AliciaSolidFIX.gltf", aiProcess_Triangulate | aiProcess_LimitBoneWeights);

    for (size_t i = 0; i < scene->mNumMeshes; i++)
    {
        auto mesh = scene->mMeshes[i];
        for (size_t k = 0; k < mesh->mNumAnimMeshes; k++)
        {
            auto animMesh = mesh->mAnimMeshes[k];
            std::cout << animMesh->mName.C_Str() << std::endl;
        }
    }

    return 0;
}
```
1. Run this code with the path directed to a broken model
2. See empty lines(names are empty/whitespace)

**Expected behavior**
Each line should have the proper name of the animMesh, as seen inside the gltf File

**Platform (please complete the following information):**
 - OS: Android

**Additional context**
Models to test with https://drive.google.com/file/d/18xKZ-ZsITdNvlD1IAdobExMYeeY0EOzS/view?usp=drive_link

Futuremappermydud:
https://pastebin.com/30SDn05U This paste shows the output of the following code (Has more logging)

```
#include <assimp/Importer.hpp>
#include <assimp/postprocess.h>
#include <assimp/scene.h>
#include <iostream>

int main()
{
    Assimp::Importer importer;

    const aiScene* scene = importer.ReadFile("C:/Users/Owens/OneDrive/Desktop/AliciaSolidFIX.gltf", aiProcess_Triangulate | aiProcess_LimitBoneWeights);

    for (size_t i = 0; i < scene->mNumMeshes; i++)
    {
        auto mesh = scene->mMeshes[i];
        std::cout << "mesh " << i << ": " << mesh->mName.C_Str() << std::endl;
        for (size_t k = 0; k < mesh->mNumAnimMeshes; k++)
        {
            auto animMesh = mesh->mAnimMeshes[k];
[...]

PR #5356, opened 2023-11-28
Fix target names not being imported on some gLTF2 models

Futuremappermydud:
This loads target names from the first primitive if available. Names from the mesh root will still overwrite the ones from the primitive.
- closes https://github.com/assimp/assimp/issues/5354


kimkulling:

PR #5359, opened 2023-11-29
fixed several issues

MarkaRagnos0815:
1. fixed issue in DefaultIOSystem::Exists on none windows platforms. 
- fopen returns a valid file pointer for directories
- using stat instead

2. removed utf8 headers from public headers because these are not needed and enforce the user to have these headers available after cmake install

3. fixed issue in q3bsp importer where tries are unitialized (tested with pk3 file from test set)


code/Common/DefaultIOSystem.cpp
```diff
@@ -99,12 +99,12 @@ bool DefaultIOSystem::Exists(const char *pFile) const {
         return false;
     }
 #else
-    FILE *file = ::fopen(pFile, "rb");
-    if (!file) {
+	struct stat statbuf;
```
turol, line 102: Inconsistent whitespace

kimkulling (approved):
Looks fine.

kimkulling:

PR #5360, opened 2023-11-30
Improve acc file loading

IOBYTE:
Add warning and support for tolerating a common problem where objects have wrong kid count. Add support for empty texture layers.


IOBYTE:
This also fixes a problem where degenerate triangles were being deleted causing a validation error for meshes with no vertices.

kimkulling (approved):
Looks fine.

kimkulling:

Issue #5332, opened 2023-11-18
Bug: X loader regression introduced with commit f844c33 on 3 Nov 2023


tellypresence:
Appears to be a recent regression in X loader

Loading bundled model [`test/models-nonbsd/X/dwarf.x`](https://github.com/assimp/assimp/blob/master/test/models-nonbsd/X/dwarf.x) exhibits strange artifacts unless `aiProcess_PreTransformVertices` flag is applied during load (which of course destroys rigging/animation)

Problems observed when loaded without `aiProcess_PreTransformVertices` flag
- Other than (most of) the head, helmet and left shoulder pad, the body mesh is not visible.
- Animation for the body mesh seems to be correct
- The axe is fully visible, but no longer animates (used to move with the body)

Note that this problem can also be observed in bundled IRR scenes which load the dwarf perserving bones/animation

<table>
<tr><th>Screenshots</th><th>Notes</th></tr>
<tr><td>
<img src="https://github.com/assimp/assimp/assets/26958898/61765fad-2806-4f39-9b37-097221c2d793" alt="x_loader_regression_normal_load_bones_n_anim_preserved_Screenshot_20231118-092548" width=120 />
<img src="https://github.com/assimp/assimp/assets/26958898/5003ba97-40d7-4d70-87b6-8f8be25ed3b0" alt="x_loader_regression_normal_load_bones_n_anim_preserved_Screenshot_20231118-092557" width=120 />
<img src="https://github.com/assimp/assimp/assets/26958898/079c9c5b-464a-4aa2-b7ce-6e6dac90be30" alt="x_loader_regression_normal_load_bones_n_anim_preserved_az_recorder_20231118_112405" width=120 />

</td><td>Loaded without

<tt>aiProcess_PreTransformVertices</tt> flag 
(bones/animation preserved)

Notes:
 - The axe is no longer animated
 - double sided rendering was not 
enabled here, but the left shoulder 
pad can be viewed from both top
and bottom</td></tr>
<tr><td>
<img src="https://github.com/assimp/assimp/assets/26958898/dfd5ebcd-accf-4656-8d22-e05fc5b89a9a" alt="x_loader_regression_pretransform_verts_bones_n_anim_lost_Screenshot_20231118-093548" width=120 />
</td><td>Loaded with

<tt>aiProcess_PreTransformVertices</tt> flag
(bones/animation lost)</td></tr>
</table>

### Bisect
[...]

PR #5372, opened 2023-12-09
Fix X importer breakage introduced in commit f844c33 

tellypresence:
Comment out new conditional which caused massive breakage as detailed in #5332


kimkulling (approved):
Thanks for figuring this out.

kimkulling:

PR #5382, opened 2023-12-15
Expose aiGetEmbeddedTexture to C-API

sacereda:
This should allow to use aiScene::GetEmbeddedTexture from C.

I think this function can be useful mostly because now some embedded textures doesn't have the '*' prefix unless you set AI_CONFIG_IMPORT_FBX_EMBEDDED_TEXTURES_LEGACY_NAMING on import, but maybe there is something I'm missing.


kimkulling:

PR #5397, opened 2023-12-24
Fix PyAssimp under Python >= 3.12 and macOS library search support

Th3T3chn0G1t:
Python 3.12 removed the `distutils` module (https://docs.python.org/3.10/library/distutils.html) which was being used to check for Anaconda library paths - conditionally disabling this code depending on the Python version is a quick-fix, though modern Python has cleverer package resolution techniques which we could use in the future.

This ties into the fact that PyAssimp failed to search the macOS `DYLD_LIBRARY_PATH` (which acts similarly to `LD_LIBRARY_PATH`) when looking for Assimp.

Hopefully these tide over PyAssimp until a better library resolution mechanism can be worked out.


kimkulling (approved):
Looks fine for me. Thanks for your help.

kimkulling:

PR #5414, opened 2024-01-09
A fuzzed stride could cause the max count to become negative and henc…

FlorianBorn71:
A fuzzed stride could cause the max count to become negative and hence wrap around uint

We have fuzzing tests that artificially break files. We want assimp to crash in a controlled way.
One case we now came across is that the fuzzer changed the stride of one buffer view:
            "byteStride": 32769,
This caused the division maxSize / stride to become 0 and with the -1 wrap around to 0xffffffff for the uint.


FlorianBorn71:
Not a big deal in reality, but it's always good to harden the code

kimkulling:
Merged, thanks a lot for you contribution.

PR #5415, opened 2024-01-10
More GLTF loading hardening

FlorianBorn71:
This small PR does not add any features or change in behavior. 
It just makes loading of fuzzed gltf files more secure because it throws more exceptions for index out-of-range cases.
We came across these cases as part of our automated input file fuzzing tests.


kimkulling (approved):
Some minor format issues, the rest looks fine.

kimkulling:

PR #5419, opened 2024-01-11
`Blendshape`->`Geometry` in FBX Export

### LINKED ISSUE #5418: Bug: Assimp FBX Blendshape export incorrect property separator
**Describe the bug**
When exporting an fbx mesh with blendshapes from assimp and attempting to load into blender, it will not load correctly, reporting that there is an incorrect `class`, which when debugging I found to be `Blendshape` instead of `Geometry`.

**To Reproduce**

Load and unload an FBX Binary file with blendshapes.

**Expected behavior**
I expect this not to fail.

 - OS: Windows 11
 - Version: head of master branch?

**Additional context**
This is due to line 1753 on FBXExporter, which is `Blendshape` instead of `Geometry`.
This change fixes importing into blender (altho I haven't tested this on other DCC tools)

JulianKnodt:
When loading a mesh exported from assimp into Blender, it warns that it has an incorrect class. While debugging, I traced it back to this being `Blendshape` where `Geometry` was expected. This is likely because this node describes a `Geometry`, which is used as a blendshape. I'm not sure if any other DCC tools or places to import it expect `Blendshape` instead (i.e. was this code ever tested?), but it fixes its use in Blender.

Fixes #5418 

Test Mesh:
[headblendshape.zip](https://github.com/assimp/assimp/files/13899594/headblendshape.zip)
Note this is actually not the mesh I originally tested on, but the original was proprietary.

### DIFF
--- a/code/AssetLib/FBX/FBXExporter.cpp
+++ b/code/AssetLib/FBX/FBXExporter.cpp
@@ -1748,7 +1748,7 @@ void FBXExporter::WriteObjects ()
         int64_t blendshape_uid = generate_uid();
         mesh_uids.push_back(blendshape_uid);
         bsnode.AddProperty(blendshape_uid);
-        bsnode.AddProperty(blendshape_name + FBX::SEPARATOR + "Blendshape");
+        bsnode.AddProperty(blendshape_name + FBX::SEPARATOR + "Geometry");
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine.

kimkulling:

PR #5421, opened 2024-01-16
Add bounds checks to the parsing utilities.

kimkulling:
There are a lot of fuzzer findings caused by out-of-bounds-aceesses caused by the internal parsing utilities for strings and character buffer. In this pull-request, I have added an end check for each access to a string parser function.

I expect there will be many issues caused by bugs I have introduced. Please help me review this PR and fix it!

Thanks, Kim


turol:
Having one massive commit like this is very bad for both reviewability and history. Please split it up. First have one commit which adds the new `aiBuffer` type and then small commits which introduce it in the loaders one by one.

kimkulling:
Done!

Issue #3411, opened 2020-10-01
Vertex colors incorrect converting DXF to glTF 


aarongarvey:
I'm a new user to Assimp, so I'm looking for some assistance in determining if something is wrong with Assimp, my workflow, or my source DXF file.

I have a 3D DXF file in which I require to be converted to glTF format.  Assimp performs the conversion fine without error on the command line using `assimp export test.dxf test.gltf` however when I view the resulting glTF file, the vertex colors used on the exported model do not match the original source DXF model.

The first link shows how the DXF model is viewed within Sketchfab when the DXF model is imported.
[Original DXF Colors](https://drive.google.com/file/d/19AqevoAZZpN9N6zJmHeneUjnPWotcfr8/view?usp=sharing)

The second link shows how the converted glTF model is viewer within Sketchfab (note: I get the same result when viewing the converted glTF in three.js and babylon as well)
[Converted glTF colors](https://drive.google.com/file/d/1_1bTb9cCqlpzBvIdwhZNjktcipX5lsV0/view?usp=sharing)

[Link to source DXF and glTF files](https://drive.google.com/file/d/1a-uG52MMSe2hLhU_j4VHAL46qD6L6IIV/view?usp=sharing)

Is anyone able to help steer me in the right direction on this?

kimkulling:
Closed by https://github.com/assimp/assimp/pull/5426

Issue #5422, opened 2024-01-16
Bug: Assimp incorrectly places 3dface inserts


seanth:
**Describe the bug**
When viewing a dxf file consisting of two 3d faces--a base rectangle and a small sphere--Sketchup and Autodesk render the image correctly. Assimp exports stls and objs that place the sphere in a very different location

**To Reproduce**
1. Install assimp
2. run the command 'assimp export <inputfile>.dxf <outputfile>.stl
3. open resulting stl 
4. see the sphere is not centered on the rectangle

**Expected behavior**
expected behaviour is that inserts should be inserted at the correct coordinate

**Screenshots**
![sketchup-correct](https://github.com/assimp/assimp/assets/371465/9faf7123-8c56-4b08-97ef-072e4bbc1cd4)
![autodesk-correct](https://github.com/assimp/assimp/assets/371465/97283634-1892-4579-9bc8-b37deb4ed504)
![assimp-incorrect](https://github.com/assimp/assimp/assets/371465/f96dd643-b15b-4848-b12b-0410c6af1f29)

**Platform (please complete the following information):**
 - OS: Mac OS 12.6 and 13.5.1
 - Browser: not applicable
 - Version: Version 5.3 -shared -st (GIT commit 0)

**Additional context**
example dxf file used in the above screenshots
[inputFile.dxf.zip](https://github.com/assimp/assimp/files/13955962/inputFile.dxf.zip)

seanth:
After doing a lot of experimentation, I'm fairly certain that when scaling is applied to inserts, the scaling is also being applied to the positional arguments. So not only is the scale changing the xyz scale of the inserted block, it is shifting the xyz position of the block.

seanth:
I think I have resolved the issue

https://github.com/seanth/assimp/commit/7c9e1d71b83f42d6be94253270295e42cbba7f1c

Changing the order of operations so placement of the inserted block happens before scaling seems to fix the problem when the position xyz values are multiplied by the scaling values.

I will test more before submitting a pull request

seanth:
Fixed by #5426

seanth:
Fixed by https://github.com/assimp/assimp/pull/5426

PR #5426, opened 2024-01-17
Fix for issues #5422, #3411, and #5443 -- DXF insert scaling fix and colour fix

seanth:
This is a very simple one line change.

In the previous version, if a dfx had multiple inserts that had different scales, assimp would place the insert in xyz space in locations that were incorrect. Outputs from assimp would have files that would be very different from what apps like sketchup or the online autodesk viewer would produce.

The bug was because scaling of the insert was being done before placement, resulting in the placement being placement•scale. For example, the insert would be placed at (pos.x•scale.x),(pos.y•scale.y),(pos.z•scale.z) instead of pos.x, pos.y, pos.z

Moving the line of code responsible for determining placement to before the line of code dealing with scaling fixes this.

I suspect this was an edge-case as the code has been in use for 10 or 12 years now and there hasn't been a bug report, but I suspect it explains why a subset of dxf viewers/converters out there mangle my files in exactly the same way. I think they relied on code from assimp

### DIFF
--- a/code/AssetLib/DXF/DXFLoader.cpp
+++ b/code/AssetLib/DXF/DXFLoader.cpp
@@ -372,6 +372,10 @@ void DXFImporter::ExpandBlockReferences(DXF::Block& bl,const DXF::BlockMap& bloc
                 // XXX order
                 aiMatrix4x4 trafo, tmp;
                 aiMatrix4x4::Translation(-bl_src.base,trafo);
+                //Need to translate position before scaling the insert
+                //otherwise the position ends up being the position*scaling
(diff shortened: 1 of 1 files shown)


seanth:
Updated AIC table resolves issues #3411 and #5443 

Image in sketchup:
Screenshot 2024-01-26 at 15 42 36

Image of resulting .obj made with assimp using the changes in this pr:

Screenshot 2024-01-26 at 16 22 41

kimkulling (approved):
Looks fine.

kimkulling:

PR #5432, opened 2024-01-22
CalcTangents: zero vector is invalid for tangent/bitangent

JensEhrhardt-eOPUS:
This fix helps when the calculated tangent of a face is a zero vector


turol:
Comparing floats with `==` is dangerous. What happens if the values are negative zeroes instead of positive zeroes? What if they're sufficiently small (around `FLT_MIN`) that calculating the length still ends up zero?

JensEhrhardt-eOPUS:
committed a version that also checks for small vectors

JensEhrhardt-eOPUS:
length of (0.5, 0.5, 0.5) == 0.866 is smaller than any normalized 3D vector

JulianKnodt:
er sorry why not just check the length is sufficiently different from 1? i.e. `Length(Vector) - 1 < EPSILON`, where `EPSILON` is a small number such as `1e-3`? It's a bit hard to understand the intent from reading the current code

The current implementation marks all vectors invalid if they're in the range [-0.5, 0.5], but I'm not sure if it's problematic if the vector is too large.

kimkulling (approved):
Looks fine.

kimkulling:

PR #5435, opened 2024-01-24
Fix FBX exporting incorrect bone order

### LINKED ISSUE #5256: Bug: FBX reorders bones indices on export
**Describe the bug**

Assimp will reorder bone output upon FBX export. While for most use cases this is fine (the bone order itself is not important for rendering a single mesh), but some downstream tools that take multiple meshes except their bones to be in the same order.

**To Reproduce**
1. Take any FBX model
2. Import with Assimp
3. Export same model with Assimp
4. Examine ordering of bones in editor such as Unity

**Expected behavior**
The bone order between the original and exported should be the same.

 - Any
 - Latest Version (5.24 I believe)

**Additional context**
This leads to an error when using LOD with Unity.

**Proposed Fix**

Inside of `FBXExporter.cpp`, line 1870 uses a set, along with a comparator. Remove this comparator, and convert the set to a vector, and that should be sufficient to preserve the input order.

JulianKnodt:
Closes #5256 

This should be tested on a model with redundant bone names before being merged, as that is the only case where differences may occur.


JulianKnodt:
@kimkulling would you like me to update this PR? It's been a minute since I touched it, but I think the sanitizer was a false positive

kimkulling (approved):
Looks fine for me.

JulianKnodt:
You'll have to merge if you like, I don't have permissions to do

kimkulling:

Issue #5438, opened 2024-01-25
Bug: Failed to load file: No materials remaining


stephengold:
**Describe the bug**
When an incomplete glTF model is being loaded, specifying `aiProcess_RemoveRedundantMaterials` causes a crash during post-processing, with the message "No materials remaining".

**To Reproduce**
1. Build the `assimp` command from source.

2. Unzip the "Running.gltf" file from the following archive:
[test.zip](https://github.com/assimp/assimp/files/14058247/test.zip)

3. Execute the following command:
    `bin/assimp info Running.gltf -rrm`

**Expected behavior**
If the model being loaded contains no materials, then RemoveRedundantMaterials post-processing should have no effect. The model should load as if the option hadn't been specified.

**Screenshots**
```console
$ bin/assimpd info Running.gltf -rrm
param = -rrm
Launching asset import ...           OK
Validating postprocessing flags ...  OK
100 %ERROR: Failed to load file: No materials remaining
assimp info: Unable to load input file Running.gltf
```

**Platform (please complete the following information):**
 - OS: Linux Mint 21.3
 - Browser: FireFox
 - Assimp version: GIT commit ac29847d

**Additional context**
not applicable

kimkulling:
With the latest master I get the following error message: 

`100 %ERROR: Failed to load file: No materials remaining`

And there is no crash.

kimkulling:
So all not used materials get removed correctly from my point of view. There is a noe hierarchy, but no meshes at all. So all not used materials will be removed.

kimkulling:
This is just a validation message from the model validation and from ym point of view this is correct.

stephengold:
> So all not used materials get removed correctly from my point of view. There is a noe hierarchy, but no meshes at all. So all not used materials will be removed.

@kimkulling If no materials are used, and all not-used materials get removed, then all materials should be removed. That won't happen if the post-processing aborts.

> This is just a validation message from the model validation and from ym point of view this is correct.

@kimkulling The message came from post-processing, not from model validation.

kimkulling:
Which behavior would you expect?

stephengold:
For a model with zero meshes, I expect that RRM post-processing should result in a model with zero materials.

kimkulling:
Good point, fix is on its way.

stephengold:
After the fix (GIT commit 3990ec80) loading the test model results in a segmentation violation:
```console

$ bin/assimpd info ~/Downloads/Running.gltf -rrm
[...]

PR #5439, opened 2024-01-25
RemoveRedundantMaterials:  avoid crash when all materials are redundant or unused

stephengold:
After analyzing the materials, if we find that all of them were redundant or unused then:
1. the rebuilding and remapping steps are skipped, 
2. `pScene->mMaterials` is set to `nullptr`,
3. `pScene->mNumMaterials` is set to zero, and
4.  resources are freed.

Note that the file's copyright notice has been updated.

### DIFF
--- a/code/PostProcessing/RemoveRedundantMaterials.cpp
+++ b/code/PostProcessing/RemoveRedundantMaterials.cpp
@@ -3,7 +3,7 @@
 Open Asset Import Library (assimp)
 ---------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


stephengold:
This PR is intended to solve issue #5438 .

stephengold:
PR #5451 addresses the same issue.
Since #5451 has been integrated, this PR is no longer useful.
There's no point in resolving the merge conflicts caused by #5451.
I am closing this PR.

However, I wonder why #5451 was preferred over this PR.

PR #5441, opened 2024-01-25
FBX Blendshape `FullWeight: Vec<Float>` -> `FullWeight: Vec<Float>`

### LINKED ISSUE #5440: Bug: FBX Blendshape Exporter exports double where binary FBX expects float
**Describe the bug**
FBX blendshape exports currently exports a stub for the full-weights that are allowed for each blendshape,
which is the maximum lerp allowed between each blendshape. Right now it is exported as float, which is fine for ascii, but the FBX SDK expects it to be a double.

**To Reproduce**
1. Import FBX with Blendshape
2. Export FBX as Binary
3. Import into FBX official converter
4. See that it is corrupted (altho super generic error)

**Expected behavior**
Should be valid both as ascii and binary.

**Screenshots**

image
Line 1810

Affects all platforms.

**Additional context**
Will submit a PR to fix soon.

JulianKnodt:
Closes #5440

### DIFF
--- a/code/AssetLib/FBX/FBXExporter.cpp
+++ b/code/AssetLib/FBX/FBXExporter.cpp
@@ -1807,7 +1807,7 @@ void FBXExporter::WriteObjects ()
         p.AddP70numberA("DeformPercent", 0.0);
         sdnode.AddChild(p);
         // TODO: Normally just one weight per channel, adding stub for later development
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine, curious to see if this has an impact of other users. But let's give this a try.

kimkulling:

PR #5445, opened 2024-01-30
Fix identity matrix check

fvbj:
Adds an extra epsilon value to check the matrix4x4 identity. The method is also used to export to GLTF/GLTF2 format to check node transformation matrices. The epsilon value can be set using AI_CONFIG_CHECK_IDENTITY_MATRIX_EPSILON with the default value set to 10e-3f for backward compatibility of legacy code.


fvbj:
Link to the issue https://github.com/assimp/assimp/issues/5455

kimkulling:

PR #5451, opened 2024-02-01
Kimkulling/fix bahavior of remove redundat mats issue 5438

kimkulling:
- closes https://github.com/assimp/assimp/issues/5438


stephengold:
It appears that issue #5461 resulted from this fix.

PR #5472, opened 2024-02-22
Fixed GLTF read reference file

ahcorde:
There is a bug in the GTLF reference file reader. 

`Read` uses [fread](https://www.tutorialspoint.com/c_standard_library/c_function_fread.htm), this function return the number of bytes readed. not really sure why this condition was compared with 1.

### DIFF
--- a/code/AssetLib/glTF/glTFAsset.inl
+++ b/code/AssetLib/glTF/glTFAsset.inl
@@ -242,7 +242,7 @@ inline bool Buffer::LoadFromStream(IOStream &stream, size_t length, size_t baseO
 
     mData.reset(new uint8_t[byteLength], std::default_delete<uint8_t[]>());
 
-    if (stream.Read(mData.get(), byteLength, 1) != 1) {
+    if (stream.Read(mData.get(), byteLength, 1) != byteLength) {
         return false;
     }
     return true;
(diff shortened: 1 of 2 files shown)


jwnimmer-tri:
The `::read` function is specified to return `nmemb`, not the number of bytes.

See e.g. https://github.com/ros2/rviz/pull/1001 for related.

ahcorde:
> The [fread function](https://man7.org/linux/man-pages/man3/fread.3.html) is specified to return `nmemb`, not the number of bytes.
> 
> See e.g. [ros2/rviz#1001](https://github.com/ros2/rviz/pull/1001) for related.

I just print the return value from the `Read` method and it's the same as the number of bytes, not sure why this is not returning 1

jwnimmer-tri:
> I just print the return value from the `Read` method and it's the same as the number of bytes, not sure why this is not returning 1.

I suppose whichever concrete implementation of `IOStream& stream` you are using in your experiment has a bug where it is mistakenly returning the number of bytes.

kimkulling:
As far as I know we are using rdead to read unsigned chars. So even  if we are reading in a struct with different members the stream willl return its size in bytes. And that is the reason for thr return values.

kimkulling:
Ok, the 1 was correct. Thanks for your contribution. But I will close this one.

PR #5480, opened 2024-02-29
Eliminate non-ascii comments in clipper

adfwer233:
This PR updated some comments in clipper, eliminating the non-ascii characters.

When compiling with MSVC, these non-ascii characters cause warning C4819 which may be treated as error. Deleting these non-ascii makes me build assimp with MSVC successfully in my environment : )


kimkulling (approved):
Looks fine.

kimkulling:

PR #5490, opened 2024-03-10
Fix compilation for MSVC14.

LukasBanana:
This PR fixes Assimp compilation for older compilers with lack of C++17 support such as MSVC14.

- `std::min`/`max` were not defined in *StackAllocator.inl*; Also added explicit template arguments to break macro expansion if *Windows.h* is included prior and `NOMINMAX` macro is not present.
- Made `static_assert` statements compatible with C++11 in *ProcessHelper.cpp*.
- Removed unused `string_view` include in *ObjFileParser.cpp*.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5498, opened 2024-03-15
updated minizip to last version

mosfet80:
Version 1.3.1 has these key changes:

Reject overflows of zip header fields in minizip
Fix bug in inflateSync() for data held in bit buffer Add LIT_MEM define to use more memory for a small deflate speedup Fix decision on the emission of Zip64 end records in minizip Add bounds checking to ERR_MSG() macro, used by zError() Neutralize zip file traversal attacks in miniunz
Fix a bug in ZLIB_DEBUG compiles in check_match()
Version 1.3 has these key changes:

Building using K&R (pre-ANSI) function definitions is no longer supported. Fixed a bug in deflateBound() for level 0 and memLevel 9. Fixed a bug when gzungetc() is used immediately after gzopen(). Fixed a bug when using gzflush() with a very small buffer. Fixed a crash when gzsetparams() is attempted for a transparent write. Fixed test/example.c to work with FORCE_STORED.
Fixed minizip to allow it to open an empty zip file. Fixed reading disk number start on zip64 files in minizip. Fixed a logic error in minizip argument processing.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #5516, opened 2024-03-27
Add initial macOS support to C4D importer

AlexTMjugador:
The [latest Cineware SDK explicitly asserts macOS support in its documentation](https://developers.maxon.net/docs/cw/22_004/page_general_information.html), but Assimp's C4D importer only works with Windows MSVC targets. Let's improve its portability by refactoring importer code to not depend on MSVC-only data types and quirks, and add support for linking against the universal macOS static libraries provided in the Cineware SDK.

Note that the C4D importer still cannot support Linux platforms because Maxon does not provide the necessary precompiled Cineware libraries for that platform. Windows MinGW targets are also out of the question as the MinGW toolchain uses compiled libraries in a different format.

### DIFF
--- a/code/AssetLib/C4D/C4DImporter.cpp
+++ b/code/AssetLib/C4D/C4DImporter.cpp
@@ -46,10 +46,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // no #ifdefing here, Cinema4D support is carried out in a branch of assimp
 // where it is turned on in the CMake settings.
 
-#ifndef _MSC_VER
-#   error C4D support is currently MSVC only
-#endif
-
 #include "C4DImporter.h"
 #include <memory>
 #include <assimp/IOSystem.hpp>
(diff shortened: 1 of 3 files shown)


code/AssetLib/C4D/C4DImporter.cpp
```diff
@@ -305,7 +301,7 @@ void C4DImporter::RecurseHierarchy(BaseObject* object, aiNode* parent) {
 
     // based on Cineware sample code
     while (object) {
-        const LONG type = object->GetType();
```
kimkulling, line 308: Does this work for32 bit and 64 bit?

kimkulling (changes requested):

code/AssetLib/C4D/C4DImporter.cpp
AlexTMjugador: Yeah, the variable type has been updated to `Int32`, matching the return type of the Cineware SDK, which consistently utilizes a 32-bit length across all platforms. (It's worth noting that while the portability to 32-bit platforms is accounted for, Cineware binaries are only available for 64-bit platforms.)

kimkulling:

PR #5521, opened 2024-03-31
Respect merge identical vertices in ObjExporter

### LINKED ISSUE #5513: Bug: Export OBJ `JoinIdenticalVertices = false` not respected
**Describe the bug**
When exporting an OBJ file with `JoinIdenticalVertices = false`, it will still merge vertices.
This is because it does not consider the face index at all when exporting the mesh.
Instead, the index can be added to the `vertexData` struct if and only if `JoinIdenticalVertices = false` (and the comparator should also be updated).

**Expected behavior**
Should not merge identical vertices iff their index is not the same.

JulianKnodt:
Fixes #5513


kimkulling (approved):
Looks fine.

kimkulling:

PR #5535, opened 2024-04-08
CMake: Allow linking draco statically if ASSIMP_BUILD_DRACO_STATIC is set.

alexrp:
In [Silk.NET](https://github.com/dotnet/Silk.NET) we're packaging Assimp as a native dependency. We'd like to enable Draco in our Assimp build but would also like to avoid having to ship the extra shared library. This patch adds an option `ASSIMP_BUILD_DRACO_STATIC` which allows building and linking Draco statically into Assimp.

This only works when using the bundled Draco since there doesn't seem to be an obvious way to make it work with `find_package()`.

I'm only able to test this on Linux. Hopefully it just works on Windows and macOS in CI...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -621,6 +621,10 @@ ELSE ()
   ADD_DEFINITIONS( -DASSIMP_BUILD_NO_C4D_IMPORTER )
 ENDIF ()
 
+if(ASSIMP_BUILD_DRACO_STATIC)
+  set(ASSIMP_BUILD_DRACO ON)
+endif()
+
 # Draco requires cmake 3.12
 IF (DEFINED CMAKE_VERSION AND "${CMAKE_VERSION}" VERSION_LESS "3.12")
   message(NOTICE "draco requires cmake 3.12 or newer, cmake is ${CMAKE_VERSION} . Draco is disabled")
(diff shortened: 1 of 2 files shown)


kimkulling (approved):
Looks fine.

kimkulling:

Brakebein:
FYI, it's a little breaking change! So far, I was building with
```
-DBUILD_SHARED_LIBS=OFF -DASSIMP_BUILD_DRACO=ON
```
After upgrading to v5.4.1 this build config was failing:

```
CMake Error at CMakeLists.txt:698 (set_target_properties):
  set_target_properties Can not find target to add properties to:
  draco_shared

CMake Error at CMakeLists.txt:379 (set_target_properties):
  set_target_properties Can not find target to add properties to:
  draco_shared
Call Stack (most recent call first):
  CMakeLists.txt:703 (TARGET_USE_COMMON_OUTPUT_DIRECTORY)

CMake Error at CMakeLists.txt:711 (INSTALL):
  INSTALL TARGETS given target "draco_shared" which does not exist.
```
I needed to explicitly set `ASSIMP_BUILD_DRACO_STATIC` to fix it.
```
-DBUILD_SHARED_LIBS=OFF -DASSIMP_BUILD_DRACO=ON -DASSIMP_BUILD_DRACO_STATIC=ON
```

Issue #5519, opened 2024-03-29
assimp is not compiled with clang


michaelzaki-12:
i am using cmake for my project, but when i use clang to compile it gives me this error 
Severity	Code	Description	Project	File	Line	Suppression State	Details
Error		unsafe buffer access [-Werror,-Wunsafe-buffer-usage]	C:\Users\User\Source\Repos\MyProject\out\build\x64-Clang-Release\piratesengine	C:\Users\User\Source\Repos\MyProject\assimp\include\assimp\fast_atof.h	178		
But I don't Get the same error when compiling with MSVC

Here is How my cmake look like

# assimp
set(BUILD_SHARED_LIBS OFF CACHE INTERNAL "" FORCE)
set(ASSIMP_BUILD_TESTS OFF CACHE INTERNAL "" FORCE)
set(ASSIMP_NO_EXPORT ON CACHE INTERNAL "" FORCE)
set(ASSIMP_BUILD_TESTS OFF CACHE INTERNAL "" FORCE)
set(ASSIMP_BUILD_ASSIMP_TOOLS OFF CACHE INTERNAL "" FORCE)

add_subdirectory(assimp)

include_directories(${CMAKE_SOURCE_DIR}/assimp/include)
target_link_libraries( MyProject PRIVATE ${OPENGL_LIBRARIES} glfw assimp)

i am Using windows 11 with visual studio 2022.

**Additional context**

kimkulling:
I am still struggling with the clang build for windows. Hopefully we will be able to see the issue there.

pmp-p:
maybe related with clang 20 wasm i get ```
/opt/python-wasm-sdk/src/assimpjs-wasm/assimp/code/Common/SceneCombiner.cpp:987:14: error: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'aiFace' [-Werror,-Wnontrivial-memaccess]
  987 |     ::memcpy(dest, old, sizeof(Type) * num);
      |              ^
/opt/python-wasm-sdk/src/assimpjs-wasm/assimp/code/Common/SceneCombiner.cpp:1096:5: note: in instantiation of function template specialization 'Assimp::GetArrayCopy<aiFace>' requested here
 1096 |     GetArrayCopy(dest->mFaces, dest->mNumFaces);
      |     ^
/opt/python-wasm-sdk/src/assimpjs-wasm/assimp/code/Common/SceneCombiner.cpp:987:14: note: explicitly cast the pointer to silence this warning
[...]

PR #5537, opened 2024-04-09
Add windows clang to CI

kimkulling:
- Add windows clang build step
- Help to evaluate https://github.com/assimp/assimp/issues/5519


test/unit/utIOStreamBuffer.cpp
```diff
@@ -86,7 +84,7 @@ TEST_F( IOStreamBufferTest, open_close_Test ) {
     const auto dataCount = dataSize / sizeof(*data);
 
     char fname[]={ "octest.XXXXXX" };
-    auto* fs = MakeTmpFile(fname);
+    FILE *fsm = fopen(fname, "w+");
```
turol, line 87: This is not even remotely correct. `MakeTmpFile` securely creates a file which did not exist previously.

kimkulling: Right!It was a test to see if this will influence the failing unittest somehow. Any ideas what is going wrong there?

turol: The `fopen` thing? That's MS braindeadedness, just add `_CRT_SECURE_NO_WARNINGS` preprocessor macro somewhere in CMakelists.txt

kimkulling: No, the failed unittest with the error code 127 when opening the tempfile.

turol: Looks like it assumes that MSVC is the same as win32 and GCC/clang isn't.

https://github.com/assimp/assimp/blob/f63625256c4e6f18fca8e7dc857b47a02320e867/test/unit/UnitTestFileGenerator.h#L53

kimkulling: Fixed

PR #5542, opened 2024-04-11
Reverts the changes introduced by commit ad766cb in February 2022.

johannesugb:
ASSIMP's way of calculating bitangents before commit ad766cb was just fine, but unfortunately it was changed by that commit in February 2022, breaking compatibility to previous code.

It introduces two problems:
- It orthogonalizes bitangents w.r.t. tangents (which is generally wrong for texture-mapped meshes).
- It changes the handedness, which is arguably not totally wrong, but breaks compatibility to previous code which uses ASSIMP.

More detailed explanation why the old code was just fine is given in Q&A #5512 

**Suggestion: Just revert to the old code, which this pull request suggests.**


johannesugb:
@kimkulling are these checks supposed to work or not? 🙃

turol:
Yes they're supposed to work. You forgot a `;`

johannesugb:
> Yes they're supposed to work. You forgot a `;`

OMG, shame! 🙈

kimkulling:
After reading all the explanations I do agree that we should use the original code.

PR #5559, opened 2024-04-30
fix ASE loader crash when *MATERIAL_COUNT or *NUMSUBMTLS is not specified or is 0

Garux:
code was doing vector[0u - 1] dereference in this case
[example model](https://github.com/assimp/assimp/files/15169736/crack.zip)


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5557, opened 2024-04-30
Bug: error: ‘void* __builtin_memmove(void*, const void*, long unsigned int)’ forming offset 8 is out of the bounds [0, 8] [-Werror=array-bounds=]


jiapei100:
**Describe the bug**

```console
[ 35%] Building CXX object code/CMakeFiles/assimp.dir/AssetLib/Blender/BlenderScene.cpp.o
cd ....../assimp/build/code && /usr/bin/c++ -DASSIMP_BUILD_DLL_EXPORT -DASSIMP_BUILD_NO_C4D_IMPORTER -DASSIMP_BUILD_NO_M3D_EXPORTER -DASSIMP_BUILD_NO_M3D_IMPORTER -DASSIMP_BUILD_NO_OWN_ZLIB -DASSIMP_IMPORTER_GLTF_USE_OPEN3DGC=1 -DMINIZ_USE_UNALIGNED_LOADS_AND_STORES=0 -DOPENDDLPARSER_BUILD -DRAPIDJSON_HAS_STDSTRING=1 -DRAPIDJSON_NOMEMBERITERATORCLASS -Dassimp_EXPORTS -I....../assimp/build/include -I....../assimp/build -I....../assimp/include -I....../assimp/code -I....../assimp/. -I....../assimp/code/../contrib/pugixml/src -I....../assimp/code/../contrib/utf8cpp/source -I....../assimp/code/../contrib/rapidjson/include -I....../assimp/code/../contrib -I....../assimp/code/../contrib/openddlparser/include -I....../assimp/code/../include -I....../assimp/build/code/../include -I/usr/include/minizip -fvisibility=hidden -fno-strict-aliasing -Wall -std=c++17 -fPIC -O3 -DNDEBUG -std=gnu++17 -fPIC -Wno-dangling-reference -Wall -Werror -MD -MT code/CMakeFiles/assimp.dir/AssetLib/Blender/BlenderScene.cpp.o -MF CMakeFiles/assimp.dir/AssetLib/Blender/BlenderScene.cpp.o.d -o CMakeFiles/assimp.dir/AssetLib/Blender/BlenderScene.cpp.o -c ....../assimp/code/AssetLib/Blender/BlenderScene.cpp
In file included from /usr/include/c++/13/bits/specfun.h:43,
                 from /usr/include/c++/13/cmath:3699,
                 from ....../assimp/include/assimp/vector2.h:53,
                 from ....../assimp/include/assimp/types.h:64,
[...]

PR #5577, opened 2024-05-15
Temporary fix for #5557 GCC 13+ build issue -Warray-bounds

dbs4261:
Added cmake code to remove warnings for array-bounds and stringop-overflow specifically for MDLLoader.cpp on GCC 13 and above.
See #5557


kimkulling (approved):
Lets test it.

Issue #3133, opened 2020-04-05
[Assimp viewer] Unable to open the same asset (file) after it was closed.


malortie:
**General testing info**

Branch: master
Commit: 085212e58ab0cf6dc96256e1577e9e31d0a8f3e2
Tested on: 32 bits and 64 bits

**Description**

Attempting to open the same asset after it was closed does not seem to work. 

**Steps to reproduce this issue**

- Launch Assimp viewer
- In **Menu / Toolbar -> Viewer**, click **Open Asset**, browse and load file 
`test/models/OBJ/spider.obj`
- Ensure that you can see the loaded model.
- In **Menu / Toolbar -> Viewer**, click **Close Asset**.
- Ensure that the model has been unloaded.
- In **Menu / Toolbar -> Viewer**, click **Open Asset**, browse and load file 
`test/models/OBJ/spider.obj`

After performing these steps, you should be able to see the message "Nothing to display" as if no asset was loaded.

tellypresence:
@malortie Is this still a problem?

malortie:
See https://github.com/assimp/assimp/pull/5582#issuecomment-2119297677

This issue can be closed.

Issue #3136, opened 2020-04-05
[Assimp viewer] Animation frame seems to advance even when playback is stopped.


malortie:
**General testing info**

Branch: master
Commit: 085212e58ab0cf6dc96256e1577e9e31d0a8f3e2
Tested on: 32 bits and 64 bits

**Description**

When stopping and then resuming an animation, the track bar does not seem to resume exactly at the position at the moment the animation was stopped.

**Steps to reproduce this issue**

- In **Menu / Toolbar -> Viewer**, click **Open Asset**, browse and load file
`test/models-nonbsd/MD5/Bob.md5mesh`
- Ensure that you can see the loaded model.
- Select the first animation of the model and click **Play**.
- Halfway through the animation, click **Stop**. 
- Wait about a second or two, and click **Play** to resume the animation.

After performing these steps, you should see the track bar set to where it would have been as if the animation would have not been stopped.

tellypresence:
@malortie Is this still a problem?

malortie:
See https://github.com/assimp/assimp/pull/5582#issuecomment-2119297677

This issue can be closed.

PR #5582, opened 2024-05-17
Assimp viewer fixes

JLouis-B:
The PR fixes a few bugs in the assimp viewer, on statistics, the "close asset", and the animations UI


kimkulling (approved):
I would recommend to use strncmp instead of strcpy, But I will refactor that lateron.

kimkulling:

JLouis-B:
Ok, thanks 👍 
By the way, you can close these bugs as fixed :
https://github.com/assimp/assimp/issues/3133
https://github.com/assimp/assimp/issues/3136

I'm not sure I understand the point about strncmp, it's just used to compare strings (and strcpy to copy), isn't it?
And I was wondering, the last time I downloaded assimp, there seemed to be a new viewer based on Qt. Has it been removed?

Issue #5592, opened 2024-05-23
Bug: GCC-14 fails to build when default symbol visibility is set to hidden 


dbs4261:
**Describe the bug**
GCC 14 will not build the OBJ loader because of a maybe-uninitialized error

**To Reproduce**
Generate with cmake: `cmake -S source -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc-14 -DCMAKE_CXX_COMPILER=/usr/bin/g++-14`
Build with cmake: `cmake --build build --target all --parallel $(nproc)`

**Error Message**
```c++
FAILED: code/CMakeFiles/assimp.dir/AssetLib/Obj/ObjFileParser.cpp.o 
[...]

PR #5593, opened 2024-05-23
Fix for #5592 Disabled maybe-uninitialized error for AssetLib/Obj/ObjFileParser.cpp

dbs4261:
Fix for #5592 Disabled maybe-uninitialized error for AssetLib/Obj/ObjFileParser.cpp as it seems to be spurious. Also fixed error message for previous GCC build issue modification.


kimkulling (approved):
Looks fine.

PR #5596, opened 2024-05-25
Fix compile bug on gcc under release mode

Jojonobody2:
There's a bug when using c++ (gcc) on linux when compiling in release mode


kimkulling (approved):
Looks fine.

kimkulling:
... but does not work.

PR #5600, opened 2024-05-27
Converted a size_t to mz_uint that was being treated as an error

BradlyLanducci:
Original error during building 

```assimp\contrib\zip\src\zip.c(1239,20): error C2220: the following warning is treated as an error```

```assimp\contrib\zip\src\zip.c(1239,20): warning C4267: 'function': conversion from 'size_t' to 'mz_uint', possible loss of data```


ShanuPatel:
Had the same issue casting fixed the issue but do you thing this type of cast is safe i heard that its unsafe

kimkulling (approved):
Looks fine.

kimkulling:

Issue #5601, opened 2024-05-28
Bug: Compile error in Visual Studio 2022 in Debug mode


Beosar:
I had to change lines 5270-5274 in contrib/zip/src/miniz.h because building assimp on Visual Studio in Debug mode would fail with a warning which is treated as an error. This is the correct code:

```
static MZ_FORCEINLINE size_t
mz_zip_array_range_check(const mz_zip_array *pArray, size_t index) {
  MZ_ASSERT(index < pArray->m_size);
  return index;
}
```

PR #5606, opened 2024-05-31
Build: Fix compilation for VS-2022 debug mode - warning

kimkulling:
- closes https://github.com/assimp/assimp/issues/5601


Beosar:
The commit did not fix the warning, you missed the change to the function parameter: `mz_uint index` should be changed to `size_t index` on line 5271.

PR #5608, opened 2024-06-06
fix PLY binary export color from float to uchar

michaelsctts:
Exporting PLY files in binary format leads to file corruption. The PLY header sets the color properties to uchar, but 16 bytes representing RGBA are written as floats instead of 4 bytes.


kimkulling (approved):
Looks fine

kimkulling:

PR #5630, opened 2024-06-20
Fix collada uv channels - temporary was stored and then updated.

StepanHrbek:
So all information on uv channels was ignored.


StepanHrbek:
Solves issue #5631

kimkulling (commented):
Looks fine.

kimkulling:

PR #5634, opened 2024-06-24
remove ASE parsing break

Garux:
remove ASE parsing break added in c1968823adfb8c997f98671becca51fc54614da7 : original intent was to keep parsing

crash case (iMaterialCount = 0) is handled by 47dbabadcd683724be26109c757ad86f4645d280


kimkulling (approved):
Makes sense.

kimkulling:
Good catch, I haven't seen this possibility for a crash.

kimkulling:

Issue #5631, opened 2024-06-20
Bug: Collada uv channel numbers are ignored.


StepanHrbek:
**Describe the bug**
Collada file can specify uv channel number but Assimp ignores it.
For example <bind_vertex_input semantic="CHANNEL1" input_semantic="TEXCOORD" input_set="0"/> describes channel with string "CHANNEL1" and number 0, where the string is used only when parsing file, and the number should be propagated into aiScene unchanged. Assimp reads the number, but then forgets it.

**To Reproduce**
1. Import any .dae with at least one textured mesh and non trivial value in input_set=, it could look like <bind_vertex_input semantic="CHANNEL1" input_semantic="TEXCOORD" input_set="5"/> 
2. After import, query uv channel number with aimaterial->Get(_AI_MATKEY_UVWSRC_BASE, ...) and see that it's not 5.

**Expected behavior**
Assimp should not ignore the number in input_set=.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/5630

tellypresence:
Please attach a demo .dae model file which had been affected by this problem, to make it easier to check if this fix gets broken in future

StepanHrbek:
https://github.com/StepanHrbek/LightsprintSDK/blob/main/data/scenes/koupelna/koupelna4.dae contains material with set="0" and set="1" and mesh with input_set="0" input_set="1".
It means material wants uv channels 0,  1 and mesh contains uv channel 0, 1.

Before #5631, Assimp imported mesh with 0, 1 but material with 1, 2, so texturing is broken.
#5631 fix makes Assimp import both with 0, 1.

PR #5636, opened 2024-06-24
Allow empty slots in mTextureCoords

### LINKED ISSUE #5632: Bug: Collada uvs from channels a,b,c are imported into Assimp channels 0,1,2.
**Describe the bug**
Collada: Now that we properly import uv channel numbers used by textures (merged today), it becomes apparent that second half of problem is importing data into correct uv channels. We properly find out that mesh contains data in channels a, b and c (example). But then d6aacefa1ef1219b85028b4038ef7e861c2e9462 changes a,b,c to 0,1,2 and we store data into wrong uv channels.

**To Reproduce**
1. Import and render .dae with texture mapped mesh that uses uv channel 1.
2. See that render is wrong, material wants data from channel 1 but data are stored in channel 0.

**Expected behavior**
Collada files should be imported with original uv channel numbers in both meshes and materials.

**Additional context**
I have problem locally fixed by reverting "empty slots are not allowed" paragraph of  d6aacefa1ef1219b85028b4038ef7e861c2e9462,
I'll send patch.
I'll also document that empty slots in mTextureCoords are ok... are they?
Also, I don't use GetNumUVChannels(), but I see it's implemented wrong. If we allow empty slots, it needs small fix.
Finally, I'm concerned about .dae files that contain channel numbers higher than AI_MAX_NUMBER_OF_TEXTURECOORDS. Do they exist? Are they common? (My approach in LightsprintSDK was to make array of channels resizable, so I never had to figure out how large numbers exist in wild.)

StepanHrbek:
This fixes #5632 by reverting undocumented assumption introduced in 2008.
It does not change API, it just explicitly states that empty slots are possible, after 2008 commits made them impossible without any trace of explanation why.
I understand if you don't want to change 16 years old status quo, there is risk of breaking userspace. But then the assumption about ordering of slots should be documented and fix of #5631 reverted.

### DIFF
--- a/code/PostProcessing/ValidateDataStructure.cpp
+++ b/code/PostProcessing/ValidateDataStructure.cpp
@@ -371,20 +371,7 @@ void ValidateDSProcess::Validate(const aiMesh *pMesh) {
         ReportWarning("There are unreferenced vertices");
     }
 
-    // texture channel 2 may not be set if channel 1 is zero ...
-    {
-        unsigned int i = 0;
-        for (; i < AI_MAX_NUMBER_OF_TEXTURECOORDS; ++i) {
-            if (!pMesh->HasTextureCoords(i)) break;
-        }
-        for (; i < AI_MAX_NUMBER_OF_TEXTURECOORDS; ++i)
-            if (pMesh->HasTextureCoords(i)) {
-                ReportError("Texture coordinate channel %i exists "
-                            "although the previous channel was nullptr.",
-                        i);
-            }
-    }
-    // the same for the vertex colors
+    // vertex color channel 2 may not be set if channel 1 is zero ...
     {
         unsigned int i = 0;
         for (; i < AI_MAX_NUMBER_OF_COLOR_SETS; ++i) {
--- a/code/AssetLib/Collada/ColladaLoader.cpp
+++ b/code/AssetLib/Collada/ColladaLoader.cpp
@@ -625,16 +625,14 @@ aiMesh *ColladaLoader::CreateMesh(const ColladaParser &pParser, const Mesh *pSrc
     }
 
     // same for texture coords, as many as we have
-    // empty slots are not allowed, need to pack and adjust UV indexes accordingly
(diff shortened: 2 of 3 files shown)


StepanHrbek:
Sorry, 3 older commits somehow slipped in, only 2 commits from today are relevant.

code/PostProcessing/ValidateDataStructure.cpp
```diff
@@ -371,6 +371,7 @@ void ValidateDSProcess::Validate(const aiMesh *pMesh) {
         ReportWarning("There are unreferenced vertices");
     }
 
+#if 0 // no documentation found on why is it necessary
```
kimkulling, line 374: Could you remove the code, please?

kimkulling (changes requested):
Just one small finding.

code/PostProcessing/ValidateDataStructure.cpp
StepanHrbek: I removed 3 unwanted commits and that #if 0 block.
It's first time I'm using hard reset on public branch, is this how things should be done?

StepanHrbek:
test/models/Collada/kwxport_test_vcolors.dae contains material with set="1" and mesh with input_set="1".
It means material wants uv channel 1 and mesh contains uv channel 1.

Before #5631, Assimp accidentally worked by importing both numbers as 0.
#5631 fix changes material number to 1.
#5632 fix changes also mesh number to 1.

It would be good to have test that checks that meshes have all uv channels requested by materials.
It would start failing on kwxport_test_vcolors.dae after #5631 fix, will stop failing after #5632 fix.

code/PostProcessing/ValidateDataStructure.cpp
kimkulling: You can just change it on your branch. We will merge this as one commit. So no hard reset will be necessary.

kimkulling (approved):

kimkulling:

PR #5639, opened 2024-06-25
Update zip.c

ThatOSDev:
Now works with GCC without failing the compilation.


kimkulling:

PR #5644, opened 2024-06-29
Fixes some uninit bool loads

TinyTinni:
This PR fixes some bool loads which are not initialized. With ubsan and the "option -fsanitize=bool", this results in a runtime error during test execution.

When running the unittest suite with activated UBSan with Clang 17.0.6, I get 2 runtime errors.

- `mRemoveEmptyBones` is not initialized when running the test "LimitBoneWeightTest". It is not initialized due to missing `SetupProperties` call. I decided to initialize the option with default off, in case someone else calls "processMesh" without setting up the properties first. I decided for default true value, is this seems to be the default value in SetupProperties, when it is not defined
```
this->mRemoveEmptyBones = pImp->GetPropertyInteger(AI_CONFIG_IMPORT_REMOVE_EMPTY_BONES, 1) != 0;
```

- uninit load in "utMetadata.copy_test". Here, in the "bool" test case, the recieving bool is not initialized. It gets theoretically loaded in the "GetAiType" overloaded function (due to call-by-copy) and is never used. In Release mode, it probably just get optimized away, but in Debug mode, a false warning of UBSan can be annoying. I also added call-by-ref functions for the other pod types, in case UBSan adds these checks for the other types.

I hope you like it.

edit: [1] reference to ubsan with "-fsanitize=bool" option: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

### DIFF
--- a/code/PostProcessing/LimitBoneWeightsProcess.cpp
+++ b/code/PostProcessing/LimitBoneWeightsProcess.cpp
@@ -53,7 +53,8 @@ namespace Assimp {
 
 // ------------------------------------------------------------------------------------------------
 // Constructor to be privately used by Importer
-LimitBoneWeightsProcess::LimitBoneWeightsProcess() : mMaxWeights(AI_LMW_MAX_WEIGHTS) {
+LimitBoneWeightsProcess::LimitBoneWeightsProcess() :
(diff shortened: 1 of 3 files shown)


kimkulling (approved):
Looks fine for me!

kimkulling:

PR #5648, opened 2024-07-01
feat: add option for creating XCFramework and configure minimum iOS target

AKosmachyov:
P.S. I noticed that arm64 does not work with the iOS simulator on Apple Silicon. I'm looking for ways to fix this


kimkulling (approved):
Cannot test this by my own, but it looks right.

kimkulling:

PR #5649, opened 2024-07-01
Fix: StackAllocator.h linkage fix

bedwardly-down:
I'm currently porting over the Thunder game engine to work on the BSD platforms (with its developer's help) and ran into a peculiar issue (see number 8 here: https://github.com/thunder-engine/thunder/issues/770). This fix will require std:c++17 set for certain build environments, so if there's a better way to go about it, I'm all ears.

### DIFF
--- a/code/Common/StackAllocator.h
+++ b/code/Common/StackAllocator.h
@@ -88,6 +88,11 @@ private:
 
 } // namespace Assimp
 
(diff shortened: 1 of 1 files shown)


bedwardly-down:
When I get home, I’ll fix this. I forgot to put it in a separate topic branch.

PR #5658, opened 2024-07-08
Fix a fuzz test heap buffer overflow in mdl material loader

sgayda2:
This bug was discovered using a fuzz test. With this patch the fuzz test passes.

This is likely not the only issue in this file, but i don't have enough expertise or knowledge to improve this further.


kimkulling (approved):
Looks fine for me!

sgayda2:
I don't have write access to the repo, so I am not clear on how to proceed to merge this

kimkulling:

PR #5675, opened 2024-07-18
FBX: add metadata of ainode as properties

fuhaixi:
Not sure if it's the correct way. But better than nothing.


kimkulling (approved):
Great, I did not know that this kind of usage is possible for fbx.

kimkulling:

PR #5681, opened 2024-07-23
Fix casting typo in D3MFExporter::writeBaseMaterials

ochafik:
color channels < 1.0f were zeroed out (first cast to int, then * 255).


kimkulling:

Issue #5689, opened 2024-07-29
Linker Error: p2t::Point 


AMZN-Gene:
Unable to compile master branch

**To Reproduce**
1. `git clone https://github.com/assimp/assimp.git src`
1. ```cmake -S ./src -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release  -DASSIMP_BUILD_ASSIMP_TOOLS=ON -DASSIMP_WARNINGS_AS_ERRORS=OFF src/CMakeLists.txt```
1. `cmake --build ./src --config release`
1. See compiler error
```
  assimp.vcxproj -> D:\3p-package-source\package-system\assimp\temp\src\lib\Release\assimp-vc143-mt.lib
assimp-vc143-mt.lib(IFCOpenings.obj) : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl
 p2t::Point::Point(double,double)" (__imp_??0Point@p2t@@QEAA@NN@Z) referenced in function "bool __cdecl Assimp::IFC::Tr
yAddOpenings_Poly2Tri(class std::vector<struct Assimp::IFC::TempOpening,class std::allocator<struct Assimp::IFC::TempOp
ening> > const &,struct Assimp::IFC::TempMesh &)" (?TryAddOpenings_Poly2Tri@IFC@Assimp@@YA_NAEBV?$vector@UTempOpening@I
FC@Assimp@@V?$allocator@UTempOpening@IFC@Assimp@@@std@@@std@@AEAUTempMesh@12@@Z) [D:\3p-package-source\package-system\a
ssimp\temp\src\tools\assimp_cmd\assimp_cmd.vcxproj]
assimp-vc143-mt.lib(IFCOpenings.obj) : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl
 p2t::CDT::CDT(class std::vector<struct p2t::Point *,class std::allocator<struct p2t::Point *> > const &)" (__imp_??0CD
T@p2t@@QEAA@AEBV?$vector@PEAUPoint@p2t@@V?$allocator@PEAUPoint@p2t@@@std@@@std@@@Z) referenced in function "bool __cdec
l Assimp::IFC::TryAddOpenings_Poly2Tri(class std::vector<struct Assimp::IFC::TempOpening,class std::allocator<struct As
[...]

PR #5693, opened 2024-07-31
Added ADD condition in poly2tri dll_symbol.h to only define macros fo…

mkuritsu:
I found out that the issue #5689 was caused by the recent update of poly2tri library in #5682,  the added [dll_symbol.h](https://github.com/assimp/assimp/commit/571ba09dc779e453d0074bfa169aa46003e15296#diff-e72e15e95c897b107e2f693545f2656e72b7164728708fc52b817cc2d8d2971b) file defines the macro P2T_COMPILER_DLLEXPORT to allow to exporting/importing functions for DLL but the only check that it performs to define those is the platform, this causes even when building assimp as a static-link library to still have the DLL exports. I made this change to also check if ASSIMP_BUILD_DLL_EXPORT, I'm not sure if this would be an acceptable change to solve this issue or if would be better to create a different macro for poly2tri or even open an issue to get a specific macro defined in the library.

### DIFF
--- a/contrib/poly2tri/poly2tri/common/dll_symbol.h
+++ b/contrib/poly2tri/poly2tri/common/dll_symbol.h
@@ -31,11 +31,11 @@
 
 #pragma once
 
-#if defined(_WIN32)
+#if defined(_WIN32) && defined(ASSIMP_BUILD_DLL_EXPORT)
 #  pragma warning( disable: 4273)
 #  define P2T_COMPILER_DLLEXPORT __declspec(dllexport)
 #  define P2T_COMPILER_DLLIMPORT __declspec(dllimport)
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) && defined(ASSIMP_BUILD_DLL_EXPORT)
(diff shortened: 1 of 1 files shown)


AMZN-Gene (commented):
Tested fixes compile errors locally.

AMZN-Gene (approved):
The new cmake change works for me locally as well

AMZN-Gene:
Any maintainers able to push this thru approval?
@mkuritsu @tellypresence

mkuritsu:
Just to better describe my change, in the original commit (and comment on this PR) I had mistaked the use of the define in poly2tri. With that I reverted my previous commit and made a new one that adds P2T_STATIC_EXPORTS to the cmake compile definitions, since the issue pointed remains the same in my original comment, that is, when assimp is being built as a static lib the dll export symbols are still present in poly2tri, therefore adding the define makes those not present.

kimkulling (approved):
Looks fine!

kimkulling:
I was in vacation. So I will merge it!

kimkulling:

PR #5710, opened 2024-08-08
Fix MSVC PDBs and permit them to be disabled if required

### LINKED ISSUE #4845: Bug: Do not add `/Zi` in `CMAKE_CXX_FLAGS_RELEASE`
**Describe the bug**
Please do not add `/Zi` in the CMAKE_CXX_FLAGS_RELEASE. Why forces to create debug information when a developer may not need them.

I had to remove the option because of a sccache bug with cmake + ninja + .pdb files:
https://gitlab.kitware.com/cmake/cmake/-/issues/20222#note_1169887 

**To Reproduce**
1. Build assimp with sccache, msvc, ninja
2.
 
```
D:\a\f3d-superbuild\f3d-superbuild\build\superbuild\assimp\src\code\Common\Compression.cpp: fatal error C1041: cannot open program database 'D:\a\f3d-superbuild\f3d-superbuild\build\superbuild\assimp\build\code\CMakeFiles\assimp.dir\vc140.pdb'; if multiple CL.EXE write to the same .PDB file, please use /FS
```

**Expected behavior**
No build failure and no .pdb file

 - OS: Windows

### LINKED ISSUE #5702: MSVC: Release build PDBs are mostly empty and unusable
#4896 removed the `/Zi` switch from Release builds, but did not remove the linker PDB generation switches.
PDB files **are** still created, but they _only_ contain STL library and Windows header debugging information.

It is therefore certain that many projects using assimp are now impossible to support, despite them carefully storing their PDBs.

`/Zi` places the intermediate object file debugging information into an intermediate PDB.

As per https://github.com/ccache/ccache/issues/1040 and issue #4845, ccache/sccache currently don't support using `/Zi`.

It seems that ccache users are currently recommended to use `/Z7` to embed the debugging information into the intermediate object files.
This is not a general solution as `/Z7` is a legacy option that results in larger and less well optimised binaries.

Note that removing the symbol information makes it impossible to debug field crashes - you can't even get a stack trace on 64bit platforms without that PDB!

On Mac/Linux, debugging symbol information is usually embedded into the binary, but can be extracted into a symbol file.
On Windows it is usually stored in PDBs, but can be embedded.

Unfortunately the CMake `RelWithDebugInfo` configuration is broken as it disables optimizations by default under MSVC: https://gitlab.kitware.com/cmake/cmake/-/issues/20812

This means most people don't use it.

There are two options:

1. Revert commit 9e1de3e
2. Actually disable PDBs for the `Release` configuration, and add workarounds to the cmake bug so that `RelWithDebugInfo` produces the properly optimized binaries and complete PDBs.

In both cases, add an ASSIMP_USE_CCACHE cmake option that replaces `/Zi` with `/Z7`.
This would also become a place to add any other special things ccache might need.

My preference is option 1.
It is the safest option and is the most friendly to future developers, as the `Release` build always sounds like the correct one to ship.

I will have some time tomorrow to do this.

RichardTea:
Fix MSVC PDBs so they are no longer empty and can be used and indexed in default Release config.

Turning off the existing option ASSIMP_INSTALL_PDB will now fully disable creation of PDBs.
When disabled, a message is also printed indicating this, and that it will not be possible to debug the release binary.

Fixes #5702 
Actually fixes #4845 (the previous attempt commit did _not_ in fact disable PDB creation)

Note: Does not support s/ccache.
I don't use this so I can't test whether using /Z7 would fix that issue, and the reporter of #4845 apparently does not need to debug their binaries.


kimkulling (approved):
Looks fine!

kimkulling:

PR #5721, opened 2024-08-17
Fix potential memory leak in SceneCombiner for LWS/IRR/MD3 loader

### LINKED ISSUE #4231: Bug: Issue 39595 in oss-fuzz: assimp:assimp_fuzzer: Null-dereference READ in Assimp::LWSImporter::BuildGraph
Status: New
Owner: ----
CC: kim.k...@googlemail.com 
Labels: Restrict-View-Commit ClusterFuzz Stability-Memory-AddressSanitizer Reproducible Engine-libfuzzer OS-Linux Proj-assimp Reported-2021-10-04
Type: Bug

New issue 39595 by ClusterFuzz-External: assimp:assimp_fuzzer: Null-dereference READ in Assimp::LWSImporter::BuildGraph
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39595

Detailed Report: https://oss-fuzz.com/testcase?key=4548424335753216

Project: assimp
Fuzzing Engine: libFuzzer
Fuzz Target: assimp_fuzzer
Job Type: libfuzzer_asan_assimp
Platform Id: linux

Crash Type: Null-dereference READ
Crash Address: 0x000000000450
Crash State:
  Assimp::LWSImporter::BuildGraph
  Assimp::LWSImporter::InternReadFile
  Assimp::BaseImporter::ReadFile
  
Sanitizer: address (ASAN)

Crash Revision: https://oss-fuzz.com/revisions?job=libfuzzer_asan_assimp&revision=202104300607

Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=4548424335753216

Issue filed automatically.

See https://google.github.io/oss-fuzz/advanced-topics/reproducing for instructions to reproduce this bug locally.
When you fix this bug, please
  * mention the fix revision(s).
  * state whether the bug was a short-lived regression or an old bug in any stable releases.
  * add any other useful information.
This information can help downstream consumers.

If you need to contact the OSS-Fuzz team with a question, concern, or any other feedback, please file an issue at https://github.com/google/oss-fuzz/issues. Comments on individual Monorail issues are not monitored.

This bug is subject to a 90 day disclosure deadline. If 90 days elapse
without an upstream patch, then the bug report will automatically
become visible to the public.

TinyTinni:
Maybe fixes #4231

(fuzzer says, it is a NULL read, but when trying the file, I get a memory leak. Also, in the issue, you also tried to fix some memory leak there. Therefore, not really sure, a bit confusing).

Short explanation about the fix:

`SceneCombiner::MergeScenes` guarantees by the documentation, that the `master` parameter is cleaned up afterwards:
>  @ param master Master scene. It will be deleted afterwards. All
>   other scenes will be inserted in its node graph.

This should also include, when the destination is null (case 1. the first inserted delete).
  
For this bug, it should also clean up, if the scene is copied.

`CopySceneFlat` does not clean up the source according to the documentation:
>   @ param src Source scene - remains unmodified.

Therefore, the cleanup duty should be in `MergeScenes`.

---------------

I also checked other usage of `MergeScene` outside of LWS loader.

It is also used in
- IRRLoader: here, the `tempscene` which is used for the master parameter is also not cleaned up -> no risk of double free + also potential leak fix [link to the line](https://github.com/assimp/assimp/blob/master/code/AssetLib/Irr/IRRLoader.cpp#L1334)
- MD3Loader: here, the `master` is also not cleaned up, just in error case, which is not possible after `MergeScene` -> no risk of double free + also potential leak fix [link to the line](https://github.com/assimp/assimp/blob/master/code/AssetLib/MD3/MD3Loader.cpp#L622)

For the later 2 issues, I couldn't find any fuzzer bug issue but clang-analyzer complains there too about a potential leak.


TinyTinni:
Also deletes the sourcelist.

>   @ param src Non-empty list of scenes to be merged along with their
>   corresponding attachment points in the master scene. The function
>   deletes the input scenes afterwards. There may be duplicate scenes.

turol:
Why `set`? Why not `unordered_set`? Does iteration order matter?

Why do you check for `nullptr`? It's safe to call `delete` on a null pointer. If you really want the check you should do it while inserting to avoid a tiny bit of useless work.

TinyTinni:
> Why `set`? Why not `unordered_set`? Does iteration order matter?
> 
> Why do you check for `nullptr`? It's safe to call `delete` on a null pointer. If you really want the check you should do it while inserting to avoid a tiny bit of useless work.

Thanks for the feedback, I fixed it.

kimkulling (approved):
Looks not bad. Hard to find I guess.

kimkulling:

PR #5728, opened 2024-08-23
Fix to check both types of slashes in GetShortFilename

imdongye:
when blender export fbx then embedded texture path have slash and back slash. so GetShortFilename have to check both types of slashes


kimkulling (approved):
Looks fine.

kimkulling:

PR #5733, opened 2024-08-28
Fix copying private data when source pointer is NULL

vjf:
When copying a scene, it was checking the *destination* for a NULL pointer instead of the *source*. The copy went ahead despite the source being a NULL pointer, causing random core dumps.


kimkulling (approved):
Looks fine.

PR #5743, opened 2024-08-31
Prevents PLY from parsing duplicate defined elements

TinyTinni:
Issue: #5729

Hi, 
this is my take on trying to fix the issue.

Prevents the PLY parser from parsing elements which are already defined.

For example when the following occurs in the header:
```
element vertex 6
property ....
element vertex 8
element vertex 5
```

The crash occurs, when the duplicate vertex is bigger than the previous vertex declaration.

Since I cannot find and document which defines how the file format should behave, I choose to throw an error, preventing from loading the mesh since it is ambigious.

It is also compatible with the preivous fix for this issue.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5748, opened 2024-09-02
Bug: FBX Importer allocates zero-length arrays to aiNode::mChildren


RichardTea:
FBXConverter.cpp lines 357 to 366:

```
if (nodes.empty()) {
    parent->mNumChildren = 0;
    parent->mChildren = nullptr;
}

parent->mChildren = new aiNode *[nodes.size()]();
parent->mNumChildren = static_cast<unsigned int>(nodes.size());
for (unsigned int i = 0; i < nodes.size(); ++i) {
    parent->mChildren[i] = nodes[i].mOwnership.release();
}
```

While it is permitted by the C++ standard to allocate a zero-length array using `new`, this produces a pointer that is non-null but cannot be dereferenced. The only legal action is to `delete` it.

Minor effect: Client applications will crash if they check `mChildren != nullptr` and then dereference the array pointer, instead of checking for `mNumChildren != 0`.

Significant effect: It causes a lot of tiny allocations as every single leaf Node in an FBX file gets a zero-size allocation, consuming a lot of heap book-keeping space for no useful purpose.

tellypresence:
FBX import issues tracked via epic #6156

RichardTea:
This issue was fixed by the linked merged request

PR #5749, opened 2024-09-02
Zero-length mChildren arrays should be nullptr

RichardTea:
Fixes issue #5748

The documentation says that `aiNode::mChildren == nullptr` when `mNumChildren == 0`.

Verify this is the case in the Scene Validation post-processing step so that unit tests should spot this logic error in the future.
Found and fixed in FBX, 3DS, COB and SMD.

No other cases spotted by the current unit tests.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5449, opened 2024-02-01
Bug: broken fbx animation


tigert1998:
**Describe the bug**
The FBX animation support is broken in the latest master branch. However, animation still works in [v5.3.1](https://github.com/assimp/assimp/releases/tag/v5.3.1). So the latest several commits must have made animation broken.

**To Reproduce**
My model loader code is available at https://github.com/tigert1998/tiger-game-engine/blob/main/engine/src/model.cc.
You could build the `instancing-demo` CMake target to test.
The two models with broken animation (fbx version) are (these two both work in v5.3.1):
- https://sketchfab.com/3d-models/silver-dragonkin-mir4-89ead4e87cdc4b70840f748383f0998f
- https://sketchfab.com/3d-models/tarisland-dragon-high-poly-ecf63885166c40e2bbbcdf11cd14e65f

**Expected behavior**

**Screenshots**

**Platform (please complete the following information):**
 - OS: Windows
 - Version: latest master

**Additional context**

tellypresence:
Confirm broken animation.  
<table>
<tr><th colspan=2>Model <tt>tarisland-dragon-high-poly</tt></th></tr>
<tr><th>Animation</th><th>Screenshot</th></tr>
<tr><td><tt>Qishilong_skill06</tt></td><td><img alt="tarisland-dragon-high-poly_Qishilong_skill06_anim" src="https://github.com/assimp/assimp/assets/26958898/5a3a3db6-287b-4140-ba91-d592f1b6679e" width=240></td></tr>
</table>

JulianKnodt:
Can you double check if this is still broken? I broke the FBX animation with a change, and it was recently fixed.

tellypresence:
Built against `feb861f` (latest) but unfortunately no improvement, animations still completely broken

tellypresence:
In fact I just built against [release 5.3.1](https://github.com/assimp/assimp/archive/refs/tags/v5.3.1.zip) and get the same results.  So either there's a problem in my implementation, or the fbx animation was broken earlier than reported

tellypresence:
Here are examples of fbx files where animation wasn't working in release 5.3.1; models available from [renderpeople.com](https://renderpeople.com/free-3d-people/), under "3D Animated People" click "Free Download" button, then click the "FBX" button and download zip

<table>
<tr><th>Model</th><th>Assimp version</th><th>Screenshot</th><th>Assimp version</th><th>Screenshot</th><th>Notes</th></tr>
[...]

PR #5643, opened 2024-06-28
FBX Import - Restored Absolute Transform Calculation

lxw404:
This fixes https://github.com/assimp/assimp/issues/5642 by restoring the absolute transformation calculation which is now required due to this change: https://github.com/assimp/assimp/pull/5349.

Issue #5642, opened 2024-06-28
Bug: FBX - Skinned Mesh Imports With Wrong Transforms


lxw404:
## Describe the bug
Currently FBX models with skinned meshes will be imported with invalid transforms due to this change: https://github.com/assimp/assimp/pull/5349 which did not revert this change: https://github.com/assimp/assimp/commit/8a1dc87b4ca.

This results in the pose calculating against invalid absolute transforms.

## To Reproduce
Import a skinned FBX mesh away from the origin of the scene such as this test one:
[zan.fbx.zip](https://github.com/Yellow-Dog-Man/Resonite-Issues/files/13877779/zan.fbx.zip)

## Expected behavior
The skinned mesh should import without pose distortion.

## Screenshots

**Blender With Armature**
 Left            |  Front
:-------------------------:|:-------------------------:
![blender_InwUAfCIYr](https://github.com/assimp/assimp/assets/52231149/3471b761-d803-441a-ab36-387567211ac2) | ![blender_2R9L6Ra3pD](https://github.com/assimp/assimp/assets/52231149/f6bfe9ef-4a30-4106-8cbe-4edc8ec42e70)

**Current behavior**
 Left            |  Front
:-------------------------:|:-------------------------:
![Resonite_z5ogi1cqZN](https://github.com/assimp/assimp/assets/52231149/eed3e714-b793-4787-89b8-4fde8f0a4c82)  |  ![Resonite_0x4TgoYrX2](https://github.com/assimp/assimp/assets/52231149/fb0271dc-963b-4d5c-b620-406e10f1e71e)

**Fixed behavior:**
Left            |  Front
:-------------------------:|:-------------------------:
![Resonite_Ch2lVFIbqZ](https://github.com/assimp/assimp/assets/52231149/248da659-edf2-47c3-989a-330dad4ea7b0) | ![Resonite_2MUYL1UySv](https://github.com/assimp/assimp/assets/52231149/aa51bf73-0b59-424d-b687-979c23094316)

## Platform (please complete the following information):
 - Windows 10

lxw404:
I believe https://github.com/assimp/assimp/issues/5449 is related to this and should be fixed by: https://github.com/assimp/assimp/pull/5643.

kimkulling:
Closed by https://github.com/assimp/assimp/pull/5643

lxw404:
@kimkulling This is still an issue in current master due to: https://github.com/assimp/assimp/pull/5643 not being merged.

PR #5751, opened 2024-09-03
FBX Import - Restored Absolute Transform Calculation

lxw404:
This fixes: https://github.com/assimp/assimp/issues/5449 and https://github.com/assimp/assimp/issues/5642 by restoring the absolute transformation calculation which is now required due to this change: https://github.com/assimp/assimp/pull/5349.

The original pull request: https://github.com/assimp/assimp/pull/5643 was not properly merged so I have created this pull request again to hopefully fix this ***critical*** issue in the FBX importer.  Please let me know if there are any concerns and I will address them as soon as possible.

### DIFF
--- a/code/AssetLib/FBX/FBXConverter.cpp
+++ b/code/AssetLib/FBX/FBXConverter.cpp
@@ -312,6 +312,8 @@ void FBXConverter::ConvertNodes(uint64_t id, aiNode *parent, aiNode *root_node) 
 
                 child->mParent = last_parent;
                 last_parent = child.mNode;
+
+                new_abs_transform *= child->mTransformation;
             }
 
             // attach geometry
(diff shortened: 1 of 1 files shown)


RichardTea:
Does this issue affect static models, or only animations?

lxw404:
@RichardTea Any model which is rigged to an armature (regardless of whether it is animating) is affected by this, as shown in: https://github.com/assimp/assimp/issues/5642.  Static unrigged models are unaffected.

kimkulling:
You need to update the branch

lxw404:
@kimkulling This branch has been kept in sync every day for the past 7 days. Please let me know if you need any assistance.  (You also have write access to the repo.)

kimkulling:
Why is my button missing to update it by myself. One option would be to do this via a separate branch, but this is a lot of work.

kimkulling:
Ah, now. We are on the way.

kimkulling:

PR #5752, opened 2024-09-03
Allow usage of pugixml from a superproject

diiigle:
When using CMake subprojects with an existing build of pugixml this prevents Assimp from using its own copy.

We had symbol collisions with a large, statically-linked application that uses pugixml in several places in the source tree.


kimkulling (approved):
Looks fime for me.

PR #5770, opened 2024-09-09
Fix conditional check in `SkipSpaces` function to prevent out-of-bound access

cla7aye15I4nd:
#### Description:
This PR resolves a potential out-of-bound access issue in the `SkipSpaces` function within the `ParsingUtils.h` file. Previously, the condition that checked for whitespace (`' '` or `'\t'`) was evaluated before verifying whether the pointer `in` had reached the end. This could lead to dereferencing the pointer when it is already at the boundary, resulting in undefined behavior.

The update reorders the condition so that the boundary check `in != end` is evaluated before accessing the value of `*in`. This ensures the function operates safely when processing input up to the boundary.

#### Changes:
- Modified `SkipSpaces` function to check for `in != end` before dereferencing `*in`.

### DIFF
--- a/include/assimp/ParsingUtils.h
+++ b/include/assimp/ParsingUtils.h
@@ -103,7 +103,7 @@ AI_FORCE_INLINE bool IsSpaceOrNewLine(char_t in) {
 // ---------------------------------------------------------------------------------
 template <class char_t>
 AI_FORCE_INLINE bool SkipSpaces(const char_t *in, const char_t **out, const char_t *end) {
-    while ((*in == (char_t)' ' || *in == (char_t)'\t') && in != end) {
+    while (in != end && (*in == (char_t)' ' || *in == (char_t)'\t')) {
         ++in;
     }
     *out = in;


include/assimp/ParsingUtils.h
```diff
@@ -103,7 +103,7 @@ AI_FORCE_INLINE bool IsSpaceOrNewLine(char_t in) {
 // ---------------------------------------------------------------------------------
 template <class char_t>
 AI_FORCE_INLINE bool SkipSpaces(const char_t *in, const char_t **out, const char_t *end) {
-    while ((*in == (char_t)' ' || *in == (char_t)'\t') && in != end) {
+    while (in >= end && (*in == (char_t)' ' || *in == (char_t)'\t')) {
```
JulianKnodt, line 106: Should this be `in < end`?
This condition seems wrong

cla7aye15I4nd: Sure, have updated and add more similar changes

kimkulling:
Do you have a testcase for validation? It would be really great to test this in a proper way. Let me know if you cannot generate one.

kimkulling (approved):
Looks fine. Any kind of testcase planned?

cla7aye15I4nd:
I do not have testcases

turol:
Looks like it hit an infinite loop. It timed out after 6 hours. The timeout could be reduced to avoid wasting CPU time.

cla7aye15I4nd:
I believe we may have found a new bug. I made the following change to the code:

```c
template <class char_t>
AI_FORCE_INLINE bool SkipSpaces(const char_t *in, const char_t **out, const char_t *end) {
    if (in > end) {
        return false; // gdb stops here
    }
    while (in < end && (*in == (char_t)' ' || *in == (char_t)'\t')) {
        ++in;
    }
```

After running `/bin/unit`, I noticed that GDB stops at the line marked with a comment. I'm not entirely sure if this is the expected behavior. If it isn't, this could indicate an issue that needs to be addressed.

tellypresence:
@cla7aye15I4nd Please resolve conflicts so PR remains viable

kimkulling:
Will get closed here: https://github.com/assimp/assimp/pull/6590

PR #5775, opened 2024-09-10
FBX Blendshape: export float & same # verts

### LINKED ISSUE #5773: Bug: FBX Export Blendshape should always have same # vertices as mesh
**Describe the bug**
When exporting an FBX file, if the input and output vertex positions are the same, assimp doesn't actually write a vertex. downstream applications expect a 1-1 correspondence, and this will often break other tools. This is inside `FBXExporter.cpp`, which checks if `pDiff < 1e-8`. This condition should be removed.

**To Reproduce**
Create a blendshape which is the same as the original mesh.

**Expected behavior**
Should export correctly.

Latest assimp commit

**Additional context**
Since blendshapes are rare and they often will move most vertices, and because the exporter is also not heavily used, I suspect this was never hit before.

JulianKnodt:
Previously it was possible that the number of vertices between the mesh and morph mesh varied. This may have caused problems in tools like Blender, and this ensures there's a 1-1 correspondence between vertices.

In addition, exporting doubles broke meshlab on import, so now it exports floats.

Closes #5773


JulianKnodt:
I think the problem for exporting blendshapes may sometimes arise if there are multiple meshes for a single node, which I think is not handled in many places in the current implementation. For now I'll just switch this to a draft, since I don't think this PR fixes everything.

kimkulling:

Issue #5772, opened 2024-09-09
Bug: FBX Blendshapes should not require normals


JulianKnodt:
**Describe the bug**
FBX Blendshapes currently require normals to be fully parsed. I've encountered some blendshapes which did not come with normal features. They should be able to be safely ignored. This requires modifying FBXMeshGeometry's ShapeGeometry constructor, and the callsites in FBXConverter.cpp to only access the normals if they are not empty.

**To Reproduce**
Sorry, can't make the mesh I'm testing on public.

**Expected behavior**
The blendshape mesh without normals should be successfully loaded.

**Platform (please complete the following information):**
 Should occur on all platforms.

**Additional context**
This is about a 20 line change, feel free to implement it, otherwise I'll get to it at some point.

PR #5776, opened 2024-09-10
FBX Blendshapes: Do not require normals

JulianKnodt:
Some blendshapes don't have normal transformations, but Assimp currently requires that normals be present. This removes that, and won't populate the normal field.

Fixes: #5772


kimkulling (approved):
Looks fine.

kimkulling:

PR #5801, opened 2024-10-07
Fix a bug in the assbin loader that reads uninitialized memory

qingyouzhao:
The loader was not checking the return value of the Read() function, which could cause it to read past the end of the file. Reading pass the end of allocated memory could lead to undefined behavior.

This PR fixes the bug by checking the return value of the Read() function and returning false if the function does not read the expected 19 bytes for `"ASSIMP.binary-dump."`


code/AssetLib/Assbin/AssbinLoader.cpp
```diff
@@ -91,7 +91,12 @@ bool AssbinImporter::CanRead(const std::string &pFile, IOSystem *pIOHandler, boo
     }
 
     char s[32];
-    in->Read(s, sizeof(char), 32);
+    const size_t read = in->Read(s, sizeof(char), 32);
+
+    pIOHandler->Close(in);
```
feuerste, line 96: Why do we close the handler here and below?

code/AssetLib/Assbin/AssbinLoader.cpp
```diff
@@ -91,7 +91,12 @@ bool AssbinImporter::CanRead(const std::string &pFile, IOSystem *pIOHandler, boo
...
+    if (read < 19) {
+      return false;
+    }
 
     pIOHandler->Close(in);
```
feuerste, line 102: How about doing this instead?
```
    char s[32];
    const size_t read = in->Read(s, sizeof(char), 32);

    pIOHandler->Close(in);

    const char *id = "ASSIMP.binary-dump.";
    if (read < strlen(id)) {
      return false;
    }
    return strncmp(s, id, strlen(id)) == 0;
```

feuerste (changes requested):
Thanks for your fix Qingyou!

code/AssetLib/Assbin/AssbinLoader.cpp
qingyouzhao: Good catch. I think we should close here before returning.

code/AssetLib/Assbin/AssbinLoader.cpp
qingyouzhao: I am not sure about the string style of Assimp so I would keep the way `strncmp(s, "ASSIMP.binary-dump.", 19) == 0` as is instead of creating a variable like `const char *id = "ASSIMP.binary-dump.";`

feuerste (approved):

code/AssetLib/Assbin/AssbinLoader.cpp
kimkulling: Good catch. I will merge.

kimkulling (approved):
Looks fine.

code/AssetLib/Assbin/AssbinLoader.cpp
kimkulling: I guess the best way is just using a scoped pointer for the filestream and let it get closed automatically. I will solve this afterward when this one is merged.

kimkulling:

PR #5805, opened 2024-10-08
+Add vertex duplication during face normal generation

diiigle:
`aiProcess_GenNormals` PostProcess now duplicates vertices if they are referenced by more than one polygon which results in the correct faceted appearance. I tried to keep an eye on performance, but haven't done any profiling.

> /** Generates normals for all faces of all meshes.
     *
     * This is ignored if normals are already there at the time this flag
     * is evaluated. Model importers try to load them from the source file, so
     * they're usually already there. Face normals are shared between all points
     * of a single face, so a single point can have multiple normals, which
     * forces the library to duplicate vertices in some cases.
     * #aiProcess_JoinIdenticalVertices is *senseless* then.
     *
     * This flag may not be specified together with #aiProcess_GenSmoothNormals.
     */

According to the documentation, this should have already been the case. Maybe it got removed over the years, or was never implemented in the first place. 

Alternatively, I have considered:

1. Guard it by an optional `aiProcess_GenNormals_DuplicateVertices` flag
2. Implement it as a separate PostProcess such that it can be used in other contexts

Let me know if you prefer either one of the alternative implementations.


kimkulling:

stillhated:
there is a leak here. have been unsuccessful finding it.

PR #5807, opened 2024-10-10
Work In Progress: USD Referencing

AMZN-Gene:
Passing USD prims along with render nodes to search for USD references. 
Saving reference path inside node->metaData['ref']

_Looking for feedback on starting to implementing USD references... not sure if this is even the correct approach.
Instead of loading reference geometry directly into the scene, I want the end user to decide what to do with the reference. 
In my case, O3DE will load each usd separately, create a procedural prefab for each, and then create a hierarchy of nested prefabs for USDs which reference other USDs._

### DIFF
--- a/code/AssetLib/Step/STEPFile.h
+++ b/code/AssetLib/Step/STEPFile.h
@@ -920,6 +920,10 @@ private:
 #pragma warning(pop)
 #endif // _MSC_VER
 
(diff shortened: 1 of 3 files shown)


tellypresence:
@AMZN-Gene Please resolve conflicts so PR remains viable

PR #5810, opened 2024-10-11
Store current exception when caught in ASSIMP_CATCH_GLOBAL_EXCEPTIONS

mischmit:
In all other instances where we set mErrorString inside a catch block we also set mException. I think that this was an oversight.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5826, opened 2024-10-16
Fixed bug in DefaultLogger::set

chefrolle695:
assigning s_pNullLogger to the parameter "logger" is definitely wrong. However, custom logger previously set from user must not be deleted. The user itself must handle allocation / deallocation.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5834, opened 2024-10-22
Bug: M3D voxel-based models unsupported (attempting import fails due to zero meshes detected)


tellypresence:
(Note: .m3d format has been deprecated, but the importer still works for some models so reporting problems for future reference)

### Overview
Latest M3D project supports voxels, but our old `m3d.h` header file doesn't contain this code.  Need to surgically update our version of the header to support voxel-based models

### Models which fail to load

Error reports reflect assimp 20 Oct 2024 commit [`e4ef891`](https://github.com/assimp/assimp/commit/e4ef89182d0ae696c1014d1fcfdc860650f7a93d)

<table>
<tr><th>Category</th><th>Model</th><th>Error message</th></tr>
<tr><td rowspan=1>Geometry</td><td><a href="https://bztsrc.gitlab.io/model3d/models/notre.m3d"><tt>Notre Dame</tt></a></td><td><tt>Validation failed: aiScene::mNumMeshes is 0.  At least one mesh must be there</tt></td></tr>
</table>

### Models which load OK
Models without textures, bones or animations, tend to have more success...
<table>
<tr><th>Category</th><th>Model</th><th>Screenshot</th></tr>
<tr><td rowspan=1>Animal</td><td><a href="https://bztsrc.gitlab.io/model3d/models/dragon_vrip_res3.m3d"><tt>Stanford Dragon</tt></a></td><td><img src="https://github.com/user-attachments/assets/3cb8a8c9-87ad-4258-bbee-fcaaaee5d029" alt="m3d_dragon" width=180 /></td></tr>
</table>

tellypresence:
@bztsrc Wondering if the version of the "Notre Dame" .m3d model available for download from your website somehow became corrupted?  Most of the models indicate errors with textures or bones, rather than zero meshes  -- if you wouldn't mind please having a look when you have time it would be very helpful, thank you

bztsrc:
Dear Steve,

Assimp developers put outright errors in my C++ wrapper class code and
later they have used that as an excuse to remove M3D from Assimp. Therefore
it has been not updated for a loooooong time in Assimp. The Notre Dame
model in question is not corrupted, it's just simply doesn't have any
meshes as it contains voxels instead (you can verify this with the on-line
Model3D validator <https://bztsrc.gitlab.io/model3d/validator/> or with the
m3dconv command line tool).

Now the M3D SDK has the capability to convert voxels into triangle mesh on
load automatically and transparently (check it out in the on-line Model3D
[...]

PR #5835, opened 2024-10-22
Simplify re-enabling M3D build support

tellypresence:
### Overview
When M3D format was deprecated, `#defines` were used to disable directly in `CMakeLists.txt`.  This had the desired effect of disabling the package globally, but with the caveat that in order to enable M3D support, the assimp file needed to be manually edited or patched, since it requires `#undefining` preprocessor variables which can be difficult to get working and maintain.

By introducing CMake "options" to control the state of the `#defines`, the configuration can easily be controlled in user's wrapper Cmake files.

The default behavior is respected, i.e. M3D will only be supported by opting-in and enabling the CMake option(s).

#### Details
Changes to CMake files only; could have cleaned up some of the `#if` statements from .cpp files since they won't be built unless M3D support has been explicitly requested, but better to treat the cpp/h files as externally owned files as they were authored outside of assimp

### Usage
After this PR is merged, to enable M3D importer and/or exporter simply override the default "off" option states in your own CMake scripts e.g.
```
option(ASSIMP_BUILD_M3D_IMPORTER "" on)
option(ASSIMP_BUILD_M3D_EXPORTER "" on)
```

#### Screenshots
Once M3D support has been enabled in assimp, some .m3d models can be loaded i.e. the old code still works to some extent, with caveats as detailed in
- #5834
- #5836
- #5837

<table>
<tr><th>Model</th><th>Screenshot</th></tr>
<tr><td><a href="https://github.com/assimp/assimp/blob/master/test/models/M3D/WusonBlitz2.m3d"><tt>test/models/M3D/WusonBlitz2.m3d</tt></a></td><td>m3d_wuson</td></tr>
</table>

### DIFF
--- a/.github/workflows/inno_setup
+++ b/.github/workflows/inno_setup
@@ -45,7 +45,7 @@ jobs:
           buildDirectory: '${{ github.workspace }}/build/'
 
       - name: Compile .ISS to .EXE Installer
-        uses: Minionguyjpro/Inno-Setup-Action@v1.2.2
+        uses: Minionguyjpro/Inno-Setup-Action@v1.2.5
         with:
           path: packaging/windows-innosetup/script_x64.iss
           options: /O+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,9 +56,12 @@ if(ASSIMP_BUILD_USE_CCACHE)
   endif()
 endif()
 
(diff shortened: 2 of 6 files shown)


kimkulling (approved):
Looks fine, let's try it out.

PR #5844, opened 2024-10-25
Try to resolve image paths by replacing backslashes or forward slashes in EmbedTexturesProcess

david-campos:
I was trying to use the `EmbedTextures` processing in a project on Linux and noticed that one of the files that was loading file before for me when I had a more rudimentary load now was missing all the textures (at least, they were not embedded).
The file in question was an Wavefront OBJ that loaded the material from an adjacent MTL file.

The issue was that the file was created for windows and had backslashes `\` instead of forward slashes `/`. So assimp was failing to resolve the path.

This PR makes a change such that the EmbedTexturesProcess class, if everything else fails, will check for the alternative separator and, if the path contains any, will try to resolve again replacing the alternative by the native separator.

I assume it will not be rare for people to find files with the paths for the wrong system in them, as I hit this, so I think if everything else fails it is worth a try to check if the author meant the back or forward slashes they wrote to actually be directory separators.

I used `std::filesystem` to get the native separator. I hope this is fine.


turol:
No, we don't currrently require `filesystem`. You could just normalize all backslashes to forward slashes since Windows also supports that. There's also `getOsSeparator()` in IOSystem that returns the current OS separator.

david-campos:
Ok, will change to use that...

kimkulling (approved):
Looks fine, thanks for your work.

kimkulling:

PR #5873, opened 2024-11-11
Obj: Fix Sonarcube findings

kimkulling:
- Fix possible nullptr dereferencings
- Remove assert tests


code/AssetLib/Obj/ObjFileImporter.cpp
```diff
@@ -777,8 +779,11 @@ void ObjFileImporter::createMaterials(const ObjFile::Model *pModel, aiScene *pSc
 //  Appends this node to the parent node
 void ObjFileImporter::appendChildToParentNode(aiNode *pParent, aiNode *pChild) {
     // Checking preconditions
-    ai_assert(nullptr != pParent);
-    ai_assert(nullptr != pChild);
+    if (pParent == nullptr || pChild == nullptr) {
+        ai_assert(nullptr != pParent);
+        ai_assert(nullptr != pChild);
```
turol, line 784: This looks wrong. You checked that one of the pointers is `nullptr` and then assert that they're not.

PR #5876, opened 2024-11-13
material.h: Fixed compilation of C programs (Fixes #5875)

### LINKED ISSUE #5875: Bug: Broken C compilation aiGetMaterialFloat() and aiGetMaterialInteger() should be static inline
**Describe the bug**
The public functions `aiGetMaterialInteger()` and `aiGetMaterialFloat()` are declared as `inline` rather than `static inline`.

This breaks compilation of C programs (unresolved external symbols) when linking against the library.

**To Reproduce**
Try to build a C sample using these functions.

**Expected behavior**
Compiling C programs using assimp should not break in published releases (this made it to the Arch Linux repos).
These functions should be declared `static inline`.

 - Arch derived distros

**Additional context**
Perhaps sample C programs regression tests are needed to avoid such breakages?

Thank you!

jerstlouis:
- `getMaterialInteger()` and `getMaterialFloat()` were declared as inline while they should be `static inline`

### DIFF
--- a/include/assimp/material.h
+++ b/include/assimp/material.h
@@ -1557,7 +1557,7 @@ ASSIMP_API C_ENUM aiReturn aiGetMaterialFloatArray(
 * @return Specifies whether the key has been found. If not, the output
 *   float remains unmodified.*/
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


kimkulling (commented):
Loks fine! Sorry, I haven't seen your PR before generating my own one.

PR #5878, opened 2024-11-16
Do not create GLTF Mesh if no faces

### LINKED ISSUE #5868: Bug: Exporting GLBhas validation errors
**Describe the bug**
GLB exported from assimp is missing the ``POSITION'' key in its attribute buffer.
This manifested when I was attempting to load a mesh exported from assimp into trimesh (see this [issue](https://github.com/mikedh/trimesh/issues/2316)). This mesh loads fine in Blender, but it seems that is more due to luck, as the official GLTF validator reports multiple errors.

**To Reproduce**
Currently I don't have a public method to replicate this.

**Expected behavior**
The exported GLB should pass validation

**Screenshots**

 - Windows
 - Version: I believe latest as of 11/1/24

**Additional context**

JulianKnodt:
Previously GLTF meshes were created even if there were no faces, which lead to GLTFs which were not technically valid.

I believe this fixes the problem I was encountering, but I've not been able to test it because of some Mac compile errors. Will test on a windows machine at some point.

Fixes #5868


kimkulling (approved):
Looks fine!

kimkulling:

PR #5884, opened 2024-11-20
pbrt: Validate mesh in WriteMesh before AttributeBegin call

lijenicol:
Currently, if a mesh is ignored, there is an AttributeBegin directive that isn't closed with its respective AttributeEnd. These changes should fix that issue by moving the validation check to before the AttributeBegin write.


kimkulling (approved):
Looks fine.

PR #5888, opened 2024-11-22
[FBX] Allow export multi materials per node

JulianKnodt:
Previously it was assumed that each node would export a single material. This removes that assumption, and also allows for exporting multiple meshes with a single node. Previously they would each be seprated into their own node.

Input FBX:
Screenshot 2024-11-21 at 5 23 02 PM

Current output FBX:
image


kimkulling:

Issue #5890, opened 2024-11-23
Bug: obj-parsing fails if a comment is at the end of a line of a definition with variable number of elements to read.


scschaefer:
**Describe the bug**
The parser fails and stops parsing if a comment is at the end of a line defining an element with a variable number of possible elements following the token. For example, when defining a vertex like `v 0.0576127 0.0488792 0.2423 # End point line 1`, the function `getNumComponentsInDataDefinition` in `ObjFileParser.cpp` does not take the `#` comment token into account and continues parsing the line and incorrectly determines that there are 4 components.
Furthermore, the function `getFace` in `ObjFileParser.cpp` also does not stop parsing the current line when it comes across a `#` comment token.
Last but not least, the function `SkipLine` in `ParsingUtils.h` also doesn't skip the rest of the line if there is a `#` comment token.

**To Reproduce**
Steps to reproduce the behaviour:
1. Compile the latest master of assimp
2. Download and build the following sample code or try loading the included linien.obj file with a current version of assimp: 
[AssimpObjCommentBug.zip](https://github.com/user-attachments/files/17879218/AssimpObjCommentBug.zip)
3. Assimp will fail with the exception `DeadlyImportError("OBJ: Invalid face index.")`

**Expected behaviour**
The mentioned functions should stop parsing/skip a line once it reaches the comment character `#` and continue with the next line.

**Platform (please complete the following information):**
 - OS: Windows 10 Pro x64 - 22H2
 - Compiler: MSVC 17.12.1 latest
 - assimp-version: master (commit 6fad13c8)

kimkulling:
Closed by https://github.com/assimp/assimp/pull/5891

PR #5891, opened 2024-11-23
Fix parsing of comments at the end of lines for tokens with variable number of elements. (#5890)

scschaefer:
This pull request should fix the problem described in issue #5890 by stopping the parsing/skipping the rest of a line when a `#` comment character is found.
Everything that follows after the `#` can be assumed to be part of the comment and should thus not be parsed.

I made sure to check that all currently existing unit tests are passing.

Looking forward to your feedback. 🙂

### DIFF
--- a/code/AssetLib/Obj/ObjFileParser.cpp
+++ b/code/AssetLib/Obj/ObjFileParser.cpp
@@ -303,6 +303,9 @@ size_t ObjFileParser::getNumComponentsInDataDefinition() {
         if (!SkipSpaces(&tmp, mEnd)) {
             break;
         }
+        if (*tmp == '#') {
+            break;
+        }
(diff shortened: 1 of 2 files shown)


kimkulling (approved):
Looks fine.

kimkulling:

tyler92:
It's possible to have infinite loop here:

```
// skip blank space, lines and comments
static void NextToken(const char **car, const char *end) {
    SkipSpacesAndLineEnd(car, end);
    while (*car < end && (**car == '#' || **car == '\n' || **car == '\r')) {
        SkipLine(car, end);
        SkipSpacesAndLineEnd(car, end);
    }
}
```

if a line ends with `#\n`

Issue #4356, opened 2022-01-25
Please include precompiled binaries in releases


Toxic-Cookie:
**Is your feature request related to a problem? Please describe.**
It's frustrating having to build the binaries needed manually instead of just having them prebuilt and ready to use. Assimp64.dll specifically.

**Describe the solution you'd like**
The binaries to be prebuilt and available in releases.

**Describe alternatives you've considered**
I've tried to build Assimp64.dll myself but all I get is the following after having followed every step perfectly:
![image](https://user-images.githubusercontent.com/66691163/150909107-b53ba9c1-afc7-4f54-80a8-d665fd5f1f7f.png)

**Additional context**
Most Github repos do not force you to build the binaries from source. I'm appalled that I have to with this one.

RichardTea:
It's very clear from the output that you've omitted some steps, unfortunately there's not enough information to tell which step(s).

Which compiler are you using?

assimp supports building using multiple different toolchains and versions of those compilers.

The Windows C++ dlls are (by default) named along the lines of "assimp-vc143-mt.dll", where the middle section specifies which toolchain was used for the build (vc143 means MSVC 2022) and the "mt/mtd" suffix indicates release or debug.

The application _and_ assimp must both be compiled using compatible toolchains. If not, your application **will** crash. Also, while (eg) MSVC 2017 and MSVC 2019 are ABI compatible, you still need the newer (MSVC 2019) redists installed for that to work. If you've only got the MSVC 2017 redist, your application will crash when it tries to load assimp_v142_mt.dll.

While assimp could provide some binaries for releases, it would have to be a very limited subset. Which specific Windows toolchain do you require?

arturred:
Hi, I would be happy to get MSVC 2017 x86 and x64 binary releases. I'm tracing vcpkg and issue https://github.com/microsoft/vcpkg/pull/22993 but they also have problems building assimp.

kimkulling:
I will not be able to fulfill this request with the time which is currently available.

PR #5892, opened 2024-11-24
Add CI to automatically build and attach binaries to releases

Saalvage:
I needed this for myself, so I thought I'd contribute it as well, it's also been requested in the past, see #4356.

Will automatically build and attach the binaries when a new release is created, see [an example release](https://github.com/Saalvage/assimp/releases/tag/0.0.0.1.0.0) and [an example CI run](https://github.com/Saalvage/assimp/actions/runs/11996636350).

New configurations can easily be added as requested.

Some considerations:
- Due to the way MSVC structures its builds, the windows binary archives contain a "Release" folder in which the actual DLL is found. There's also a .pdb file that is significantly larger than the DLL itself included as well. This may not be desirable, but it makes implementation somewhat easier (honestly just saves an extra step of extracting the file we want from that folder or making the files to zip conditional).
- I tried to get ccache/sccache to work, however, it seems there's some issues with caches being loaded when a new tag is created. Potentially relevant: https://github.com/hendrikmuhs/ccache-action/issues/108. Since releases get created so rarely this is probably of little to no concern.

If there's anything that should be adjusted/improved upon please let me know!

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,50 @@
+name: Build and Publish Prebuilt Binaries
+
+on:
+  release:
+    types: [created]
+    
+jobs:
+  build:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: windows-x64
+            os: windows-latest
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine for me. Lets give it a try.

PR #5895, opened 2024-11-26
Simplify JoinVerticesProcess

JulianKnodt:
Previously JoinVertices had a lot of extra complexity, simplify it greatly.

This does change the behavior slightly. Before, vertices were merged if they were within some epsilon distance of each other. This changes it so that if vertices are _exactly_ equal they will be merged.


JulianKnodt:
I need to check that this works for some more tests, will re-enable for review later

kimkulling (approved):
Looks fine, thanks a lot for your work!

PR #5898, opened 2024-11-27
Synchronize `DefaultLogger`

### LINKED ISSUE #5896: Bug: Logstream thread-safety issue in C API
**Describe the bug**
UB when attaching a logstream while concurrently executing an operation that logs.

**To Reproduce**
MRE:
```cpp
#include <assimp/cimport.h>
#include <assimp/postprocess.h>

#include <iostream>
#include <thread>

void callback(const char *msg, char *) {
    std::cout << msg << std::endl;
}

static aiLogStream logstream;

static void thread_func() {
    std::cout << "ATTACHED " << std::endl;
    aiAttachLogStream(&logstream);

    const auto *scene = aiImportFile("MODEL_OF_YOUR_CHOICE", aiProcessPreset_TargetRealtime_MaxQuality);
    aiReleaseImport(scene);
}

int main(int argc, char *argv[]) {
    logstream.callback = &callback;

    for (int i = 0; i < 1000; i++) {
        std::jthread y(&thread_func);
        std::jthread x(&thread_func);

        x.join();
        y.join();
    }

    return 0;
}
```

**Platform:**
 - OS: Windows 10
 - Version 3267fde

**Additional context**
The issue is that while `aiAttachLogStream` is properly synchronized to prevent two streams from being attached concurrently, the actual logging happening in e.g. `DefaultLogger::WriteToStreams` is not, meaning that while thread A is attaching a logstream (modifying `m_StreamArray`) thread B may be logging something (thus iterating over `m_StreamArray`).

I see two ways in which this could be solved:
1. Move synchronization to `DefaultLogger`, this would also be the only way that the C++ API may be thread-safe.
2. Only ever attaching a single logstream from the C API and having it effectively replicate `DefaultLogger` while being synchronized. It would simply forward all messages to the callbacks registered via the C API.

Saalvage:
Fixes #5896.

It appears sensical that a singleton like this needs to be thread-safe.


kimkulling (approved):
Looks fine.

kimkulling (approved):
Looks fine.

kimkulling:

Issue #5901, opened 2024-11-28
Bug: Compile error if ASSIMP_DOUBLE_PRECISION enable


hankarun:
**Describe the bug**
When ASSIMP_DOUBLE_PRECISION is set to TRUE, compiler error happens in the material header file.

**To Reproduce**
Just enable the variable.

**Expected behavior**
It needs to compile

**Platform (please complete the following information):**
 - OS: Windos Visual Studio 2022
 - Version: master and 5.4.3

**Additional context**
There are some missing ai_real definitions in the material function definitions.
Related #5850

kimkulling:
Closed by https://github.com/assimp/assimp/pull/5902

PR #5902, opened 2024-11-28
Fix compiler error when double precision is selected,

hankarun:
This commit fixes compiler error, when double precision flag is set in cmake. Some material functions expecting float causes compiler error, float parameters changed to ai_real. #5901


kimkulling (approved):
Is fine. I guess the idea having not all of the material parameters is a failure.  We do not have enough resources to change the underlying concept.

I will merge this one.

kimkulling:

hankarun:
You are welcome. It was a simple change. As you said it is a little awkward that double precision affects material properties. I think material needs to revert from real to float for all functions and add another getPropertyDouble if there is none.

PR #5920, opened 2024-12-10
Add arm64-simulator support to iOS build script

DwayneCoussement:
Execute the build script and confirm the creation of `libassimp.xcframework`, it should include support for iOS devices, as well as simulators both on x86_64 and arm64


kimkulling (approved):
Looks fine.

kimkulling:

PR #5922, opened 2024-12-13
Replace C# port with maintained fork

Saalvage:
There have been no updates to AssimpNet in over two years, which means it is stuck on an old version of Assimp and fails to take advantage of some more modern additions to .NET (namely the built-in math types). I've created a fork of the project that updates it to the latest Assimp release, adds better interop with modern C# and includes ARM64 binaries for MacOS. I'm also looking to maintain this fork for the foreseeable future with bug fixes and updates to the latest Assimp versions.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5925, opened 2024-12-18
Fix docs for aiImportFileExWithProperties to not talk about the importer keeping the Scene alive

david-campos:
The docs talked about the importer, but no importer exists in the context of calling this function. It seems like the docs may have been accidentally copied from somewhere else.
I copied the docs from another import function within the same file. This seems reasonable because the implementations of both functions suggest they work the same way: the function creates the importer and sets it into the scene private data so that the later release call can release the importer.

Please let me know if I got it wrong.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5913, opened 2024-12-06
Bug: Memory Leak in ReadFileFromMemory Function in Assimp Library


YancyLii:
**Describe the bug**
A memory leak is observed when calling the ReadFileFromMemory function in the Assimp library. Specifically, the MemoryIOSystem object, created inside the ReadFileFromMemory function, is not properly deallocated, leading to a direct leak of 40 bytes. This happens because the MemoryIOSystem object is set as the IO handler using SetIOHandler, but it is not freed after it is replaced, causing the memory to remain allocated throughout the execution.

**To Reproduce**
[Test_assimp_leak1.tar.gz](https://github.com/user-attachments/files/18034352/Test_assimp_leak1.tar.gz)
1. Download the tar.gz file and decompress it, then (sudo) execute shell script
2. ./Test_assimp leak-ad451fcec032bb949d0dfd887b4eadc1947dfca0
3. Observe the error message

**Expected behavior**
The MemoryIOSystem object created in the ReadFileFromMemory function should be properly managed and deallocated after it is no longer needed. Specifically, it should be freed when replaced by the SetIOHandler function, ensuring no memory leak occurs.
**Screenshots**
![image](https://github.com/user-attachments/assets/d837e99f-f90e-4908-bdae-072c5996d821)

**Platform (please complete the following information):**
 - OS: ubuntu

**Additional context**
This issue is triggered when using fuzz testing with the LLVMFuzzerTestOneInput function, where a memory leak is detected by the AddressSanitizer tool.
The leak occurs due to improper handling of the memory allocated for the MemoryIOSystem object, which is not freed when the IO handler is reset.
The ReadFileFromMemory function in Importer.cpp is where the MemoryIOSystem is instantiated, and the memory leak can be traced to the lack of cleanup after SetIOHandler.

david-campos:
> Expected behavior
> The MemoryIOSystem object created in the ReadFileFromMemory function should be properly managed and deallocated after it is no longer needed. Specifically, it should be freed when replaced by the SetIOHandler function, ensuring no memory leak occurs.

I am not sure this is the core issue, `SetIOHandler` does delete the previous handler when a new one is set :thinking:

PR #5926, opened 2024-12-18
SetIOHandler delete previous handler when setting to nullptr

david-campos:
If `nullptr` is passed, a default handler is created and set, but the previous one is not released. This looks like a bug, as it makes the function inconsistent and really hard to use without leaking.

I suspect this is the cause for https://github.com/assimp/assimp/issues/5913, but I have to be able to repro it yet to confirm.

A new `ImporterImpl` is created with a `nullptr` mIOHandler` initially, the `ReadFileFromMemory` would store that `nullptr` into `io`, changes it, then sets it back.

```
    IOSystem* io = pimpl->mIOHandler;

// ...
        // prevent deletion of the previous IOHandler
        pimpl->mIOHandler = nullptr;
        SetIOHandler(new MemoryIOSystem((const uint8_t*)pBuffer,pLength,io));
// ...
        SetIOHandler(io);
```

Notice that if `io` is `nullptr` this will leak.

A deeper question I am not sure I can ask is if the behavior of creating a default IOHandler is expected: when the class is created, the `mIOHandler` is `nullptr`, but after the final call to `SetIOHandler(nullptr)` the `mIOHandler` will actually be the default one. Is this change of state expected for the `ReadFileFromMemory` function?

EDIT: please let me know if I should add the case exposed in the referenced issue in some unit test somewhere for the address sanitizer in CI to catch it(?)

### DIFF
--- a/code/Common/Importer.cpp
+++ b/code/Common/Importer.cpp
@@ -298,13 +298,15 @@ void Importer::SetIOHandler( IOSystem* pIOHandler) {
     ai_assert(nullptr != pimpl);
 
     ASSIMP_BEGIN_EXCEPTION_REGION();
+    if (pimpl->mIOHandler) {
+        delete pimpl->mIOHandler;
+    }
     // If the new handler is zero, allocate a default IO implementation.
     if (!pIOHandler) {
         // Release pointer in the possession of the caller
         pimpl->mIOHandler = new DefaultIOSystem();
         pimpl->mIsDefaultHandler = true;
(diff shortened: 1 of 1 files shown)


david-campos:
Ok, judging by the failing tests. It seems some tests expect handles set externally to never be deleted.
This makes the function inconsistent, should the function _never_ delete the previous handler instead(?)
Having it delete it when the set handler is not `nullptr` only makes this function error-prone.

PR #5927, opened 2024-12-18
AssbinImporter::ReadInternFile now closes stream before throwing

### LINKED ISSUE #5913: Bug: Memory Leak in ReadFileFromMemory Function in Assimp Library

david-campos:
While one of the throws in the function did indeed close the open stream, several didn't.

Let me know how I can add a unit test matching the fixed bug.

Fixes #5913


kimkulling (approved):
Looks fine.

kimkulling:

PR #5928, opened 2024-12-18
Prefix MTL textures with the MTL directory path

### LINKED ISSUE #2355: {OBJ} mtllib with a path == can't find texture file
I have an OBJ with materials stored in a directory, so mtllib has a path on it:

`mtllib foo_materials/foo.mtl`

In the .mtl file there are materials like this:
```
newmtl This_is_a_foo_material.jpg
Ns 7.843137
Ka 0.000000 0.000000 0.000000
Kd 0.460400 0.440400 0.480400
Ks 0.000000 0.000000 0.000000
Ni 1.300000
d 0.000000
illum 2
map_Kd foo.jpg
```

Other tools open this file fine.

With assimp, it processes everything alright, but when I want to use the materials, I have no way to get the `foo_materials/` part of the path to find the actual files.

```cpp
aiString name = material->GetName();   // gives me "This_is_a_foo_material.jpg"
      
aiString texturePath;

material->GetTexture( aiTextureType_DIFFUSE, 0, &texturePath );

// texturePath contains "foo.jpg"
```

I would have expected `texturePath` to include the full path (i.e. `foo_materials/foo.jpg`).

Am I missing something?

david-campos:
Path to textures defined in MTL files are relative to the MTL file rather than to the OBJ, so we need to prefix them with the MTL file directory path.

:construction:  **Draft PR: missing tests**

Fixes #2355

### DIFF
--- a/code/AssetLib/Obj/ObjFileMtlImporter.cpp
+++ b/code/AssetLib/Obj/ObjFileMtlImporter.cpp
@@ -44,6 +44,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "ObjFileMtlImporter.h"
 #include "ObjFileData.h"
 #include "ObjTools.h"
+#include <assimp/DefaultIOSystem.h>
 #include <assimp/ParsingUtils.h>
 #include <assimp/fast_atof.h>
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine for me.

Issue #5929, opened 2024-12-18
[Regression] Bug in PR 5802 causing CI "ubuntu-latest-g++-build-and-test" check failures


tellypresence:
PR #5802 (merged 18 Nov 2024, commit [`41cc2f6`](https://github.com/assimp/assimp/commit/41cc2f66a4c3f89b0885f3fbcc49ff981147e161)) introduced bug in line 1643 of [`glTF2Importer.cpp`](https://github.com/assimp/assimp/blame/master/code/AssetLib/glTF2/glTF2Importer.cpp)

Other PRs will now fail `ubuntu-latest-g++-build-and-test` check due to error e.g.
```
  Error: /home/runner/work/assimp/assimp/code/AssetLib/glTF2/glTF2Importer.cpp:1643:28: 
      error: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ 
          output may be truncated copying between 0 and 3 bytes from a string of length 3 [-Werror=stringop-truncation]
   1643 |                     strncpy(tex->achFormatHint, ext, len);
        |                            ^
  cc1plus: all warnings being treated as errors
```

Sample CI errors can be viewed [here](https://github.com/assimp/assimp/actions/runs/12402209985/job/34623261357?pr=5830) and [here](https://github.com/assimp/assimp/actions/runs/12402318384/job/34623593570?pr=5857#logs)

kimkulling:
https://github.com/assimp/assimp/pull/5931/

PR #5931, opened 2024-12-19
glTF importers: Avoid strncpy truncating away the ' \0' character

david-campos:
In build https://github.com/assimp/assimp/actions/runs/12403918024/job/34628244451?pr=5928 I hit some building errors because this strncpy functions were copying using the string length, which would remove the null terminator character. I just added a `+1` to ensure the null character is accounted for.


david-campos:
This fixes issue #5929

kimkulling:
Merged, thanks a lot for your conrtibution.

PR #5932, opened 2024-12-19
Support x3d importer breakage investigation

tellypresence:
X3D importer broke when migrating from IRRXML to pugixml

This PR provides latest code for both flows to allow comparison and hopefully figuring out where things went wrong

(Have modified the legacy (IRRXML-based) source files to match more closely the newer pugi-xml files, to make diff'ing more convenient)


kimkulling:
X3D works, but with some glitchs.

PR #5938, opened 2024-12-23
Disable logs for fuzzer by default

tyler92:
Using a logger in the fuzzer generates a large amount of output in the OSS-Fuzz project. For instance, [the latest run](https://oss-fuzz-build-logs.storage.googleapis.com/log-ae10e9c1-7390-4e01-a26d-6533a51a4529.txt) produced approximately 5 GB of output, which is generally not very useful This patch disables logging by default (`_DEBUG` is not set in the OSS-Fuzz environment).


kimkulling (approved):
Seems to  be ok for me.

kimkulling:

PR #5945, opened 2025-01-04
Updated Material.cpp to Add Missing Texture Types to String

crazyjackel:
In pull request #5101, user @Sanchikuuus added MayaTextures from the FBX Converter as a valid AiTextureType.

These same AiTextureTypes were not added to the material.cpp file which would result in 'Bug' string outputs when investigating Maya-exported FBXs.

Noticed whilst doing a code read-through.


kimkulling (approved):
Looks fine for me.

kimkulling:
Thanks for the monitoring. And many thanks for your contribution.

PR #5950, opened 2025-01-09
Added GLTF Extension KHR_materials_anisotropy

luho383:
This adds the Extension according to https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md

It also adds an new Texture Type and Material Key into the core Material definitions.


kimkulling:

PR #5954, opened 2025-01-10
Add customExtension support to the scene

BurntRanch:
Before this, top-level extension data defined in the top-level would not be exposed in assimp.

This exposes the top-level extension data inside of the scenes Metadata.


kimkulling (approved):
Looks fine.

kimkulling:

PR #5955, opened 2025-01-10
Change Collada "If else" chains to switches

MichaelHinrichs:


kimkulling (approved):
Looks fine, nice refactoring.

turol:
You can't `switch` on non-primitive types like strings in C++. You could use an array though.

kimkulling:
Will not work with this kind of c++ version. I will close the PR.

PR #5959, opened 2025-01-12
Optimise fetching tinyusd library

Pichas:
For now when you build Assimp with TinyUSD you cannot use options from the TinyUSD lib.
Also it compiles some souces twice and Assimp doesn't depend on TinyUSD lib file.

Here is the example of current build:
image
As you can see we have `exe` file (and Assimp as well) before the tinyUSD lib is built.

This PR fixes these issues.
image
As toy can see I have the same project but with right order of libraries and less files.
Also now I can use options like
    `option(TINYUSDZ_WITH_MODULE_USDA_WRITER "Build with USDA writer feature" OFF)`
to avoid compilation of features I don't need


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #5967, opened 2025-01-18
Fix aiTextureType enum order

### LINKED ISSUE #5966: Bug: aiTextureType_UNKNOWN should be last member of enum aiTextureType
There are several places in the code where the assumption is that  `aiTextureType_UNKNOWN` will always be the last item in the `enum`, for purposes of iterating over all texture types, e.g.
 - [`code/Pbrt/PbrtExporter.cpp`](https://github.com/assimp/assimp/blob/79d451bf4b2efbca20e9f13422e0b7399bd1b77d/code/Pbrt/PbrtExporter.cpp#L422) lines 422 and 605
```
for (int tt = 1; tt <= aiTextureType_UNKNOWN; tt++) {
                       ^^^^^^^^^^^^^^^^^^^^^
```
 - [`code/AssetLib/FBX/FBXExporter.cpp`](https://github.com/assimp/assimp/blob/79d451bf4b2efbca20e9f13422e0b7399bd1b77d/code/AssetLib/FBX/FBXExporter.cpp#L564) line 564
```
for (size_t tt = aiTextureType_DIFFUSE; tt < aiTextureType_UNKNOWN; ++tt) {
                                             ^^^^^^^^^^^^^^^^^^^^^
```
- [`port/jassimp/jassimp-native/src/jassimp.cpp`](https://github.com/assimp/assimp/blob/79d451bf4b2efbca20e9f13422e0b7399bd1b77d/port/jassimp/jassimp-native/src/jassimp.cpp#L1323) line 1323
```
for (int ttInd = aiTextureType_DIFFUSE; ttInd < aiTextureType_UNKNOWN; ttInd++)
                                                ^^^^^^^^^^^^^^^^^^^^^
```

But at some point additional texture type members started being added to `enum aiTextureType` after `aiTextureType_UNKNOWN`, so that those new texture types will be ignored by any loops using `aiTextureType_UNKNOWN` as the upper range limit.

To fix, move `aiTextureType_UNKNOWN` back to the end of the `enum` and modify `AI_TEXTURE_TYPE_MAX`

tellypresence:
Fixes #5966 

Reorder so `aiTextureType_UNKNOWN` is last item in `aiTextureType` enum; this will fix code which uses `aiTextureType_UNKNOWN` as the upper range limit when iterating over texture types

### DIFF
--- a/include/assimp/material.h
+++ b/include/assimp/material.h
@@ -339,7 +339,7 @@ enum aiTextureType {
 #endif
 };
 
-#define AI_TEXTURE_TYPE_MAX aiTextureType_MAYA_SPECULAR_ROUGHNESS
+#define AI_TEXTURE_TYPE_MAX aiTextureType_UNKNOWN
 
 // -------------------------------------------------------------------------------
 /**
--- a/include/assimp/material.h
+++ b/include/assimp/material.h
@@ -292,14 +292,6 @@ enum aiTextureType {
     aiTextureType_DIFFUSE_ROUGHNESS = 16,
     aiTextureType_AMBIENT_OCCLUSION = 17,
 
-    /** Unknown texture
-     *
-     *  A texture reference that does not match any of the definitions
-     *  above is considered to be 'unknown'. It is still imported,
-     *  but is excluded from any further post-processing.
-    */
-    aiTextureType_UNKNOWN = 18,
-
     /** PBR Material Modifiers
     * Some modern renderers have further PBR modifiers that may be overlaid
     * on top of the 'base' PBR materials for additional realism.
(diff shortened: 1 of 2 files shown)


turol:
This is an ABI change. You need to bump major version.

tellypresence:
> This is an ABI change. You need to bump major version.

Bumped

tellypresence (commented):
This PR could be simplified by leaving order alone and setting new var `AI_TEXTURE_SHADING_MODE_MAX` to the highest index, then replacing usages where `aiTextureType_UNKNOWN` is treated as the max index...would be more robust as more developers add new texture types

PR #5972, opened 2025-01-22
fix: reduce gltf2 export time

1323236654:
when i export a GLTF file from other formats, I find that it takes me a long time to export GLTF. I had a model with over 500,000 nodes and it took over 12 hours to export the model.
   I found a for loop in FindUniqueID () with an index starting at 0. If I have more than 500,000 nodes, this for loop will run 1+2+3+4+.... Half a million timesso. So I changed this function to record the current index with a special key(the first id string). Now I exported the same model in just 8 minutes!!


kimkulling (approved):
Looks fine.

code/AssetLib/glTF2/glTF2Asset.h
```diff
@@ -1267,6 +1267,7 @@ class Asset {
     size_t mBodyOffset;
     size_t mBodyLength;
     IdMap mUsedIds;
+    std::map<std::string, int, std::less<>> mUsedNamesMap;
```
turol, line 1270: Why not `std::unordered_map`?

code/AssetLib/glTF2/glTF2Asset.inl
```diff
@@ -2205,17 +2207,13 @@ inline std::string Asset::FindUniqueID(const std::string &str, const char *suffi
...
-    buffer.resize(id.size() + 16);
-    int offset = ai_snprintf(buffer.data(), buffer.size(), "%s_", id.c_str());
-    for (int i = 0; it != mUsedIds.end(); ++i) {
-        ai_snprintf(buffer.data() + offset, buffer.size() - offset, "%d", i);
-        id = buffer.data();
-        it = mUsedIds.find(id);
-    }
+    auto key = id;
+    id += "_" + std::to_string(mUsedNamesMap[key]);
+    mUsedNamesMap[key] = mUsedNamesMap[key] + 1;
```
turol, line 2216: This does two lookups. Use iterator instead.

kimkulling: Good point

kimkulling:

PR #6001, opened 2025-02-13
Fixing PyAssimp misalignment errors with certain structures

fishguy6564:
QuatKey and Bone structures are missing certain fields, thus causing corruptions such as grabbing rotation frame data for animations. I have also included the "aiProcess_PopulateArmatureData" bit flag as it was missing from the current PyAssimp distribution.


kimkulling (approved):
Looks fine.

kimkulling:

Issue #5908, opened 2024-12-05
Clang 20 - Emscripten 3.1.73 - Assimp compile error: 'memcpy' is a pointer to non-trivially copyable type 'aiFace'


danoli3:
Emscripten with Clang 20 (latest 3.1.73) does not compile.

Note Clang 19 does work (3.1.63 or before).

```
assimp/code/Common/SceneCombiner.cpp:987:14: error: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'aiFace' [-Werror,-Wnontrivial-memaccess] ::memcpy(dest, old, sizeof(Type) * num);

assimp/code/Common/SceneCombiner.cpp:987:14: error: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'aiMeshMorphKey' [-Werror,-Wnontrivial-memaccess]
  987 |     ::memcpy(dest, old, sizeof(Type) * num);
      |              ^
```

**To Reproduce**
1.  Checkout Emscripten 3.1.73
2. Build Assimp via CMake
2. See error

**Expected behavior**
Build success 

**Screenshots**

**Platform (please complete the following information):**
 - OS: Emscripten WASM
 - Version 3.1.73

**Additional context**
```
[...]

PR #6029, opened 2025-02-26
Use std::copy to copy array and remove user destructor to make sure is_trivially_copyable in order to avoid -Wno-error=nontrivial-memcall

cielavenir:
Addresses https://github.com/assimp/assimp/issues/5908. Tested locally using 20.1.0 RC.

@danoli3 

----

@kimkulling The error message suggests casting to void, but I really don't know it is correct fix.

### DIFF
--- a/code/Common/SceneCombiner.cpp
+++ b/code/Common/SceneCombiner.cpp
@@ -1095,10 +1095,6 @@ void SceneCombiner::Copy(aiMesh **_dest, const aiMesh *src) {
 
     // make a deep copy of all faces
     GetArrayCopy(dest->mFaces, dest->mNumFaces);
-    for (unsigned int i = 0; i < dest->mNumFaces; ++i) {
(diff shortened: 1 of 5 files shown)


cielavenir:
@turol 

- SceneCombiner.cpp now uses std::copy
- contrib/Open3DGC removes user destructor to make sure is_trivially_copyable

No weird cast anymore.

kimkulling (approved):
Looks fine for me.

cielavenir:
unfortunately I'm not sure how to remove this address-sanitizer issue (leak)

cielavenir:
wow it works magically, https://github.com/cielavenir/assimp/actions/runs/13623600242

cielavenir:
oh actually the former part was a duplicate of https://github.com/assimp/assimp/pull/5987 but 5987 misses 063158a110d22027141d3244162aaa8a276584fd, so this is still useful

PR #6062, opened 2025-03-23
Fix Cinema4D Import

krishty:
#5623 broke the Cinema4D import by forgetting to rename a constant. This commit fixes the invalid name.


kimkulling (approved):
Looks exellent! :-)

PR #6065, opened 2025-03-23
Harmonize Importer #includes

krishty:
Some importers referred to their headers via `#include "AssetLib/[format]/[header].h"`, others via `#include "[header].h"`, others mixed both (e.g. IRR). This is a matter of taste, but it should at least be done the same way everywhere.

Most importers seem to prefer `#include "[header].h"`, so this commit enforces it.

Cross-referencing files from other importers is still done through `AssetLib/[format]`.

Assimp headers are not affected and I would advise against changing them, as this could lead to subtle collisions with user includes.


kimkulling (approved):
Looks fine.

kimkulling:
Merged, thanks for your contributions.

PR #6074, opened 2025-03-29
Add std::string_view overloads for exporting and importing

karjalanp11rakka:
This pull request adds overloads with `std::string_view` parameters instead of `std::string` for the export and import functions' paths. This improves flexibility for users who use `std::string_view` for their paths.


karjalanp11rakka:
Forgot to add definitions to to

turol:
This also looks wrong. The actual `ReadFile` expects the string to be NUL-terminated. This is not necessarily the case for `string_view`.

PR #6078, opened 2025-03-31
Updated Inner Cone formula for Spot Lights in GLTF

crazyjackel:
Noticed an easy TODO whilst going through the code.

Updated the Inner Cone formula such that exponent values of 0 result in no inner cone and large exponent values result in a inner cone being most of the cone angle-wise.

GLTF 1.0 spec for KHR_materials_common is rather lack-luster in detailing falloffExponent. https://github.com/KhronosGroup/glTF/tree/main/extensions/1.0/Khronos/KHR_materials_common


turol:
Can `l.falloffExponent` be -1.0? If so, this will end up dividing by zero.

crazyjackel:
> Can `l.falloffExponent` be -1.0? If so, this will end up dividing by zero.

The spec and the schema lists it as a number. So spec-wise, it can be -1.0, practically, I have no clue what this would actually mean. The spec is very undetailed on this field, but gltf 1.0 is mostly deprecated anyway, just need to provide a way to convert to gltf 2.0 or the rare case wherein you want to use older unmaintained software with newer models.

I think using absolute values should be fine for the algorithm.

`ail->mAngleInnerCone = l.falloffAngle * (1.0f - 1.0f / (1.0f + std::abs(l.falloffExponent)));`

kimkulling:

Issue #5701, opened 2024-08-06
Bug: Inconsistent Material API when using DOUBLE_PRECISION


larioteo:
**Describe the bug**
This bug was initially discovered two years ago, and now I had the luck to discover it again. Either way, I have more details on the bug, so I thought, let's start a new fresh issue for it.

I added below the code lines that I tracked down. As you see in `material.h`, there is `float*` specified as argument, while in the `material.inl` and `MaterialSystem.cpp`, `ai_real*` is specified as argument. So the project is not able to compile under double precision.

Also, this should be more of an issue if you look at the definition of `aiColor4D`:
![image](https://github.com/user-attachments/assets/98f2095c-38aa-4a07-9ec1-de52544716a4)
which is always a float. So the API is rather broken in my opinion. I first replaced the `float*` to `ai_real*`, which corrupted the stack..., now I know why.

Simple replacing it with `float*`, doesn't work that easily. For that, I would need a lot more time to understand everything, which is not feasible, sorry.

Link to similar issue: https://github.com/assimp/assimp/issues/4685
(There was also a fix attached below, which never got merged, I don't know if it is the right solution to this problem.)

**To Reproduce**
Simply build with ASSIMP_DOUBLE_PRECISION enabled using VS2022 17.11 Preview 6. I am using a custom premake5 build for assimp, but it's clear that the fault is in the code, not the build itself.

**Expected behavior**
It should compile if you enable double precision.

**Platform (please complete the following information):**
 - OS: Windows 11
 - Version: 23H2

**Additional context**
Affected code:
- aiGetMaterialFloatArray (material.h): https://github.com/assimp/assimp/blob/a37d748c172a294004b99461f8c3febff68b1426/include/assimp/material.h#L1525C1-L1525C52
- aiGetMaterialFloat (material.h): https://github.com/assimp/assimp/blob/a37d748c172a294004b99461f8c3febff68b1426/include/assimp/material.h#L1552
- aiMaterial::Get() (material.inl): https://github.com/assimp/assimp/blob/a37d748c172a294004b99461f8c3febff68b1426/include/assimp/material.inl#L181
- aiGetMaterialFloatArray (MaterialSystem.cpp): https://github.com/assimp/assimp/blob/a37d748c172a294004b99461f8c3febff68b1426/code/Material/MaterialSystem.cpp#L89

sagaceilo:
[...]

PR #6088, opened 2025-04-10
Fix the function aiGetMaterialColor when the flag ASSIMP_DOUBLE_PRECISION is enabled

Riv1s-sSsA01:
# fixed problem
The function `aiGetMaterialColor` has been failed when the flag `ASSIMP_DOUBLE_PRECISION` is enabled.

# the problem cause
The floating-point type macro `ai_real` indicates `double` when the flag is enabled.
On the other hand, the members' type of `aiColor4D` is always `float`.

# related issue
#5701


Riv1s-sSsA01:
I resubmitted for my another branch.

PR #6098, opened 2025-04-14
fix: missing OS separator in outfile

Latios96:
The outfile was missing an OS separator when a PBRT file was exported in a path that was not the current directory. Writing to /foo/bar.pbrt was writing to foobar.pbrt instead. mPath and mFile are now handled like it's described in PbrtExporter.h


kimkulling:
Merged, thanks a lot for yor contribution.

PR #6099, opened 2025-04-14
fix: write textures to folder of PBRT file

Latios96:
Previously, the textures folder was created in the current directory and not in the directory next to the PBRT file


kimkulling:
After applying this PR the unittests are broken. Would be great if you could have a look onto that issue.

kimkulling:
Thanks for all your hard work. The merge is on its way!

kimkulling:

PR #6104, opened 2025-04-15
Use unique pointer to fix possible leak

kimkulling:
- closes https://github.com/assimp/assimp/issues/6053


turol:
This is not entirely the same. The old code called `m_pIO->Close` but the new one will use `operator delete`. I think there's an open bug report that we should do more of closing instead of deleting.
It should be possible to add a deleter to `unique_ptr` and if that doesn't work then add our own wrapper type that also takes the `m_pIO` pointer and calls `Close`.

Issue #6114, opened 2025-04-18
FBX Importer: AI_CONFIG_IMPORT_FBX_PRESERVE_PIVOTS=false, Wrong Animation


Nor-s:
**Describe the bug**

import FBX created with a dummy node name.
- when not simple animation

However, `AI_CONFIG_IMPORT_FBX_PRESERVE_PIVOTS=false` setting 
prevents the creation of dummy node names.

As a result, animations and characters are not mapped properly, which leads to problems.

**To Reproduce**
1. mixamo animation download 
2. `Assimp::Importer importer; importer.SetPropertyBool(AI_CONFIG_IMPORT_FBX_PRESERVE_PIVOTS, false);`
3. import mixamo animation
4. see below image (test program:https://github.com/Nor-s/Anim)
    - anim channel name contains AssimpFBX string
    - But model is no, so animation is wrong

![Image](https://github.com/user-attachments/assets/7c50116e-f1bf-4f94-a6ad-205f883f81c1)

![Image](https://github.com/user-attachments/assets/05b1247d-4725-424c-877d-fbb6838e2a18)

**Expected behavior**

No dummy node in animation
and always create simple anim node When `AI_CONFIG_IMPORT_FBX_PRESERVE_PIVOTS=false`

**Platform (please complete the following information):**
 - Windows11
 - OpenGL

tellypresence:
Please attach the `.fbx` 3d model file for investigation

Nor-s:
> Please attach the `.fbx` 3d model file (or link) for investigation

https://github.com/Nor-s/assimp-export-test/blob/main/assets/Capoeira%20(1).fbx

I used this file.

Nor-s:
It seems to be the same problem.

https://github.com/assimp/assimp/issues/5909#issue-2720970516

volcoma:
Same problem here.
https://github.com/assimp/assimp/issues/4005

There seems to be a pull request to fix this. 
https://github.com/assimp/assimp/pull/6115

Since there are so many issues related to this. Please take a look at the PR.

tellypresence:
FBX import issues tracked via epic #6156

kimkulling:
Closed by https://github.com/assimp/assimp/pull/6115

Issue #4005, opened 2021-07-28
When improt Fbx files,set AI_CONFIG_IMPORT_FBX_PERSERVE_PIVOTS to false,the animations will add many nodes.


JohnVlBetter:
When improt Fbx files,set AI_CONFIG_IMPORT_FBX_PERSERVE_PIVOTS to false,the animations will add many nodes.such as Group1_$AssimpFbx$_Rotation,Group1_$AssimpFbx$_Translation....... I don't understand why one node's transformation will be split into multiple nodes?

tellypresence:
@JohnVlBetter Did the above workaround fix your problem?

volcoma:
@tellypresence this is just a workaround until a proper fix is made inside assimp. There are several bugs reported about this, and also pull request with a proper fix.

tellypresence:
@volcoma Thank you for following up, I think you are referring to issue #6114 and PR #6115, will leave this open

PR #6115, opened 2025-04-18
Fix: Add "preservePivots" condition when importing FBX animation

Nor-s:
### Abstract

-  #6114

The animation importer part has no handling for the "preservePivots" setting.

This setting prevents dummy nodes when processing nodes in the model.

However, animations are not, so if you use this option, the animation will not work with the model.

See comments below.

https://github.com/assimp/assimp/blob/ac5988422a7f1898b25c13fb5fb068c646a4c544/code/AssetLib/FBX/FBXConverter.cpp#L2957C1-L2984C73

---

https://github.com/assimp/assimp/blob/ac5988422a7f1898b25c13fb5fb068c646a4c544/code/AssetLib/FBX/FBXConverter.cpp#L747C49-L747C51

`GenerateTransformationNodeChain`
- Processing in this function

https://github.com/assimp/assimp/blob/ac5988422a7f1898b25c13fb5fb068c646a4c544/code/AssetLib/FBX/FBXConverter.cpp#L870C4-L875C1

*Note: This function creates dummy node names.
- `NameTransformationChainNode`

### DIFF
--- a/code/AssetLib/FBX/FBXConverter.cpp
+++ b/code/AssetLib/FBX/FBXConverter.cpp
@@ -2958,7 +2958,7 @@ void FBXConverter::GenerateNodeAnimations(std::vector<aiNodeAnim *> &node_anims,
     // be invoked _later_ (animations come first). If this node has only rotation,
     // scaling and translation _and_ there are no animated other components either,
     // we can use a single node and also a single node animation channel.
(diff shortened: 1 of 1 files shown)


volcoma:
This is very nice to be handled from inside assimp. Here is the same issue discussed - i have posted a workaround from outside. but it would be great if this is fixed in the library.
https://github.com/assimp/assimp/issues/4005

PR #6120, opened 2025-04-23
glTF2: Avoid lookups for generating IDs

### LINKED ISSUE #3444: glTF2 Exporter: Horrible performance with any significant number of asset ids
**Describe the bug**
glTF2 Asset::FindUniqueID() is ridiculously slow.
Export takes a whole minute instead of a fraction of a second.

**To Reproduce**
Export a model with e.g. a significant large number of meshes.

**Expected behavior**
Asset::FindUniqueID() should complete in O(log(n)) time.

This loop (https://github.com/assimp/assimp/blob/master/code/AssetLib/glTF2/glTF2Asset.inl#L1726):
```c++
    int offset = ai_snprintf(buffer.data(), buffer.size(), "%s_", id.c_str());
    for (int i = 0; it != mUsedIds.end(); ++i) {
        ai_snprintf(buffer.data() + offset, buffer.size() - offset, "%d", i);
        id = buffer.data();
        it = mUsedIds.find(id);
    }
```

loops through every single entry in the map of ids, every single time an id is added, and do a lookup as well on each and every single entry. This is a huge problem.

All that is needed as a fix is something as simple as:

```c++
    int offset = ai_snprintf(buffer.data(), buffer.size(), "%s_%d", id.c_str(), lastId++);
    id = buffer.data();

    /* This is ridiculously slow and overkill...
    for (int i = 0; it != mUsedIds.end(); ++i) {
        ai_snprintf(buffer.data() + offset, buffer.size() - offset, "%d", i);
        id = buffer.data();
        it = mUsedIds.find(id);
    }*/
```
with `int lastId;` added as a member of `class Asset`.

Also the `__GNUC___` check had an extra underscore afterwards, is that intended?

https://github.com/assimp/assimp/blob/master/code/AssetLib/glTF2/glTF2Asset.h#L88

AFAIK the proper preprocessor definition is with only two underscores:
```
#if _MSC_VER > 1500 || defined(__GNUC__)
#       define ASSIMP_GLTF_USE_UNORDERED_MULTIMAP
```

jerstlouis:
- Fixes #3444


code/AssetLib/glTF2/glTF2Asset.h
```diff
@@ -1266,9 +1266,8 @@ class Asset {
     size_t mSceneLength;
     size_t mBodyOffset;
     size_t mBodyLength;
-    IdMap mUsedIds;
-    std::map<std::string, int, std::less<>> mUsedNamesMap;
     Ref<Buffer> mBodyBuffer;
+    std::map<std::string, int, std::less<>> lastUsedID;
```
turol, line 1270: Why `map`? Why not `unordered_map`? Is iteration order necessary for some reason?

jerstlouis: >  Is iteration order necessary for some reason?

It should not be

> Why map? Why not unordered_map?

C++ is not my native language -- that's why. I code in [eC](https://ec-lang.org).

I can amend the PR to use `unordered_map` if you'd like or feel free to do it.

jerstlouis: Also because I followed the example of `mUsedNamesMap` being used previously, where an `unordered_map` would probably have been fine in that case as well.

kimkulling:
Finally! Thanks a lot for your conrtribution. Got completly out of my focus.

PR #6127, opened 2025-04-30
Refactor

kimkulling:


tellypresence:
Can a separate PR please be raised for the CMS importer change?

(Assimp has a continuing problem with regressions and avoiding mixing changes to multiple formats in a single PR would be a big help)

Issue #6081, opened 2025-03-31
[3MF] Importer broken at commit b7b6601 (4 Feb 2024), models crash on load


tellypresence:
3MF importer was broken at 4 Feb 2024 commit [`b7b6601`](https://github.com/assimp/assimp/commit/b7b6601a759fe724219d8eda93e5ed158adc045c), models crash on load

Bisect details
<table>
<tr><th>Git commit</th><th>Model</th><th>Screenshot</th><th>Notes</th></tr>
<tr><td rowspan=3>4 Feb 2024<br><a href="https://github.com/assimp/assimp/tree/b7b6601a759fe724219d8eda93e5ed158adc045c"><tt>b7b6601</tt></a></td><td><a href="https://github.com/assimp/assimp/files/8867483/3d.zip"><tt>2_3mf.3mf (in 3d.zip)</tt></a></td><td>N/A</td><td>Crash on load</td></tr>
<tr><td><a href="https://github.com/assimp/assimp/files/8867483/3d.zip"><tt>butterfly_3mf.3mf (in 3d.zip)</tt></a></td><td>N/A</td><td>Crash on load</td></tr>
<tr><td><a href="https://github.com/3MFConsortium/3mf-samples/blob/master/examples/material/multipletextures.3mf"><tt>multipletextures.3mf</tt></a></td><td>N/A</td><td>Crash on load</td></tr>
<tr><td rowspan=3>4 Feb 2024<br><a href="https://github.com/assimp/assimp/tree/bc6673b4b2884903139b78639ce81762ca710bc8"><tt>bc6673b</tt></a></td><td><a href="https://github.com/assimp/assimp/files/8867483/3d.zip"><tt>2_3mf.3mf (in 3d.zip)</tt></a></td><td><img src="https://github.com/user-attachments/assets/bc844cc5-963c-4697-99cc-f01ae199d690" alt="2_3mf.3mf" width=120 ></td><td>Geometry loaded OK (no textures is known separate issue)</td></tr>
<tr><td><a href="https://github.com/assimp/assimp/files/8867483/3d.zip"><tt>butterfly_3mf.3mf (in 3d.zip)</tt></a></td><td><img src="https://github.com/user-attachments/assets/63bd6c6a-5a03-4f22-891b-d5dbc21f2dbb" alt="butterfly_3mf.3mf" width=120 ></td><td>Geometry loaded OK (no textures is known separate issue)</td></tr>
[...]

PR #6131, opened 2025-05-04
[Broken 3D formats] Make broken 3D formats opt-in via CMake options

### LINKED ISSUE #6132: Provide CMake options to support easy opt-out of broken 3D formats
Currently 3MF and OpenGEX formats are broken, will fail to load models and may crash.

CMake options should be provided to make it simple for users to opt-out of building these formats.

(Also, documentation should be kept updated to reflect status of broken 3D formats)

tellypresence:
Closes #6132

3MF and OpenGEX are completely broken; they should not be built by default.

Developers wishing to investigate and fix simply need to turn on the corresponding CMake options in `CMakeLists.txt`
- `ASSIMP_BUILD_3MF_IMPORTER_CMAKEOPT`
- `ASSIMP_BUILD_OPENGEX_IMPORTER_CMAKEOPT`

(Also update documentation to reflect that not all formats currently work)


tellypresence:
> Just checked a couple of 3mf-files and all of them worked. What kind of problems do you mean with this format?

Please try the 3 models in
- #6081

kimkulling:
I will.

kimkulling:
Both are working on the current master. I saw one issue with the embedded textures. But it is possible to load the referenced 3mf files in the viewer with the latest master. There was an issue I fixed some months ago. I will check this tomorrow.

kimkulling:
There is already an option to disable importers / exporters.

Issue #4030, opened 2021-08-13
Bug: Fail loading .glb file with member "extras"


lekhachoa:
I wrote a simple method using provided function from docs to read gltf/glb file. Most of time it works very well. Sometimes I meet error while loading files

**Describe the bug**

Error opening .glb file with assimp (5.0.1). Shows error:

> Member "extra" was not of type "object" when reading meshes[0]

**To Reproduce**

Sample files for testing:
https://drive.google.com/drive/folders/1msySjZgHt-rZrD9-RwLdEmwYWaryQJqy?usp=sharing

**Platform**
 - OS: Ubuntu 20.04.2

What should I do to fix this problem?

stephengold:
I verified that this is still an issue as of c44e3427 .
Compare with #5253, which appears to be a duplicate of this issue.

tellypresence:
@lekhachoa Is this still a problem?

stephengold:
It's still a problem for *me* as of 2025-04-30, Git hash 275eca95:

```console
$ ./assimpd info issue5253.gltf 
Launching asset import ...           OK
Validating postprocessing flags ...  OK
50 %ERROR: Failed to load file: Member "extras" was not of type "object" when reading nodes[0] (nodes[0])
assimp info: Unable to load input file issue5253.gltf
$ 
```

Has any attempt been made to solve this issue?

kimkulling:
Yes, PR is on its way.

PR #6133, opened 2025-05-05
Replace exception by error in log

kimkulling:
- closes https://github.com/assimp/assimp/issues/4030


stephengold:
Thank you, @kimkulling !

PR #6134, opened 2025-05-05
Fix a out of bound buffer access in ParsingUtils GetNextLine

qingyouzhao:
In the while loop, when `!IsLineEnd(*buffer)` is true, eventually `*_out++=*buffer++;` writes the last byte before `end` and sets `_out` to `end`. In the current implementation if end is `_out + BufferSize`, the next line `*out = (char_t)'\0';` writes out of bounds of `out[BufferSize]`. This change prevents this out of bounds error by setting `end `one byte before to preserve a byte for writing the trailing `'\0'` byte.


kimkulling:
Very good catch, thanks a lot for the PR!

kimkulling (approved):
Looks fine for me.

PR #6135, opened 2025-05-05
Fix a bug where string erases throws out of range

qingyouzhao:
Currently when there isn't any `' '` character inside `pOut->szName`, pos will be `std::string::npos` according to [`find_last_of` documentation](https://en.cppreference.com/w/cpp/string/basic_string/find_last_of). Calling `erase` with  `std::string::npos` throws according to [`erase` documentation](https://en.cppreference.com/w/cpp/string/basic_string/erase).

This PR adds a check before calling erase.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6139, opened 2025-05-08
+ Only the recognition of KTX2 compressed images as texture objects within the glb model is currently handled.

copycd:
Currently, when importing a glb file with a KTX2 image, the texture object is not recognized.


kimkulling (approved):
Looks correct from my point of view.

kimkulling:

PR #4980, opened 2023-02-25
bugfix remove duplicated data.

Jackie9527:
There are some problems when using assimp to process some fbx files exported from UE.
After analysis, it is confirmed that the problems are caused by duplicate data, which could be solved by eliminating duplicate data.

PR #6146, opened 2025-05-11
Fixed regression that scrambled FBX blendshape order

lxw404:
At some point the default data type of blendshape data in FBX data changed from `std::vector` to `std::unordered_set` which scrambles the order of blendshapes when imported/exported.

This pull request reverts the datatype back to `std::vector` to preserve the order properly.

### DIFF
--- a/code/AssetLib/FBX/FBXDeformer.cpp
+++ b/code/AssetLib/FBX/FBXDeformer.cpp
@@ -45,6 +45,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #ifndef ASSIMP_BUILD_NO_FBX_IMPORTER
 
+#include <algorithm>
+
 #include "FBXParser.h"
 #include "FBXDocument.h"
 #include "FBXMeshGeometry.h"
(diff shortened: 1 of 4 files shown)


turol:
This looks like it might lead to quadratic behavior in some cases. Does the order need to be preserved or is deterministic order sufficient? In that case `std::map` (a binary tree) should work.

lxw404:
The absolute original order should be preserved.  In 3D modeling software like Blender for example, when you are importing/exporting any model with blendshapes, the absolute blendshape order is _always_ preserved.  This allows the designer full control to organize them however they want, putting specific blendshapes for facial shapes together, alphabetizing visemes, etc.

turol:
This was changed in https://github.com/assimp/assimp/pull/4980. It claims to fix a bug with duplicate data. We should get a testcase for that and make sure it doesn't break before merging this.

kimkulling (approved):
Looks fine for me from the code perspective. But I cannot merge it without a rebase from your side. And as mentioned before we will need some testcases for it. So please provide this.

lxw404:
> This was changed in #4980. It claims to fix a bug with duplicate data. We should get a testcase for that and make sure it doesn't break before merging this.

I understand this concern, and it should already be handled by this implementation.

You can validate [here](https://github.com/assimp/assimp/pull/6146/files#diff-72e5eb2a9ba9d06fb0f637b70a4012db290fe43efad2dbed89cf911e6e92c9b1R150), [here](https://github.com/assimp/assimp/pull/6146/files#diff-72e5eb2a9ba9d06fb0f637b70a4012db290fe43efad2dbed89cf911e6e92c9b1R178), and [here](https://github.com/assimp/assimp/pull/6146/files#diff-b8c8ad13691342e341f5494588705ef1648b2e760e35b8d28448a33d7548abdaR74), that every time data is about to be added to the `vector`, a check is done to prevent duplicate data which was previously handled implicitly by an `insert` to an `unordered_set`.

Unless the original author of https://github.com/assimp/assimp/pull/4980 has an actual test case for their original issue, I'm not sure what to make of the request.

lxw404:
It has been over 3 months without a response from the original author of that issue.

Major 3D modeling software does not even allow exporting blendshapes with the same names.  The original issue was probably a problem with the Unreal Engine fbx exporter which I cannot replicate the behavior of in the most recent version.

@kimkulling do you have any suggestions on what you need to see if the above code is not convincing?

kimkulling:
Merged by https://github.com/assimp/assimp/pull/6623

PR #6148, opened 2025-05-11
Fix memory leak

Marti2203:
Fixes the following issue detected by OSS-Fuzz:
* https://issues.oss-fuzz.com/issues/389327770

This fix was generated by CodeRover-S, an LLM agent for fixing security vulnerabilities. More details can be found at:
* https://arxiv.org/pdf/2411.03346
* https://github.com/AutoCodeRoverSG/auto-code-rover


Marti2203:
Fixed in https://github.com/assimp/assimp/pull/6166

PR #6152, opened 2025-05-11
Update contrib/zip to fix data loss warning

### LINKED ISSUE #6160: MSVC warning: conversion from 'ssize_t' to 'mz_uint', possible loss of data
When building with MSVC with /W4, the following warning gets triggered

```
conversion from 'ssize_t' to 'mz_uint', possible loss of data
```
comming from `contrib/zip/src/zip.c`

**Better solution**
Make the potential data loss explicit with an explicit cast.

limdor:
Without these changes, the warning "Type conversion may result in loss of data" might be triggered because `i` is of type `ssize_t` but the type of the parameter is `mz_uint` that is an alias for `unsinged int`. This was already fixed in the original repository
see
https://github.com/kuba--/zip/blob/d7a2252a537926cbdef8512741a322a183fcfd09/src/zip.c#L481
and
https://github.com/kuba--/zip/blob/d7a2252a537926cbdef8512741a322a183fcfd09/src/zip.c#L538

Fixes #6160


kimkulling (approved):
Looks fine.

kimkulling:

PR #6153, opened 2025-05-11
Fix warning abut inexistent warning

### LINKED ISSUE #6151: Warning C4619 (#pragma warning : there is no warning number 'number') with Visual Studio 2015 and newer

In the following line the warning 4351 is diabled. However, this warning no longer exists in VS 2015 and newer

https://github.com/assimp/assimp/blob/536cd59807aa80515722ad176c9fb61b996e1355/include/assimp/defs.h#L177

same fore

https://github.com/assimp/assimp/blob/536cd59807aa80515722ad176c9fb61b996e1355/include/assimp/mesh.h#L55

**Better solution**

Only disable warning 4351 if MSVC is older than 2015

limdor:
These warnings are not existing before Visual Studio 2015

Fixes #6151

### DIFF
--- a/code/AssetLib/Blender/BlenderIntermediate.h
+++ b/code/AssetLib/Blender/BlenderIntermediate.h
@@ -116,7 +116,7 @@ namespace Blender {
         mywrap arr;
     };
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && _MSC_VER < 1900
 #   pragma warning(disable:4351)
 #endif
 
@@ -171,7 +171,7 @@ namespace Blender {
         // original file data
         const FileDatabase& db;
     };
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && _MSC_VER < 1900
 #   pragma warning(default:4351)
 #endif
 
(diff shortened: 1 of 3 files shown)


kimkulling (approved):
Looks fine.

code/AssetLib/Blender/BlenderIntermediate.h
```diff
@@ -116,7 +116,7 @@ namespace Blender {
         mywrap arr;
     };
 
-#ifdef _MSC_VER
+#if _MSC_VER < 1500
```
TinyTinni, line 119: Why 1500 here and not 1900? Possible typo?

limdor: Correct, that is a typo. It should be 1900.

limdor:
@kimkulling I had to push an update because there was an issue on the preprocessing macro that was missing the `defined(_MSC_VER)`. Can you retrigger the tests?
The problem is that if `_MSC_VER` is not defined, then the value when comparing is `0`. In places were the comparison is `_MSC_VER > x` it works even when it is not defined because it evaluates to false, but for cases where is `_MSC_VER < x` it is problematic because when `_MSC_VER ` is not defined it will evaluate to true.

PR #6158, opened 2025-05-13
Fix for CanRead function for FBXImporter

jsharam:
I am having an issue where I cannot load FBX files from memory since upgrading to 5.4.3 (from 5.2.4). Having stepped through the code I think I've traced the problem back to the "CanRead" function in the FBXImporter.cpp. It looks like the problem was introduced in this commit:

https://github.com/assimp/assimp/commit/0b0ec713f67d75bfd26317cc3e064acc518fe209

It looked like some sort of typo to me - I'm not sure why that magic token would be switched out for newline characters. If this is incorrect then I'm open to other ways of solving it. I've verified that switching back to "fbx" does indeed allow my app to function correctly again. This effectively reverts the small part of that commit to allow reading of FBX from memory.


kimkulling:

PR #6186, opened 2025-05-15
Fix issue compiling when assimp added as subdirectory

plemanski:
When assimp is added as a subdirectory to a parent project, and the /utf-8 flag is already set, the cl will throw the following error `cl : Command line error D8016 : '/utf-8' and '/source-charset:utf-8>' command-line options are incompatible`

Because the /utf-8 flag is equivalent to `/source-charset:utf-8 /execution-charset:utf-8` setting the `/source-charset:utf-8` is redundant.


kimkulling (approved):
Looks fine.

kimkulling:

Osyotr:
This change is not correct because `CMAKE_CXX_FLAGS` is of type `STRING` therefore the check should be `if(NOT CMAKE_CXX_FLAGS MATCHES "/utf-8")`.
Proposal: add `/utf-8` unconditionally.

PR #6200, opened 2025-05-26
Fix HUNTER_ERROR_PAGE

deccer:
Looks like docs.hunter.sh is outdated and redirects to some dude's personal blog now which seems totally unrelated.


deccer:
Never heard of this package manager either BTW, This thing must have gone through various peoples' repos, dead links everywhere in the official docs too.

Issue #6236, opened 2025-06-23
Bug: Memory leak when texture transforming


HandsomeXi:
**Describe the bug**
Memory leak when texture transforming
**To Reproduce**

**Expected behavior**

**Screenshots**

**Platform (please complete the following information):**

**Additional context**

HandsomeXi:
the pr:  https://github.com/assimp/assimp/pull/6237

tellypresence:
@HandsomeXi Please add "Closes #6236" to description for PR #6237 so the issue and PR are properly associated

HandsomeXi:
@tellypresence Thanks, done.

PR #6237, opened 2025-06-23
bugfix: Fixed the memory leak when texture transforming (#6236)

HandsomeXi:
Closes https://github.com/assimp/assimp/issues/6236

### DIFF
--- a/code/PostProcessing/TextureTransform.cpp
+++ b/code/PostProcessing/TextureTransform.cpp
@@ -194,7 +194,6 @@ inline static const char* MappingModeToChar(aiTextureMapMode map) {
 void TextureTransformStep::Execute( aiScene* pScene) {
     ASSIMP_LOG_DEBUG("TransformUVCoordsProcess begin");
 
-
     /*  We build a per-mesh list of texture transformations we'll need
      *  to apply. To achieve this, we iterate through all materials,
      *  find all textures and get their transformations and UV indices.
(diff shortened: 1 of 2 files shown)


tellypresence:
@HandsomeXi Please check the sonarCloud report and modify PR so the checks are all green

HandsomeXi:
@tellypresence My initial implementation was `delete[] mesh->mTextureCoords[n]`, but ci reported an error `Rewrite the code so that you no longer need this "delete"`. Later I changed it to `std::unique_ptr<aiVector3D[]> oldTextureCoords(mesh->mTextureCoords[n]);`, trying to avoid using delete, but ci still reported errors `Use "std::array" or "std::vector" instead of a C-style array. `and `Use "std::make_unique" to construct "std::unique_ptr".`. What confuses me is that the project uses a lot of new/delete operations, pointers and C-style array, is this inspection rule incompatible with the project style? How should I release this memory?

HandsomeXi:
@tellypresence This problem bothered me, as the memory leak could reach several GB when gltf models were loaded and destroyed frequently.

kimkulling:
Looks fine for me. Sonarcube always shouts about direct meory access via new / delete. But the STL uses new / delete dramatically often and makes the heap muddy. So for me ignoring deleting them would be fine.

Of course this way it is much cleaner.

kimkulling:

PR #6239, opened 2025-06-23
Clone meshlab/tinyusdz repos in default location

### LINKED ISSUE #6238: meshlab and tinyusdz clone locations mix source code and build files
Two external 3rd party repos are cloned when opting in for 3D formats:
- `tinyusdz` repo will be cloned for `.usd*` support
- `meshlab` repo will be cloned for `.wrl` (VRML) support

In theory, it makes sense to only clone the source files a single time regardless of build targets; the way CMake works, if default location is used, the source files will be cloned in a matrix of build type (e.g. debug and release) multiplied by ABI (e.g. x86, x86_64 etc).  So it might be the source files would be cloned 4 times in this example (debug/x86, debug/x86_64, release/x86, release/x86_64).  This seemed wasteful, and in order to address this situation, cmake var `FETCHCONTENT_BASE_DIR` was overridden.

While this approach might save some disk space, it also causes build files to be mixed in with source files, which breaks things when building for multiple targets.  For example, after building for android (which uses `Ninja` generator), if attempting to build for iOS (which uses `XCode` generator) will encounter breaking error e.g.
```
**** Cloning meshlab repo, git tag ad55b47a9b0700e7b427db6db287bb3a39aa31e7

CMake Error: Error: generator : Xcode
Does not match the generator used previously: Ninja
Either remove the CMakeCache.txt file and CMakeFiles directory
or choose a different binary directory.
```
This situation entailed that every time when switching back and forth between build targets, would need to manually move or delete build folders, which is error-prone, tedious and avoidable.

**Better solution**
Instead, leave `FETCHCONTENT_BASE_DIR` as default value.  This change does mean the repo will be cloned per-target, but saves a lot of time and trouble when building for multiple platforms using different generators.

tellypresence:
Closes #6238 

This will prevent conflicts when building for multiple target platforms which use different build generators (e.g. both android and iOS)

### DIFF
--- a/code/AssetLib/VRML/VrmlConverter.hpp
+++ b/code/AssetLib/VRML/VrmlConverter.hpp
@@ -45,7 +45,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 
 #if !defined(ASSIMP_BUILD_NO_VRML_IMPORTER)
-#include "contrib/meshlab/autoclone/meshlab_repo-src/src/meshlabplugins/io_x3d/vrml/Parser.h"
+#include "Parser.h"
 #endif // #if !defined(ASSIMP_BUILD_NO_VRML_IMPORTER)
 
 namespace Assimp {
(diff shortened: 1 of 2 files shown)


kimkulling:

PR #6244, opened 2025-07-01
fix: texture paths issues with fbx export

pjoe:
Was experiencing a number of issues when importing a `.glb` (with  albedo, ORM and normal map) and exporting to FBX:

- After opening FBX in Blender material was using wrong textures (some empty ones), even though the real textures was included in the file
- After opening in 3dsmax it would fail to decode textures (because their file name didn't include extension

This PR fixes this by:

- Adding filename extension if missing to texture path from format hint
- Using proper texture path when assigning to materials

### DIFF
--- a/code/AssetLib/FBX/FBXExporter.cpp
+++ b/code/AssetLib/FBX/FBXExporter.cpp
@@ -1582,6 +1582,7 @@ void FBXExporter::WriteObjects () {
         }
     }
 
+    std::map<std::string, std::string> tpath_by_image;
     // FbxVideo - stores images used by textures.
     for (const auto &it : uid_by_image) {
         FBX::Node n("Video");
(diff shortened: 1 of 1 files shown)


pjoe:
Anything more I can do to get this mergeable?

kimkulling:
Hi, sorry I was in vacation. I am not allowed to rebase your branch onto the latest master. it would be great if you could prepare this. create another fork of your branch and merge this. 

Or another option would be to

pjoe:
Okay, I think I managed to rebase the existing branch 😄

pjoe:
@kimkulling Is this rebase okay, or is there something else I should do?

kimkulling:
I will merge it here: https://github.com/assimp/assimp/pull/6405 

Sorry for all the delays.

PR #6250, opened 2025-07-08
Fix to fail in the function fast_atoreal_move when the flag ASSIMP_DOUBLE_PRESICION is enabled

Riv1s-sSsA01:
# Fixed Problem
Fail in the function `fast_atoreal_move` when the flag `ASSIMP_DOUBLE_PRESICION` is enabled.

# Failure Causes
## ai_real
The type `ai_real` is `double` when the flag `ASSIMP_DOUBLE_PRESICION` is enabled.
So, the below function call fails.
```
fast_atoreal_move<float>(hoge, (float &)ai_real_value)
```

## aiColor
The member of _`aiColornD`_ (e.g. `aiColor4D`) is `float` even when the flag `ASSIMP_DOUBLE_PRESICION` is enabled.
So, the below call fails.
```
fast_atoreal_move<ai_real>(hoge, (ai_real &)color.r)
```

# Refactoring
This PR includes refactoring.
The purpose is to avoid reccurences of the same problems.

# Related PR
#5660

### DIFF
--- a/code/AssetLib/AC/ACLoader.cpp
+++ b/code/AssetLib/AC/ACLoader.cpp
@@ -75,6 +75,8 @@ static constexpr aiImporterDesc desc = {
     "ac acc ac3d"
 };
 
+static constexpr auto ACDoubleSidedFlag = 0x20;
+
 // ------------------------------------------------------------------------------------------------
 // skip to the next token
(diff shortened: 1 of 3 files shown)


kimkulling:

PR #6251, opened 2025-07-10
fix(GLTF): in the importers, use the bufferView name as the embedded image file name if the image name is empty.

H-EAL:
Some exporters will save the embedded image name in the bufferView associated with the image and not in the image object itself.
This PR first checks if the image has a valid name otherwise it falls back to the bufferView name.

Any Sketchfab auto converted glb file will showcase the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved glTF and glTF 2.0 imports by preserving meaningful names for embedded textures when image metadata does not include a filename.
  * Embedded textures now use available buffer view names as a fallback, making imported assets easier to identify and manage.
  * This prevents otherwise unnamed embedded textures from losing useful naming information during import.


kimkulling (approved):
Looks fine and correct.

kimkulling:
It seems that there is the possiblility of a nullptr dereferencing error in your code. I have prepared a possible solution for that.

H-EAL:
> It seems that there is the possiblility of a nullptr dereferencing error in your code. I have prepared a possible solution for that.

```
        tex->mFilename = img.name;
        if (img.name.empty() && img.bufferView->name != nullptr) {
            tex->mFilename = img.bufferView->name;
        }
```
Shouldn't be the check against `img.bufferView` ? As far as I can see `name` is a `std::string`, but `bufferView` can definitely be invalid, so it should be:

```
        tex->mFilename = img.name;
        if (img.name.empty() && img.bufferView) {
            tex->mFilename = img.bufferView->name;
        }
```

kimkulling:
Sure, thanks for the hint!

H-EAL:
Hi @kimkulling, I don't know if you noticed but the build didn't pass, there's still a lingering `!= nullptr` that needs to be removed.

kimkulling:

Issue #5635, opened 2024-06-24
Bug: OBJ fails to load when the .mtl file contains "aniso" or "anisor"


TenHenEggs:
**Describe the bug**
Trying to import an OBJ file with its .mtl containing "aniso" or "anisor" fails with the message:
`Cannot parse string "niso" as a real number: does not start with digit or decimal point followed by digit.`

**To Reproduce**
Try to import [this](https://github.com/tinyobjloader/tinyobjloader/blob/release/models/pbr-mat-ext.mtl) model from [tinyobjloader](https://github.com/tinyobjloader/tinyobjloader)

**Expected behavior**
Model is successfully imported

**Platform**
 - OS: Windows
 - Version 5.4.0
 - Installed via vcpkg

kimkulling:
That is right. I have just checked the implementation and the parser is just ignoring ansio.

SharpieLaBeouf:
I am having the same issue. I have exported my wavefront model from Blender using the PBR Extensions export option for [wavefront materials](https://en.wikipedia.org/wiki/Wavefront_.obj_file#Physically-based_rendering). I have received this error when attempting to load using ASSIMP.

Assimp Scene Invalid: Cannot parse string "niso" as a real number: does not start with digit or decimal point followed by digit.

GitHubDragonFly:
@TenHenEggs and @SharpieLaBeouf maybe also check the issue https://github.com/assimp/assimp/issues/5554

It might show you possible ways of fixing OBJ + MTL issues.

tellypresence:
@TenHenEggs Is this still a problem?

TenHenEggs:
@tellypresence 
On release 5.4.3 bug still happens.
I also managed to compile master ac59884, and it doesn't solve the issue either.

turol:
Link points to .mtl and not the .obj. Can you link to both for ease of downloading?

TenHenEggs:
Sure,
[pbr-mat-ext.mtl](https://github.com/tinyobjloader/tinyobjloader/blob/release/models/pbr-mat-ext.mtl)
[pbr-mat-ext.obj](https://github.com/tinyobjloader/tinyobjloader/blob/release/models/pbr-mat-ext.obj)

Zip file with both files:
[pbr-mat-ext.zip](https://github.com/user-attachments/files/19822308/pbr-mat-ext.zip)

turol:
I think this is the relevant piece of code:
[...]

PR #6253, opened 2025-07-13
Add BOM skip and fix mtl parsing

paulbauriegel:
I have a obj + mtl 3D file that was failing to load in open3D and open3D is using this library. There where 2 issues I found when trying to load with this library. 

1.  In the OBJ file if the first line starts with a invisible BOM char like: \uFEFFmtllib model.mtl 

```bash
paul@mac assimp % head -n 5 ../0FABEEBE-68BE-4F36-A268-CAE2853D08A8/model.obj | hexdump -C
00000000  ef bb bf 6d 74 6c 6c 69  62 20 6d 6f 64 65 6c 2e  |...mtllib model.|
00000010  6d 74 6c 0d 0a 67 20 5f  30 0d 0a 76 20 30 2e 36  |mtl..g _0..v 0.6|
00000020  38 35 36 34 30 20 35 2e  35 36 37 38 33 30 20 2d  |85640 5.567830 -|
00000030  31 2e 33 39 30 32 37 30  0d 0a 76 20 30 2e 36 38  |1.390270..v 0.68|
00000040  35 32 37 30 20 35 2e 35  33 38 30 33 30 20 2d 31  |5270 5.538030 -1|
00000050  2e 33 36 34 34 33 30 0d  0a 76 20 30 2e 36 38 39  |.364430..v 0.689|
00000060  37 37 30 20 35 2e 35 36  33 35 37 30 20 2d 31 2e  |770 5.563570 -1.|
00000070  33 39 30 32 37 30 0d 0a                           |390270..|
```

The parsing of the textures does not work because it will not properly read the line mtllib model.mtl to to the leading characters. So in ObjFileParser::parseFile I added a check that skips the BOM if there.

2. Then once it was trying to load the MTL I saw a couple of error because assimp is not handling all the attributes of the materials properly. Specifically there was a collision for ao where it was expecting something else. But also other attributes it didn't handle. I didn't change the whole material definition but a made a few adjustment so that it works. Please let me know what you think.


kimkulling:

nickykitchingman:
Whilst these changes allow the import of models with these PBR attributes without error, none of them are actually included in the resulting material. This stems from all the new comparisons similar to `keyword == "aniso"`. The `keyword` string includes the following space character so these comparisons will always fail. The usual pattern in this parser seems to be to compare the two strings against the length of the attribute, e.g., `strlen("aniso")`. It does not seem like these changes were actually tested. #5635 was also closed by this branch, yet these changes do not load the attributes such as `aniso` provided in the example model referenced by the issue.

PR #6265, opened 2025-07-17
[Draft] Introduce MikkTSpace Tangent calculation

kimkulling:
## Summary by CodeRabbit

* **New Features**
  * Added an opt-in tangent post-process that generates tangents/bitangents using MikkTSpace (controlled by `AI_CONFIG_POSTPROCESS_USE_MIKKTSPACE_TANGENTS`, default off).

* **Bug Fixes**
  * Improved tangent post-process activation so it only runs when both the import requests tangent-space calculation and the feature is explicitly enabled.

* **Chores**
  * Updated the build to compile MikkTSpace, enabled Clipper/Poly2Tri source blocks, and added/organized unit tests and test build wiring for MikkTSpace tangent generation.


KeyC0de:
So I wonder what tangent calculation method was used up to now; I was clearly mistaken when I thought MikkTSpace was used already.

PR #6267, opened 2025-07-19
test/unit/utProfiler.cpp: fix `gcc-16` build failure

trofi:
Upcoming `gcc-16` improved detection of unused variables as:

    test/unit/utProfiler.cpp: In member function 'virtual void utProfiler_addRegion_success_Test::TestBody()':
    test/unit/utProfiler.cpp:71:22: error: variable 'j' set but not used [-Werror=unused-but-set-variable=]
       71 |         volatile int j=0;
          |                      ^

The change is intentional accoring to Andrew Pinski and is not a faalse positive.

### DIFF
--- a/test/unit/utProfiler.cpp
+++ b/test/unit/utProfiler.cpp
@@ -70,6 +70,7 @@ TEST_F( utProfiler, addRegion_success ) {
     for ( int i=0; i<10; i++ ) {
         volatile int j=0;
         j++;
+        (void)j; /* avoid gcc-16 warnings about unused variable */
     }
     myProfiler.EndRegion( "t1" );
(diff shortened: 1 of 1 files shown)


pinskia:
The change happened due to https://gcc.gnu.org/PR44677 (https://gcc.gnu.org/r16-2258-g0eac9cfee8cb0b21d ).  Unlike clang, gcc does not treat volatile variables here (increment) any different from non volatile variables.

jakubjelinek:
Note, both gcc and clang do warn on void foo () { volatile int j = 0; j = 2; j = 4; } so not sure why ++j; case is different.

kimkulling (approved):
Looks fine.

kimkulling:

Issue #6276, opened 2025-07-23
Bug: Memory hog when running assimp info on fbx files in s390x


pranavkaruvally:
Hi,

There was an error reported in Debian saying that while running the test suites for assimp it is taking an extremely high memory on s390x machines ([https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1107477](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1107477))
Logs: [https://ci.debian.net/packages/a/assimp/testing/s390x/61147071/#L11086](https://ci.debian.net/packages/a/assimp/testing/s390x/61147071/#L11086)

So, isolating from the test bed, this issue is produced when running `assimp info` on any `.fbx` files.
Eg: `assimp info /home/pranav/github/assimp/test/models/FBX/animation_with_skeleton.fbx`

When I tested this out personally, it was filling up 64 GB of RAM and an additional swap space of 20 GB (afterwards the process was killed by the OOM reaper).

Edit:
The link to the 3D file is [https://github.com/assimp/assimp/blob/master/test/models/FBX/animation_with_skeleton.fbx](https://github.com/assimp/assimp/blob/master/test/models/FBX/animation_with_skeleton.fbx)

tellypresence:
@pranavkaruvally Please attach 3D model file so others can reproduce and investigate, thank you

tellypresence:
FBX import issues tracked via epic #6156

PR #6277, opened 2025-07-23
Fix endiannes issues regarding floats on s390x (#6276)

pranavkaruvally:
Added some missing endianness swaps for parsing floats.
For the structure that represents the IEEE 754 standard, when bit fields are set in a structure and that structure's data is written through `memcpy`, the ordering of the members do matter and should be in reverse order for big endian machines.

### DIFF
--- a/code/AssetLib/FBX/FBXParser.cpp
+++ b/code/AssetLib/FBX/FBXParser.cpp
@@ -365,10 +365,14 @@ float ParseTokenAsFloat(const Token& t, const char*& err_out)
         }
 
         if (data[0] == 'F') {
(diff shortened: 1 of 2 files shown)


tellypresence:
@pranavkaruvally Please add "Closes #6276" in description so the issue becomes associated with this PR, thank you

pranavkaruvally:
> @pranavkaruvally Please add "Closes #6276" in description so the issue becomes associated with this PR, thank you

Hi @tellypresence 
Can you please mention what other improvement is to be made?
Thanks

kimkulling:

PR #6278, opened 2025-07-24
Restore merging of near-identical vertices for JoinIdenticalVertices

mwpenny:
Commit 12f33099964c50877ff9bc29957212e020deb6be changed the behavior of the JoinIdenticalVertices option:

* Before the change, vertices equal within an epsilon value were merged
* After the change, vertices must be exactly equal to be merged

This is a breaking change and can cause issues in programs which do not expect the additional vertices (e.g., due to floating point imprecision). In particular, I experienced it with my project's asset conversion pipeline after upgrading assimp (e.g., broken UV mapping, among other things).

This PR reinstates the epsilon behavior while preserving the overall code simplification of the breaking commit.


include/assimp/Vertex.h
```diff
@@ -183,29 +183,6 @@ struct Vertex {
         return *this;
     }
 
-    bool operator < (const Vertex & o) const {
```
mwpenny, line 186: I moved the updated comparison logic to `JoinVerticesProcess.cpp` since it is specific to that, and I wanted to avoid causing any confusion or surprises. Pre 12f33099964c50877ff9bc29957212e020deb6be assimp did not define `Vertex::operator<`.

code/PostProcessing/JoinVerticesProcess.cpp
```diff
@@ -100,6 +100,48 @@ void JoinVerticesProcess::Execute( aiScene* pScene) {
...
+
+        if ((a.normal - b.normal).SquareLength() > squareEpsilon) {
+            return (a.normal < b.normal);
+        }
+
+        // We just test the other attributes even if they're not present in the mesh.
+        // In this case they're initialized to 0 so the comparison succeeds.
+        // By this method the non-present attributes are effectively ignored in the comparison.
+
+        if ((a.tangent - b.tangent).SquareLength() > squareEpsilon) {
```
mwpenny, line 120: I added back the checks for `tangent` and `bitangent` since the pre 12f33099964c50877ff9bc29957212e020deb6be logic checked for them. Per the comment (also restored from before), these will have no effect if they are not set as there will be no difference between them.

mwpenny:
Hi there. Do you have any feedback, and are there any steps I'm missing on my side? Thanks!

mwpenny:
Friendly ping @kimkulling. Is there anything else needed before this regression fix can be merged?

I understand that maintaining a project of this size can be quite busy, so please let me know if there's anything I can do to make it easier.

For example, I fixed the problem but preserved the code simplifications of the original commit, but I could just do a revert instead if that's simpler.

Thanks for your work on this very useful project.

kimkulling:

PR #6281, opened 2025-07-24
OBJ: avoid extra scanning while reading faces in ObjFileParser.cpp

ov:
The code in `ObjFileParser::getFace()` tries to be smart about end of buffer and scans for it for every face index it reads. This takes us to something close to `O(N^2)` where `N` is the file size, so I wonder if we really need that there.

This patch replaces the whole thing with a simpler `atoi()` call that passes all my internal tests. If there is a file that fails we might consider scan ahead, but limit that scan to something like 10 symbols or so as otherwise we do a lot of unnecessary work.

## Summary by CodeRabbit

* **Refactor**
  * Simplified OBJ face-index parsing for a more direct extraction approach, reducing internal complexity while preserving existing behavior.
* **Chores**
  * Updated copyright year in source headers.


kimkulling:
Great finding, thanks a lot for your help!

PR #6282, opened 2025-07-24
OBJ.MTL: avoid possible conflict between `anisotropy` and `adobe_xxx` params

ov:
Adobe added a couple of custom parameters to OBJ MTL files. Here's an example MTL file I have:

```mtl
newmtl cone_Mat
Pr 0.503
Ni 1.600
Ka 0 0 0
Pm 0
d 1.0
Ke 0.0
Kd 0.816 0.816 0.816
Tr 0
illum 7
Ks 1 1 1
adobe_interior_color 1.0 1.0 1.0
adobe_density 0
Ns 16
adobe_translucence 0
```

Note there's a couple of `adobe_xxx` parameters there. They conflict with the standard `anisotropy` parameter in the MTL format parser, as it only checks the first letter `a` and assumes this is `anisotropy`, while it might be not...

The patch fixes that by checking one more letter for `n`. That's definitely not the best way to do that, but other things are done the same way in that parser, so maybe we can survive with that for now.

## Summary by CodeRabbit

* **Style**
  * Improved formatting and consistency in the OBJ/MTL material parsing flow, including clearer block scoping around conditional property handling.
  * Refined token-based float parsing to make “missing value” cases explicit and easier to reason about.
* **Chores**
  * Updated a buffer size constant to use compile-time `constexpr`.
  * Applied minor header whitespace/documentation cleanup.


kimkulling:

PR #6284, opened 2025-07-25
fix stream reading checks in glTF2 binary loader

ov:
This patch fixes the number of errors while checking what `Stream::Read()` returns. The calling code expects it to return the number of records read, while the method returns the number of bytes read. This effectively breaks GLB format loading, the patch fixes that.


ov:
nevermind, my bad :)

PR #6298, opened 2025-07-30
Fixed a `container-overflow` error

DonggeLiu:
Fix a `container-overflow` error in `ODDLParser::OpenDDLParser::parseIntegerLiteral` by swapping the order of conditions in a while loop to ensure the end-of-buffer check happens before dereferencing the pointer. This prevents reading past the end of the buffer when lookForNextToken returns the end pointer.

https://oss-fuzz.com/testcase-detail/4980126616780800

https://issues.oss-fuzz.com/issues/42527625


kimkulling:

PR #6307, opened 2025-08-04
Fix import of large GLB files by providing a proper helper for size_t reading

ov:
I have a huge GLB file (about 3.8 Gb) here which I can load with Blender, but cannot load with Assimp. I keep getting "Mesh XXX has no faces" error on import.

Here's the mesh in question:

```json
    {
      "name": "Minion Wingnut Posed.stl",
      "primitives": [
        {
          "attributes": {
            "NORMAL": 201,
            "POSITION": 200
          },
          "indices": 202,
          "material": 16,
          "mode": 4
        }
      ]
    },
```

So its `indices` field points to accessor 202:

```json
    {
      "bufferView": 202,
      "byteOffset": 0,
      "componentType": 5125,
      "count": 4433256,
      "max": [4433255],
      "min": [0],
      "type": "SCALAR"
    },
```

...which points to buffer view 202. Here are buffer views 201 and 202:

```json
    {
      "buffer": 0,
      "byteLength": 53199072,
      "byteOffset": 2107625916,
      "target": 34962
    },
    {
      "buffer": 0,
      "byteLength": 53199072,
      "byteOffset": 2160824988,
      "target": 34962
    },
```

Note that buffer's 201 offset is still below 2^31, while 202 crosses that, so it feels like `byteOffset` is somehow parsed as `int` instead of `unsigned int`.

I traced it down to `ReadHelper` template in `glTFCommon.h` which is overloaded for `uint64_t`, but not for `size_t`.  The non-overloaded version is used instead leading to `val.GetInt()` being used which breaks `byteOffset` and fails the GLB loading.

The patch fixes that for the model I have. Obviously I cannot share it due to its size and it also contains the customer's data, but you can easily verify that the wrong `ReadHelper` is used for `byteLength` parsing.


ov:
**update:** it appears that `uint64_t != size_t` only on Mac, so I added a simple `ifdef` around the specialization, effectively making the patch Mac-only.

It might help to actually check if `size_t == uint64_t` or not and specialize in that case, but for the life of me I cannot make such `enable_if/is_same` magic. Another possible option is to replace `size_t(0)` parameters of `ReadMember()` with `uint64_t(0)` to avoid any ambiguity for the compiler regardless of the platform.

tellypresence:
If this is supposed to be associated with issue 6310, please add "Closes #6310" to description of PR above

kimkulling:
closes https://github.com/assimp/assimp/issues/6310

kimkulling:

PR #6309, opened 2025-08-04
Clean msvc versions into code/CMakeLists.txt

mosfet80:
Removed old msvc versions


kimkulling:
There are still supported.

PR #6326, opened 2025-08-15
Fixed a `heap-use-after-free` in `SortByPTypeProcess`.

DonggeLiu:
The process incorrectly deleted original meshes from the scene when cleaning up after an error if those meshes had been added to its output list (`outMeshes`).
The fix ensures proper ownership transfer by nullifying the original mesh pointer in the scene (`pScene->mMeshes[i] = nullptr;`) when the mesh is moved to outMeshes. This prevents the scene destructor from attempting to delete the mesh again later, while allowing the error cleanup path in `SortByPTypeProcess` to correctly delete all meshes it owns (both newly created and transferred originals).


kimkulling:
merged, thanks a lot for your contribution.

PR #6329, opened 2025-08-22
[CMake] fix zlib linking

ZeunO8:
When linking zlib, assimp should use `${ZLIB_LIBRARIES}` for the zlib library name.

`-lz` is not valid in some cases, and instead should be `-l${ZLIB_LIBRARIES}`


ZeunO8:
This change is added from the `v6.0.2` tag, not the latest `main`

ZeunO8:
Just thinking post PR submission... what about `ZLIB_LIBRARY` instead of `_LIBRARIES`

kimkulling (approved):
Looks fine.

PR #6345, opened 2025-09-06
Fix hasTangentsAndBitangents method in jassimp's AiMesh.java

fifth-light:
Original code `m_tangents != null && m_tangents != null;` triggers error prone.

I am building jassimp with Bazel, and it's built-in error prone alarms:
```
external/+_repo_rules+jassimp/jassimp/src/jassimp/AiMesh.java:317: error: [IdentityBinaryExpression] A binary expression where both operands are the same is usually incorrect; the value of this expression is equivalent to `m_tangents != null`.
        return m_tangents != null && m_tangents != null;
                                  ^
    (see https://errorprone.info/bugpattern/IdentityBinaryExpression)
```

### DIFF
--- a/Build.md
+++ b/Build.md
@@ -110,5 +110,5 @@ You need to have pip installed:
 pip install pyassimp
 ```
 
-### Get the SDK from itchi.io
-Just check [itchi.io](https://kimkulling.itch.io/the-asset-importer-lib)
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine.

kimkulling:

PR #6359, opened 2025-09-23
Fix an invalid offsetof warning-as-error

sliedes:
The code already disables the warning on clang. It's also caused on GCC in some circumstances (particularly when building with `-D_GLIBCXX_DEBUG`).


kimkulling:

PR #6384, opened 2025-10-28
Removed unnecessary copy constructor declaration in aiVector3t

### LINKED ISSUE #6381: Bug:Deprecated copy assignment operator in vector3.h
**Description**
The aiVector3t class has an explicitly declared default copy constructor, and an implicitly declared copy assignment operator. This causes the deprecated-copy warning to be triggered on Clang (and quite possible GCC), which in turn may result in build errors if the warnings as errors flag is on.

**To Reproduce**
- Create a sample program
- Make sure the vector3.h header is included (in my case, I included [assimp/Importer.hpp](https://github.com/assimp/assimp/blob/master/include/assimp/Importer.hpp))
- Build with clang with -Werror -Wall -Wextra flags
- The build will result in an error

**Expected behavior**
The build should not fail

**Fix**
Simply removing the explicit copy constructor declaration makes the warning disappear. Alternatively, explicitly declaring the copy assignment operator would also fix the issue.

 - OS: Windows 10
 - Version: 22H2

EddieBreeg:
Fixes #6381

### DIFF
--- a/include/assimp/vector3.h
+++ b/include/assimp/vector3.h
@@ -81,6 +81,14 @@ public:
     /// @param  _xyz  The value for x, y and z.
     explicit aiVector3t (TReal _xyz ) : x(_xyz), y(_xyz), z(_xyz) {}
 
+    /// @brief  The copy constructor.
+    /// @param  o The instance to copy from.
+    aiVector3t( const aiVector3t& o ) = default;
+
+    /// @brief  The copy assignment operator.
+    /// @param  o The instance to copy.
+    aiVector3t& operator=(const aiVector3t& o) = default;
+
     /// @brief  combined operators
     /// @brief  The copy constructor.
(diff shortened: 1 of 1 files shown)


turol:
I think it would be better to add the explicit copy assignment operator instead.

EddieBreeg:
I could do that, of course, but my reasoning was that explicitly declaring special functions just to default them doesn't do anything to make the API clearer. Arguably, for a class such as this one, we all know what these default functions do, and adding unnecessary declarations just adds noise  for no added value?
I'd love to hear your thoughts if you have another perspective though!

turol:
Personal preference. Explicit is better than implicit.

EddieBreeg:
Fair enough, I'm right there with you in principle. I can't say I see a point in this specific case but I'll add the explicit declaration anyhow.

kimkulling:
That was the reason to add them years ago.

PR #6391, opened 2025-11-13
Fix apt-get install command in Dockerfile

kimkulling:


kimkulling:
Will create a security risk.

PR #6393, opened 2025-11-14
Added static cast for conversion from uLong to unsigned int for compi…

johncurley:
On macOS 12.7.6 clang the code was failing to compile due to the conversion from uLong to int.


kimkulling:
The change is gone? Is it correct?

johncurley:
Seems to be working now. I'm guessing it came up from a -Werror flag being present but it's probably good to address.

PR #6397, opened 2025-11-18
Added missing scene extras properties for glTF2Importer

Orleonn:
Hi.
I wanted to import my custom scene properties via the glTF2 improter, but I found that the importer ignores them.


kimkulling:

PR #6404, opened 2025-11-28
Fixed DXFHelper using float instead of ai_real

JakobPer:
The DXF importer did not use ai_real for importing floating point values which caused it to not use double precision when it was enabled. DXF supports double values though.

I also added some CMake presets for VisualStudio to make building easier there.

 * Fixed DXFHelper using float instead of ai_real

## Summary by CodeRabbit

* **New Features**
  * Added double-precision build configuration option for enhanced numerical accuracy in specific use cases.

* **Improvements**
  * Enhanced precision handling in DXF file format processing.


kimkulling (approved):
Looks fine.

kimkulling:

PR #6411, opened 2025-12-05
Only build collada test when importer/exporter is build

negril:
The collada export/import tests are build even if `ASSIMP_BUILD_COLLADA_IMPORTER` and/or `ASSIMP_BUILD_COLLADA_EXPORTER` are disabled causing tests to fail.

A similar check already exists for `ASSIMP_BUILD_USD_IMPORTER` and should probably also be done for others.

See-also: https://bugs.gentoo.org/962559

## Summary by CodeRabbit

* **Chores**
  * Improved build configuration to conditionally run tests based on enabled features, enhancing build efficiency.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -147,8 +147,6 @@ SET( IMPORTERS
   unit/utBlendImportMaterials.cpp
   unit/utBlenderWork.cpp
   unit/utBVHImportExport.cpp
-  unit/utColladaExport.cpp
-  unit/utColladaImportExport.cpp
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks good!

turol:
This actually disabled the collada tests unconditionally. Since collada is built by default unlike usd there is no option for it and `ASSIMP_BUILD_COLLADA_IMPORTER` defaults to false.

I intend to add some new collada tests soons. If you want to have the ability to disable these tests please send in a proper fix or I will just enable them all unconditionally again.

PR #6421, opened 2025-12-10
Fix incorrect FBX connection order

mwpenny:
PR #6120 fixed a macro which previously prevented `std::unordered_multimap` from being used during FBX import when compiled with GCC (50e661a337e9d5bd21551c9535b82b500e6897f4).

This caused `FBX::Element` properties to be stored in arbitrary order, but the order of connections ("C" properties) in FBX files matters.
* Insertion order is computed while [iterating the map](https://github.com/assimp/assimp/blob/75386282c15611c365d91ac17db01d9a64644920/code/AssetLib/FBX/FBXDocument.cpp#L505)
* Stored insertion order is how `Connection`s are [compared](https://github.com/assimp/assimp/blob/75386282c15611c365d91ac17db01d9a64644920/code/AssetLib/FBX/FBXDocument.h#L985-L1002)
* The comparison function is used to sort the `Connection`s during [retrieval](https://github.com/assimp/assimp/blob/75386282c15611c365d91ac17db01d9a64644920/code/AssetLib/FBX/FBXDocument.cpp#L551)

The main issue I saw was incorrect materials.

This PR changes the type of `FBX::ElementMap` to always use `std::multimap` (i.e., ordered). This was the behavior for years under GCC while the broken macro was in use. Unordered containers are still used elsewhere in the FBX importer to benefit from their performance.

Another way to fix this, for example, is to store the ordering in the `Element`. Let me know if you prefer that. I erred on the side of simplicity and history.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal data structure implementation for FBX asset parsing to improve consistency and performance characteristics.


kimkulling:

mwpenny:
Happy to help. Thanks for the fast response, and for this project.

Question: is there a rough release schedule/cadence documented somewhere? Or a list of remaining work before the next release? I took a look at the DevOps backlog board but couldn't find it.

There has been a lot of great work since the previous release. It would be nice to be able to get the latest improvements and fixes via a package manager. That would also help with downstream build automation for library consumers.

Thanks again.

PR #6432, opened 2025-12-20
Fix UTF-8 flag check in CMakeLists.txt

ljgdsq:
assimp’s MSVC UTF‑8 detection uses IN_LIST on CMAKE_CXX_FLAGS, which is a string, not a list. The condition always fails, so assimp adds /source-charset:utf-8 even when a parent project already sets /utf-8

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration for UTF-8 source charset handling in MSVC builds.


kimkulling:

PR #6472, opened 2026-02-10
FBXExporter: Fix stack-use-after-scope in WriteObjects

oliverchang:
The FBX exporter was unconditionally dereferencing a map iterator (`tp_elem`) even when the key was not found (i.e., `tp_elem == tpath_by_image.end()`). This resulted in a stack-use-after-scope error when accessing `tp_elem->second` to populate "FileName" and "RelativeFilename" nodes, as dereferencing the end iterator of the map accessed invalid stack memory (the map's sentinel).

The code already correctly initialized a local `tfile_path` variable based on whether the iterator was valid. This patch updates the `AddChild` calls to use `tfile_path` instead of dereferencing the potentially invalid iterator.

Fixes: https://issues.oss-fuzz.com/issues/465494996

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected texture path handling in FBX exports to use remapped paths instead of originally discovered paths, ensuring accurate texture references in exported files.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -3,7 +3,7 @@ Open Asset Import Library (assimp)
 
 Open Asset Import Library is a library that loads various 3D file formats into a shared, in-memory format. It supports more than __40 file formats__ for import and a growing selection of file formats for export.
 
-### Current project status
+## Current project status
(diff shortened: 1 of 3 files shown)


kimkulling:

PR #6537, opened 2026-02-17
HL1MDLLoader: fix segfault in destructors

oliverchang:
In `HL1MDLLoader::read_animations`, the arrays for `scene_->mAnimations` and `scene_animation->mChannels` were allocated using `new T*[count]`. This performs default initialization, which leaves the pointer elements with indeterminate (garbage) values.

If an exception (such as a `DeadlyImportError` from malformed input) is thrown during the loop populating these arrays, the `aiScene` or `aiAnimation` destructors are invoked during stack unwinding. These destructors iterate through the allocated arrays and call `delete` on each element. Because the arrays contained garbage values for indices not yet reached by the loader, the destructor attempted to delete invalid memory addresses, leading to a segmentation fault.

This patch changes the allocations to use value-initialization (`new T*[count]()`), ensuring all pointers are initialized to `nullptr`. Since `delete nullptr` is a safe no-op, the destructors can now safely clean up partially initialized objects during an exception.

Fixes: https://issues.oss-fuzz.com/issues/483188619

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed improper pointer initialization in animation channel allocation, ensuring memory is properly initialized to null values.

### DIFF
--- a/code/AssetLib/FBX/FBXExporter.cpp
+++ b/code/AssetLib/FBX/FBXExporter.cpp
@@ -1766,8 +1766,8 @@ void FBXExporter::WriteObjects () {
             // can't easily determine which texture path will be correct,
             // so just store what we have in every field.
             // these being incorrect is a common problem with FBX anyway.
-            tnode.AddChild("FileName", tp_elem->second);
-            tnode.AddChild("RelativeFilename", tp_elem->second);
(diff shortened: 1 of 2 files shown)


kimkulling:

PR #6540, opened 2026-02-20
Obj: Fix heap-buffer-overflow in getFace via vertical tabs

bwendling:
The `ObjFileParser::getFace` method failed to recognize the vertical tab character (`\v`, 0x0b) as a separator. While the `IsSpaceOrNewLine` utility handles most whitespace (space, tab, CR, LF, FF), it excludes `\v`.

When encountering a vertical tab, the parser fell through to an `else` block that calls `::atoi(&(*m_DataIt))`. Because `atoi` treats `\v` as whitespace per the C standard, it skips the character and continues reading. If `\v` is located at the end of the buffer (e.g., followed by a newline at the buffer boundary), `atoi` can read past the allocated memory, triggering a heap-buffer-overflow.

This fix explicitly checks for `\v` and treats it as a separator, resetting the position counter and preventing the invalid `atoi` call.

Verified with AddressSanitizer and confirmed that all 584 existing unit tests pass.

Fixes: https://issues.oss-fuzz.com/issues/476180586

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where OBJ model files containing certain whitespace characters would fail to parse correctly during face index processing.

### DIFF
--- a/code/AssetLib/Obj/ObjFileParser.cpp
+++ b/code/AssetLib/Obj/ObjFileParser.cpp
@@ -467,7 +467,7 @@ void ObjFileParser::getFace(aiPrimitiveType type) {
                 ASSIMP_LOG_ERROR("Obj: Separator unexpected in point statement");
             }
             iPos++;
-        } else if (IsSpaceOrNewLine(*m_DataIt)) {
+        } else if (IsSpaceOrNewLine(*m_DataIt) || *m_DataIt == '\v') {
             iPos = 0;
         } else {
             //OBJ USES 1 Base ARRAYS!!!!


bwendling:
While this patch fixes the issue for the vertical tab (`\v`), there might be similar issues with other control characters, like `\a`, `\b`, and `\e` (GCC only).

I also noticed that the `Skip*` methods check `in != end` after accessing what `in` points to. This could lead to accessing out-of-bounds memory if `in == end`.

bwendling:
> 🧹 Nitpick comments (1)
> > code/AssetLib/Obj/ObjFileParser.cpp (1)> `470-470`: **Consider fixing `IsSpaceOrNewLine` in `ParsingUtils.h` rather than patching only `getFace`.**
> > > The fix correctly prevents the heap-buffer-overflow by stopping `\v` from reaching `::atoi`. However, `\v` (vertical tab) is standard C whitespace, and the omission is in `IsSpaceOrNewLine` itself, which only includes space, tab, `\r`, `\n`, `\f`, and null. A single addition there would fix all call sites consistently. As it stands, other parse paths in this file still don't recognize `\v`:
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@code/AssetLib/Obj/ObjFileParser.cpp` at line 470, The root issue is that
> > > IsSpaceOrNewLine (in ParsingUtils.h; also called IsLineEnd in some variants)
> > > omits the vertical-tab character '\v', causing callers like getFace,
> > > copyNextWord, getObjectName, and getNewMaterial to mis-handle '\v'; fix it by
> > > adding '\v' to the set of whitespace/line-end characters in IsSpaceOrNewLine so
> > > all call sites consistently treat vertical tab as whitespace, then revert the
> > > local special-case change in getFace that added an explicit '\v' check; ensure
> > > unit/parsing tests cover the '\v' case.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> Nitpick comments:
> In `@code/AssetLib/Obj/ObjFileParser.cpp`:
> - Line 470: The root issue is that IsSpaceOrNewLine (in ParsingUtils.h; also
> called IsLineEnd in some variants) omits the vertical-tab character '\v',
> causing callers like getFace, copyNextWord, getObjectName, and getNewMaterial to
> mis-handle '\v'; fix it by adding '\v' to the set of whitespace/line-end
> characters in IsSpaceOrNewLine so all call sites consistently treat vertical tab
> as whitespace, then revert the local special-case change in getFace that added
> an explicit '\v' check; ensure unit/parsing tests cover the '\v' case.
> ```

We considered doing exactly this for the patch, but decided against it because we weren't certain that it wouldn't cause other issues. For instance, `\v` isn't typically considered an "end-of-line" white-space character. Perhaps there should be an `IsControl()` method that's integrated into the other methods somehow?

kimkulling:
Merged, thank you for your contribution!

PR #6547, opened 2026-02-27
Fix duplicate KEYS chunk overwrite in B3DImporter

EasyT-T:
### Problem

When loading `.b3d` models, the `B3DImporter` assigns keyframes from each `KEYS` chunk directly to the `aiNodeAnim` structure. If a model contains multiple `KEYS` chunks for the same node (which is common output from popular `.b3d` authoring tools), later chunks silently overwrite the keyframes defined by earlier ones, causing animation data loss.

### Root Cause

In `B3DImporter.cpp`, the handler for `KEYS` chunks replaces the existing position/rotation/scaling key arrays on `aiNodeAnim` rather than appending to them.

### Fix

Changed the importer to accumulate keyframes across multiple `KEYS` chunks instead of overwriting. When a subsequent `KEYS` chunk is encountered for the same node animation, the new keyframes are now appended to the existing arrays.

### References

- Affected files: `code/AssetLib/B3D/B3DImporter.h`, `code/AssetLib/B3D/B3DImporter.cpp`

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of B3D animation keyframe processing for better organization and maintainability; animation input parsing now uses an intermediate keyframe container.
  * No user-facing behavior changes: existing animation import and playback results remain unchanged.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Exporter.cpp
+++ b/code/AssetLib/glTF2/glTF2Exporter.cpp
@@ -190,7 +190,7 @@ void SetAccessorRange(Ref<Accessor> acc, void *data, size_t count,
 
     // Allocate and initialize with large values.
     for (unsigned int i = 0; i < numCompsOut; i++) {
-        acc->min.push_back(std::numeric_limits<double>::min());
+        acc->min.push_back(std::numeric_limits<double>::max());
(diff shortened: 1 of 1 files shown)


kimkulling:

PR #6556, opened 2026-03-03
Cleanup the aiColor3D type

Pichas:
This MR fixes next issues for aiColo3D type:
* respect Rule of Zero. This allows compiler to generate proper copy/move constructors and assign operators. It also makes the type to be trivially copyable/movable. This also allows to use functions that checks type traits like `std::vector<T>::emplace_back() noexcept` in right way if aiColo3D is a field in another type.
* fixes security issue with accessing memory after the object via const `operator[]`
* add epsilon for comparison operators to avoid issues with float types

## Summary by CodeRabbit

* **Breaking Changes**
  * Color type no longer supports implicit copy/assignment — code relying on automatic copying must be adjusted.

* **Updates**
  * Color equality, inequality and ordering now use a small floating-point tolerance for more stable comparisons.
  * Black/dark detection uses the same tolerance to reduce false positives from tiny numeric noise.
  * Component indexing and constructors adjusted for more predictable, non-throwing behavior.

* **Documentation**
  * Improved multi-line brief documentation for several public types.


kimkulling:

PR #6575, opened 2026-03-16
Fix a container-overflow vulnerability in `Assimp::ColladaLoader::B…

bwendling:
…uildMeshesForNode`

The root cause was an unchecked access to the `newMats` vector using `matIdx`. When a material name (`meshMaterial`) is not found in `mMaterialIndexByName`, `matIdx` defaults to 0.  If `newMats` is empty (which happens if the material library is empty or failed to load), accessing `newMats[0]` results in a container-overflow.

The fix involves adding a bounds check `matIdx < newMats.size()` to the condition guarding the access to `newMats`. Additionally, based on maintainer feedback, I added a warning log `ASSIMP_LOG_WARN` when the index is out of bounds, to inform the user about the missing material definition or broken reference.

I verified the fix using the provided reproduction command. The container-overflow is no longer triggered. I also built and ran the unit tests. Since `bin/unit` was missing, I manually built the unit tests using `cmake.real` and `ninja` in `/src/assimp/build_tests` with `ASSIMP_BUILD_TESTS=ON`, `ASSIMP_BUILD_ZLIB=ON`, and `ASSIMP_WARNINGS_AS_ERRORS=OFF` (to bypass a gtest compilation warning), and updated `run_tests.sh` to point to the built binary. All 584 tests passed.

Fixes: https://issues.oss-fuzz.com/issues/483102958

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when loading Collada files with invalid material indices. The application now gracefully handles such cases by logging a warning and continuing without applying problematic texture mappings, preventing potential crashes.


bwendling:
## Notes

The original CodeMender patch was to modify the if conditional like this:
```
if (table && !table->mMap.empty() && matIdx < newMats.size()) {
```
I reran CodeMender asking whether `newMats.size() == 0` is a valid value. After testing, CodeMender came up with the patch below. Both will solve the issue, so I leave the choice to you.

Also, `matIdx` is used around line number 545, but it's while analyzing a different container. The assumption of the code seems to be that `mMaterialIndexByName` and `mMeshIndexByID` are always in the same order. Is that a valid assumption?

## Code Review Issue

https://github.com/assimp/assimp/pull/6575#issuecomment-4070041383 says that it's a "code smell" issue. I'm not sure how to address that with the patch.

kimkulling:

Issue #4746, opened 2022-10-03
Bug: Collada Animation Clips


pohldaniel:
**Describe the bug**
The Collada Parser seems not handling AnimationClips in a expected way

**To Reproduce**
1. Go to ['...'](https://github.com/pohldaniel/OpenGL_Games/tree/master/06Voyager/res/models/dragon_dae)
2. Load the dragon.dae with assimp and check the parsed "animation name" from the aiAnimation

**Expected behavior**
The AnimationClips should be loaded correct

**Platform (please complete the following information):**
 - Windows 10 (x64/ x86) Visual Studio 2015

**Additional context**
I am pretty sure that the "ReadAnimationClipLibrary( ... ) function out of the file "ColladaParser.cpp | line: 397" should written down in the following way.

void ColladaParser::ReadAnimationClipLibrary(XmlNode &node /*NODE: library_animation_clips*/) {
    if (node.empty()) {
        return;
    }
	for (XmlNode &animationClip : node.children()) {

		std::string animName;
		if (!XmlParser::getStdStrAttribute(animationClip, "name", animName)) {
			if (!XmlParser::getStdStrAttribute(animationClip, "id", animName)) {
				animName = std::string("animation_") + ai_to_string(mAnimationClipLibrary.size());
			}
		}

		std::pair<std::string, std::vector<std::string>> clip;
		clip.first = animName;

		for (XmlNode &instanceAnimation : animationClip.children()) {
			const std::string &currentName = instanceAnimation.name();
			if (currentName == "instance_animation") {
				std::string url;
				readUrlAttribute(instanceAnimation, url);
				clip.second.push_back(url);
			}	
		}

		if (clip.second.size() > 0) {
			mAnimationClipLibrary.push_back(clip);
		}
	}
}

tellypresence:
Looks like above 3d model link got moved to [`dragon.dae`](https://github.com/pohldaniel/OpenGL_Games/blob/master/34Animation/res/models/dragon/dragon.dae)

tellypresence:
[`dragon.dae`](https://github.com/pohldaniel/OpenGL_Games/blob/master/34Animation/res/models/dragon/dragon.dae) animation seems correct
<img src="https://github.com/user-attachments/assets/194eae8f-97bd-4a42-966d-dcf936025e84" width=180 >

pohldaniel:
[...]

PR #6589, opened 2026-03-22
Fix collada animation clips

pohldaniel:
This fix is realted to the issue https://github.com/assimp/assimp/issues/4746 and should fix them :).

## Summary by CodeRabbit

* **Bug Fixes**
  * Animation-clip parsing now processes each clip individually, resolving per-clip names/IDs correctly.
  * Instance animations are collected only for their respective clips, preventing cross-clip aggregation.
  * Unnamed clips receive a stable fallback name to avoid missing or mislabelled animations during import.
  * Improved overall import accuracy for animation clips and their associated animations.


kimkulling (approved):
Looks correct. Thanks a lot for your work.

PR #6593, opened 2026-04-07
Added Gaussian Import to PlyLoader

Dany-RD:
## Summary by CodeRabbit

* **New Features**
  * Added support for Gaussian Splat attributes in PLY import: per-vertex Gaussian data (diffuse DC, rest coefficients, scale, rotation, and opacity) are now detected in compatible source files and exposed on meshes when available.


turol:
This is an API change so needs a major version bump if merged. Do we really need this? Are there other formats for GS than .ply so this is actually useful to someone?

kimkulling:
This will break ABI compatibility.

PR #6599, opened 2026-04-15
validate aiString length to prevent out-of-bounds read/write

jmestwa-coder:
## Summary

Validate `aiString` length during Assbin import to prevent out-of-bounds memory access from malformed input.

---

## Problem

`aiString::length` is read from input and used to fill a fixed-size buffer (`AI_MAXLEN`) without bounds checking.

This allows malformed `.assbin` data to trigger out-of-bounds writes and undefined behaviour.

---

## Fix

- Reject `aiString::length >= AI_MAXLEN`
- Validate read size (`bytesRead == length`)
- Fail early with `DeadlyImportError`

---

## Key Points

- Centralized fix in `Read<aiString>()` (covers all Assbin string reads)
- No behavior change for valid files
- Prevents memory corruption from untrusted input

---

## Test

- Export valid `.assbin`
- Locate node chunk via structure markers (no fixed offsets)
- Corrupt `aiString` length
- Verify import fails with expected error

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened ASSBIN import validation to reject malformed string lengths and prevent oversized node name payloads from being processed.
  * Improved handling of incomplete/truncated string data and ensured imported strings are safely terminated.
  * Strengthened stream cleanup during import to reduce risk of resource leaks on error paths.

* **Tests**
  * Added a unit test covering rejection of ASSBIN blobs with invalid (oversized) node name length fields.


kimkulling (approved):
Looks good. I love the way you are testing the out-of-bounds access while reading an aiString. Thanks for your great work!

kimkulling:
The sanitizer is not happy. Could you please take a look?

turol:
Needs some RAII here:

https://github.com/assimp/assimp/blob/17c12da558d23d70e4c728b30958e95cd616cbcb/code/AssetLib/Assbin/AssbinLoader.cpp#L680

This bug is not caused by this PR but merely exposed.

jmestwa-coder:
Thanks for the feedback.

I’ve updated the code to use RAII for IOStream so it is properly closed on all paths, including exceptions. This resolves the sanitizer issue without changing existing logic.

jmestwa-coder:
@kimkulling , I resolved the merge conflicts and updated the branch.

jmestwa-coder:
noticed the last master merge left some damage on this branch, pushed a fix:

- restored the license header in AssbinLoader.cpp (merge had pasted code over it)
- test now expects the "String length too large" message that came in from #6606, instead of my old wording

kimkulling:

PR #6600, opened 2026-04-20
Fix out-of-bounds read in StreamReader::IncPtr

uwezkhan:
While going through StreamReader, I noticed that IncPtr allows negative offsets without checking if the pointer stays within bounds.

This can lead to the cursor moving before the start of the buffer, which may cause out-of-bounds reads when parsing malformed files.

Added proper checks to ensure the pointer always stays within the valid buffer range before updating it.

No impact on valid inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stream reading now validates pointer bounds before moving, preventing forward/backward seeks that would go past valid data and improving stability.
  * Direction-aware checks reject attempts to move before the buffer or beyond the read limit, avoiding corrupted reads.
  * Clarified error reporting when an invalid internal pointer state is detected, making failures easier to diagnose.


kimkulling (commented):
Looks fine for me.

kimkulling:

PR #6601, opened 2026-04-20
Fix integer truncation in StreamReader size calculations

uwezkhan:
While reviewing StreamReader, I noticed that some size-related methods use C-style casts that may truncate values on 64-bit systems.

This patch replaces those casts with safer static_cast usage to avoid unintended truncation and improve correctness when handling larger files.

The changes are minimal and do not modify existing APIs or behavior for valid inputs.

## Summary by CodeRabbit

* **Documentation**
  * Clarified parameter naming and return descriptions in internal API docs for greater accuracy.

* **Refactor**
  * Improved internal handling of position/size calculations and pointer-to-size conversions for safer, more consistent behavior; no change to public signatures or runtime behavior.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6602, opened 2026-04-20
Fix glTF2 occlusion strength import/export and normal scale export

Valakor:
When trying to use assimp to import glTF2 files I ran into some issues with a handful of glTF-specific property keys:
- Occlusion strength was being imported and exported with the custom property key `"$tex.file.strength"` rather than using the GLTF-specific key `AI_MATKEY_GLTF_TEXTURE_STRENGTH` defined as `"$tex.strength"` that already existed making it confusing and difficult to access this material property.
- Normal scale was being imported as the existing GLTF-specific key `AI_MATKEY_GLTF_TEXTURE_SCALE` defined as `"$tex.scale"` but exported by looking for the custom property key `"$tex.file.scale"` meaning round-trip import -> export would not retain this property.

## Summary by CodeRabbit

* **Refactor**
  * Optimized glTF2 texture property handling in export and import operations. Texture scale and strength values are now accessed more efficiently, improving the overall robustness of glTF2 asset processing.

### DIFF
--- a/code/AssetLib/glTF2/glTF2Exporter.cpp
+++ b/code/AssetLib/glTF2/glTF2Exporter.cpp
@@ -556,12 +556,6 @@ void glTF2Exporter::GetMatTexProp(const aiMaterial &mat, unsigned int &prop, con
     mat.Get(textureKey.c_str(), tt, slot, prop);
 }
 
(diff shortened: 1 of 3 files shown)


kimkulling:

PR #6603, opened 2026-04-20
Fix aiGetMaterialProperty in the presence of material keys that are prefixes of other keys

Valakor:
The existing code compares each property key in a material to the input via `strncmp(prop->mKey.data, pKey, strlen(pKey))`. This will incorrectly return longer keys whose prefix matches the searched-for key. For example: searching for `"$tex.mapping"` when the material contains `"$tex.mappingname"` will return the "mappingname" property despite that not being the requested key.

I haven't done an exhaustive search of all material keys to find common prefixes, but I know this affects at least these two glTF properties that contain the common prefix `AI_MATKEY_MAPPING`: `AI_MATKEY_GLTF_MAPPINGNAME` and `AI_MATKEY_GLTF_MAPPINGID`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Material property lookup now uses a full-name comparison to avoid partial/prefix matches.
  * This ensures properties are selected precisely, preventing incorrect property application and improving material consistency.


kimkulling:
Good catch, thanks for your input. I guess using strcpy will create some security flaws.

Valakor:
The codebase seems to use `strcmp` fairly ubiquitously which is why I changed it to that, but it would be easy to instead do something like `strncmp(prop->mKey.data, pKey, std::max(prop->mKey.length, strlen(pKey)))` if that is preferred.

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6618, opened 2026-04-29
Add overflow checks for texture allocations in binary importers

uwezkhan:
While looking through the texture loading code, I noticed that some importers allocate texture buffers using width * height without checking whether that multiplication can overflow.

If a malformed file provides very large texture dimensions, the multiplication can wrap and lead to an undersized allocation before the texture data is processed.

This patch adds overflow checks before allocating texture buffers in the affected importers, using the same validation approach already present in other texture-loading code.

Valid inputs are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened texture import validation to reject zero-dimension or malformed textures and prevent integer-overflow, avoiding invalid memory allocations.
  * Ensured pixel data and material payloads are fully and safely read; imports now fail gracefully on truncated or unexpected input, reducing crashes and corrupted assets.
* **Refactor**
  * Improved internal buffer ownership and read logic to enhance stability and memory-safety during imports.


uwezkhan:
Pushed a fix for the texture read path.

The merge with master had left ReadBinaryTexture with a duplicated, half-merged read block, so I collapsed it back to a single branch. Both paths now verify the length returned by stream->Read:

- mHeight == 0 (compressed): reads mWidth bytes and throws on a short read.
- mHeight != 0: reads mWidth*mHeight texels and throws on a short read.

The width*height overflow guard is unchanged. Before, a truncated file left partially initialized aiTexel data; now it fails with a clear error. Valid files are unaffected.

kimkulling (changes requested):
There are some static code analysis finding, which are suspicious. Could you take a look onto them?

uwezkhan:
Took a look, the findings were real. The uLongf > SIZE_MAX guards in the compressed path were always false (uLongf can never exceed SIZE_MAX), which is what the analyzer flagged as suspicious. On top of that, the merge with master had left delete[] calls on what are now std::vector buffers plus Close() calls on the already-managed stream, which is what broke the builds.

Pushed a cleanup: the vectors own the buffers now, so the manual delete[]/Close() and the dead size guards are gone. Short reads and zlib failures still throw DeadlyImportError. I also cleared the remaining sonar finding in M3DImporter by filling the texel buffer directly instead of a temporary vector plus std::copy, and restored zero-initialization of the scene pointer arrays that the merge dropped (before: a throw mid-read could free indeterminate pointers; after: they start as nullptr).

Behavior for valid files is unchanged. Full unit suite passes locally.

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6628, opened 2026-05-05
Prevent Heap Buffer Overflow in MaterialSystem String Assignments

metsw24-max:
This patch addresses a classic memory safety issue by removing an unsafe `strcpy` call in `code/Material/MaterialSystem.cpp` (`aiMaterial::AddBinaryProperty`) and replacing it with a strictly bounded memory copy. 

Previously, the length of `pKey` was checked using `ai_assert(AI_MAXLEN > pcNew->mKey.length)`. In production environments built with optimizations (`NDEBUG`), this assertion expands to nothing. Consequently, the subsequent `strcpy(pcNew->mKey.data, pKey)` operation executes blindly. If the incoming key exceeds `AI_MAXLEN`, it writes past the allocated bounds of the `mKey.data` heap buffer.

The patch resolves this by:
1. Evaluating the length of the incoming key.
2. Statically bounding the length to `AI_MAXLEN - 1` using `std::min<size_t>()`.
3. Executing a bounded `memcpy` operation instead of an unbounded `strcpy`.
4. Manually appending the null-terminator at the exact copied length.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of material property handling by preventing issues with excessively long property identifiers, ensuring proper data integrity and always terminating property names.
  * Added a warning when a property name is too long and gets truncated.

### DIFF
--- a/code/Material/MaterialSystem.cpp
+++ b/code/Material/MaterialSystem.cpp
@@ -513,7 +513,10 @@ aiReturn aiMaterial::AddBinaryProperty(const void *pInput,
 
     const size_t keyLen = ::strlen(pKey);
     pcNew->mKey.length = static_cast<ai_uint32>(std::min<size_t>(keyLen, AI_MAXLEN - 1));
-    memcpy(pcNew->mKey.data, pKey, pcNew->mKey.length);
+    if (keyLen >= AI_MAXLEN) {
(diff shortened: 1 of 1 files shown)


kimkulling:

PR #6643, opened 2026-05-10
Improve parser robustness and defensive validation

metsw24-max:
This PR introduces several parser robustness and defensive validation improvements across ASSIMP importer and shared utility code paths.

The changes focus on:
- integer overflow validation,
- safer allocation handling,
- improved bounds checking,
- defensive runtime validation,
- guaranteed null-termination after bounded string copies,
- and cleanup of unsafe string handling patterns.

The goal is to improve resilience when processing malformed or adversarial input while preserving existing behavior for valid assets.

## Main Improvements

- Added overflow validation before allocation-related arithmetic
- Improved UTF conversion bounds handling
- Replaced debug-only assertions with runtime validation
- Ensured explicit null-termination after bounded copies
- Hardened logging buffer handling
- Improved validation of file-controlled dimensions and lengths

## Testing

- Successful Release build
- Successful compilation of unit tests
- Successful execution of importer/exporter test suites
- No observed regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent invalid or overflowing texture dimensions during AMF loading.
  * Improved handling of texture format hints to ensure proper string termination.
  * Fixed UTF-8 conversion range handling for certain Unicode encodings.
  * Prevented oversized log messages from causing errors; messages are now safely truncated.
  * Improved memory cleanup during mesh and scene construction.


metsw24-max:
Pushed a follow-up for the valid points from the automated review.

**Format hint truncation** — the M3D and USD copies were bounded to 7 chars, which would clip 8-character hints like `rgba8888`. Switched both to `HINTMAXTEXTURELEN - 1` so the full hint survives and it's still null-terminated.

**AMF texture** — moved the dimension and `width * height` overflow checks ahead of the division/multiplication that used them, so we don't divide by zero or wrap before validating.

**Assbin texture** — the `width * height * 4` byte count could still overflow a 32-bit value even after the pixel-count guard, so the count is now computed in 64-bit and rejected against `UINT_MAX / 4`.

Also restored a few trailing newlines I'd stripped by accident. The rest (unaligned reads) looks pre-existing rather than something this change introduces, so I've left it out to keep the diff focused.

metsw24-max:
Merged master in, since the Assbin and MaterialSystem hunks here were superseded by fixes that landed in the meantime (those two files now carry no delta). Also swapped the remaining strncpy hint copies for ai_snprintf to clear the SonarCloud hotspot, and applied the analyser's auto/const suggestions in BaseImporter. Full unit suite passes locally.

PR #6654, opened 2026-05-14
Migrate OBJ parser from fixed-size stack buffer to std::string

metsw24-max:
Migrate OBJ Parser to Safe Buffers Programming Model

This pull request eliminates stack-based buffer overflow vulnerabilities (CWE-121) in the Wavefront OBJ file parser by implementing the Safe Buffers Programming Model. The migration replaces a fixed-size 4096-byte stack buffer with `std::string`, adds compiler-level enforcement via `-Wunsafe-buffer-usage`, and maintains complete backward compatibility with all existing OBJ import/export functionality.

### Buffer Migration

| Before | After |
|--------|-------|
| `char mBuffer[4096]` (fixed-size stack buffer) | `std::string mBuffer` (dynamic, bounds-checked) |
| Manual bounds checking with risk of overflow | Automatic capacity management via STL |
| `std::fill_n(mBuffer, Buffersize, '\0')` | `mBuffer.clear()` |

### Function Modernization

The `copyNextWord()` function has been refactored from an unsafe pointer-based implementation to a safe string operation:

**Before:**
```cpp
void ObjFileParser::copyNextWord(char *pBuffer, size_t length) {
    size_t index = 0;
    while (mDataIt != mDataItEnd && !IsSpaceOrNewLine(*mDataIt)) {
        if (index >= length - 1) return;
        pBuffer[index++] = *mDataIt++;
    }
    pBuffer[index] = '\0';
}

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory safety in the OBJ parser and more robust handling of multi-line/continued lines for correct OBJ parsing.

* **Tests**
  * Updated unit tests to validate multi-line continuations and the revised parser behavior.

* **Chores**
  * Enabled an additional Clang compiler warning to surface unsafe buffer usage during builds.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/assimp/assimp/pull/6654?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #6647, opened 2026-05-12
validate mesh face buffer allocation arithmetic

metsw24-max:
This PR introduces integer overflow hardening in importer parsing paths that process untrusted asset data.

The patch improves the safety of allocation-related arithmetic by validating calculations before they are used for buffer sizing or memory allocation. The goal is to prevent invalid allocation sizes caused by malformed or adversarial model files.

### Checked Arithmetic
- Added safe arithmetic helpers for addition and multiplication
- Replaced unchecked arithmetic operations with validated calculations
- Added overflow detection before allocation-related computations are performed

### Safer Buffer Size Handling
- Hardened face buffer size calculations
- Added validation for mesh face and index accumulation logic
- Improved handling of large or malformed metadata values

### Error Handling Improvements
- Added consistent failure handling using `DeadlyImportError`
- Invalid or overflowed calculations now fail safely instead of continuing with unsafe values

## Security Benefits

These changes improve parser robustness when handling untrusted asset files and reduce the risk of:
- integer overflows
- invalid allocation sizes
- parser instability
- unsafe memory operations caused by malformed input

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved 3D mesh import stability with overflow protection, so complex/non-triangular geometry now fails gracefully with clearer error messages instead of risking crashes or corrupted results.

* **Refactor**
  * Modernized the native import pipeline for scenes, materials, animations, lights, and cameras, with more consistent buffer/data handling to improve reliability across a wider range of assets.

PR #6655, opened 2026-05-16
Safe allocator and elimination of unsafe C/C++ buffer patterns

metsw24-max:
This PR hardens the jassimp JNI bridge (`port/jassimp/jassimp-native/src/jassimp.cpp`) against integer overflow and allocation failures when converting scenes for Java.

**Problem**
Buffer sizes were computed with unchecked multiplications (`count * sizeof(T)`), buffers were managed with raw `malloc`/`free` and `memcpy`, and there was no upper bound on direct-buffer allocations driven by file content.

**Fix**
- size arithmetic goes through the existing `SafeMultiply`/`SafeAdd` helpers, with a 1 GiB cap (`JASSIMP_MAX_DIRECT_ALLOC`) on any single direct allocation, checked before the Java side allocates the matching buffer
- buffer ownership moved to `std::vector` (RAII), `memcpy` replaced with `std::copy`; `JavaIOStream` now owns the model buffer via a moved vector instead of a raw pointer it had to free
- the model buffer allocation and the scene conversion are wrapped in try/catch so `std::bad_alloc` cannot cross the JNI boundary
- oversized copies are rejected in `copyBuffer`/`copyBufferArray`; zero-length copies are a no-op, so channels that legitimately carry no data (an animation with only rotation keys, for example) still import
- the `jobject` locals handed to `SmartLocalRef` are initialised, so the destructor no longer reads an indeterminate handle when the constructing call fails early; `NewDirectByteBuffer`/`NewIntArray` results are null-checked

**Notes**
- This PR originally also touched the OBJ parser, its tests and CMakeLists; those changes were superseded by #6654 and have been dropped after merging master, so the diff is now confined to jassimp.cpp.
- The merge also repairs three lines in this file that recent refactors on master left uncompilable (two `{nullptr]` member initialisers and an empty `!=` comparison in the import entry point). The port isn't built by CI, so these hadn't been caught.
- Because master's copy of the file doesn't parse, Sonar treats the whole file as new code and pre-existing findings show up against this PR. The two it rated as bugs are addressed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when importing models with large or malformed data.
  * Added safeguards to prevent excessive memory usage and arithmetic overflows during file, mesh, material, animation, and scene-data processing.
  * Improved handling of invalid, empty, or unreadable model data.
  * Corrected progress-handler validation during imports.
  * Prevented memory-allocation failures from escaping during model imports.


metsw24-max:
Pushed a fix for the red build jobs. The `-Wunsafe-buffer-usage` flag was the cause: GCC doesn't recognise it at all, and under Clang the existing `-Werror` turned it into thousands of failures across pre-existing code. Dropped the flag rather than churn the whole tree, so the build-and-test jobs should go green now.

metsw24-max:
Merged master to clear the conflict. The OBJ parser and CMake parts of this PR had been superseded by #6654, so the diff is now just the jassimp.cpp hardening, rebased onto the SafeMultiply/SafeAdd helpers that landed with #6647 rather than keeping a duplicate helper. One thing worth flagging: the recent refactors on master left three lines in this file uncompilable (two `{nullptr]` initialisers and a dropped `nullptr` in the import entry point) — the port isn't built by CI, so it hadn't surfaced. Those are fixed here too, and I've updated the PR description to match the narrower scope.

metsw24-max:
Picked up the two outstanding review points: the SafeAdd overflow branch now logs and returns false instead of throwing, so nothing can escape the JNI entry point uncaught, and the mesh-ref size check in loadSceneNode now runs before NewIntArray. Also merged current master to refresh CI; the earlier Windows failures were six-hour runner hangs mid-build on a month-old run and master builds those jobs fine, so they should clear with this push.

metsw24-max:
Pushed a follow-up.

**Regression**: the zero-length rejection in `copyBuffer` broke channels that legitimately carry no data. A node animation with only rotation keys gets empty position and scaling buffers on the Java side, and the import failed on them. Empty copies are a no-op again, oversized ones are still rejected.

**Sonar**: the gate was failing on the `delete` in `JavaIOSystem::Close` and on `SmartLocalRef` comparing an uninitialised handle when the constructing call fails early. Master's copy of this file doesn't parse (the `{nullptr]` lines), so Sonar treats the whole file as new code and pre-existing findings land on this PR. Both are addressed: the stream now has a single owner for its buffer and the wrapped `jobject` locals are initialised.

**Review points**: size checks run before the Java side allocates the matching buffers, the conversion in `aiImportFile` is wrapped in a `bad_alloc` handler, and the two JNI allocations are null-checked. The double-precision remark applies to the whole port rather than this loop, since vertices and keys are copied as raw float layouts too, so I've left that alone.

Also re-authored the last two commits, which had picked up the wrong git identity.

metsw24-max:
Pushed a fix for the Sonar gate: the reordered copy helpers had grown an identical tail, so that's one shared helper now, and the double free it reports on the stream delete is the same finding it raises on DefaultIOSystem::Close on master, so Close hands the stream to a unique_ptr instead.

PR #6661, opened 2026-05-20
Fix binary STL import on big-endian hosts

### LINKED ISSUE #5509: Bug: binary STL files are not being recognized in s390x/Ubuntu Noble (big endian)
**Describe the bug**
Using the s390x architecture (big endian) on Ubuntu Noble, when using the function `aiImportFileExWithProperties(..)` to open an STL binary file, assimp returns the message: `Failed to determine STL storage representation for ...` which appears when assimp can not determine if the STL file is a binary file or an ascii file in the code:
```c++
    if (IsBinarySTL(mBuffer, mFileSize)) {
        bMatClr = LoadBinaryFile();
    } else if (IsAsciiSTL(mBuffer, mFileSize)) {
        LoadASCIIFile(mScene->mRootNode);
    } else {
        throw DeadlyImportError("Failed to determine STL storage representation for ", pFile, ".");
    }
``` 
which is stored in the [STLLoader.cpp](https://github.com/assimp/assimp/blob/072320df4fc064f3bb690e24524e6d3fc090f4bd/code/AssetLib/STL/STLLoader.cpp#L72-L87). Note that this same test does not fail in the other supported architectures: amd64 or arm64 to name some of them.

If I tried to printout the values used by the STLLoader.cpp file code linked before, I get in the relevant Dart test:
```
p25: Running main() from /usr/src/googletest/googletest/src/gtest_main.cc
25: [==========] Running 1 test from 1 test suite.
25: [----------] Global test environment set-up.
25: [----------] 1 test from MjcfParserTest
25: [ RUN      ] MjcfParserTest.RoboticsFetch
25: Warning [Body.cpp:443] [MjcfParser] Unsupported number of <geom> in inferring <inertial>. We use only the first <geom> for now.
25: 1. File is: dart://sample/mjcf/openai/robotics/stls/fetch/base_link_collision.stl
25:  -------------- fileSize: 
25: 236084
25:  ---- faceCount: 
25: 1880227840
25:  --- expectedBinaryFileSize: 
25: 3817078868
25:  -------------- fileSize: 
25: 236084
25:  ---- faceCount: 
25: 1880227840
25:  --- expectedBinaryFileSize: 
25: 3817078868
```
Values are not the expected.

**To Reproduce**
1. I've been debugging a problem on the Dart library and using the following Dockerfile.
```Dockerfile
FROM --platform=s390x ubuntu:noble AS s390_dart_base
ENV LANG C
ENV LC_ALL C
ARG DEBIAN_FRONTEND=noninteractive

RUN apt-get update && \
    apt-get install -y vim ubuntu-dev-tools cmake build-essential git software-properties-common 

WORKDIR /root
RUN sed -i -e 's/Types: deb/Types: deb deb-src/g' /etc/apt/sources.list.d/*.sources \
    && cat /etc/apt/sources.list.d/ubuntu.sources \
    && apt-get update
RUN ls *.deb && dpkg -i libfcl*.deb
RUN apt-get build-dep dart -y
RUN apt-get update \
    && apt-get source dart
RUN cd dart-6.1* \
    && dpkg-buildpackage -j6 || true
```
2. run test with `make -j1 test ARGS\+=--verbose ARGS\+="-R test_MjcfParser"`

An easier approach could be just take the STL files from the Dart project and create a minimal test program. The [base_link_collision.stl](https://github.com/dartsim/dart/blob/main/data/mjcf/openai/robotics/stls/fetch/base_link_collision.stl) was the one that first failed on my test.

**Expected behavior**
The file is detected as a binary STL file.
xplain your problem.

 - OS:  Ubuntu Noble / s390 architecture

**Additional context**
Note that this is not the only problem related to assimp on s390, see #4788

Scottcjn:
## Summary

Fixes #5509. Binary STL is rejected on big-endian hosts with `Failed to determine STL storage representation`.

Binary STL stores the facet count, the geometry floats and the per-facet attribute field little-endian on disk. `STLLoader.cpp` read them host-endian, so on a big-endian host the facet count decoded to garbage, the `84 + faceCount*50 == fileSize` check in `IsBinarySTL()` failed, and the file fell through to the unknown representation error.

## Fix

Byte-swap the on-disk values with `AI_SWAP4` / `AI_SWAP2` from `assimp/ByteSwapper.h`:

- the facet count, in both `IsBinarySTL()` and `LoadBinaryFile()`
- the normal and the three vertex float triples per facet
- the 16 bit attribute field

`STLExporter.cpp` already uses `AI_SWAP4` to write these same fields, so only the read side was missing. The macros expand to nothing on little-endian builds, so there is no behaviour change there.

The four near identical read and convert blocks in the facet loop are folded into a `ReadBinaryVec3` helper. That is what SonarCloud was flagging as duplicated new code, and it leaves the loop shorter than it was before the patch.

## Test

`importBinarySTLFromFileTest` imports the existing `Spider_binary.stl` and checks the facet and vertex counts plus the first facet's normal and vertex, so both the count path and the per float geometry path are covered.

## Verification

x86_64 against current master: configure and build clean, all 8 `utSTLImporterExporter` tests pass.

Big-endian behaviour was reproduced on x86_64, not on real hardware. Forcing the `AI_SWAP*` macros to be active on a little-endian build produces the same single spurious swap that unpatched code performs on a big-endian host. In that configuration the import fails with exactly the error from #5509:

```
scene=NULL
error="Failed to determine STL storage representation for .../Spider_binary.stl."
```

With the macros in their normal state the same file loads. Reading the file's count field big-endian gives 1476722688 rather than 1368, which is what breaks the size check.

I do not have a big-endian machine or an s390x cross toolchain available, so this has not been run on real big-endian hardware.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved binary STL file handling for reliable vector, facet count, and color data reading across different system architectures.
  * Enhanced detection and import of binary STL files.

* **Tests**
  * Added validation for successful binary STL imports, including mesh structure, vertex and face counts, positions, and normals.


Scottcjn:
The SonarCloud "25.5% Duplication on New Code" failure is on the regression test in the new big-endian STL test cases — the cases mirror each other (LE input + expected output vs BE input + expected output) for readability. That's intentional duplication for test clarity, not production-code duplication.

Happy to refactor into a parameterized helper if maintainers prefer the lower duplication metric, but the current shape was chosen so each test case stands alone when debugging endianness regressions. Let me know your preference.

Scottcjn:
Rebased onto current master and pushed a rewritten version. Sorry for the long gap.

I need to correct something I said in May. I claimed the SonarCloud duplication was in the tests, mirrored little-endian and big-endian cases kept separate for readability. That was wrong. There are no such test cases, and the duplication was in production code: four near identical read and convert blocks in the facet loop in `LoadBinaryFile`. I should have looked properly before answering.

Those four blocks are now a single `ReadBinaryVec3` helper, so the duplication is gone rather than argued about, and `STLLoader.cpp` ends up shorter than it was before the patch. The now unused `theVec3F` local went with it.

I also corrected the verification claim in the description. It previously said I had cross-compiled for s390x and run it under qemu. I cannot substantiate that and I do not have an s390x toolchain, so I have replaced it with what I actually did:

- x86_64 against current master, clean build, all 8 `utSTLImporterExporter` tests pass.
- Big-endian reproduced on x86_64 by forcing the `AI_SWAP*` macros active on a little-endian build, which is the same single spurious swap unpatched code performs on a big-endian host. That configuration fails with the exact error from #5509, `Failed to determine STL storage representation`, and loads fine with the macros in their normal state.

I also checked the new test actually has teeth rather than passing by accident: with the swaps forced on, `importBinarySTLFromFileTest` fails, so a byte-swap regression cannot slip through it.

The expected values in the test are read straight out of `Spider_binary.stl`: 1368 facets, 4104 vertices, and the first facet's normal and vertex. Decoding the count big-endian gives 1476722688, which is what fails the `84 + faceCount*50 == fileSize` check.

@kimkulling no rush, but this should be in a reviewable state now if you have time.

Scottcjn:
The undefined-behavior-sanitizer job went red on my last push and it was my fault. Fixed in the commit above.

`STLLoader.cpp:405: load of misaligned address ... for type 'aiVector3f *', which requires 4 byte alignment`

A binary STL facet record is 50 bytes, so every other facet starts on an odd offset. The helper I added took the cursor as `aiVector3f *`, and the loop cast the byte pointer to that type before calling it. Forming a pointer to a misaligned address is undefined behaviour under `-fsanitize=alignment` on its own, before the `memcpy` through it ever runs, so the memcpy was not enough to make it safe. The cursor is now a plain `const unsigned char *` and the typed pointer is gone entirely, which also removes the cast and a round trip back out of it.

On verification, so this is not another claim you have to take on trust. I built this branch locally with `-fsanitize=undefined -fno-sanitize-recover=all` and ran the suite under `UBSAN_OPTIONS=halt_on_error=1`: 607 tests pass, 8 of them STL. But I also reverted just this file to the pre-fix state, rebuilt, and ran it again, and **it passed locally too**. So my machine does not reproduce the failure your CI saw, and I cannot honestly say I have demonstrated the fix against a reproduction. What I can say is that the pre-fix code definitely formed a misaligned `aiVector3f *`, your sanitizer caught it with a concrete address, and the new code cannot form one at all. Your CI is the real verdict here, not my local run.

Sorry for the noise on a PR I had just said was ready.

laurentketterle-hub:
@mainteneur this PR is ready to merge, CI green

laurentketterle-hub:
@mainteneur this PR is ready to merge, CI green

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6671, opened 2026-05-29
Add import properties support and update VectorKey structure in PyAssimp

Estradel:
The VectorKey struct in PyAssimp was missing a "`mInterpolation`" field so I added it because I was getting nonsense values in my animation channels. So far so good, it seems it has fixed my issue.

Secondly, I don't want to deal with multiple nodes (such as `_$AssimpFbx$_Translation`, `_$AssimpFbx$_PreRotation`, `_$AssimpFbx$_Rotation`) for a same node / bone of a FBX file. So I wanted to set the `IMPORT_FBX_PRESERVE_PIVOTS` property to false in the PyAssimp importer. As a result, I tried to come up with a solution to set custom importer properties when the `pyassimp.load()` function is used. I asked Gemini to help me figure out how to do it. So let me know if there's anything wrong or bad with this solution for any reasons I'm not aware of. I think it could be useful for others to be able to set these properties.

## Summary by CodeRabbit

* **New Features**
  * File loading now accepts optional import properties, including Boolean, integer, decimal, and text values, to customize how assets are imported.
  * Animation keyframe data now includes interpolation information, supporting smoother and more accurate animation playback.
  * Asset import behavior remains unchanged when no additional properties are provided, preserving existing loading workflows.

### DIFF
--- a/port/PyAssimp/pyassimp/core.py
+++ b/port/PyAssimp/pyassimp/core.py
@@ -35,7 +35,7 @@ class AssimpLib(object):
     """
     Assimp-Singleton
     """
-    load, load_mem, export, export_blob, release, dll = helper.search_library()
+    load, load_mem, export, export_blob, release, dll, create_store, set_prop_int, set_prop_float, set_prop_string, import_props, release_store = helper.search_library()
 _assimp_lib = AssimpLib()
 
 def make_tuple(ai_obj, type = None):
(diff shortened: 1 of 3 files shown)


kimkulling (approved):
I will merge it.

kimkulling:
You need to update your PR.

PR #6679, opened 2026-06-05
fix off-by-one face index check in lwo importers

alhudz:
ASAN on a crafted lwo2 file (1 point, one polygon indexing vertex 1):

    READ of size 12 at ... heap-buffer-overflow
      #1 Assimp::LWOImporter::InternReadFile LWOLoader.cpp:341
      0x... is located 0 bytes after 12-byte region

the clamp uses `index > mTempPoints.size()`, so an index equal to the point count slips through and `mTempPoints[index]` reads one past the end. same bug in the legacy LWOB path. changed both to `>=`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bounds checking for face vertex indices in LWO 3D model files, fixing off-by-one errors that could cause geometry loading issues.

### DIFF
--- a/code/AssetLib/LWO/LWOBLoader.cpp
+++ b/code/AssetLib/LWO/LWOBLoader.cpp
@@ -171,7 +171,7 @@ void LWOImporter::CopyFaceIndicesLWOB(FaceList::iterator &it,
                 uint16_t index;
                 ::memcpy(&index, cursor++, 2);
                 mi = index;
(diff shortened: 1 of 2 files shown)


kimkulling (approved):
Looks fine for me.

kimkulling:

alhudz:
Looking forward to contributing more. Thank you

dxbjavid:
Glad to be contributing alongside the team. Thank you for the merge.

PR #6680, opened 2026-06-06
Harden aiGetMaterialString against malformed serialized string proper…

netliomax25-code:
## Summary

Material string properties are deserialized from asset files and therefore originate from untrusted input. The previous implementation of `aiGetMaterialString()` relied primarily on assertions and extracted the serialized string length using `reinterpret_cast<uint32_t*>` on a byte buffer.

This change hardens material string parsing by introducing runtime validation of serialized metadata and replacing the unsafe length extraction with a `memcpy`-based read.

### MaterialSystem.cpp

* Validate minimum serialized string size before parsing.
* Validate declared string length against stored buffer size.
* Enforce `AI_MAXLEN` bounds.
* Verify null termination of serialized string payloads.
* Replace `reinterpret_cast<uint32_t*>` length extraction with `std::memcpy`.

### Regression Tests

Added coverage for malformed material string properties including:

* Truncated serialized buffers.
* Declared length larger than available payload.
* Missing null terminator.
* Randomized malformed inputs.
* Valid control cases to verify behavior remains unchanged.

## Security Rationale

Material properties may be loaded from malformed or attacker-controlled asset files.

Without runtime validation, inconsistent serialized metadata can lead to incorrect parsing behavior. In addition, reading a `uint32_t` directly from a byte buffer via `reinterpret_cast` may result in undefined behavior on architectures that require aligned access and can violate strict aliasing assumptions.

This patch improves robustness against malformed input and removes an error-prone parsing pattern while preserving behavior for valid material data.

## Compatibility

No behavioral changes for valid material string properties.

Malformed or inconsistent serialized string data now returns `AI_FAILURE` instead of relying on assertions or proceeding with invalid metadata.

## Validation

* MaterialSystem source compiled successfully.
* MaterialSystem unit tests compiled successfully.
* MaterialSystem test suite passed locally.
* Added regression coverage for malformed serialized string inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened material property string deserialization to safely handle malformed asset data by validating declared lengths, buffer layout, and null-termination before reading values.

* **Tests**
  * Added unit tests for multiple invalid string-property scenarios (length mismatch, truncation, oversized length, missing null-terminator) and a deterministic randomized validation check to confirm correct success/failure outcomes.


netliomax25-code:
Pushed a small follow-up. The fuzz test already uses a unique key per iteration (the earlier review note on reusing "testFuzz"), and I swapped the manual strlen/memcpy in the test helper for aiString::Set, which clears the SonarCloud strlen hotspot that was failing the quality gate. The hardening in aiGetMaterialString itself is unchanged.

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6682, opened 2026-06-08
fix oob read on md2 files with zero vertex count

alhudz:
Repro: load an `.md2` whose header sets `numVertices` to `0` while `numTriangles` is non-zero (`test/models/MD2/malformed_zero_numvertices.md2`).
Cause: `ValidateHeader` only rejects a zero frame count, so the per-face index clamp `iIndex = numVertices - 1` in `InternReadFile` underflows to `0xFFFFFFFF` and `pcVerts[iIndex]` is read far out of bounds.
Fix: reject `numVertices == 0` in `ValidateHeader`, next to the existing `numFrames == 0` check (this also removes the matching underflow in the `frameSize` calculation).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced 3D model file validation to detect and reject malformed files with zero vertices before processing, preventing potential crashes or import failures.

* **Tests**
  * Added test coverage to verify the import system correctly handles and rejects malformed model files.


kimkulling (approved):
.Looks fine for me.

kimkulling:

alhudz:
Thanks for the review and merge!

PR #6685, opened 2026-06-09
bound md5 weight index in remaining mesh loops

alhudz:
Repro: load any `.md5mesh` whose `vert` line sets `firstWeight` past the parsed `numweights` (e.g. `vert 0 ( 0.1875 -0.25 ) 1000000 2`).
Cause: `LoadMD5MeshFile` walks each vertex's weights as `w` in `[firstWeight, firstWeight+numWeights)` and reads `meshSrc.mWeights[w]`, but only the third loop bounds `w` against `mWeights.size()`. The weight-count loop and the weight-sum loop have no such check, so an out-of-range index is read straight off the end of the array.
Fix: add the existing `w >= mWeights.size()` guard to the two unguarded loops. Both `firstWeight` and `numWeights` come unvalidated from the file text. Regression model + test included; ASAN flagged a heap read at `MD5Loader.cpp:422` before the change, clean after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MD5 mesh loading to safely reject invalid weight references and malformed vertex or face data.
  * Added validation for MD5 camera cuts that reference invalid frame ranges.
  * Improved handling of large or unusually sized MD5 mesh data to prevent allocation and indexing errors.
* **Tests**
  * Added coverage for malformed MD5 weight data and confirmed invalid meshes are rejected during validation.


alhudz:
The sanitizer job was catching a leak from my own test: piCount was a raw new[] freed only at the end of the loop, so the throw on an invalid weight index leaked it. Switched it to a std::vector so it frees on every exit path. The malformed mesh is still rejected and the five utMD5Importer tests pass locally.

alhudz:
Rebased onto master to clear the conflict. #6700 landed the guards on the counting and apply passes in the meantime, so this now comes down to the one loop that's still unbounded: the weight-sum pass around `MD5Loader.cpp:484`. Same `w >= mWeights.size()` check as the apply loop just below it. Regression model and `importMalformedWeightIndex` kept, and the malformed mesh is still rejected with `MD5MESH: Invalid weight index`.

alhudz:
Rebased again to clear the conflict with the new MD5 overflow tests on master. Still just the one guard on the weight-sum loop; the malformed mesh is still rejected with `MD5MESH: Invalid weight index` and all ten `utMD5Importer` tests pass locally.

kimkulling:

PR #6689, opened 2026-06-09
FVFData support for multiple UV Sets

WaaghMan:
Added FVFData support for up to 3 additional UV Sets.

Attached is a sample .X model file with 3 UV Sets (one via standard means, the other 2 via FVFData)

[CarModel.zip](https://github.com/user-attachments/files/28758874/CarModel.zip)

## Summary by CodeRabbit

* **Bug Fixes**
  * X model parsing now detects and processes FVF data blocks inside mesh geometry.
  * Additional UV channels from those blocks are imported and populate extra texture-coordinate sets so multi-UV assets display correctly.
  * Added bounds checks to prevent exceeding supported texture-coordinate limits, avoiding corrupted or dropped UV data.

### DIFF
--- a/code/AssetLib/X/XFileParser.cpp
+++ b/code/AssetLib/X/XFileParser.cpp
@@ -573,16 +573,8 @@ void XFileParser::ParseDataObjectMeshFVFData(Mesh *pMesh) {
     // Number of total values in FVF Data. 2 x Vertices x UV sets
     unsigned int numCoords = ReadInt();
     
-    // Number of additional UV Channels in FVFData
-    unsigned int channels = 0;
-    if ((flags & 0x100) != 0) // D3DFVF_TEX1
-        channels++;
-
-    if ((flags & 0x200) != 0) // D3DFVF_TEX2
-        channels++;
-
-    if ((flags & 0x400) != 0) // D3DFVF_TEX3
-        channels++;
+    // Number of additional UV Channels in FVFData (D3DFVF_TEX1 | D3DFVF_TEX2 | D3DFVF_TEX3)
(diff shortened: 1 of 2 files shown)


code/AssetLib/X/XFileParser.cpp
```diff
@@ -562,6 +564,57 @@ void XFileParser::ParseDataObjectMeshTextureCoords(Mesh *pMesh) {
...
+    }
+    for (unsigned int a = 0; a < numCoords / (2 * channels); a++)
+    {
+        for (unsigned int uvSet = 0; uvSet < channels; uvSet++) 
+        {
+            std::vector<aiVector2D> &coords = pMesh->mTexCoords[baseUvSetCount + uvSet];
+
+            // They're saved as ints but they're actually floats.
+            unsigned int val = ReadInt();
+            float x = reinterpret_cast<float &>(val);
```
tellypresence: That doesn't look correct, should be able to replace with e.g.
`auto x = static_cast<float>(val);`

(using `auto` avoids repeating the cast type, which clang may squawk about)

tellypresence:
@WaaghMan Thank you for raising the PR, always the surest way to make sure things get done.

I have found that codeRabbit comments are very useful; would suggest please going through and addressing all of the codeRabbit warnings as this will increase code quality/robustness -- thank you again for contributing

WaaghMan:
Handled CodeRabbit suggestions and changed the cast to your suggestion (It's been a while since I touched C++, so I used the same cast I saw elsewhere).

Tested the code with our assets and it's working fine.

kimkulling (approved):
Looks really nice! Thanks for your work.

PR #6698, opened 2026-06-13
bound skin weight vertex index in x importer

alhudz:
1. `weight.mVertex` is read straight from the file in `XFileParser::ParseDataObjectSkinWeights` and never range-checked.
2. in `XFileImporter::CreateMeshes`, `oldWeights` is sized `sourceMesh->mPositions.size()`, but the loop writes `oldWeights[obone.mWeights[d].mVertex]`, so a vertex index past the position count is an out-of-bounds heap write. The face-vertex path just above already guards against `mPositions.size()`; the bone-weight path did not.

`repro`: a `.x` mesh with a `SkinWeights` block whose `vertexIndices` holds a value `>=` the position count (added model sets `1000000` on a 24-vertex cube).
`expected`: the out-of-range weight is skipped and the file imports.
`actual`: ASAN `heap-buffer-overflow` WRITE at `XFileImporter.cpp:373`.

The guard reverted in #5332 bounded by `obone.mWeights.size()` (the weight count, unrelated to the array) and dropped valid weights; this bounds by `oldWeights.size()` (the actual array length) instead, so only genuinely out-of-range indices are skipped. Added a malformed model and a regression test that crashes without the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed X model importer crash when processing files with malformed bone-weight vertex indices by implementing proper bounds checking during the weight remapping process.

* **Tests**
  * Added test fixture containing intentionally malformed skin-weight data to validate the importer's improved robustness when handling files with invalid vertex index references.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6704, opened 2026-06-17
bound texcoord index against face index count in ply loader

alhudz:
Repro: import a `.ply` whose face `texcoord` list is longer than twice its `vertex_indices` list (1 index, 6 texcoords reproduces it). ASAN reports a `heap-buffer-overflow` READ at `PlyLoader.cpp:627`.
Cause: the UV loop in `LoadFace` reads `mFaces[pos].mIndices[a / 2]` with `a` bounded by the texcoord list length, which is independent of `mNumIndices`, so `a / 2` walks off the end of the face index array.
Fix: stop the loop once `a / 2` reaches `mNumIndices`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential out-of-bounds indexing vulnerability when importing PLY files with malformed face texture coordinates.

* **Tests**
  * Added test coverage for handling PLY files with out-of-bounds texture coordinate data.


alhudz:
any update?

PR #6706, opened 2026-06-19
bound edge index in ndo face edge walk

alhudz:
Repro: load an `.ndo` whose edge points its next-edge link past the edge table; ASAN reports a heap read at `NDOLoader.cpp:103` while walking the face edge loop.
Cause: `ProcessFaceEdgesAndVertices` reads `obj.edges[cur_edge]` and follows the file-controlled next-edge link (`edge[4]`/`edge[5]`) with no bound on `cur_edge`. The start edge comes from `face_table` so it is valid, but the link is unvalidated, so the next iteration indexes `obj.edges` out of range. The sibling vertex read in the same loop is already guarded; the edge read was not.
Fix: bound `cur_edge` against `obj.edges.size()` before dereferencing and break out of the walk on an invalid index.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential crash when importing NDO files with malformed edge data by adding validation to prevent invalid memory access.

* **Tests**
  * Added unit tests for NDO import/export functionality to ensure robustness against edge-related malformations.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6709, opened 2026-06-20
bound 3mf property and vertex indices in xmlserializer

alhudz:
Repro: load a `.3mf` whose triangle carries a `pid` pointing at a colour/texture/material group and a `p1`/`p2`/`p3` value past the end of that group (`malformed_property_index_oob.3mf`).
Cause: `ReadObject` and `ImportTriangles` take the file `pindex`/`p1..p3` and index the resource group vectors (`mMaterialIndex`, `mTex2dCoords`, `mColors`) directly, with no bound. The per-triangle texture/colour assignment also indexes `mTextureCoords[0]`/`mColors[0]` (sized `mNumVertices`) with the file vertex index `face.mIndices[idx]`.
Expected: out-of-range property/vertex indices are ignored.
Actual: heap out-of-bounds read on the group vector (ASAN `BUS`/`READ` at `XmlSerializer.cpp:524`), and an out-of-bounds write on the per-vertex arrays when the triangle vertex index exceeds the vertex count.
Fix: bound each property index against its group size and each vertex index against `mNumVertices` before the access. Regression model + test added.

## Bug Fixes
* Improved import stability for 3MF 3D model files through enhanced validation of mesh attributes and resource indices.
* Added protective checks to prevent crashes when processing 3D models with invalid or out-of-bounds property references.
* Increased robustness when handling malformed or corrupted 3D model files during import operations.


kimkulling (approved):
Looks fine.

kimkulling:

PR #6715, opened 2026-06-26
Collada: read &lt;input&gt; offset without aliasing into size_t (fixes big-endian import)

### LINKED ISSUE #4788: Bug: Open simple COLLADA file on BigEndian architecture fails (s390x or ppc64)
**Describe the bug**
Loading the COLLADA example from [1.5 documentation (Appendix A)](https://www.khronos.org/files/collada_spec_1_5.pdf) fails on s390x architecture with the following error description: `Expected different index count in <p> element.`

It works on any other architectures like amd64 or arm64.

**To Reproduce**
Build and run the program at: https://salsa.debian.org/mzf/assimp-collada-test

**Expected behavior**
The importer loads the given COLLADA file without error, like on amd64 architecture.

**Screenshots**
N/A

 - OS: Linux Debian sid
 - Version: Assimp version 5.2.5

**Additional context**
s390x is a Big Endian architecture, which may be the root of the issue. I can do additional check on this architecture to ease debugging, do not hesitate to ask.

This is the forwarding of the Debian issue [1023212](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1023212)

Thanks!

Scottcjn:
Fixes #4788.

`ColladaParser::ReadInputChannel` read the per-index `<input>` `offset` attribute by casting `channel.mOffset` (a `size_t`) to `unsigned int&` and writing the 4-byte value through it:

```cpp
XmlParser::getUIntAttribute(node, "offset", (unsigned int &)channel.mOffset);
```

On a 64-bit big-endian host the 4-byte write lands in the high word of the 8-byte `size_t`, so `offset="1"` is read back as `4294967296`. `numOffsets` in `ReadPrimitives` then explodes and the import fails with a `std::bad_alloc` or the `Expected different index count in <p> element` error reported in the issue. (The cast is also undefined behavior in general; it only happens to work on little-endian.)

This reads the attribute into a real `unsigned int` and assigns it to `mOffset`, which is correct on both endiannesses.

### Testing

On real POWER8 hardware, importing the bundled `test/models/Collada/cube_triangulate.dae` (Collada-importer build):

- **ppc64 big-endian** (gcc 13.3): fails before this change (`std::bad_alloc`); imports correctly after (1 mesh, 12 faces).
- **ppc64le** (gcc 10) and **x86_64**: imports `1 mesh, 12 faces` both before and after — unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when importing COLLADA files with input channel offsets, helping prevent incorrect data handling during import.


kimkulling (approved):
Looks fine for me.

kimkulling:
Good catch, thanks for your help here. I am wondering why our sanitizer has not found this issue.

kimkulling:

PR #6717, opened 2026-06-28
Fix heap-buffer-overflow in LWO2 variable-sized index reader

Alb3e3:
## Motivation

`LWOImporter::ReadVSizedIntLWO2()` (in `code/AssetLib/LWO/LWOLoader.h`) decodes a variable-sized index by reading 2 bytes — or 4 bytes when the first byte is `0xFF` — and advancing the file cursor, **without ever checking the cursor against the end of the file buffer**.

A malformed LWO2 file can declare a polygon whose index count (`numIndices`, up to `0x3FF`) runs past the end of the chunk. `CountVertsAndFacesLWO2()` then calls `ReadVSizedIntLWO2()` repeatedly, walking the cursor off the end of the heap allocation and producing an out-of-bounds read.

The helper is unguarded at all of its call sites (`CountVertsAndFacesLWO2`, `CopyFaceIndicesLWO2`, the VMAP/VMPA paths, and the envelope readers), so the same overflow is reachable from several chunk types.

### AddressSanitizer report (before the fix)

```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
  #0 Assimp::LWOImporter::ReadVSizedIntLWO2(unsigned char*&)   LWOLoader.h:457
  #1 Assimp::LWOImporter::CountVertsAndFacesLWO2(...)          LWOLoader.cpp:817
  #2 Assimp::LWOImporter::LoadLWO2Polygons(unsigned int)       LWOLoader.cpp:792
  #3 Assimp::LWOImporter::LoadLWO2File()                       LWOLoader.cpp:1558
  #4 Assimp::LWOImporter::InternReadFile(...)                  LWOLoader.cpp:195
```

## Implementation

Add a single bounds check inside `ReadVSizedIntLWO2()`: each byte read is now guarded against `mFileBufferEnd` (the same end pointer the loader already uses elsewhere, e.g. in `GetS0`), and a truncated index raises `DeadlyImportError`. Because every call site funnels through this one helper, the single guard closes all of them.

The body was also flattened slightly (a small `readByte()` lambda) so the bounds check isn't duplicated six times. `#include <assimp/Exceptional.h>` is added to the header so `DeadlyImportError` is available where the inline function is defined.

## Impact

- Well-formed LWO/LWO2 files are unaffected — the existing bundled `test/models/LWO` corpus still imports identically.
- Malformed/truncated files now fail with a clean `DeadlyImportError` instead of reading out of bounds.

## How it was found / verified

Found with AddressSanitizer + libFuzzer, fuzzing the LWO importer in isolation. After the fix the original reproducer no longer triggers ASan, the bundled LWO test models still parse, and ~80s of additional coverage-guided fuzzing surfaced no further crash in this path.

## AI tool disclosure

Per the project's [AI Tool Use Policy](https://github.com/assimp/assimp/blob/master/AIToolPolicy.md): this fix was prepared with AI assistance (fuzzing harness setup and an initial draft of the guard). I reviewed and verified the change myself — the root-cause analysis, the bounds logic, and the before/after testing are mine, and I'm happy to answer questions during review. The commit carries an `Assisted-by:` trailer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness when importing LWO2 files by adding guarded bounds checks during variable-sized index parsing.
  * Avoided potential crashes from truncated or malformed inputs that could previously cause out-of-bounds reads.
  * When the file ends unexpectedly, the importer now reports a clearer, LWO2-specific error instead of risking invalid memory access.

### DIFF
--- a/code/AssetLib/LWO/LWOLoader.h
+++ b/code/AssetLib/LWO/LWOLoader.h
@@ -290,7 +290,6 @@ private:
     /// @param idx Absolute index of the first vertex
     /// @param data Value of the VMAP to be assigned - read numRead
     ///        floats from this array.
-    */
     void DoRecursiveVMAPAssignment(VMapEntry *base, unsigned int numRead,
             unsigned int idx, float *data);
 
--- a/code/AssetLib/LWO/LWOLoader.h
+++ b/code/AssetLib/LWO/LWOLoader.h
@@ -46,6 +46,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "LWOFileData.h"
 #include <assimp/BaseImporter.h>
+#include <assimp/Exceptional.h>
 #include <assimp/material.h>
 #include <assimp/DefaultLogger.hpp>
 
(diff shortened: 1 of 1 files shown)


kimkulling (approved):
Looks fine.

Alb3e3:
Thanks for the review and the comment cleanup! Small heads-up: the `/** … */` → `///` conversion left a stray `*/` at `LWOLoader.h:293` (just above `DoRecursiveVMAPAssignment`). A bare `*/` is a syntax error, so it was failing the build across all platforms. I've pushed a one-line commit removing the orphaned terminator — CI should go green now. Verified locally with a clean Release build of the library on this branch.

PR #6718, opened 2026-06-28
Fix heap-buffer-overflow write in MD5 MakeDataUnique

Alb3e3:
## Motivation

`MD5Importer::MakeDataUnique()` (`code/AssetLib/MD5/MD5Loader.cpp`) duplicates shared vertices so each triangle owns its three vertices. It sized the output array to `faces * 3` but started the append cursor `iNewIndex` at the *original* vertex count:

```cpp
const unsigned int iNewNum   = meshSrc.mFaces.size() * 3;
unsigned int       iNewIndex = meshSrc.mVertices.size();   // original count
meshSrc.mVertices.resize(iNewNum);                         // only faces*3
...
meshSrc.mVertices[iNewIndex] = meshSrc.mVertices[face.mIndices[i]];  // append
```

The implicit assumption is `original_verts + duplicates == faces*3`, which only holds when every declared vertex is referenced and there are at least as many triangle corners as vertices. A malformed `.md5mesh` that declares **more vertices than triangle corners** (or leaves vertices unreferenced) makes `iNewIndex` start at/above the resized size, so the first duplicate write lands past the end of the heap allocation.

### AddressSanitizer report (before the fix)

```
ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 16
  #0 __asan_memcpy
  #1 Assimp::MD5Importer::MakeDataUnique(Assimp::MD5::MeshDesc&)  MD5Loader.cpp:251
  #2 Assimp::MD5Importer::LoadMD5MeshFile()                      MD5Loader.cpp:426
  #3 Assimp::MD5Importer::InternReadFile(...)                    MD5Loader.cpp:165
```

## Implementation

- Size the vertex array to the true worst case — `numOrigVerts + faces*3` (every corner could append one duplicate) — and `resize()` back down to the number actually produced (`iNewIndex`) at the end. Well-formed meshes are trimmed to exactly the same count as before, so their output is unchanged.
- Validate face indices against **`numOrigVerts`** (the size `abHad` is allocated with) rather than the resized array. The previous check used the post-resize size, which could let an index in `[numOrigVerts, faces*3)` through and then read `abHad` out of bounds — this closes that too.
- Guard the weights-per-vertex estimate against a zero original vertex count (avoids a divide-by-zero / runaway `reserve`).

## Impact

- Well-formed MD5 files import identically — verified against the bundled `test/models/MD5` corpus.
- Malformed files now either fail with a clean `DeadlyImportError` (bad index) or import within bounds instead of corrupting the heap.

## How it was found / verified

AddressSanitizer + libFuzzer on the MD5 importer. After the fix the reproducer no longer triggers ASan, the bundled MD5 models still import, and a follow-up coverage-guided run found no further overflow in this path.

## AI tool disclosure

Per the [AI Tool Use Policy](https://github.com/assimp/assimp/blob/master/AIToolPolicy.md): prepared with AI assistance (fuzzing + an initial draft). I reviewed and verified the analysis and the fix myself and can answer questions during review; the commit carries an `Assisted-by:` trailer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes when importing certain MD5 models that require vertex duplication.
  * Improved validation and memory safety during MD5 imports, preventing out-of-bounds access and handling vertex and face data more reliably.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6723, opened 2026-07-03
bound cut frame indices in MD5 camera loader

zayeem06:
LoadMD5CameraFile indexes the parsed frame array with cut values read straight from the .md5camera file, so a cut that points past the last frame walks past the vector and reads out of bounds while building the animation keys; validate each cut range against the frame count before the read. A crafted file is added under test/models/MD5/invalid with a regression test that now rejects it instead of over-reading.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation of MD5 camera cut ranges to prevent out-of-bounds frame access during camera animation import.
  * Updated camera keyframe construction to use validated frame bounds and keep position/rotation key timestamps synchronized.
* **Tests**
  * Added an invalid MD5 camera-cut fixture and a unit test confirming the importer rejects the file cleanly (returns no scene).

### DIFF
--- a/code/AssetLib/MD5/MD5Loader.cpp
+++ b/code/AssetLib/MD5/MD5Loader.cpp
@@ -741,13 +741,21 @@ void MD5Importer::LoadMD5CameraFile() {
         aiNodeAnim *nd = anim->mChannels[0] = new aiNodeAnim();
         nd->mNodeName.Set("<MD5Camera>");
 
-        nd->mNumPositionKeys = nd->mNumRotationKeys = *(it + 1) - (*it);
+        const unsigned int firstFrame = *it;
+        const unsigned int lastFrame = *(it + 1);
+        // Cut indices come straight from the file and are used to index into
+        // frames; reject ranges that run past the end (or wrap) before use.
+        if (lastFrame < firstFrame || lastFrame > frames.size()) {
+            throw DeadlyImportError("MD5CAMERA: Cut references a frame out of range");
+        }
+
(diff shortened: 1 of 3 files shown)


zayeem06:
Good catch on the mid-loop throw. Since `mAnimations` is allocated before the build loop, a later cut throwing would leave trailing slots uninitialized and the scene destructor would delete garbage pointers. Moved the range check into a pre-pass over the cuts so it runs before any allocation, which means no half-built scene can exist. Rebuilt and the MD5 tests (including the invalid-cut regression) still pass.

code/AssetLib/MD5/MD5Loader.cpp
```diff
@@ -729,6 +729,16 @@ void MD5Importer::LoadMD5CameraFile() {
             cuts.push_back(static_cast<unsigned int>(frames.size() - 1));
     }
 
+    // Cut indices come straight from the file and are used to index into
+    // frames; reject any range that runs past the end (or wraps) before we
+    // allocate or read anything, so a mid-loop throw can't leave the scene
+    // holding half-initialized animations.
+    for (std::vector<unsigned int>::const_iterator it = cuts.begin(); it != cuts.end() - 1; ++it) {
```
turol: Use `auto` for the iterator type.

turol (changes requested):

code/AssetLib/MD5/MD5Loader.cpp
zayeem06: Done, switched it to auto.

kimkulling (approved):
Looks fine for me.

turol (approved):

PR #6730, opened 2026-07-08
bound uv point count to face size in sib ReadUVs

zayeem06:
**Heap overflow walking the UV index block in ReadUVs**

The FTVS chunk gives each face its own point count, which `ReadUVs` uses to step through that face's slice of `mesh->idx`. It never checks that count against the face's real point count (stored by `ReadFaces` at `mesh->idx[faceStart[faceIdx]]`), so a larger value walks `idx` past the face block and off the end of the vector, then feeds the garbage `idx[UV]` into `mesh->uv[id]`.

I kept the fix at the same layer as the existing `faceIdx` check since the count is read right there. Every other reader (ConnectFaces, CalculateNormals, ReadCreases) already uses the stored count, so ReadUVs was the only one trusting the file value.

Reproduced with a crafted .sib under ASAN (heap-buffer-overflow read at SIBImporter.cpp:273); with the guard it's rejected cleanly. Added a regression test plus the model; the existing heffalump.sib still imports unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of imported UV data to reject inconsistent face/UV counts instead of reading invalid mesh data.
  * Importing malformed SIB files now fails cleanly when UV information exceeds the available face size.
* **Tests**
  * Added coverage for an invalid SIB file case to confirm bad UV face data is rejected during import.


zayeem06:
Looks like #6755 already landed the same bounds check plus the leak cleanup on parse failure, so this one is redundant now. Happy to close it out.

PR #6734, opened 2026-07-15
check 32-bit overflow in MDC ValidateSurfaceHeader frame checks

zayeem06:
ValidateSurfaceHeader bounds-checks the surface frame-index tables with ulOffsetFrameBaseFrames + ulNumBaseFrames * 2, but that expression is evaluated in 32-bit so a large offset wraps below iMax and passes the check, unlike the sibling checks just above it that multiply by sizeof() and stay 64-bit. A crafted .mdc whose surface sets ulOffsetFrameBaseFrames near UINT_MAX then reaches InternReadFile, which reads an int16 at pcSurface + that offset (MDCLoader.cpp:294) far outside the file buffer. Multiplying by sizeof(int16_t) keeps the arithmetic in size_t so the check can no longer wrap; valid files are unaffected and a regression model plus test are included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of MDC model frame-data boundaries to prevent invalid files from being processed incorrectly.

* **Tests**
  * Added coverage confirming that MDC files with invalid frame offsets are rejected gracefully.


kimkulling (approved):
Looks fine for me.

zayeem06:
Thanks for the review. Branch is up to date with master and all checks are green, so it's ready whenever you want to merge it.

PR #6735, opened 2026-07-16
bound face texcoord count to vertex index count in ply LoadFace

zayeem06:
LoadFace reads the vertex_index list and the texcoord list of a face as two separate properties, but the texcoord loop indexes mIndices[a / 2] while a runs over the texcoord list length, and the file sets those two lengths independently. A face declaring one vertex index and six texcoord values makes the loop read past the end of the one-element mIndices array, which asan reports as a heap-buffer-overflow read at PlyLoader.cpp:627 against the allocation made a few lines above it; the stale value then picks which texture coordinate gets written. The check belongs here because this is the only point where both list lengths are known, so it rejects the face before the walk starts rather than after the read. Worth verifying that the bound allows quads, since (iNum / 3) == 2 also admits 7 and 8 coordinate values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed PLY face texture-coordinate data by adding validation to prevent cases with excessive per-face texcoord values.
  * Invalid face texcoord definitions are now rejected safely, reducing the risk of crashes or incorrect imports.

* **Tests**
  * Added new unit tests covering both failure and success scenarios for face texcoord parsing, including invalid texcoord list metadata and a valid quad face with matching texcoord data.

### DIFF
--- a/code/AssetLib/Ply/PlyLoader.cpp
+++ b/code/AssetLib/Ply/PlyLoader.cpp
@@ -623,6 +623,9 @@ namespace Assimp {
                     // should be 6 coords
                     auto p = GetProperty(instElement->alProperties, iTextureCoord).avList.begin();
                     if ((iNum / 3) == 2) { // X Y coord
+                        if (iNum > mGeneratedMesh->mFaces[pos].mNumIndices * 2) {
+                            throw DeadlyImportError("Invalid .ply file: Too many texture coordinates for face");
+                        }
                         for (unsigned int a = 0; a < iNum; ++a, ++p) {
                             unsigned int vindex = mGeneratedMesh->mFaces[pos].mIndices[a / 2];
                             if (vindex < mGeneratedMesh->mNumVertices) {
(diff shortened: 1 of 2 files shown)


zayeem06:
Added the quad regression case: 4 vertex indices with 8 texcoord values imports fine and keeps its texture coords, so the bound does allow quads. That was the open question in the description, so it's good to have it pinned down by a test. Full utPLYImportExport suite is green (16/16).

On hoisting the check above the `(iNum / 3) == 2` filter: I don't think that one holds. With 9 values `iNum / 3` is 3, so the loop never runs and nothing is read, meaning there's no out-of-bounds access to guard against there. Those lists are already ignored today, and moving the check out would start rejecting files that currently import cleanly, which is a behavior change wider than the memory safety issue I was fixing. Inside the branch the bound is exact: iNum is 6, 7 or 8, the highest index touched is (iNum - 1) / 2, and `iNum <= mNumIndices * 2` keeps that below mNumIndices.

Happy to widen it if you'd rather reject those malformed lists outright, but that felt like a separate call from the overflow fix.

zayeem06:
any update?

PR #6737, opened 2026-07-17
don't step past the terminator in TokenMatchI

zayeem06:
TokenMatchI advances with in += len + 1 once the token matches, and IsLineEnd counts '\0' as a line end, so a token sitting at the very end of the buffer matches and leaves the cursor one byte past the terminator. TextFileToBuffer reserves exactly size + 1 bytes, so that lands one byte past the heap allocation: a .csm ending in the token $rate walks off the end and CSMLoader.cpp:138 hands the stale pointer to fast_atoreal_move, which asan flags as a heap-buffer-overflow read 0 bytes after the buffer. MD3 and Unreal reach the same helper the same way. The sibling TokenMatch already handles this with an explicit in[len] != '\0' check, so I mirrored it here rather than fixing each caller.

Worth verifying that stopping on the terminator instead of skipping it is right for the callers: every one of them re-checks for '\0' or goes through a bounded SkipSpaces afterwards, so the extra byte was never being consumed for anything. The added tests cover both the EOF case and the ordinary trailing-space case, and the first one fails on master.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token parsing at the end of input to prevent advancing beyond the available buffer.
  * Preserved correct handling of whitespace and following numeric text during parsing.

* **Tests**
  * Added coverage for tokens ending at the input boundary.
  * Added coverage for tokens followed by spaces and numeric values.


zayeem06:
any update?

kimkulling (approved):
Looks great, approved.

kimkulling:

PR #6739, opened 2026-07-19
zero-initialize assbin bone, material and channel arrays before filling

zayeem06:
ReadBinaryMesh, ReadBinaryMaterial and ReadBinaryAnim read an element count out of the .assbin file, store it into mNumBones/mNumProperties/mNumChannels, and only then allocate the pointer array and fill it entry by entry. The per-entry readers throw on a bad chunk magic or on EOF, so a file that announces 64 entries and supplies none leaves the array uninitialized while the count already claims 64; unwinding destroys the half-built scene and ~aiMesh, aiMaterial::Clear and ~aiAnimation each walk the count and delete every slot, so operator delete runs on uninitialized heap bytes. Under ASAN a crafted file crashes in each of the three destructors with 0xbebebebebebebebe in the register holding the freed pointer.

ReadBinaryScene already memsets all six of its own pointer arrays for this reason and ReadBinaryNode bumps mNumChildren only after each child is read, so this just applies the existing idiom to the three nested arrays that were missed. Worth checking the fourth case, node->mChildren at line 262, which I left alone because its count is incremented per successful read and is never ahead of the array contents.

The regression test corrupts an exported blob in memory the same way the neighbouring rejectOversizedNodeNameLengthInAssbin test does; it segfaults in ~aiMaterial without the patch. Full unit suite is 602/602 both under ASAN and in a -Werror release build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading binary assets by initializing allocated data before use.
  * Strengthened handling of malformed binary files, including invalid size and count values.

* **Tests**
  * Added and refactored unit coverage for corrupted binary data to confirm imports fail safely with clear error messages.

### DIFF
--- a/code/AssetLib/Assbin/AssbinLoader.cpp
+++ b/code/AssetLib/Assbin/AssbinLoader.cpp
@@ -458,6 +458,7 @@ void AssbinImporter::ReadBinaryMesh(IOStream *stream, aiMesh *mesh) {
     // write bones
     if (mesh->mNumBones) {
         mesh->mBones = new C_STRUCT aiBone *[mesh->mNumBones];
+        memset(mesh->mBones, 0, mesh->mNumBones * sizeof(aiBone *));
         for (unsigned int a = 0; a < mesh->mNumBones; ++a) {
             mesh->mBones[a] = new aiBone();
             ReadBinaryBone(stream, mesh->mBones[a]);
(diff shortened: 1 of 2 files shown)


zayeem06:
Pushed a small follow-up: the two little-endian blob helpers were copy-pasted into both tests, which is what tripped the SonarCloud duplication gate. They're now file-scope statics in an anonymous namespace and both tests call them, so the new test only adds its own logic.

I kept the memset in the loader rather than switching to `new T*[n]()`. ReadBinaryScene right above already zeroes its six arrays that way, so this stays consistent with the surrounding code instead of introducing a second idiom in the same file. Happy to change it if you'd rather have the value-initialized form.

Still 602/602 locally, and the ASAN job is green.

zayeem06:
Renamed the two blob helpers to PascalCase to match the naming convention. I also pulled the export-spider-to-bytes boilerplate that both corruption tests were repeating into a shared ExportSpiderToAssbinBytes() helper, which should clear the SonarCloud duplication gate. Both regression tests still fail cleanly (String length too large / Magic chunk identifiers are wrong) and the Assbin suite is green locally.

Issue #3286, opened 2020-06-17
Multi-layer dxf file reading problem


Cheng1087:
When I use assimp to read a multi-layer dxf file, only some layers can be read. I can't get the curve data in the file.
As shown in the figure below, I can only read part of the yellow straight line, and cannot get the data of the curve and border.
I want to know if I am using it wrong or if assimp cannot get this data.
![企业微信截图_15923828966879](https://user-images.githubusercontent.com/48266038/84875667-db6c4b80-b0b8-11ea-8ca5-b72f61bfd7aa.png)
![diff](https://user-images.githubusercontent.com/48266038/84982491-6b69ce00-b169-11ea-8770-d414afe39923.png)

jubalh:
Is there any plan/roadmap for this?

tellypresence:
@Cheng1087 Is this still a problem?

tellypresence:
Tracked by epic #6097, if someone picks this up please reopen

PR #6741, opened 2026-07-21
Fix DXF polyline handling to respect vertex chain.

rolandhill:
This PR fixes DXF polylines to emit lines while respecting chained vertexes. Previously it would emit disjoint pairs (0,1), (2,3), ..., which drops every second segment and loses the final vertex when the vertex count is odd.

This could be a contributor to #3286.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DXF polyline importing by correctly connecting consecutive vertices into continuous line segments.
  * Updated handling of closed polylines to preserve the connection between the final and first vertices.


kimkulling (approved):
Looks fine, approved.

kimkulling:

PR #6744, opened 2026-07-22
fix off-by-one in ms3d group material index check

zayeem06:
The group material index read from a .ms3d file is rejected with `g.mat > pScene->mNumMaterials`, so a group whose index equals the material count slips through and lands in aiMesh::mMaterialIndex pointing one past the end of aiScene::mMaterials. Exporting such a scene reads that slot, e.g. PbrtExporter::WriteMesh, where asan reports a heap-buffer-overflow read right after the material pointer array. ValidateDataStructure already uses `>=` for this same field, so the loader is the odd one out; the check is corrected there and a crafted model plus a regression test are added under the existing MS3D model directory.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of MS3D model material references.
  * Invalid material indexes are now rejected instead of being accepted during model loading, preventing malformed models from loading incorrectly.

* **Tests**
  * Added coverage for valid MS3D imports.
  * Added regression coverage confirming models with invalid material indexes fail safely.


zayeem06:
any update?

PR #6747, opened 2026-07-27
Reject SIB faces whose point count exceeds the chunk data

### LINKED ISSUE #6712: Bug:SIB importer integer overflow in ReadFaces() leads to heap-buffer-overflow write
# Heap-buffer-overflow in SIB importer due to integer overflow in `numPoints * 3`

## Summary

A heap-buffer-overflow vulnerability exists in Assimp's SIB importer.

The issue is located in the SIB face parsing logic:

```text id="of2hbx"
code/AssetLib/SIB/SIBImporter.cpp
```

When parsing polygon face data from a `.sib` file, the importer reads a 32-bit unsigned integer named `numPoints`, which represents the number of points in a polygon. This value is fully controlled by the input file.

The importer later calculates the required index buffer size using:

```cpp id="2c23oy"
numPoints * N
```

where `N` is the number of index channels per point:

```cpp id="9usypc"
enum {
    POS,
    NRM,
    UV,
    N
};
```

Here, `N` is equal to `3`, representing position, normal, and texture-coordinate indices.

When `numPoints` is large enough, the multiplication overflows in 32-bit unsigned arithmetic. For example:

```text id="1vit3s"
numPoints = 0x55555556
N         = 3

0x55555556 * 3 = 0x100000002
```

After 32-bit truncation, the result becomes:

```text id="ko9xp5"
0x00000002
```

As a result, `std::vector::resize()` allocates a much smaller buffer than required. The subsequent loop still iterates using the original `numPoints` value and writes three indices per iteration, causing an out-of-bounds write into the heap buffer.

A crafted 44-byte `.sib` file is sufficient to trigger the issue.

## Affected Component

```text id="hku9md"
Assimp SIB importer
code/AssetLib/SIB/SIBImporter.cpp
```

The vulnerable function is:

```text id="k2p3rh"
ReadFaces()
```

## Vulnerability Type

```text id="74jcn7"
CWE-190: Integer Overflow or Wraparound
CWE-787: Out-of-bounds Write
CWE-122: Heap-based Buffer Overflow
```

The root cause is an unchecked integer multiplication used to calculate the vector resize size.

## Reproduction Environment

Tested environment:

```text id="s1cydl"
OS: Ubuntu 22.04.5 LTS x86_64
Kernel: Linux 6.8.0-124-generic
Compiler: g++ 11.4.0
CMake: 3.22+
Python: 3.10+
```

Build Assimp with AddressSanitizer enabled:

```bash id="lk377p"
git clone https://github.com/assimp/assimp.git --depth 1
cd assimp

mkdir -p build_asan
cd build_asan

cmake .. \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DASSIMP_BUILD_TESTS=OFF \
  -DBUILD_SHARED_LIBS=OFF \
  -DASSIMP_BUILD_ASSIMP_TOOLS=ON

make -j$(nproc) assimp assimp_cmd
```

## Proof of Concept

The following Python script generates a minimal malformed `.sib` file that triggers the vulnerability.

Create `gen_poc.py`:

```python id="m82c06"
#!/usr/bin/env python3
import os
import struct

def generate_poc():
    os.makedirs("poc", exist_ok=True)

    # SIB is a chunk-based binary format:
    # Tag:  4 bytes
    # Size: 4 bytes, little-endian
    # Data: Size bytes

    # VRTS sub-chunk: define one vertex
    # 3 floats = 12 bytes
    vrts_tag = b"VRTS"
    vrts_size = 12
    vrts_data = struct.pack("<fff", 0.0, 0.0, 0.0)
    vrts_chunk = vrts_tag + struct.pack("<I", vrts_size) + vrts_data

    # FACS sub-chunk: malformed face data
    # numPoints = 0x55555556
    # numPoints * 3 = 0x100000002
    # uint32_t truncation -> 0x00000002
    facs_tag = b"FACS"
    facs_size = 8
    numPoints = 0x55555556
    index = 0
    facs_data = struct.pack("<II", numPoints, index)
    facs_chunk = facs_tag + struct.pack("<I", facs_size) + facs_data

    # Top-level SHAP chunk
    shap_data = vrts_chunk + facs_chunk
    shap_tag = b"SHAP"

    with open("poc/overflow.sib", "wb") as f:
        f.write(shap_tag)
        f.write(struct.pack("<I", len(shap_data)))
        f.write(shap_data)

    print("PoC file generated: poc/overflow.sib")
    print(f"File size: {4 + 4 + len(shap_data)} bytes")
    print(f"numPoints = 0x{numPoints:08X} = {numPoints}")
    print(f"overflow check: 0x{numPoints:08X} * 3 = 0x{numPoints * 3:016X}")
    print(f"uint32_t truncation: 0x{(numPoints * 3) & 0xFFFFFFFF:08X}")

if __name__ == "__main__":
    generate_poc()
```

Generate the PoC file:

```bash id="qiptnm"
python3 gen_poc.py
```

Expected output:

```text id="yc63r8"
PoC file generated: poc/overflow.sib
File size: 44 bytes
numPoints = 0x55555556 = 1431655766
overflow check: 0x55555556 * 3 = 0x0000000100000002
uint32_t truncation: 0x00000002
```

Run Assimp on the generated file:

```bash id="eov5aw"
./build_asan/bin/assimpd info poc/overflow.sib
```

## Observed Result

AddressSanitizer reports a heap-buffer-overflow while parsing the crafted `.sib` file.

The write occurs in the SIB face parsing loop after the index buffer has been resized using the overflowed result of `numPoints * N`.

The vulnerable write path is:

```cpp id="o65c6e"
idx[POS] = p;
idx[NRM] = ptIdx;
idx[UV]  = ptIdx;
```

## Root Cause Analysis

The vulnerable logic is in `ReadFaces()`:

```cpp id="r6s8u6"
static void ReadFaces(SIBMesh *mesh, StreamReaderLE *stream) {
    uint32_t ptIdx = 0;

    while (stream->GetRemainingSizeToLimit() > 0) {
        uint32_t numPoints = stream->GetU4();

        size_t pos = mesh->idx.size() + 1;

        mesh->idx.resize(pos + numPoints * N);

        mesh->idx[pos - 1] = numPoints;
        uint32_t *idx = &mesh->idx[pos];

        mesh->faceStart.push_back(static_cast<uint32_t>(pos - 1));
        mesh->mtls.push_back(0);

        for (uint32_t n = 0; n < numPoints; n++, idx += N, ptIdx++) {
            uint32_t p = stream->GetU4();

            if (p >= mesh->pos.size()) {
                throw DeadlyImportError("Vertex index is out of range.");
            }

            idx[POS] = p;
            idx[NRM] = ptIdx;
            idx[UV] = ptIdx;
        }
    }
}
```

The multiplication:

```cpp id="aril7v"
numPoints * N
```

is performed before the result is used in `std::vector::resize()`.

Since `numPoints` is a 32-bit unsigned integer and `N` is `3`, a value such as `0x55555556` causes the multiplication to overflow and wrap around.

The vector is therefore resized to a much smaller size than required. The subsequent loop still uses the original `numPoints` and writes three `uint32_t` values per point, causing heap memory corruption.

## Expected Behavior

Assimp should reject malformed `.sib` files where `numPoints` is unreasonably large or where the multiplication required for index buffer allocation would overflow.

The importer should validate the face point count before using it in allocation or loop bounds.

## Actual Behavior

Assimp accepts the oversized `numPoints` value, performs an overflowing size calculation, allocates an undersized vector, and writes past the end of the allocated heap buffer.

## Impact

A crafted `.sib` file can cause Assimp, or an application using Assimp to import untrusted assets, to crash due to heap memory corruption.

The confirmed impact is denial of service through a heap-buffer-overflow write. Since the issue is an out-of-bounds write reachable through a crafted file parser input, it should be treated as a security-relevant memory safety issue.

## Suggested Fix

The importer should perform explicit overflow and sanity checks before resizing the index vector.

For example:

```cpp id="xvfw9x"
if (numPoints == 0) {
    throw DeadlyImportError("Invalid SIB face point count.");
}

if (numPoints > stream->GetRemainingSizeToLimit() / sizeof(uint32_t)) {
    throw DeadlyImportError("SIB face point count exceeds available data.");
}

if (numPoints > (std::numeric_limits<size_t>::max() - pos) / N) {
    throw DeadlyImportError("SIB face index buffer size overflow.");
}

mesh->idx.resize(pos + static_cast<size_t>(numPoints) * N);
```

It would also be safer to compute the required size in `size_t` or `uint64_t` and validate it before calling `resize()`:

```cpp id="xoyzil"
const size_t count = static_cast<size_t>(numPoints);
const size_t required = count * N;

if (count != 0 && required / count != N) {
    throw DeadlyImportError("SIB face index count overflow.");
}

if (required > std::numeric_limits<size_t>::max() - pos) {
    throw DeadlyImportError("SIB index vector size overflow.");
}

mesh->idx.resize(pos + required);
```

Additionally, the parser should reject face definitions whose declared `numPoints` cannot be satisfied by the remaining bytes in the `FACS` chunk.

## Additional Notes

The issue is triggered using Assimp's own command-line tool and does not require modifying Assimp source code or directly calling internal parser APIs.

The malformed `.sib` file is only 44 bytes and contains a `SHAP` chunk with one `VRTS` sub-chunk and one malformed `FACS` sub-chunk.

https://github.com/user-attachments/assets/2e3c5bb9-4423-4e82-9742-03952e5cb823

Alb3e3:
## Motivation

Fixes #6712. `ReadFaces()` in `code/AssetLib/SIB/SIBImporter.cpp` sizes the index array from a file-supplied point count:

```cpp
uint32_t numPoints = stream->GetU4();
size_t pos = mesh->idx.size() + 1;
mesh->idx.resize(pos + numPoints * N);   // N == 3
```

`numPoints` and `N` are both 32-bit, so `numPoints * N` wraps before it is widened to `size_t`. A face claiming `0x55555556` points multiplies to `0x100000002`, truncates to `2`, and the array is sized for two entries — while the loop below writes three entries per point:

```cpp
for (uint32_t n = 0; n < numPoints; n++, idx += N, ptIdx++) {
    idx[POS] = p;
    idx[NRM] = ptIdx;
    idx[UV] = ptIdx;
}
```

The very first iteration already writes past the end. A 48-byte file is enough; on a debug build the import aborts:

```
Info,  T0: Found a matching importer for this file format: Silo SIB Importer.
free(): invalid size
Aborted (core dumped)
```

## Implementation

Every point of a face is stored as exactly one 4-byte index, so a face can never have more points than the remaining chunk data can hold. Reject counts above that bound with a `DeadlyImportError` before sizing the array, and widen the multiplication to `size_t`.

The bound is exact rather than conservative — a legitimate face with `numPoints` points always has `4 * numPoints` bytes of index data left in the chunk — so no valid file is affected. Because the check also caps `numPoints` at a quarter of the chunk size, `numPoints * N` can no longer wrap on 32-bit platforms either.

## Impact

- Valid files import unchanged; the bundled `test/models/SIB/heffalump.sib` still imports (covered by the existing `utSIBImporter.importTest`).
- A crafted point count now fails the import with `SIB: Invalid face point count.` instead of corrupting the heap.

## Verification

New test `utSIBImporter.faceWithOverflowingPointCount` builds the 48-byte SIB stream inline and imports it through `ReadFileFromMemory`. It aborts with the heap corruption quoted above on `master` and passes with the fix.

```
./build-test/bin/unit --gtest_filter="utSIBImporter.*"  ->  [  PASSED  ] 2 tests
./build-test/bin/unit                                    ->  602 tests from 116 test suites, [  PASSED  ] 602 tests
```

The first commit adds only the failing test, the second contains the fix.

## Scope note

`ReadUVs()` walks `mesh->idx` from a face start for a separately supplied point count, which looks like it can read past that face's index range. That is a distinct issue and is not addressed here.

## AI tool disclosure

Per the [AI Tool Use Policy](https://github.com/assimp/assimp/blob/master/AIToolPolicy.md): prepared with AI assistance. I reviewed and verified the root cause, the reproducer and the fix myself, ran the test suite locally, and can answer questions during review; the commits carry an `Assisted-by:` trailer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness when importing malformed SIB files, including rejecting faces with invalid or excessively large point counts.
  * Reduced the risk of incorrect memory allocation during face processing.
  * Improved resource cleanup after importing materials, meshes, lights, and scene data.
* **Tests**
  * Added regression coverage for malformed SIB face data, including zero-point faces and oversized index allocations.

### DIFF
--- a/code/AssetLib/SIB/SIBImporter.cpp
+++ b/code/AssetLib/SIB/SIBImporter.cpp
@@ -127,7 +127,6 @@ struct SIB {
     // so a copy would hand the same pointers to two destructors.
     SIB(const SIB &) = delete;
     SIB &operator=(const SIB &) = delete;
-
     ~SIB() {
         // Anything still held here was not handed over to the aiScene, which happens
         // when parsing was aborted by an error.
@@ -878,6 +877,11 @@ void SIBImporter::InternReadFile(const std::string &pFile, aiScene *pScene, IOSy
     if (pScene->mNumLights) {
         memcpy(pScene->mLights, &sib.lights[0], sizeof(aiLight *) * pScene->mNumLights);
     }
+
+    // The scene owns them now, so make sure they are not deleted twice. The light list
+    // is still needed below and is released after the nodes have been built.
+    sib.mtls.clear();
+    sib.meshes.clear();
 
     // The scene owns them now, so make sure they are not deleted twice. The light list
     // is still needed below and is released after the nodes have been built.
--- a/code/AssetLib/SIB/SIBImporter.cpp
+++ b/code/AssetLib/SIB/SIBImporter.cpp
@@ -251,9 +251,17 @@ static void ReadFaces(SIBMesh *mesh, StreamReaderLE *stream) {
     while (stream->GetRemainingSizeToLimit() > 0) {
         uint32_t numPoints = stream->GetU4();
 
+        // Every point of the face is stored as a single 4-byte index, so a face can
+        // never have more points than the remaining chunk data can hold. Rejecting
+        // larger counts here also keeps the index array size below the point where
+        // numPoints * N would wrap around.
+        if (numPoints > stream->GetRemainingSizeToLimit() / 4) {
+            throw DeadlyImportError("SIB: Invalid face point count.");
+        }
+
         // Store room for the N index channels, plus the point count.
         const size_t pos = mesh->idx.size() + 1;
-        mesh->idx.resize(pos + numPoints * N);
+        mesh->idx.resize(pos + static_cast<size_t>(numPoints) * N);
         mesh->idx[pos - 1] = numPoints;
         uint32_t *idx = &mesh->idx[pos];
 
--- a/code/AssetLib/SIB/SIBImporter.cpp
+++ b/code/AssetLib/SIB/SIBImporter.cpp
@@ -254,7 +254,7 @@ static void ReadFaces(SIBMesh *mesh, StreamReaderLE *stream) {
         // never have more points than the remaining chunk data can hold. Rejecting
         // larger counts here also keeps the index array size below the point where
         // numPoints * N would wrap around.
-        if (numPoints > stream->GetRemainingSizeToLimit() / 4) {
+        if (numPoints == 0 || numPoints > stream->GetRemainingSizeToLimit() / 4) {
             throw DeadlyImportError("SIB: Invalid face point count.");
         }
 
--- a/test/unit/utSIBImporter.cpp
+++ b/test/unit/utSIBImporter.cpp
@@ -96,7 +96,7 @@ TEST_F(utSIBImporter, faceWithOverflowingPointCount) {
     // the index array. 0x55555556 * 3 wraps around to 2 in 32-bit arithmetic, so the
     // array is sized for two entries while the face loop writes many more. Such a face
     // must be rejected instead.
-    static const unsigned char sibData[] = {
+    static const unsigned char kSibData[] = {
         // 'SHAP' chunk, 0x28 bytes of payload
         0x53, 0x48, 0x41, 0x50, 0x28, 0x00, 0x00, 0x00,
         // 'VRTS' chunk: a single vertex at the origin
@@ -108,7 +108,24 @@ TEST_F(utSIBImporter, faceWithOverflowingPointCount) {
     };
 
     Importer importer;
-    const aiScene *scene = importer.ReadFileFromMemory(sibData, sizeof(sibData), 0, "sib");
+    const aiScene *scene = importer.ReadFileFromMemory(kSibData, sizeof(kSibData), 0, "sib");
+    EXPECT_EQ(nullptr, scene);
+}
+
+TEST_F(utSIBImporter, faceWithZeroPointCount) {
+    static const unsigned char kSibData[] = {
+        // 'SHAP' chunk, 0x20 bytes of payload
+        0x53, 0x48, 0x41, 0x50, 0x20, 0x00, 0x00, 0x00,
+        // 'VRTS' chunk: a single vertex at the origin
+        0x56, 0x52, 0x54, 0x53, 0x0c, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        // 'FACS' chunk: one face claiming zero points
+        0x46, 0x41, 0x43, 0x53, 0x04, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00
+    };
+
+    Importer importer;
+    const aiScene *scene = importer.ReadFileFromMemory(kSibData, sizeof(kSibData), 0, "sib");
     EXPECT_EQ(nullptr, scene);
 }
 
--- a/test/unit/utSIBImporter.cpp
+++ b/test/unit/utSIBImporter.cpp
@@ -91,6 +91,24 @@ TEST_F(utSIBImporter, UvPointCountExceedingFace) {
 
     Importer importer;
     const aiScene *scene = importer.ReadFileFromMemory(kSibData, sizeof(kSibData), 0, "sib");
+TEST_F(utSIBImporter, faceWithOverflowingPointCount) {
+    // The point count of a face is multiplied by the number of index channels to size
+    // the index array. 0x55555556 * 3 wraps around to 2 in 32-bit arithmetic, so the
+    // array is sized for two entries while the face loop writes many more. Such a face
+    // must be rejected instead.
+    static const unsigned char sibData[] = {
+        // 'SHAP' chunk, 0x28 bytes of payload
+        0x53, 0x48, 0x41, 0x50, 0x28, 0x00, 0x00, 0x00,
+        // 'VRTS' chunk: a single vertex at the origin
+        0x56, 0x52, 0x54, 0x53, 0x0c, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        // 'FACS' chunk: one face claiming 0x55555556 points
+        0x46, 0x41, 0x43, 0x53, 0x0c, 0x00, 0x00, 0x00,
+        0x56, 0x55, 0x55, 0x55, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
+    };
+
+    Importer importer;
+    const aiScene *scene = importer.ReadFileFromMemory(sibData, sizeof(sibData), 0, "sib");
     EXPECT_EQ(nullptr, scene);
 }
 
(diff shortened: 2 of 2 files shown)


Alb3e3:
The address-sanitizer job flagged a leak that the new regression test exposes: when SIB parsing throws, the materials/meshes/lights collected in the importer's local `SIB` struct are never released (`1132 byte(s) leaked in 4 allocation(s)`). That is pre-existing behaviour on every SIB error path — the new test is simply the first one to reach it.

Pushed `d0d39451b`, which gives the struct a destructor and clears the lists once the `aiScene` has taken ownership, so nothing is deleted twice.

Verified locally with an ASan/LSan build of the SIB importer:

```
# before
==625601==ERROR: LeakSanitizer: detected memory leaks
SUMMARY: AddressSanitizer: 1132 byte(s) leaked in 4 allocation(s).

# after, crafted file (error path)
scene=(nil) err=SIB: Invalid face point count.        # no ASan output

# after, test/models/SIB/heffalump.sib (success path)
scene=0x7399a4de0080 err=                             # no ASan output, no double free
```

Alb3e3:
Pushed `2f79aef74`, which addresses the SonarCloud quality gate and one more case in the same code path.

- Renamed the test payloads to `kSibData` per the constant naming convention (CodeRabbit's note on `test/unit/utSIBImporter.cpp`).
- Rejected a zero face point count alongside the existing upper bound. With `numPoints == 0` the index array is resized so that `pos` lands exactly on `mesh->idx.size()`, which makes the following `&mesh->idx[pos]` an out-of-bounds `operator[]` on the vector. Added `faceWithZeroPointCount` to cover it.

Local verification on this branch:

```
$ ./bin/unit --gtest_filter='utSIBImporter.*'
[  PASSED  ] 3 tests.

$ ./bin/unit
603 tests from 116 test suites ran. (12362 ms total)
[  PASSED  ] 603 tests.
```

Note on the red Windows jobs: `windows-latest-cl.exe` and `windows-hunter-latest-cl.exe` both ended at exactly 6h00m, i.e. the Actions job timeout rather than a build failure. The same two jobs are timing out the same way on unrelated open PRs (#6739, #6744), so it looks like a runner issue rather than anything in this branch.

This work was done with AI assistance; I reviewed, built and tested every change locally before pushing.

Alb3e3:
@kimkulling Apologies — this PR has been failing to build since 27 July and I only just caught it. My fault for not watching the checks.

`test/unit/utSIBImporter.cpp` was missing a closing brace on the `UvPointCountExceedingFace` test, so the two tests after it were parsed as local classes nested inside its `TestBody()`:

```
error: local class ... shall not have static data member ... test_info_
error: expected '}' at end of input
error: unused variable 'scene' [-Werror=unused-variable]
```

Fixed and force-pushed `3fc6dbd6`. The missing lines were the assertion plus the brace, which resolves the unused-variable error at the same time:

```c
    const aiScene *scene = importer.ReadFileFromMemory(kSibData, sizeof(kSibData), 0, "sib");
    EXPECT_EQ(nullptr, scene);
}
```

Verified locally: the file now compiles clean with no warnings, and `unit --gtest_filter='utSIBImporter.*'` gives **4/4 passing** (`importTest`, `UvPointCountExceedingFace`, `faceWithOverflowingPointCount`, `faceWithZeroPointCount`), each rejecting with `SIB: Invalid face point count.` as intended.

The importer change itself was never the problem — it was only the test file.

PR #6749, opened 2026-07-27
Reject MD5 element indices that wrap when the container is grown

### LINKED ISSUE #6710: Bug:Assimp MD5 importer integer overflow causes ASan crash via crafted .md5mesh file
# Integer overflow in MD5 importer leads to out-of-bounds vector access

## Summary

An integer overflow vulnerability exists in Assimp's MD5 mesh parser.

The issue is located in the MD5 importer, mainly in:

```text
code/AssetLib/MD5/MD5Parser.cpp
```

When parsing `vert`, `tri`, and `weight` definitions from a `.md5mesh` file, the parser reads index values using `strtoul10()`. This function returns an `unsigned int` but does not check for integer overflow during decimal parsing.

A crafted `.md5mesh` file can provide an extremely large index value such as:

```text
4294967295
```

which is equal to:

```text
0xFFFFFFFF
```

The value is accepted as a valid unsigned integer. Later, the parser uses the index in expressions such as:

```cpp
resize(idx + 1)
```

When `idx` is `0xFFFFFFFF`, `idx + 1` wraps around to `0` in 32-bit unsigned arithmetic. This unexpectedly resizes the target `std::vector` to zero elements. The parser then continues to access the vector using the original large index, causing an out-of-bounds vector access and a crash detected by AddressSanitizer.

## Affected Component

```text
Assimp MD5 importer
code/AssetLib/MD5/MD5Parser.cpp
include/assimp/fast_atof.h
```

The vulnerable integer parsing helper is:

```cpp
inline unsigned int strtoul10(const char* in, const char** out = 0) {
    unsigned int value = 0;
    for (;;) {
        if (*in < '0' || *in > '9')
            break;
        value = (value * 10) + (*in - '0');
        ++in;
    }
    if (out)
        *out = in;
    return value;
}
```

The multiplication and addition in this parser do not check for overflow.

## Vulnerability Type

```text
CWE-190: Integer Overflow or Wraparound
CWE-787: Out-of-bounds Write
CWE-125: Out-of-bounds Read
```

The root cause is integer overflow / wraparound. The observed result is an invalid vector access and process crash.

## Reproduction Environment

Tested on:

```text
Ubuntu 22.04 / Ubuntu 24.04
```

Build Assimp with AddressSanitizer:

```bash
git clone https://github.com/assimp/assimp.git --depth 1
cd assimp

mkdir build_asan
cd build_asan

cmake .. \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DASSIMP_BUILD_TESTS=OFF \
  -DBUILD_SHARED_LIBS=OFF \
  -DASSIMP_BUILD_ASSIMP_TOOLS=ON

make -j$(nproc) assimp
make -j$(nproc) assimp_cmd
```

The ASan-enabled command-line tool is generated as:

```text
build_asan/bin/assimpd
```

## Valid Control File

The following valid file is used as a baseline.

Create `valid.md5mesh`:

```text
MD5Version 10
commandline ""
numJoints 1
numMeshes 1
joints {
 "origin" -1 ( 0.000000 0.000000 0.000000 ) ( 0.000000 0.000000 0.000000 )
}
mesh {
 shader "test"
 numverts 3
 vert 0 ( 0.0 0.0 ) 0 1
 vert 1 ( 1.0 0.0 ) 0 1
 vert 2 ( 0.0 1.0 ) 0 1
 numtris 1
 tri 0 0 1 2
 numweights 1
 weight 0 0 1.000000 ( 0.0 0.0 0.0 )
}
```

Run:

```bash
./build_asan/bin/assimpd info valid.md5mesh
```

Expected result:

```text
Launching asset import ... OK
Importing file ... OK
Meshes: 1
Faces: 1
Vertices: 1
Bones: 1
Materials: 1
```

## Proof of Concept A: Vertex Index Overflow

Create `poc_vert.md5mesh`:

```text
MD5Version 10
commandline ""
numJoints 1
numMeshes 1
joints {
 "origin" -1 ( 0.000000 0.000000 0.000000 ) ( 0.000000 0.000000 0.000000 )
}
mesh {
 shader "test_shader"
 numverts 0
 numtris 0
 numweights 0
 vert 4294967295 ( 0.500000 0.500000 ) 0 1
}
```

Run:

```bash
./build_asan/bin/assimpd info poc_vert.md5mesh
```

Observed result:

```text
ERROR: AddressSanitizer: SEGV on unknown address
The signal is caused by a READ memory access.

#0 fast_atoreal_move
#1 Assimp::MD5::MD5MeshParser::MD5MeshParser
   code/AssetLib/MD5/MD5Parser.cpp:380
#2 Assimp::MD5Importer::LoadMD5MeshFile
#3 Assimp::MD5Importer::InternReadFile
#4 Assimp::BaseImporter::ReadFile
#5 Assimp::Importer::ReadFile
#6 ImportModel
#7 Assimp_Info
#8 main
```

## Proof of Concept B: Triangle Index Overflow

Create `poc_tri.md5mesh`:

```text
MD5Version 10
commandline ""
numJoints 1
numMeshes 1
joints {
 "origin" -1 ( 0.000000 0.000000 0.000000 ) ( 0.000000 0.000000 0.000000 )
}
mesh {
 shader "test_shader"
 numverts 3
 numtris 0
 numweights 0
 vert 0 ( 0.0 0.0 ) 0 1
 vert 1 ( 1.0 0.0 ) 0 1
 vert 2 ( 0.0 1.0 ) 0 1
 tri 4294967295 0 1 2
}
```

Run:

```bash
./build_asan/bin/assimpd info poc_tri.md5mesh
```

Observed result:

```text
ERROR: AddressSanitizer: SEGV on unknown address

#0 Assimp::MD5::MD5MeshParser::MD5MeshParser
   code/AssetLib/MD5/MD5Parser.cpp:401
```

## Proof of Concept C: Weight Index Overflow

Create `poc_weight.md5mesh`:

```text
MD5Version 10
commandline ""
numJoints 1
numMeshes 1
joints {
 "origin" -1 ( 0.000000 0.000000 0.000000 ) ( 0.000000 0.000000 0.000000 )
}
mesh {
 shader "test_shader"
 numverts 0
 numtris 0
 numweights 0
 weight 4294967295 0 0.500000 ( 0.000000 0.000000 0.000000 )
}
```

Run:

```bash
./build_asan/bin/assimpd info poc_weight.md5mesh
```

Observed result:

```text
ERROR: AddressSanitizer: SEGV on unknown address

#0 Assimp::MD5::MD5MeshParser::MD5MeshParser
   code/AssetLib/MD5/MD5Parser.cpp:420
```

## Root Cause Analysis

The root cause is that `strtoul10()` parses decimal strings into an `unsigned int` without checking whether the intermediate value exceeds the representable range.

The vulnerable logic is:

```cpp
value = (value * 10) + (*in - '0');
```

Since `value` is an unsigned 32-bit integer, arithmetic overflow wraps around silently.

In this case, the crafted value:

```text
4294967295
```

is accepted as:

```text
0xFFFFFFFF
```

Then the parser performs a resize operation based on:

```cpp
idx + 1
```

For `idx = 0xFFFFFFFF`, the result wraps around:

```text
0xFFFFFFFF + 1 = 0x00000000
```

This causes the vector to be resized to zero elements. The parser then attempts to access the vector using the original index `0xFFFFFFFF`, leading to an out-of-bounds access.

The issue can be triggered through multiple MD5 mesh declarations:

```text
vert 4294967295 ...
tri 4294967295 ...
weight 4294967295 ...
```

## Expected Behavior

Assimp should reject malformed `.md5mesh` files that contain out-of-range indices.

The parser should detect integer overflow during numeric parsing and should validate that parsed indices are within the expected range before resizing or indexing vectors.

## Actual Behavior

Assimp accepts the oversized index value, performs an overflowing `idx + 1` calculation, resizes the vector incorrectly, and then accesses the vector out of bounds.

## Impact

A crafted `.md5mesh` file can crash Assimp or any application that uses Assimp to import untrusted MD5 mesh files.

The confirmed impact is denial of service through a crash. Since the crash is caused by invalid memory access in a parser reachable from crafted input, this should be treated as a security-relevant memory safety issue.

## Suggested Fix

The parser should reject out-of-range integer values during parsing.

Possible fixes include:

1. Replace `strtoul10()` with a checked parser that detects overflow.
2. Use a wider integer type, such as `uint64_t`, during parsing and validate the result before casting to `unsigned int`.
3. Before using `idx + 1`, check whether `idx == UINT_MAX`.
4. Validate parsed indices against the declared counts, such as `numverts`, `numtris`, and `numweights`.
5. Reject sparse or excessively large indices that would cause unreasonable vector allocation or invalid access.

For example:

```cpp
if (idx == std::numeric_limits<unsigned int>::max()) {
    throw DeadlyImportError("MD5 index value is too large");
}
```

or:

```cpp
uint64_t parsed = parse_uint64_checked(token);
if (parsed > std::numeric_limits<unsigned int>::max()) {
    throw DeadlyImportError("MD5 index value exceeds unsigned int range");
}
```

Additionally, before resizing:

```cpp
if (idx > max_allowed_index || idx + 1 < idx) {
    throw DeadlyImportError("Invalid MD5 index");
}
```

## Additional Notes

The issue is triggered using Assimp's own command-line tool and does not require modifying Assimp source code or directly calling internal parser APIs.

The malformed file is a valid-looking MD5Mesh text file with only the index value changed to `4294967295`.

https://github.com/user-attachments/assets/3dc0bc2d-8dc9-491e-ba9a-98d1b0001d22

Alb3e3:
## Motivation

Fixes #6710. The MD5 mesh parser grows its containers with the index the file supplies:

```cpp
const unsigned int idx = ::strtoul10(sz, &sz);
if (idx >= desc.mVertices.size())
    desc.mVertices.resize(idx + 1);

VertexDesc &vert = desc.mVertices[idx];
```

`strtoul10()` has no overflow check, so `4294967295` parses to `0xffffffff`. `idx + 1` then wraps to `0`, `resize(0)` empties the container instead of growing it, and the next line addresses it with the original index — an out-of-bounds read and write.

The same pattern exists three times, for `vert`, `tri` and `weight`:

```
code/AssetLib/MD5/MD5Parser.cpp:374   desc.mVertices.resize(idx + 1);
code/AssetLib/MD5/MD5Parser.cpp:397   desc.mFaces.resize(idx + 1);
code/AssetLib/MD5/MD5Parser.cpp:417   desc.mWeights.resize(idx + 1);
```

All three reproduce from a ~20 line `.md5mesh`. On a debug build with hardened libstdc++ each aborts in `std::vector::operator[]`:

```
... [with _Tp = Assimp::MD5::VertexDesc ...]: Assertion '__n < this->size()' failed.
... [with _Tp = aiFace ...]:                  Assertion '__n < this->size()' failed.
... [with _Tp = Assimp::MD5::WeightDesc ...]: Assertion '__n < this->size()' failed.
```

## Implementation

Move the three sites onto a small helper that rejects the index that cannot be grown past and does the arithmetic in `size_t`:

```cpp
template <typename T>
static void ResizeForIndex(std::vector<T> &container, unsigned int idx, unsigned int line) {
    if (idx == std::numeric_limits<unsigned int>::max()) {
        MD5Parser::ReportError("Index out of range", line);
    }
    if (idx >= container.size()) {
        container.resize(static_cast<size_t>(idx) + 1);
    }
}
```

Only the wrapping value is rejected; every other index keeps the existing tolerant behaviour of growing the container, so files that declare their counts late or list elements out of order still load. Large-but-representable indices continue to fail the way they do today, through a failed allocation that the importer reports as an import error.

The error goes through `MD5Parser::ReportError`, so it carries the line number like the other parser diagnostics: `[MD5] Line 7: Index out of range`.

## Impact

- Valid files are unaffected: `SimpleCube.md5mesh`, `BoarMan.md5mesh` and `Bob.md5mesh` still import.
- A crafted index now fails the import cleanly instead of accessing the container out of bounds.

## Verification

Three fixtures under `test/models/MD5/invalid/` and three tests in `utMD5Importer`, following the existing `InvalidBoneIndex.md5mesh` convention in that directory. Each test aborts on `master` with the assertion quoted above and passes with the fix.

```
./build-test/bin/unit --gtest_filter="utMD5Importer.*"  ->  [  PASSED  ] 9 tests
./build-test/bin/unit                                    ->  604 tests from 116 test suites, [  PASSED  ] 604 tests
```

The first commit adds only the fixtures and tests, the second the fix.

## AI tool disclosure

Per the [AI Tool Use Policy](https://github.com/assimp/assimp/blob/master/AIToolPolicy.md): prepared with AI assistance. I reviewed and verified the root cause, the fixtures and the fix myself, ran the suite locally, and can answer questions during review; the commits carry an `Assisted-by:` trailer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when importing MD5 meshes with invalid or overflowing vertex, triangle, and weight indices.
  * Invalid mesh files are now rejected safely instead of risking incorrect processing.

* **Tests**
  * Added regression coverage for overflowing indices across vertices, triangles, and weights.

### DIFF
--- a/code/AssetLib/MD5/MD5Parser.cpp
+++ b/code/AssetLib/MD5/MD5Parser.cpp
@@ -53,6 +53,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <assimp/mesh.h>
 #include <assimp/DefaultLogger.hpp>
 
+#include <limits>
+
 using namespace Assimp;
 using namespace Assimp::MD5;
 
@@ -91,6 +93,20 @@ AI_WONT_RETURN void MD5Parser::ReportError(const char *error, unsigned int line)
     char szBuffer[1024];
     ::ai_snprintf(szBuffer, 1024, "[MD5] Line %u: %s", line, error);
     throw DeadlyImportError(szBuffer);
+}
+
+// ------------------------------------------------------------------------------------------------
+// Grow a container so that a file-supplied index can be used to address it
+template <typename T>
+static void ResizeForIndex(std::vector<T> &container, unsigned int idx, unsigned int line) {
+    // The index is parsed without an overflow check, and idx + 1 wraps around to zero
+    // for the largest one, which would shrink the container instead of growing it.
+    if (idx == std::numeric_limits<unsigned int>::max()) {
+        MD5Parser::ReportError("Index out of range", line);
+    }
+    if (idx >= container.size()) {
+        container.resize(static_cast<size_t>(idx) + 1);
+    }
 }
 
 // ------------------------------------------------------------------------------------------------
@@ -370,8 +386,7 @@ MD5MeshParser::MD5MeshParser(SectionArray &mSections) {
                     AI_MD5_SKIP_SPACES(&sz, elem.end, elem.iLineNumber);
                     const unsigned int idx = ::strtoul10(sz, &sz);
                     AI_MD5_SKIP_SPACES(&sz, elem.end, elem.iLineNumber);
-                    if (idx >= desc.mVertices.size())
-                        desc.mVertices.resize(idx + 1);
+                    ResizeForIndex(desc.mVertices, idx, elem.iLineNumber);
 
                     VertexDesc &vert = desc.mVertices[idx];
                     if ('(' != *sz++)
@@ -393,9 +408,7 @@ MD5MeshParser::MD5MeshParser(SectionArray &mSections) {
                 else if (TokenMatch(sz, "tri", 3)) {
                     AI_MD5_SKIP_SPACES(&sz, elem.end, elem.iLineNumber);
                     const unsigned int idx = strtoul10(sz, &sz);
-                    if (idx >= desc.mFaces.size()) {
-                        desc.mFaces.resize(idx + 1);
-					}
+                    ResizeForIndex(desc.mFaces, idx, elem.iLineNumber);
 
                     aiFace &face = desc.mFaces[idx];
                     if (face.mNumIndices != 3) {
@@ -413,8 +426,7 @@ MD5MeshParser::MD5MeshParser(SectionArray &mSections) {
                     AI_MD5_SKIP_SPACES(&sz, elem.end, elem.iLineNumber);
                     const unsigned int idx = strtoul10(sz, &sz);
                     AI_MD5_SKIP_SPACES(&sz, elem.end, elem.iLineNumber);
-                    if (idx >= desc.mWeights.size())
-                        desc.mWeights.resize(idx + 1);
+                    ResizeForIndex(desc.mWeights, idx, elem.iLineNumber);
 
                     WeightDesc &weight = desc.mWeights[idx];
                     weight.mBone = strtoul10(sz, &sz);
--- a/test/models/MD5/invalid/OverflowingTriangleIndex.md5mesh
+++ b/test/models/MD5/invalid/OverflowingTriangleIndex.md5mesh
@@ -1,0 +1,22 @@
+MD5Version 10
+commandline "crafted regression asset - triangle index overflow"
+
+numJoints 1
+numMeshes 1
+
+joints {
+"origin" -1 ( 0 0 0 ) ( 0 0 0 ) //
+}
+
+mesh {
+shader ""
+
+numverts 1
+vert 0 ( 0 0 ) 0 1
+
+numtris 1
+tri 4294967295 0 0 0
+
+numweights 1
+weight 0 0 1 ( 0 0 0 )
+}
--- a/test/models/MD5/invalid/OverflowingVertexIndex.md5mesh
+++ b/test/models/MD5/invalid/OverflowingVertexIndex.md5mesh
@@ -1,0 +1,22 @@
+MD5Version 10
+commandline "crafted regression asset - vertex index overflow"
+
+numJoints 1
+numMeshes 1
+
+joints {
+"origin" -1 ( 0 0 0 ) ( 0 0 0 ) //
+}
+
+mesh {
+shader ""
+
+numverts 1
+vert 4294967295 ( 0 0 ) 0 1
+
+numtris 1
+tri 0 0 0 0
+
+numweights 1
+weight 0 0 1 ( 0 0 0 )
+}
--- a/test/models/MD5/invalid/OverflowingWeightIndex.md5mesh
+++ b/test/models/MD5/invalid/OverflowingWeightIndex.md5mesh
@@ -1,0 +1,22 @@
+MD5Version 10
+commandline "crafted regression asset - weight index overflow"
+
+numJoints 1
+numMeshes 1
+
+joints {
+"origin" -1 ( 0 0 0 ) ( 0 0 0 ) //
+}
+
+mesh {
+shader ""
+
+numverts 1
+vert 0 ( 0 0 ) 0 1
+
+numtris 1
+tri 0 0 0 0
+
+numweights 1
+weight 4294967295 0 1 ( 0 0 0 )
+}
--- a/test/unit/ImportExport/utMD5Importer.cpp
+++ b/test/unit/ImportExport/utMD5Importer.cpp
@@ -68,6 +68,28 @@ TEST(utMD5Importer, importInvalidBoneIndex) {
     ASSERT_EQ(nullptr, scene);
 }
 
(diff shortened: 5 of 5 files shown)


kimkulling (approved):
Looks fine for me.

kimkulling:
Merged,, thank you for your contribution.

PR #6752, opened 2026-07-29
bound face uv and color indices in ase BuildUniqueRepresentation

zayeem06:
BuildUniqueRepresentation checks each face's position index against mPositions before copying, but it reads amTexCoords and mVertexColors using the per-face UV and color indices with no check, and those come straight from the *MESH_TFACE and *MESH_CFACE triples where the parser only validates the face number, not the index values. An ASE file that names a UV or vertex-color index past the channel size gives a heap out-of-bounds read at the two copy sites in that loop. I added the same bounds guard the position index already has before both reads, so a bad index is rejected the way an invalid vertex index already is. The new test/models/ASE/invalid/InvalidUVIndex.ase overreads on the current tree and imports cleanly after the change; worth confirming the throw is the behavior you want here rather than skipping the coordinate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ASE model import validation for invalid UV and vertex color references.
  * Invalid texture-coordinate indices now produce a clear import failure instead of risking incorrect data access.

* **Tests**
  * Added coverage for ASE files containing invalid UV indices.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6754, opened 2026-08-01
fix anim problem in fbx exporter

liuchunlin:
1. fix fbx rot anim error caused by Euler Angles Gimbal Lock problem and angle cross border problem like 179° to -179°
2. fbx anim curve set interp type as Linear (fbxsdk cannot parse without a valid interp type)
3. fbx skin bind-matrix export improve (current skeleton pose may diff with bind-pose)
4. fix unknown file extension of embedded texture in fbx exporter (caused by unsigned integer)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved animation export to reduce rotational discontinuities and gimbal-lock artifacts.
  * Corrected bone bind-pose transform data in exported files.
  * Fixed embedded texture format detection for paths using different separators.
  * Animation keys now correctly use linear interpolation for smoother, more predictable playback.
  * Improved handling of angle transitions around rotation limits to preserve consistent animation results.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6755, opened 2026-08-01
Bounds-check SIB UV point counts against the face

Alb3e3:
## Motivation

`ReadUVs()` in the SIB importer reads a per-face point count directly from the UV (`FTVS`) chunk and loops that many times over the face's index data, but never checks it against the number of points the face was actually built with in `ReadFaces()`:

```cpp
uint32_t pos = mesh->faceStart[faceIdx];
uint32_t *idx = &mesh->idx[pos + 1];

for (uint32_t n = 0; n < numPoints; n++, idx += N) {   // numPoints is attacker-controlled
    uint32_t id = idx[UV];
    mesh->uv[id].x = stream->GetF4();
    mesh->uv[id].y = stream->GetF4();
}
```

`faceIdx` is validated, but `numPoints` is not. A `FTVS` chunk that claims more points than its face has walks the `idx` cursor past the end of that face's index data and off the end of `mesh->idx`, producing an out-of-bounds read of `idx[UV]` (and then an out-of-bounds write into `mesh->uv` through the garbage index).

### AddressSanitizer report on current `master` (before the fix)

```
ERROR: AddressSanitizer: heap-buffer-overflow READ of size 4
  #0 ReadUVs             code/AssetLib/SIB/SIBImporter.cpp:273
  #1 ReadShape           code/AssetLib/SIB/SIBImporter.cpp
  #2 ReadScene           code/AssetLib/SIB/SIBImporter.cpp
  #3 SIBImporter::InternReadFile(...)
0x... is located 8 bytes before a 4-byte region
allocated by:
  #7 ReadFaces           code/AssetLib/SIB/SIBImporter.cpp:236
```

The reproducer is a hand-built 0x4c-byte `.sib`: one vertex, one face with a single point, and a `FTVS` chunk that claims that face has two points.

## Fix

The face's real point count is stored at `mesh->idx[faceStart[faceIdx]]` (the slot `ReadFaces()` writes the count into). Reject a UV chunk whose `numPoints` exceeds it, which bounds the loop to the face's own index data and also keeps every `id` within `mesh->uv`.

## Impact

- Malformed/crafted SIB files now fail with a clean `DeadlyImportError` instead of reading and writing out of bounds.
- Valid files are unaffected: the bundled `heffalump.sib` model still imports identically (`utSIBImporter.importTest` passes), and the full unit suite passes (602/602).

## Verification

- ASan build of current `master`: reproduced the overflow, then confirmed the patched build rejects the same file cleanly.
- `test/unit`: `utSIBImporter.*` pass, full `unit` suite 602/602.
- Added `utSIBImporter.uvPointCountExceedingFace` as a regression test.

This fix was developed with AI assistance; I reviewed, built, and tested every change locally, including reproducing the crash under AddressSanitizer before and after the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed UV data during SIB file import.
  * Invalid files now fail safely instead of risking import errors from inconsistent point counts.
  * Improved cleanup when SIB imports are interrupted or fail, reducing the risk of retained resources.

* **Tests**
  * Added regression coverage verifying that malformed UV data is rejected safely during import.


Alb3e3:
Rebased onto current master and fixed the red `adress-sanitizer` job.

The regression test added here makes `ReadUVs()` throw, and the SIB importer had no cleanup on that path: `InternReadFile()` collects materials, meshes and lights in a local `SIB` struct and only hands the pointers to the `aiScene` once parsing has finished, so an exception dropped all of them. That is what CI reported:

```
SUMMARY: AddressSanitizer: 1132 byte(s) leaked in 4 allocation(s)
```

The leak is pre-existing, not something this change introduces — the new test is just the first one to abort a SIB parse mid-way. Fixed by giving `SIB` a destructor that deletes whatever it still holds, and clearing the lists once the scene has taken ownership so nothing is freed twice.

That is the same commit as `d0d3945` in #6747, which hit the identical leak for the same reason. Whichever of the two lands first, the other will drop the duplicate on rebase.

Verified locally with a Debug `-DASSIMP_ASAN=ON` build (gcc, LeakSanitizer on):

- `utSIBImporter` 2/2 passed, no leak report
- full suite `608 tests from 116 test suites` — `[ PASSED ] 608 tests`, exit 0

Same AI-assistance disclosure as the original description applies: I reviewed, built and tested every change locally.

kimkulling (approved):
Good work, approved.

kimkulling:

kimkulling:
Closes https://github.com/assimp/assimp/issues/6712

PR #6756, opened 2026-08-03
GenBoundingBoxes: fix AABB for meshes with coordinates beyond +/-999999

zaczhuozhang:
### Problem

`aiProcess_GenBoundingBoxes` initializes its min/max accumulators with hardcoded `+/-999999` sentinel values:

```cpp
aiVector3D min(999999, 999999, 999999), max(-999999, -999999, -999999);
```

Any mesh containing a coordinate beyond that range gets a wildly wrong AABB, because the corresponding accumulator never moves past the sentinel. This is easy to hit in practice:

- CAD/DXF files in projected geographic coordinate systems (e.g. Swiss LV95, eastings ~2'600'000)
- millimeter-unit files for anything larger than ~1 km

Example: a small polyline at `x ≈ 2'600'000` ends up with `mAABB.mMin.x = 999999` and `mAABB.mMax.x ≈ 2'600'000`, i.e. a bounding box ~1'600'000 units across for a tiny object.

### Fix

- Initialize the accumulators with `std::numeric_limits<ai_real>::max()` / its negation.
- Skip meshes without vertices so their AABB stays at the default zero value instead of being filled with the initialization values.

### Testing

Added `utGenBoundingBoxesProcess.executeFarFromOriginTest`, which places vertices at offsets around 2.6e6 and verifies a tight AABB. It fails against the previous sentinel-based initialization and passes with this change. Existing `executeTest` still passes.

Per the AI tool policy: this change was prepared with assistance from Claude and has been reviewed by me.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bounding-box generation for meshes with vertices located far from the origin.
  * Meshes without vertices are now handled safely without producing invalid bounds.

* **Tests**
  * Added regression coverage to verify accurate minimum and maximum coordinates for distant vertices.
  * Added coverage for empty meshes, confirming their bounds default to zero.

### DIFF
--- a/code/PostProcessing/GenBoundingBoxesProcess.cpp
+++ b/code/PostProcessing/GenBoundingBoxesProcess.cpp
@@ -46,6 +46,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <assimp/postprocess.h>
 #include <assimp/scene.h>
 
+#include <limits>
+
 namespace Assimp {
 
 bool GenBoundingBoxesProcess::IsActive(unsigned int pFlags) const {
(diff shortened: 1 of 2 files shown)


kimkulling (approved):
Good catch, thanks for you great work.

kimkulling:

PR #6778, opened 2026-08-16
[PyAssimp] Drop Python 2 Compatibility

thedestroyer4312:
# Scope of Changes
- Remove backwards compatible Python 2 code shims
- Drop distutils/Anaconda search path
  - Logic error of checking `sys.version.major < 3` instead of `<=` so it never triggered on Python 3 anyways
  - Can add conda checking back in the future with `CONDA_PREFIX`, etc. if so desired 
- Delete `setup.py` as its use is deprecated
   - https://packaging.python.org/en/latest/discussions/setup-py-deprecated/
> python setup.py and the use of setup.py as a command line tool are deprecated.
> 
> This means that commands such as the following MUST NOT be run anymore:
> 
>     python setup.py install
> 
>     python setup.py develop
> 
>     python setup.py sdist
> 
>     python setup.py bdist_wheel
- Update `sample.py` to use a context manager for `load()

# Testing
Build via CMake and did `pip install -e `.

Then:
```python
import pyassimp
available_formats()
```
and `python scripts/sample.py`.

# AI Use Disclosure
Used an LLM to scan and delete the code. Have reviewed changes by hand.

## Summary by CodeRabbit

- **New Features**
  - Added modern package configuration for installing PyAssimp with `pip install .`.

- **Documentation**
  - Updated installation instructions and minimum supported Python version to Python 3.10.

- **Refactor**
  - Removed legacy Python 2 compatibility and outdated platform-specific setup behavior.
  - Updated the sample script to use modern Python syntax and automatic scene cleanup.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6783, opened 2026-08-19
fix csm $points heap overflow when a channel drops a sample

zayeem06:
The $points reader grows each channel's key buffer from a single shared counter, but a DROPOUT marker lets a channel skip a sample so its key count falls behind the others; once the shared counter doubles, that channel's realloc test stops matching and it keeps writing past a buffer that was never grown, which a crafted .csm turns into a heap overflow that ASAN flags as a write at CSMLoader.cpp:228. I switched the growth to track each channel's own capacity so a buffer is reallocated exactly when that channel fills up, and dropped the now-unused shared granularity counter. The added invalid model reproduces the overflow on the old code and the existing valid CSM models still import unchanged, so it is worth confirming the per-channel growth reads the way reviewers expect for the multi-marker case.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue that could cause CSM animation files with `DROPOUT` markers and uneven channel data to fail during loading.
  - Improved handling of channel-specific sample counts for affected animations.

- **Tests**
  - Added regression coverage for CSM files containing `DROPOUT` markers and large point sequences.


kimkulling (approved):
Looks fine.

PR #6799, opened 2026-08-22
AMF: fix uninitialized last byte in converted texture format hint

PavolUlicny:
## Summary

`AMFImporter::PostprocessHelper_GetTextureID_Or_Create` builds the format hint of a converted texture with `memcpy(converted_texture.FormatHint, templateColor, 8)`, copying 8 bytes from the 9-byte literal `"rgba0000"` into the 9-byte buffer `SPP_Texture::FormatHint`. Byte 8 stays uninitialized because the struct is default-initialized (`SPP_Texture converted_texture;`). The later copy into public scene data, `strncpy(pScene->mTextures[idx]->achFormatHint, tex_convd.FormatHint, HINTMAXTEXTURELEN)`, transfers all nine bytes because `strncpy` only stops early at a NUL, so the uninitialized byte lands in the public `aiTexture::achFormatHint`.

This regressed when #5656 replaced `strcpy(converted_texture.FormatHint, "rgba0000")` with the 8-byte `memcpy`. Before that commit the terminator was written; after it, byte 8 is garbage.

## Impact

- Every consumer of uncompressed AMF embedded textures sees a garbage trailing byte in `achFormatHint` (observed e.g. `"rgba8880G"`, varies per run).
- Because byte 8 can be non-NUL, any consumer calling `strlen`, `%s`, or `strcmp` on `achFormatHint` reads past the end of the 9-byte array. CWE-457.

## Reproduction

Minimal AMF with grayscale textures and a `texmap`, read with default postprocessing, then print `scene->mTextures[0]->achFormatHint`. Under valgrind this reports `Conditional jump or move depends on uninitialised value(s)` at the `strncpy` call in `AMFImporter::Postprocess_BuildScene`. With this patch the hint is clean (`"rgba8880"`) and valgrind is silent.

## Fix

Copy the full buffer including the terminator:

```cpp
memcpy(converted_texture.FormatHint, templateColor, sizeof(converted_texture.FormatHint));
```

Both are 9 bytes. No unit test asserts on the hint contents, so nothing else changes.

## AI usage

AI tools were used to help track down and fix this bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved texture format hint initialization to ensure the full buffer is copied correctly.
  * No user-visible behavior changes are expected.


fallenmi (approved):
The destination buffer is nine bytes, so copying its full size also initializes the NUL terminator after the eight-character AMF format hint. This prevents the uninitialized tail byte from being propagated into the public `aiTexture::achFormatHint` field.

I verified exact head `e3289d083278d84325614c184678e483d051d6cb` against exact base `21bea12f1278090c09a94c39ca090d9fb458ccd9` and current GitHub merge `4ad3341458b05516eabcbeaee9a355fbb9f7b13b`. Using the public `Assimp::Importer` API on the repository's textured `3_bananas.amf` fixture, with Clang's deterministic automatic-variable pattern initialization, base exports `rgba8888aa` and fails the termination oracle. Head and current merge export `rgba888800` in three repeated runs each. The AMF-only library also builds with warnings-as-errors at head and merge. Live CI has 12 successful checks, two expected release-job skips, and all three workflow runs completed successfully.

Disclosure: I used Codex to inspect the change and run exact-base, exact-head, and current-merge build and public-import oracles. I verified the results and live pull-request state before preparing this review.

PR #6804, opened 2026-08-23
Fix glTF KHR_texture_transform coordinate space mapping

killianguilland:
Fixes a math mapping issue where `glTF2Importer` incorrectly converted `KHR_texture_transform` offsets.
glTF applies translation after scaling and rotation, but Assimp's `TextureTransformProcess` applies translation before. 
The importer now correctly calculates the pre-transform offset so that textures end up in the exact correct position.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved texture transform handling when textures use rotation or scaling.
  - Prevented invalid texture coordinates when scaling values are zero.


turol:
Why did you remove the test case?

PR #6813, opened 2026-08-24
Fix null context logging in glTF object lookup

### LINKED ISSUE #6811: Bug: NULL dereference logging an error in `FindObjectInContext`
**Describe the bug**

A glTF file crashes the importer with a NULL dereference, because an error path logs a `const char *` that is null.

`FindObjectInContext` takes an optional `extraContext` defaulting to `nullptr` and logs it unguarded, `code/AssetLib/glTFCommon/glTFCommon.h`:

```c++
461	inline Value *FindObjectInContext(Value &val, const char * memberId, const char *context, const char *extraContext = nullptr) {
...
469	    if (!it->value.IsObject()) {
470	        ASSIMP_LOG_ERROR("Member \"", memberId, "\" was not of type \"", context, "\" when reading ", extraContext);
471	        return nullptr;
472	   }
```

`FindObject` is the two-argument wrapper, so `extraContext` stays null:

```c++
503	inline Value *FindObject(Document &doc, const char *memberId) {
504	    return FindObjectInContext(doc, memberId, "the document");
505	}
```

The logger streams that pointer into an `ostream`, which calls `strlen(nullptr)`:

```
==14==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
==14==The signal is caused by a READ memory access.
    #1 in strlen
    #5 in operator<<<const char *, nullptr> /src/assimp/include/assimp/TinyFormatter.h:135:20
    #6 in Assimp::Logger::formatMessage<char const*&>(...) /src/assimp/include/assimp/Logger.hpp:252:42
    #12 in FindObjectInContext /src/assimp/code/AssetLib/glTFCommon/glTFCommon.h:470:9
    #13 in FindObject /src/assimp/code/AssetLib/glTFCommon/glTFCommon.h:504:12
    #14 in glTF2::LazyDict<glTF2::Light>::AttachToDocument(...)
```

The four sibling functions do guard this.
`FindStringInContext`, `FindNumberInContext`, `FindUIntInContext` and `FindArrayInContext` all route their error through `throwUnexpectedTypeError`, which checks the pointer before using it (`glTFCommon.h:397`):

```c++
397	    if (extraContext && (strlen(extraContext) > 0)) {
```

`FindObjectInContext` is the only one of the five that formats its own message, and the only one that crashes.

**To Reproduce**

1. Get OSS-Fuzz and build the assimp fuzz targets with ASan:

```bash
export DOCKER_DEFAULT_PLATFORM=linux/amd64 # if on mac
git clone https://github.com/google/oss-fuzz.git
cd oss-fuzz
python3 infra/helper.py build_image assimp
python3 infra/helper.py build_fuzzers --sanitizer address assimp
```

2. Take the attached files ([Reproduce.zip](https://github.com/user-attachments/files/31382928/Reproduce.zip))

3. Run through the GLB target:

```bash
python3 infra/helper.py reproduce assimp assimp_fuzzer_glb minimal_extensions_not_object.glb # Minimized crash-020ca813184069a29526bb3dc6fa4153099fdf74
```

4. See `AddressSanitizer: SEGV on unknown address 0x000000000000`.

`minimal_extensions_not_object.glb` is a shortened version of `crash-020ca813184069a29526bb3dc6fa4153099fdf74`. It contains:

```json
{"asset":{"version":"2.0"},"extensions":0}
```

A top level `"extensions"` that is present but not an object is enough.
`LazyDict<Light>::AttachToDocument` calls `FindObject(doc, "extensions")` (`glTF2Asset.inl:410`), the member exists but fails the `IsObject()` test, and the error path logs the null pointer.

**Expected behavior**

The importer should log the error and continue, which is what the function already intends: it returns `nullptr` on the next line rather than throwing.

 - OS: OSS-Fuzz at commit `a4df12d70b5420567d893b3f53e5818a74df5db7`, default `linux/amd64` containers
 - Browser: not applicable
 - Version: assimp master at commit `21bea12f1278090c09a94c39ca090d9fb458ccd9`

**Additional context**

Suggested fix, guard the pointer the way `throwUnexpectedTypeError` already does:

```diff
-        ASSIMP_LOG_ERROR("Member \"", memberId, "\" was not of type \"", context, "\" when reading ", extraContext);
+        ASSIMP_LOG_ERROR("Member \"", memberId, "\" was not of type \"", context, "\" when reading ",
+                extraContext ? extraContext : "the document");
```

Routing this message through `throwUnexpectedTypeError` instead would also work, but that helper throws, and this call site deliberately does not.

Found by the CISPA Fandango-Team while triaging OSS-Fuzz findings for assimp.

carrerasdarren-cell:
## Summary

- guard the optional glTF object error context before formatting it
- report the expected `object` type with the same context format as sibling lookups
- add a regression fixture for a non-object top-level `extensions` member

## Testing

- reproduced the original 64-byte GLB crash under AddressSanitizer before the fix
- `build-asan/bin/unit --gtest_filter=utglTF2ImportExport.wrongObject`
- `build-asan/bin/unit --gtest_filter=utglTF2ImportExport.*` (48 tests)
- verified the original GLB no longer triggers an AddressSanitizer error

Fixes #6811

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when invalid glTF object types are encountered.
  * Error messages now correctly identify the expected type and include the full context.

* **Tests**
  * Added coverage for invalid top-level extensions data to ensure imports handle malformed assets safely.

### DIFF
--- a/code/AssetLib/glTFCommon/glTFCommon.h
+++ b/code/AssetLib/glTFCommon/glTFCommon.h
@@ -467,7 +467,13 @@ inline Value *FindObjectInContext(Value &val, const char * memberId, const char 
         return nullptr;
     }
     if (!it->value.IsObject()) {
-        ASSIMP_LOG_ERROR("Member \"", memberId, "\" was not of type \"", context, "\" when reading ", extraContext);
+        std::string fullContext = context;
+        if (extraContext && extraContext[0] != '\0') {
+            fullContext += " (";
+            fullContext += extraContext;
+            fullContext += ")";
+        }
+        ASSIMP_LOG_ERROR("Member \"", memberId, "\" was not of type \"object\" when reading ", fullContext);
         return nullptr;
    }
     return &it->value;
--- a/test/models/glTF2/wrongTypes/topLevelExtensionsNotObject.gltf
+++ b/test/models/glTF2/wrongTypes/topLevelExtensionsNotObject.gltf
@@ -1,0 +1,1 @@
+{"asset":{"version":"2.0"},"extensions":0,"scene":0,"scenes":[{"nodes":[0]}],"nodes":[{}]}
--- a/test/unit/utglTF2ImportExport.cpp
+++ b/test/unit/utglTF2ImportExport.cpp
@@ -964,7 +964,8 @@ TEST_F(utglTF2ImportExport, wrongObject) {
 #define TUPLE(x, y, z, w) tup_T(x, y, z, w)
 #endif
     TUPLE("/glTF2/wrongTypes/badObject.gltf", "object", "pbrMetallicRoughness", "materials[0]"),
-    TUPLE("/glTF2/wrongTypes/badExtension.gltf", "object", "KHR_texture_transform", "materials[0]")
+    TUPLE("/glTF2/wrongTypes/badExtension.gltf", "object", "KHR_texture_transform", "materials[0]"),
+    TUPLE("/glTF2/wrongTypes/topLevelExtensionsNotObject.gltf", "object", "extensions", "the document")
 #undef TUPLE
     };
     for (const auto &tuple : wrongTypes) {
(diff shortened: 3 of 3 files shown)


kimkulling (approved):
Looks fine.

kimkulling:

PR #6814, opened 2026-08-26
Fix lwo face index underflow when the point list is empty

zayeem06:
CopyFaceIndicesLWO2 and its LWOB counterpart clamp an out-of-range face vertex index to mTempPoints.size() - 1, which underflows to 0xFFFFFFFF when a POLS chunk is read before any PNTS chunk has filled the layer's point list. A later PNTS then satisfies the conversion guard and the poisoned index is used to read layer.mTempPoints during mesh conversion, so a small crafted LWO2 or LWOB file drives a heap out-of-bounds read that ASAN flags at LWOLoader.cpp:341. The fix clamps to 0 while the point list is still empty so the index stays valid once conversion runs; regression models and tests for both flavours are included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or unusual LWO files with face indices appearing before point data.
  * Prevented invalid index values when files contain no points, allowing affected scenes to import successfully.

* **Tests**
  * Added regression coverage for LWO2 and LWOB files with out-of-order face and point data.


kimkulling (approved):
Looks fine.

PR #6818, opened 2026-08-27
guard x importer texcoord read against undersized fvfdata uv array

zayeem06:
XFileImporter::CreateMeshes reads sourceMesh->mTexCoords[e][pf.mIndices[d]] without bounding the index, unlike the position and normal reads just above it. Every normal texcoord source is forced to match the vertex count, but ParseDataObjectMeshFVFData sizes its UV array to numCoords/(2*channels), an independent value taken from the file, so a crafted .x that declares fewer FVF coords than the mesh has vertices lets a face index run past the shorter texcoord array and read out of bounds. I added the same index bound the normal read already uses; it is worth checking against valid multi-UV .x models, and a regression model plus unit test are included with the full asan unit suite still green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed X-format models with invalid texture-coordinate indices.
  - Prevented crashes or unsafe memory access when importing affected files.
  - Valid texture coordinates continue to import with existing conversion behavior.

- **Tests**
  - Added coverage for malformed texture-coordinate data to verify safe, non-throwing imports.


fallenmi (approved):
No blocking findings.

The new bound is applied to the same untrusted face index used for positions, and it safely leaves the destination UV at its default zero value when malformed FVF data provides fewer coordinates than the mesh has vertices. Valid coordinate reads are unchanged.

I reproduced the issue against current `master` `a47827a28ae93e132f3357d662df407e19bb94a8`, exact head `88739e6423a7dc99a7d946b5f4001d455e4eafd4`, and current merge `41509670bc00cb190c33c4164cdf69051e39e906`. Under AddressSanitizer, current base reports an 8-byte heap-buffer-overflow immediately after the one-element FVF UV allocation; head and merge import the same malformed model successfully as a non-null one-mesh/three-vertex scene with zeroed missing UVs. Independent valid one- and two-channel FVF controls produce identical UV values on base, head, and merge. The owning ASan suites pass 13/13 on base and 14/14 on head and merge, including the submitted regression; the current upstream rollup is 14 successful checks plus two expected release-only skips.

Disclosure: I used OpenAI Codex to inspect the exact revisions and policy, build and run the AddressSanitizer differential oracle and valid multi-UV controls, execute the focused tests, and draft this review. I reviewed and understood the resulting diff and evidence before posting.

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6819, opened 2026-08-28
Fix @param names in the public headers that do not match the signatures

darkdi:
Reading `Logger.hpp` I stopped at `OnWarn(const char* essage)`. Every other method in that class takes `message`, and the block above it documents `message` too, so the parameter name has simply lost its first letter. That sent me through the rest of `include/assimp` looking for `@param` names that are not in the signature, and there are about twenty.

Most are copied blocks that were never adjusted: `MergeBones` still documents the `flags` and `begin` of `MergeMeshes` above it, `aiVector3LessThan` still documents the `epsilon` of `aiVector3AreEqualEpsilon`, and the axis-angle decompose documents `rotation` where the second argument is `scaling` (I took the wording for it from the Euler variant next door). The rest are arguments that were renamed while the tag stayed, plus one `@param` in `CheckMagicToken` that has no name at all. Everything here is comments except the `essage` typo.

Two places I left alone on purpose. `Exporter.hpp:199` documents `pBlob` for `Export(pScene, ...)`, and the fix there is a sentence someone has to write rather than a rename, so it seemed better as its own change. `aabb.h` looked like a hit in my scan and is fine.

I found these with a checker I wrote and read every place in the file before touching it. The commit carries an `Assisted-by` trailer.

## Summary by CodeRabbit

* **Documentation**
  * Corrected and clarified API reference parameter names across importers, parsers, scene utilities, shape generators, archive handling, metadata, and C interfaces.
  * Improved descriptions for buffer inputs, outputs, scaling data, and other function parameters.
  * Removed outdated parameter documentation to keep the generated reference accurate.


kimkulling (approved):
Good catches, looks fine for me.

kimkulling:
Merged, thank you for your contirbution.

PR #6824, opened 2026-09-02
Fix BatchLoader ownership after polling shared scenes

### LINKED ISSUE #6820: Bug:  assimp:assimp_fuzzer: Heap-use-after-free in aiScene::~aiScene
Detailed Report: https://oss-fuzz.com/testcase?key=4986524437643264

Project: assimp
Fuzzing Engine: libFuzzer
Fuzz Target: assimp_fuzzer
Job Type: libfuzzer_asan_assimp
Platform Id: linux

Crash Type: Heap-use-after-free READ 8
Crash Address: 0x51a000000688
Crash State:
  aiScene::~aiScene
  Assimp::BatchLoader::~BatchLoader
  Assimp::LWSImporter::InternReadFile
 
Sanitizer: address (ASAN)

Recommended Security Severity: High

Regressed: https://oss-fuzz.com/revisions?job=libfuzzer_asan_assimp&range=202008080609:202104020622

Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=4986524437643264

Issue filed automatically.

See https://google.github.io/oss-fuzz/advanced-topics/reproducing for instructions to reproduce this bug locally.
When you fix this bug, please
  * mention the fix revision(s).
  * state whether the bug was a short-lived regression or an old bug in any stable releases.
  * add any other useful information.
This information can help downstream consumers.

If you need to contact the OSS-Fuzz team with a question, concern, or any other feedback, please file an issue at https://github.com/google/oss-fuzz/issues. Comments on individual Monorail issues are not monitored.

This bug is subject to a 90 day disclosure deadline. If 90 days elapse
without an upstream patch, then the bug report will automatically
become visible to the public.

carrerasdarren-cell:
Fixes #6820.

## Root cause

`BatchLoader` coalesces duplicate file requests and keeps a reference count for the resulting `aiScene`. `GetImport()` can return that scene while another duplicate request remains unpolled. The caller then owns the scene, and the LWS merge path deletes it after transferring its contents, but the pending batch request still retains the same pointer. `BatchLoader` consequently deletes the already-freed scene from its destructor.

## Fix

Track whether each loaded scene has been claimed by a caller. The batch loader still deletes scenes that were never polled, while claimed scenes remain caller-owned even when duplicate request references are pending. The `GetImport()` documentation now states that ownership contract explicitly.

The regression queues the same in-memory OBJ twice, polls it once, deletes the caller-owned scene, and verifies that destroying the loader with the duplicate request still pending does not reclaim it.

## Verification

- Reproduced the public OSS-Fuzz testcase from #6820 on current `master` under AddressSanitizer before the change.
- Confirmed the same instrumented import/export harness exits cleanly after the change.
- `BatchLoaderTest.*` and `utLWSImportExport.*`: 17/17 pass under AddressSanitizer.
- Full unit suite: 632/632 pass under AddressSanitizer.

## AI tool disclosure

Prepared with AI assistance. I independently reviewed and tested the code, understand the ownership change, and take responsibility for the contribution. The commit includes an `Assisted-by: OpenAI Codex` trailer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed batch loading so scenes returned to callers are not deleted prematurely.
  - Preserved correct ownership when multiple requests reference the same loaded scene.
  - Clarified that callers are responsible for deleting returned scenes exactly once.

- **Documentation**
  - Updated API documentation to explain scene ownership and release requirements.

- **Tests**
  - Added regression coverage for scene ownership when duplicate requests remain outstanding.

### DIFF
--- a/code/Common/BaseImporter.cpp
+++ b/code/Common/BaseImporter.cpp
@@ -495,6 +495,7 @@ struct LoadRequest {
             flags(_flags),
             refCnt(1),
             scene(nullptr),
+            sceneClaimed(false),
             loaded(false),
             id(_id) {
         if (_map) {
@@ -510,6 +511,7 @@ struct LoadRequest {
     unsigned int flags;
     unsigned int refCnt;
     aiScene *scene;
+    bool sceneClaimed;
     bool loaded;
     BatchLoader::PropertyMap map;
     unsigned int id;
@@ -564,7 +566,9 @@ BatchLoader::BatchLoader(IOSystem *pIO, bool validate) {
 BatchLoader::~BatchLoader() {
     // delete all scenes what have not been polled by the user
     for (LoadReqIt it = m_data->requests.begin(); it != m_data->requests.end(); ++it) {
-        delete (*it).scene;
+        if (!(*it).sceneClaimed) {
+            delete (*it).scene;
+        }
     }
     delete m_data;
 }
@@ -611,6 +615,7 @@ aiScene *BatchLoader::GetImport(unsigned int which) {
     for (LoadReqIt it = m_data->requests.begin(); it != m_data->requests.end(); ++it) {
         if ((*it).id == which && (*it).loaded) {
             aiScene *sc = (*it).scene;
+            (*it).sceneClaimed = true;
             if (!(--(*it).refCnt)) {
                 m_data->requests.erase(it);
             }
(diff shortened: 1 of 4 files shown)


kimkulling (approved):
Looks fine for me, thanks a lot!

kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6829, opened 2026-09-06
glTF2: set aiMesh::mMethod on meshes with morph targets

### LINKED ISSUE #5303: Bug: glTF import fails to set mMethod in aiMesh
**Describe the bug**
After importing a glTF mesh with morphs, the `mMethod` field (in `aiMesh`) ought to be determined. However, it's typically `aiMorphingMethod_UNKNOWN` (0x0).

**To Reproduce**
1. Download a glTF model file with morph animation. There are at least 3 in the Khronos sample models, so pick one:
   + [AnimatedMorphCube](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphCube/glTF)
   + [AnimatedMorphSphere](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphSphere/glTF)
   + [MorphPrimitivesTest](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MorphPrimitivesTest)
2. Import the glTF model to an `aiScene`.
3. Iterate over the meshes in the `aiScene`; verify that all of them have `mesh.mMethod == 0x0`

**Expected behavior**
At least one of the meshes (specifically, all those with `mNumAnimMeshes > 0`) should have a determined morph method, either `aiMorphingMethod_VERTEX_BLEND` (0x1) or `aiMorphingMethod_MORPH_NORMALIZED` (0x2) or `aiMorphingMethod_MORPH_RELATIVE` (0x3).

If none of these methods accurately describes the morph method in the glTF spec, then another (non-zero) method should be defined in "mesh.h" .

**Screenshots**
not applicable

 - OS: Mint Linux
 - Browser: Firefox
 - Version: the 59b12344 commit of Assimp) (2023-09-14)

Chang-Jin-Lee:
Closes #5303.

## The problem

The glTF2 importer never assigns `aiMesh::mMethod`, so every mesh with morph targets comes back `aiMorphingMethod_UNKNOWN`. A caller receives a set of `aiAnimMesh` targets with no statement of how to combine them. Collada is currently the only importer that sets the field.

## Which value, and why it is not the obvious one

glTF phrases morph targets as displacements, so `aiMorphingMethod_MORPH_RELATIVE` looks like the answer. It is the wrong one here, because the field describes the data assimp hands back rather than the wording of the source format, and this importer does not hand back displacements.

`aiCreateAnimMesh` memcpy's the base mesh into each new `aiAnimMesh` (`code/Common/CreateAnimMesh.cpp`), and `ImportMeshes` then adds the target's delta on top of that copy:

```cpp
aiAnimMesh.mVertices[vertexId] += positionDiff[vertexId];
```

So each stored target is an absolute morphed position. `glTF2Exporter` is the mirror image and says so in its own comment — it subtracts the base again on the way out:

```cpp
// NOTE: in gltf it is the diff stored
pPositionDiff[vt] = pAnimMesh->mVertices[vt] - aim->mVertices[vt];
```

Absolute targets are what `aiMorphingMethod_MORPH_NORMALIZED` describes, and its blend reproduces glTF's exactly. Substituting `target[i] = base + delta[i]` into

```
base * (1 - sum(w)) + sum(w[i] * target[i])
```

gives `base + sum(w[i] * delta[i])`, which is the glTF formula. Tagging `MORPH_RELATIVE` would instead tell a caller to add the absolute targets onto the base a second time.

Nothing else changes: the assignment sits inside the existing `if (!targets.empty())`, so a mesh without morph targets still reports `UNKNOWN`.

## Testing

`importMorphingMethodIsSet` imports the `AnimatedMorphCube` fixture already in `test/models/`, asserts the method on every mesh that carries targets, and asserts meshes without targets keep `UNKNOWN`. It fails on master with `0 != 2` and passes with the change. No new assets.

```
$ bin/unit --gtest_filter="*glTF2*"
51 tests ... [  PASSED  ] 51 tests.

$ bin/unit
```

## One open question for you

`SimpleMorph`, `MorphPrimitivesTest`, `MorphStressTest` and `AnimatedMorphSphere` are also in `test/models/glTF2/` and would extend the same assertion cheaply. I kept the test to one fixture to stay small; say the word if you would rather it cover all of them.

## Summary by CodeRabbit

- **Improvements**
  - glTF animation imports now support normalized 8-bit and 16-bit integer rotation and morph-weight data.
  - Morph-target meshes now use normalized morph blending semantics for improved compatibility.
  - Unsupported animation component formats are reported as import errors.

- **Tests**
  - Added coverage verifying normalized morphing methods for meshes with and without morph targets.

---

### DIFF
--- a/code/AssetLib/glTF2/glTF2Importer.cpp
+++ b/code/AssetLib/glTF2/glTF2Importer.cpp
@@ -646,6 +646,12 @@ void glTF2Importer::ImportMeshes(glTF2::Asset &r) {
 
             std::vector<Mesh::Primitive::Target> &targets = prim.targets;
             if (!targets.empty()) {
+                // Each target below is resolved to absolute positions: aiCreateAnimMesh
+                // copies the base mesh and the target's displacement is added on top.
+                // Blending those the way aiMorphingMethod_MORPH_NORMALIZED describes,
+                // base * (1 - sum(w)) + sum(w[i] * target[i]), reproduces glTF's
+                // base + sum(w[i] * displacement[i]) exactly.
+                aim->mMethod = aiMorphingMethod_MORPH_NORMALIZED;
                 aim->mNumAnimMeshes = (unsigned int)targets.size();
                 aim->mAnimMeshes = new aiAnimMesh *[aim->mNumAnimMeshes];
                 std::fill(aim->mAnimMeshes, aim->mAnimMeshes + aim->mNumAnimMeshes, nullptr);
--- a/test/unit/utglTF2ImportExport.cpp
+++ b/test/unit/utglTF2ImportExport.cpp
@@ -1151,3 +1151,30 @@ TEST_F(utglTF2ImportExport, importMalformedSparseAccessor) {
     std::string errorString = importer.GetErrorString();
     EXPECT_NE(errorString.find("Invalid sparse accessor: missing required 'values' object."), std::string::npos);
 }
+
+TEST_F(utglTF2ImportExport, importMorphingMethodIsSet) {
+    // Issue #5303: glTF morph meshes came back aiMorphingMethod_UNKNOWN because
+    // the importer never assigned mMethod.
+    Assimp::Importer importer;
+    const aiScene *scene = importer.ReadFile(ASSIMP_TEST_MODELS_DIR "/glTF2/glTF-Sample-Models/AnimatedMorphCube-glTF/AnimatedMorphCube.gltf",
+            aiProcess_ValidateDataStructure);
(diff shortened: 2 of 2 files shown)


Chang-Jin-Lee:
Rebased onto master. This had gone `CONFLICTING` when #6831 landed, since both add a test at the end of `utglTF2ImportExport.cpp`, so it could not have been merged even if it had been reviewed. It is one commit on master now — `glTF2Importer.cpp` +6 and the one test — with 12 checks green and 2 skipped.

PR #6830, opened 2026-09-06
glTF: refuse a float array with a non-numeric entry instead of reporting success

### LINKED ISSUE #6190: Bug: GLTF importer potentially uses uninitialized variable
See gltfCommon.h

This function is called by the GLTF reader, for instance to read scale values.

```cpp
template <class T>
struct ReadHelper<Nullable<T>> {
    static bool Read(Value &val, Nullable<T> &out) {
        return out.isPresent = ReadHelper<T>::Read(val, out.value);
    }
};
```

Note that it uses the return value of the other ReadHelper to store whether the value can be used.

For vec3 and vec4 it then calls 

```cpp
template <unsigned int N>
struct ReadHelper<float[N]> {
    static bool Read(Value &val, float (&out)[N]) {
        if (!val.IsArray() || val.Size() != N) return false;
        for (unsigned int i = 0; i < N; ++i) {
            if (val[i].IsNumber())
                out[i] = static_cast<float>(val[i].GetDouble());
        }
        return true;
    }
};
```

This function has a bug. It checks whether val is a number, but if not, it just ignores the value. In the end it returns true anyway.
Since it returns false in all other error cases (val not an array, or val size not as expected), I'd say it should also return false, if any element of the array is not a number.

Chang-Jin-Lee:
Closes #6190.

## The problem

`ReadHelper<float[N]>::Read` skipped array entries that are not numbers, then returned `true` anyway:

```cpp
if (!val.IsArray() || val.Size() != N) return false;
for (unsigned int i = 0; i < N; ++i) {
    if (val[i].IsNumber())
        out[i] = static_cast<float>(val[i].GetDouble());
}
return true;
```

Its two sibling error paths — a non-array, and an array of the wrong size — both return `false`. Only this case reported a successful read of a value it had not finished reading, which is what @jankrassnigg pointed out in the issue.

## Why it is reachable

`Nullable` stores its value with no initialiser (`T value;`), and takes `isPresent` straight from the reader's return:

```cpp
template <class T>
struct ReadHelper<Nullable<T>> {
    static bool Read(Value &val, Nullable<T> &out) {
        return out.isPresent = ReadHelper<T>::Read(val, out.value);
    }
};
```

`Node` holds `Nullable<vec3> translation`, `Nullable<vec4> rotation` and `Nullable<vec3> scale`. So a node carrying

```json
"scale": [2.0, "not-a-number", 3.0]
```

came out with `isPresent == true` over a value whose middle component was never assigned, and `glTF2Importer` then fed all three components through `CopyValue` into the node transform.

## The change

The array is validated in full before any of it is written. A malformed array is refused whole, the same way a wrong size already is, and the caller's array is left untouched rather than half-updated. That second part matters for the direct, non-`Nullable` reads — the node `"matrix"` path goes through `ReadValue` and `ReadMember(obj, "translation", ...)` ignores its result — where a partial write would otherwise persist.

Valid files are unaffected: an all-numeric array takes the same path and produces the same values.

## Testing

The test needs a file that is malformed in this specific way and no shipped asset is, so it adds a three-line `malformed_node_scale.gltf` alongside the existing `malformed_sparse.gltf`. It asserts the node keeps its default transform. On master it fails deterministically — the scale comes back `3` on z — so the assertion does not depend on whatever the unassigned component happens to contain.

```
$ bin/unit --gtest_filter="*glTF*"
55 tests from 2 test suites ran. [  PASSED  ] 55 tests.

$ bin/unit
```

## Note

This is independent of #6829, which I opened earlier today against `glTF2Importer.cpp`. The two touch different files and can be reviewed in either order.

## Summary by CodeRabbit

- **Bug Fixes**
  - Malformed numeric arrays are now rejected as a whole when they contain invalid entries.
  - Invalid node scale data no longer partially updates transforms; affected nodes retain their default scale instead of applying incomplete values.
  - Import handling is now more consistent and predictable when scale arrays contain non-numeric values.

- **Tests**
  - Added coverage to verify correct handling of malformed node scale data during import.

---

### DIFF
--- a/code/AssetLib/glTFCommon/glTFCommon.h
+++ b/code/AssetLib/glTFCommon/glTFCommon.h
@@ -302,9 +302,16 @@ template <unsigned int N>
 struct ReadHelper<float[N]> {
     static bool Read(Value &val, float (&out)[N]) {
         if (!val.IsArray() || val.Size() != N) return false;
+        // Check the whole array before writing any of it. An entry that is not
+        // a number used to be skipped while Read still reported success, so a
+        // Nullable<vec3> went isPresent over a value whose skipped component
+        // was never assigned. Refuse the array whole, the way a wrong size or a
+        // non-array already is refused, and leave out untouched either way.
         for (unsigned int i = 0; i < N; ++i) {
-            if (val[i].IsNumber())
-                out[i] = static_cast<float>(val[i].GetDouble());
+            if (!val[i].IsNumber()) return false;
+        }
+        for (unsigned int i = 0; i < N; ++i) {
+            out[i] = static_cast<float>(val[i].GetDouble());
         }
         return true;
     }
--- a/test/models/glTF2/malformed_node_scale.gltf
+++ b/test/models/glTF2/malformed_node_scale.gltf
@@ -1,0 +1,23 @@
+{
+  "asset": {
+    "version": "2.0"
+  },
+  "scene": 0,
+  "scenes": [
+    {
+      "nodes": [
+        0
+      ]
+    }
+  ],
+  "nodes": [
+    {
+      "name": "BadScale",
+      "scale": [
+        2.0,
+        "not-a-number",
+        3.0
+      ]
+    }
+  ]
+}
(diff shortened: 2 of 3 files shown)


Chang-Jin-Lee:
Rebased onto master. This had gone `CONFLICTING` when #6831 landed, since both add a test at the end of `utglTF2ImportExport.cpp`, so it could not have been merged even if it had been reviewed. It is one commit on master now — `glTFCommon.h` +9/-2, the malformed model, and the one test — with 12 checks green and 2 skipped.

PR #6543, opened 2026-02-25
glTF2: Preserve interpolation type and CubicSpline tangent data in an…

EarendelArc:
…imation keys

fix(gltf2): preserve interpolation type and CubicSpline tangents

The glTF2 importer previously ignored the mInterpolation field and discarded CubicSpline tangent data, leading to incorrect animation playback.

- Implement MapInterpolation() to map glTF2 interpolation types to Assimp.
- Set mInterpolation for all aiVectorKey and aiQuatKey instances.
- For CUBICSPLINE samplers, store [in-tangent, value, out-tangent] triplets (N x 3 keys) instead of discarding tangents.
- Fixes rendering consistency for InterpolationTest.glb where STEP, LINEAR, and CUBICSPLINE rows previously rendered identically.

Affected channels: mPositionKeys, mRotationKeys, mScalingKeys.

## Summary by CodeRabbit

* **New Features**
  * Added support for cubic spline interpolation when importing animations from glTF files, enabling smoother animation playback for translation, rotation, and scaling keyframes.

Issue #5668, opened 2024-07-15
Importing GLTF doesn't specify interpolation for aiNodeAnim type


m0tholith:
**Describe the technical debt**
Animation types in the .gltf importer are not exposed (at least not to the C API).

**Better solution**
I suggest adding an `mInterpolation` field to the `aiNodeAnim` class, and using `Interpolation_LINEAR` as a default value for filetypes that don't support specifying the type of interpolation; that way different animation channels can use different interpolation methods (mimicking how .gltf specifies interpolation types).

**Desktop (please complete the following information):**
 - OS: NixOS 24.05
 - Browser: firefox
 - Version: 5.4

**Additional context**
This issue is pretty much a bump of #4890 since I'm actually trying to use the exact same file that issue's OP was trying to use.

kimkulling:
This is true for the whole interface. SO all loaders a missing this information. There is the implicite rule to use linear interpolation, which is not good enough from my point of view.

tellypresence:
Feature request issues tracked via epic #6183

Chang-Jin-Lee:
This looks resolved on master, and I think it can be closed.

`aiAnimInterpolation` landed the day after your comment here, in `da281b7f`. It went onto `aiVectorKey` and `aiQuatKey` as a per-key field, one level below the `aiNodeAnim` this issue proposed. That turned out to be the more expressive home: COLLADA specifies interpolation per key, so a channel-level field would have had to discard that.

glTF started populating it in #6543 (`8ef1461c`, February), which maps the sampler mode and keeps the CUBICSPLINE tangents. Both commits are in v6.0.5.

I checked this on current master rather than reading it off the diff — three minimal files, one per sampler mode, imported and dumped:

```
STEP         -> key 0 Step,       key 1 Step
LINEAR       -> key 0 Linear,     key 1 Linear
CUBICSPLINE  -> 6 keys, all CubicSpline (3 per keyframe: in-tangent, value, out-tangent)
```

`mInterpolation` is declared `C_ENUM aiAnimInterpolation`, so it reaches the C API as well, which was the original complaint in #4890.

One gap is left behind. `grep -rn mInterpolation test/unit/` finds nothing, so none of this has regression coverage. I am happy to send a test if that would be useful.

PR #6831, opened 2026-09-06
glTF2: cover animation interpolation with a regression test

Chang-Jin-Lee:
Follow-up to #5668, which I think can be closed — the feature it asks for is already on master. What is missing is coverage.

`da281b7f` introduced `aiAnimInterpolation` on `aiVectorKey` and `aiQuatKey`, and #6543 taught the glTF2 importer to fill it from the sampler. `grep -rn mInterpolation test/unit/` finds nothing, so that mapping has never been guarded.

The gap is easy to miss because a regression here is silent. Every sampler mode keeps importing, and `aiAnimInterpolation_Linear` is also the value a key is constructed with, so a broken mapping still yields a scene that loads and animates — just always linearly. Nothing throws and no count changes.

## What is in here

Three fixtures, one per sampler mode, each a single node with one translation channel and an inline base64 buffer, so no external asset is pulled in. The CUBICSPLINE file also pins the key count at six, since that mode stores an in-tangent and an out-tangent alongside each keyframe value and those arrive as keys of their own.

No source changes.

## Verification

I checked that the test catches the regression it names rather than only passing today. Pointing `MapInterpolation`'s `STEP` and `CUBICSPLINE` cases at `aiAnimInterpolation_Linear`:

```
expected: aiAnimInterpolation_Step (0)          actual: 1   animation_interpolation_step.gltf
expected: aiAnimInterpolation_Cubic_Spline (3)  actual: 1   animation_interpolation_cubic.gltf
```

Restoring the mapping turns it green again.

```
$ bin/unit --gtest_filter="*importAnimationInterpolationIsPreserved*"
[  PASSED  ] 1 test.

$ bin/unit
```

Independent of #6829 and #6830; this one touches only `test/`.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for glTF2 animation imports.
  * Verified that STEP, LINEAR, and CUBICSPLINE interpolation modes are preserved, along with the expected animation structure and position-key data.
  * This helps ensure imported animations retain their intended motion behavior across supported interpolation types.

---

### DIFF
--- a/test/models/glTF2/animation_interpolation_cubic.gltf
+++ b/test/models/glTF2/animation_interpolation_cubic.gltf
@@ -1,0 +1,77 @@
+{
+ "asset": {
+  "version": "2.0"
+ },
+ "scene": 0,
+ "scenes": [
+  {
+   "nodes": [
+    0
+   ]
+  }
+ ],
+ "nodes": [
+  {
+   "name": "Mover"
+  }
+ ],
+ "buffers": [
+  {
+   "byteLength": 80,
+   "uri": "data:application/octet-stream;base64,AAAAAAAAgD8AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIA/AAAAAAAAAAAAAAAAAAAAAAAAAAA="
+  }
+ ],
+ "bufferViews": [
+  {
+   "buffer": 0,
+   "byteOffset": 0,
+   "byteLength": 8
+  },
+  {
+   "buffer": 0,
+   "byteOffset": 8,
+   "byteLength": 72
+  }
+ ],
+ "accessors": [
+  {
+   "bufferView": 0,
+   "componentType": 5126,
+   "count": 2,
+   "type": "SCALAR",
+   "min": [
+    0.0
+   ],
+   "max": [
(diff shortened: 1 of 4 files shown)


kimkulling (approved):
Looks fine for me.

Chang-Jin-Lee:
Pushed a one-character-per-side fix for the red CI. `std::array` wraps a C array, so the aggregate needs the inner braces; with a single level every job rejected it:

```
test/unit/utglTF2ImportExport.cpp:1171:5: error: too many initializers for
'const std::array<Expectation, 3>'
```

Kept the `std::array` you switched to rather than reverting to the plain array, since that was the point of the edit.

Built locally with `-DASSIMP_WARNINGS_AS_ERRORS=ON` to match CI: compiles clean, `importAnimationInterpolationIsPreserved` passes, and the wider glTF suite is 55 of 55.

Chang-Jin-Lee:
One thing that is not visible from the PR page: that push put the workflows back behind the first-time-contributor gate. All three runs on `test/gltf2-animation-interpolation` from 11:20 are sitting at `action_required`, so the red result you saw is still the last one that executed. They need approving again before the green shows up here.

<sub>Assisted-by: Claude (Anthropic)</sub>

kimkulling:
Merged, thnk you for your contribution.

PR #6832, opened 2026-09-06
Fix heap-buffer-overflow reading MD2 skin names

afonsojanu:
The skin name in an MD2 file lives in a fixed 64-byte field copied directly out of the file, and nothing guarantees it actually has a NUL terminator anywhere inside it. On a malformed file where that field is filled with non-zero bytes right up to the end of the file, `strlen()` on `pcSkins->name` just keeps going past the end of the read buffer looking for a byte that isn't there.

Fixed it by bounding the scan with `strnlen()`, capped at one less than the field's own size. `MDCLoader` already does the same thing for its own fixed-size name fields (`ucName`), so this brings MD2 in line with that.

Added a regression test with a small hand-built MD2 file: one skin, name field packed entirely with non-NUL bytes, ending exactly at the file's last byte. Confirmed with an ASan build that this crashes with a heap-buffer-overflow in `strlen` on unmodified master, and imports cleanly with the fix applied. Full unit suite (666 tests) passes clean under ASan+UBSan.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed MD2 texture names to prevent reads beyond the available data.
  - MD2 files with full-length, unterminated skin-name fields now import successfully.

- **Tests**
  - Added regression coverage for importing MD2 files with unterminated texture names.


kimkulling (approved):
Looks fine for me.

kimkulling:

PR #6834, opened 2026-09-08
fix collada memory leaks when an import fails midway

YuEfSaEDU:
OSS-Fuzz #6791 reports a direct leak in `ColladaLoader::BuildHierarchy`: when a node's mesh cannot be built — in the reproducer, a skin controller references a nonexistent joints source and `CreateMesh` throws `DeadlyImportError` — the exception unwinds through `BuildHierarchy` and every `aiNode` allocated so far is lost, because the partially built hierarchy is held only through raw pointers and `pScene->mRootNode` is never assigned.

This change gives each in-flight node `unique_ptr` ownership that is released only when the node is returned to its parent, and zero-initializes the `mChildren` slot array so a partially filled array is destroyed safely. The same reproducer exposes two further leaks on the same path, fixed here as well: the temporary materials in `newMats` are owned by `unique_ptr` until `StoreSceneMaterials` transfers them to the scene, and `PostProcessRootAnimations` now deletes the sub-animations replaced by its `mAnims = temp` shallow assignment — the latter affects valid files containing `<animation_clip>`, not just malformed input.

Verified with the OSS-Fuzz reproducer added as a regression test (`utColladaImportExport.parseInvalid6791`) and a Windows debug-CRT leak harness: master leaks 156 blocks on the reproducer including eight `aiNode` objects, while with this patch the import fails cleanly with zero leaked blocks and `library_animation_clips.dae` still imports successfully. The full unit suite passes (675 tests on MSVC); LeakSanitizer coverage of the fuzz target itself runs on the Linux fuzzing infrastructure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Collada asset loading reliability when files contain invalid or incomplete node, joint, or material data.
  * Prevented memory leaks during Collada animation processing and error handling.
  * Ensured failed imports exit safely without crashes.

* **Tests**
  * Added regression coverage for invalid Collada files with unresolved joint references.


turol:
Need rebase.

PR #6839, opened 2026-09-15
Optimize XGL loader

turol:
Use unordered containers in XGL loader. Saves about 15% time in XGL unit tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved XGL model importing for mesh, material, vertex, normal, and texture-coordinate data.
  - Preserved deterministic material-based mesh ordering during import for consistent results.
  - Improved efficiency when processing XGL scene data.

- **Tests**
  - Expanded XGL import coverage across supported sample files.
  - Added validation for scene flags, meshes, materials, animations, textures, lights, cameras, skeletons, vertices, and faces.
  - Confirmed expected mesh ordering and imported scene structure across test models.


turol:
I don't think multiple materials with the same id attribute is valid. None of our current test models have that. If someone has such a thing we can look at it later.

PR #6853, opened 2026-09-16
Fix OOB read / uninitialized bytes in openddlparser Value::getFloat/getDouble

iliasabk:
Fixes the OSS-Fuzz-reported `Use-of-uninitialized-value` that surfaces in `aiMatrix4x4t<float>::Decompose` (issue #6790).

**Root cause:** in `contrib/openddlparser/code/Value.cpp`, the fallback branches used for type-mismatched values copy a fixed 4 bytes regardless of the stored size:

```cpp
float tmp;
::memcpy(&tmp, m_data, 4);   // getFloat: OOB read when m_size < 4 (bool/int8/int16/half)
...
double tmp;
::memcpy(&tmp, m_data, 4);   // getDouble: OOB read AND upper 4 bytes of tmp left uninitialized
```

A malformed OpenGEX file can declare e.g. `transform{float[16]{...}}` containing non-float literals; `setMatrix` calls `getFloat()` on those values, the out-of-bounds/uninitialized bytes flow into `aiNode::mTransformation`, and MSan reports the use at the later `Decompose` call site.

**Fix:** zero-initialize the temporary and copy at most `m_size` bytes, preserving the existing reinterpretation semantics for `m_size >= sizeof(tmp)` while eliminating the out-of-bounds read and the uninitialized tail.

---
*This PR was prepared by an LLM-based coding agent, which is available to respond to review feedback and make follow-up changes.*


turol:
Openddlparser is a separate project. Please send any changes upstream
https://github.com/kimkulling/openddl-parser

Issue #6466, opened 2026-02-03
Bug:Floating-point division by zero in glTF2Exporter / glTFExporter when mTicksPerSecond == 0


JungWooJJING:
## Description

The glTF2 exporter and glTF1 exporter divide by `aiAnimation::mTicksPerSecond` without checking for zero.
When importing a file format that does not specify a tick rate (e.g., 3DS), this field stays at its default `0.0`, causing undefined behavior (CWE-369).

`ExportAnimations()` in `glTF2Exporter.cpp` converts tick-based keyframe times to seconds by dividing by `ticksPerSecond`.
Three functions — `ExtractTranslationSampler`, `ExtractScaleSampler`, `ExtractRotationSampler` — all perform this division without a zero guard.
The FBX exporter handles this correctly with `std::fpclassify()`, but the glTF exporters do not.

## Impact

- Exported glTF files contain `inf`/`NaN` in animation timestamps → invalid glTF spec, downstream viewers (Blender, Three.js, Unity) may crash
- Platforms that trap FP exceptions → `SIGFPE` crash (DoS)
- Undefined behavior per C++ standard

## Vulnerable Code

**glTF2Exporter.cpp:1703** — no zero check:
```cpp
const float ticksPerSecond = static_cast<float>(anim->mTicksPerSecond);
```

**glTF2Exporter.cpp:1639, 1658, 1677** — division by zero:
```cpp
// ExtractTranslationSampler (line 1639)
times[i] = static_cast<float>(key.mTime / ticksPerSecond);

// ExtractScaleSampler (line 1658)
times[i] = static_cast<float>(key.mTime / ticksPerSecond);

// ExtractRotationSampler (line 1677)
times[i] = static_cast<float>(key.mTime / ticksPerSecond);
```

**glTFExporter.cpp:942** — same bug in glTF1 exporter:
```cpp
timeData[i] = static_cast<float>(nodeChannel->mPositionKeys[frameIndex].mTime / ticksPerSecond);
```

## Root Cause

`aiAnimation::mTicksPerSecond` is documented as `0 if not specified in the imported file` (`include/assimp/anim.h`).
Multiple importers leave this at the default `0.0`:

- **3DS** — format has no tick rate field, never sets `mTicksPerSecond`
- **glTF2** — explicitly sets `ai_anim->mTicksPerSecond = 0` (glTF2Importer.cpp:1527)
- **X (DirectX)**, **ASE** — may leave it at 0

The FBX exporter handles this correctly:
[...]

PR #6867, opened 2026-09-16
glTF exporters: handle aiAnimation::mTicksPerSecond == 0

iliasabk:
## Summary

Fixes a floating-point division by zero in both glTF exporters reported in #6466.

**Root cause:** `aiAnimation::mTicksPerSecond == 0` means "not specified in the imported file" (see `anim.h`). Both `glTF2Exporter::ExportAnimations()` and `glTFExporter` divided key times by it unconditionally (`key.mTime / ticksPerSecond`), producing `inf`/`NaN` in the exported sampler input accessor — flagged by UBSan as `float-divide-by-zero` and yielding non-finite glTF output.

**Fix:** treat an unspecified tick rate as seconds (divisor 1.0), matching the existing defensive handling in the FBX exporter (`FBXExporter.cpp:1046`).

**Verification:** a scene with `mTicksPerSecond == 0` and two position keys (t=0, t=2) exported to `glb2` previously wrote `NaN`/`+inf` into the buffer — now writes finite values. Both `glb` (v1) and `glb2` exporters verified.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed animation export for glTF and glTF2 files when the source does not specify a ticks-per-second rate.
  - Animation keyframe times are now interpreted as seconds in this case, preventing invalid timing caused by division by zero.
  - This improves the reliability of exported translation, rotation, and scale animations.
  - Existing behavior remains unchanged for animations with an explicitly defined rate.


kimkulling (approved):
Looks fine for me.

kimkulling:

Issue #6761, opened 2026-08-06
Bug:Blender importer computes file offset from attacker-controlled pointer values → StreamReader::SetCurrentPos pointer arithmetic overflow / UBSan SIGILL (CWE-682)


1820893135-pixel:
**Describe the bug**

In the Blender SDNA/DNA decode path, `Structure::ReadCustomDataPtr()` (BlenderDNA.inl:337) computes a read position directly from pointer fields in the .blend binary file:

```c
db.reader->SetCurrentPos(block->start + (ptrval.val - block->address.val))
```

`block->start`, `ptrval.val`, and `block->address.val` are all read from untrusted file bytes without any validation. `StreamReader::SetCurrentPos()` then executes `SetPtr(mBuffer + pos)` (StreamReader.h:255). When a malformed .blend file overwrites the mesh CustomData DATA block header with `0xFF`, the computed `pos` becomes huge, `mBuffer + pos` leaves address-space bounds, and UBSan flags it as **pointer arithmetic overflowed bounds** → `__builtin_trap` → **SIGILL (exit 132)**. The offset is entirely file-controlled, and the defect is reached early in scene parsing through a 22-frame call chain.

**To Reproduce**

1. Download the seed file from assimp's repository:
   **https://github.com/assimp/assimp/blob/master/test/models/Blender/BlenderDefault_248.blend**
   (125,148 bytes, MD5 `42c14b7788674fa8b81d3a00db1dfc11`). Click "Download raw file" or use:
   ```bash
   wget https://raw.githubusercontent.com/assimp/assimp/master/test/models/Blender/BlenderDefault_248.blend
   ```

2. Generate the POC by overwriting the mesh CustomData pointer fields with 0xFF:

   ```bash
   # Offset 0x13B40, 64 bytes to 0xFF:
   python3 -c "data=bytearray(open('BlenderDefault_248.blend','rb').read());[data.__setitem__(0x13B40+i,0xFF) for i in range(64)];open('assimp_poc4_customdata_ptr.bin','wb').write(data)"
   ```

   Or use the generator script (`generate_assimp_pocs.py`, attached) to create all 4 POC files at once:
   ```bash
   python3 generate_assimp_pocs.py BlenderDefault_248.blend
   ```

   Offset 0x13B40 sits immediately after a `DATA` file-block header (block code `DATA` at 0x13B38) and is the start of a mesh CustomDataLayer payload. This region holds the `Pointer` struct's `address` and `start` fields. The region changes from:

```
[...]

PR #6868, opened 2026-09-16
StreamReader+Blender: validate seek offsets before pointer arithmetic

iliasabk:
## Summary

Fixes a UBSan pointer-arithmetic-overflow (SIGILL on some builds) reported in #6761, reproducible via the mutation recipe in the issue (overwrite a CustomData DATA block header with 0xFF in `BlenderDefault_248.blend`).

**Root cause (two layers):**
1. `StreamReader::SetCurrentPos()` formed `mBuffer + pos` *before* any range check. A corrupted offset made the pointer leave the buffer object entirely — undefined behaviour flagged by UBSan's `pointer-overflow` before `SetPtr()`'s own check could run.
2. `Blender::LocateFileBlockForAddress()` used `lower_bound()`, which returns the first block starting *at or after* the address — for a pointer into a block's interior that is the *next* block, so `ptrval - block->address` underflowed to a huge `size_t` fed into `SetCurrentPos()`.

**Fix:**
- `SetCurrentPos()` now rejects `pos` beyond the stream extent before forming the pointer (same validate-before-arithmetic pattern `IncPtr()` already uses).
- `LocateFileBlockForAddress()` uses `upper_bound()` + decrement to find the *last block starting at or before* the address — the block that can actually contain it — and performs the end check via subtraction, which cannot overflow.

**Verification:**
- The #6761 reproducer previously trapped at `StreamReader.h:255` (`addition of unsigned offset ... overflowed`); it now fails cleanly with `DeadlyImportError: Failure resolving pointer 0x50225408`.
- Regression: 10 valid BLEND fixtures (`fleurOptonl`, `Bob`, `BlenderDefault_*`, `MirroredCube_252`, `TexturedPlane_*` …) still import correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent seeking beyond available buffered data.
  * Improved error handling for end-of-file and read-limit conditions.
  * Reduced the risk of invalid memory access and crashes when reading truncated or malformed file data.

* **Style**
  * Standardized internal formatting without changing application behavior.


kimkulling (approved):
Looks fine.

kimkulling:

PR #6726, opened 2026-07-07
Harden MDL importer metadata parsing

Alb3e3:
## Summary
- validate file-controlled HL1 and MDL7 counts before allocations and cursor arithmetic
- bounds-check HL1 compressed animation value offsets/runs and related skin, bone, sequence, and mesh indices
- fix MDL7 one-past vertex/UV checks and avoid unaligned MDL7/HL1 metadata reads

## Why
Malformed MDL files can use oversized metadata counts, invalid animation offsets, and one-past indices to drive out-of-bounds accesses, sanitizer aborts, or excessive allocations during import.

## Validation
- `git diff --check`
- `cmake --build build-asan --target assimp -j2`
- relinked `../../fuzzers/fz_mdl_metadata` with ASan/UBSan against `build-asan/lib/libassimp.a`
- `ASAN_OPTIONS=allocator_may_return_null=1:detect_leaks=0:abort_on_error=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 timeout 90 ../../fuzzers/fz_mdl_metadata -runs=0 ../../crashes/mdl ../../cmin/mdl ../../verify/mdl-current-crashes ../../crashes/hl1-mdl-metadata-final` (`Done 76 runs`)
- `ASAN_OPTIONS=allocator_may_return_null=1:detect_leaks=0:abort_on_error=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 timeout 100 ../../fuzzers/fz_mdl_metadata -artifact_prefix=../../crashes/hl1-mdl-metadata-final/ -max_total_time=60 -rss_limit_mb=2048 ../../fuzz-work/mdl-metadata-corpus ../../cmin/mdl ../../crashes/mdl ../../verify/mdl-current-crashes ../../crashes/hl1-mdl-metadata-final` (`Done 14409 runs in 61 second(s)`)

## Tooling
This patch was developed with AI assistance and reviewed/tested locally before submission.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation and error handling when loading Half-Life 1 and MDL7 model files.
  - Prevented crashes and invalid memory access caused by malformed, incomplete, or oversized model data.
  - Hardened parsing for textures, skins, meshes, bones, animations, attachments, hitboxes, and transitions.
  - Improved animation data validation and handling of invalid references.
  - Fixed MDL7 boundary checks for face, UV, vertex, frame, and grouped-frame data.
[...]

Issue #6631, opened 2026-05-06
Bug: Heap-buffer-overflow in Assimp::MDLImporter::GenerateOutputMeshes_3DGS_MDL7


TYGLS:
**Describe the bug**

Heap-buffer-overflow in `Assimp::MDLImporter::GenerateOutputMeshes_3DGS_MDL7` function at `code/AssetLib/MDL/MDLLoader.cpp:1944` caused by out-of-bounds write to `pcMesh->mBones` when the loop index `iCurrent` exceeds the allocated array size.

**To Reproduce**

1. Clone the assimp repository and build it refer to oss-fuzz.

```shell
export CC=clang \
    CXX=clang++ \
    CFLAGS='-fsanitize=address -O0 -g' \
    CXXFLAGS='-fsanitize=address -O0 -g' \
    LIB_FUZZING_ENGINE='-fsanitize=fuzzer'
```

2. Run the PoC using assimp_fuzzer:

[poc.zip](https://github.com/user-attachments/files/27425495/poc.zip)

   The PoC is provided as a zip archive. After extracting, run:

```shell
./assimp_fuzzer ./poc.mdl
```

**ASAN Report**

```
=================================================================
==2565==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x5040000005c8 at pc 0x5fb05af47230 bp 0x7ffe51596b30 sp 0x7ffe51596b28
WRITE of size 8 at 0x5040000005c8 thread T0
    #0 0x5fb05af4722f in Assimp::MDLImporter::GenerateOutputMeshes_3DGS_MDL7(Assimp::MDL::IntGroupData_MDL7&, Assimp::MDL::IntSplitGroupData_MDL7&) /home/hexijie/fuzz/project/assimp/code/AssetLib/MDL/MDLLoader.cpp:1944:63
    #1 0x5fb05af37cf8 in Assimp::MDLImporter::InternReadFile_3DGS_MDL7() /home/hexijie/fuzz/project/assimp/code/AssetLib/MDL/MDLLoader.cpp:1552:9
    #2 0x5fb05af2d498 in Assimp::MDLImporter::InternReadFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, aiScene*, Assimp::IOSystem*) /home/hexijie/fuzz/project/assimp/code/AssetLib/MDL/MDLLoader.cpp:243:13
[...]

PR #6870, opened 2026-09-16
MDL7: fix OOB write indexing mBones by weight-list position

iliasabk:
## Summary

Refs #6631 — `MDLImporter` allocates `pcMesh->mBones` with `mNumBones` entries (the count of *non-empty* per-bone weight lists) but wrote into it using `iCurrent`, the position within `aaiVWeightList` (which has `iNumOutBones` slots). When a bone earlier in the list has no vertex weights, any later non-empty slot writes past the end of the array — heap-buffer-overflow write.

The write now uses a separate output-slot counter; `iCurrent` keeps indexing `aaiVWeightList`/`apcOutBones` (both `bones_num`-sized).

Note: the original testcase is currently rejected earlier by the bounds validation added in #6726, so the crash is no longer reachable with that file — but the indexing mismatch remains a latent OOB write for any MDL7 mesh whose weight lists have gaps (e.g. vertices bound to bones {0,2} but not 1).

## Test plan

- [x] MDL7 fixtures still import (`test/models/MDL/MDL7 (3DGS A7)`)
- [x] ASan build clean

Generated with [Devin](https://devin.ai)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed skeletal mesh generation so bones with valid weight data are stored consistently with the correct names and offset matrices.
  * Improved handling of empty bone weight entries to prevent gaps, mismatches, or incorrect skeletal assignments.
  * Improved error handling during bone setup to prevent partially created bone data when processing fails.


kimkulling:

PR #6872, opened 2026-09-18
FBX: accept integer texture UV transform components

### LINKED ISSUE #6838: Bug: Dropping textures from fbx generated by Agisoft Metashape
**Describe the bug**

The FBX importer silently discards a diffuse texture that is embedded in the file and correctly connected to the material. The resulting aiMaterial has no texture property of any type, and aiScene::mNumTextures is 0, even though the file contains a valid Video object holding a 13.5 MB JPEG and a well-formed OP connection from the Texture to the material's DiffuseColor property.

The file is produced by Agisoft Metashape (photogrammetry export, binary FBX 7700, single mesh, single material, one UV channel). The same file loads with its texture in Blender, and re-exporting it from Blender produces an FBX that assimp loads correctly.

Inspecting the raw FBX node tree shows nothing unusual:

Objects:

Material 'material0'
Texture  'texture0'   (Type, Version, TextureName, Properties70, Media,
                       FileName, RelativeFilename, ModelUVTranslation,
                       ModelUVScaling, Texture_Alpha_Source, Cropping)
Video    'texture0'   (Content: 13,560,596 bytes, JPEG magic FF D8 FF;
                       RelativeFilename: "...\model-textured.jpg")

Connections:

OO  Geometry 'Model'     ->  Model 'Model'
OO  Material 'material0' ->  Model 'Model'
OP  Texture 'texture0'   ->  Material 'material0'  ["DiffuseColor"]
OO  Video 'texture0'     ->  Texture 'texture0'

To Reproduce

Export a textured model from Agisoft Metashape as binary FBX with embedded texture. ([Download model](https://cybid-my.sharepoint.com/:u:/g/personal/maciej_mucha_cybid_com_pl/IQD6bylJdOw3S4DE8d6eUJNyAYnChSmlJWn_IjNO3rTwkDI?e=YbVxIB))
Run assimp info model-textured.fbx.
Observe Textures (embed.): 0 and that material0 lists only colour/scalar properties, every entry with texture semantic n/a — no $tex.file.

Reproduced with the stock assimp info CLI (no custom post-processing flags), and independently via Importer::ReadFile with flags 0, enumerating all values of aiTextureType from 1 to AI_TEXTURE_TYPE_MAX and dumping aiMaterial::mProperties directly. No texture property is present under any semantic.

**Expected behavior**

material0 should carry a diffuse texture property referencing the embedded image (*0), and aiScene::mNumTextures should be 1 with mTextures[0] holding the compressed JPEG blob and achFormatHint set to jpg.

**Platform:**

OS: Windows (x64)
Compiler: MSVC
Version: reproduced on assimp 5.4 and on 6.0.5 (release build, official Windows installer binaries for the CLI test)

**Additional context**

Material properties reported for material0 (abridged): ?mat.name, $mat.shadingm, $clr.diffuse, $clr.emissive, $clr.ambient, $clr.specular, $mat.shinpercent, $mat.shininess, $mat.roughnessFactor, $clr.transparent, $mat.transparencyfactor, $mat.opacity, $clr.reflective, $mat.reflectivity, $mat.bumpscaling, $mat.displacementscaling, $raw.Emissive, $raw.Ambient, $raw.Shininess, $raw.Reflectivity, $raw.Diffuse, $raw.Specular. No |file entry and no $tex.* entry of any kind.

Mesh reports 1 UV channel, so the geometry side appears intact.

Possibly unrelated, but noted in case it shares a cause: assimp info reports 2 meshes both named Model, with 943990 faces each but different vertex counts (647149 and 723429), while the source FBX contains a single Geometry 'Model'.

Scene also contains 80 camera nodes (Metashape exports one per source photo), if that is relevant to the converter's traversal.

Bot has made me a helper class to patch aiScene with lost textures, from my initial tests it works well
https://pastebin.com/zRQkYSGQ

jeevan6996:
## Summary

Fixes #6838.

Metashape can write `ModelUVTranslation` and `ModelUVScaling` as binary int32 values. The FBX texture constructor currently reads them as float/double tokens, which throws and causes the texture object to be skipped, even when its connections and embedded image are valid.

Accept binary int32 values specifically for these texture UV components using the existing checked integer parser. Other representations continue through the float parser; the shared parser and property-table override behavior are unchanged.

Add small, original binary fixtures and their generator covering integer, float and double values, property overrides, and rejection of nonnumeric values. The reporter's model is not included.

## Validation

- Reproduced the missing texture on the reporter's original model before the fix. After the fix it imports one diffuse texture and the 13,560,596-byte embedded JPEG.
- Confirmed the new numeric-UV regression test fails on the unchanged importer and passes with this fix.
- All 21 FBX importer/exporter tests and 40 related material, parsing, scene, texture-transform and math tests pass locally (61 total).
- Formatted the changed C++ lines using the repository configuration; `git diff --check` passes.

Tested on macOS ARM64 with Homebrew LLVM 23.1.0 in a Debug build with the FBX importer and FBX/ASSBIN/ASSXML exporters enabled. Local dependency build workarounds were `-include ostream -Wno-error=unused-template`; these are not source changes. The full all-format suite, Windows/Linux builds and sanitizers have not been run locally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FBX texture import compatibility for UV translation and scaling values stored as integers, floating-point numbers, or overridden in material properties.
  - Embedded textures now retain the correct UV transformations across supported numeric formats.
  - Invalid, non-numeric UV values are safely rejected instead of producing unusable texture data.

- **Tests**
  - Added coverage for embedded textures, multiple UV value formats, material overrides, and invalid input handling.


kimkulling (approved):
Looks fine for me.

kimkulling:

Issue #6837, opened 2026-09-12
Proposal: native SketchUp (.skp) importer


iamahsanmehmood:
Also posted as a discussion for open conversation: https://github.com/orgs/assimp/discussions/6836

Assimp has importers for dozens of formats but none for SketchUp's native `.skp` files — currently the only path in is exporting from SketchUp to a format Assimp already reads, which requires SketchUp itself.

[OpenSKP](https://github.com/iamahsanmehmood/openskp) is an MIT-licensed, from-scratch `.skp` reader/writer with a C++17 port (`packages/cpp`) — no Trimble SDK, cross-platform including Linux and WASM. Proposing a `SkpImporter` under `code/AssetLib/Skp/` following `Assimp::BaseImporter`, mapping OpenSKP's scene output into `aiScene`/`aiMesh`/`aiMaterial`/`aiNode`, integrated with Assimp's `IOSystem`.

Happy to put together an RFC/PR if there's interest.

tellypresence:
@iamahsanmehmood Yes, please, by all means implement an importer and raise a PR.

Will be even better if you can add some lightweight `.skp` files under `test/models` and of course unit tests.

One important point, looks like [`openskp`](https://github.com/iamahsanmehmood/openskp) is your own project; keep in mind that the assimp repo CI checks are strict and catch a lot of problems, so might need to make code changes to get through PR approval.  The team is very professional and would never get personal (it's simply about code quality) but this issue:

- #2803 

is a textbook case of how _not_ to react

iamahsanmehmood:
Opened a PR: #6880

PR #6880, opened 2026-09-22
Add SkpImporter: native SketchUp (.skp) import via OpenSKP

iamahsanmehmood:
## What

Implements the `SkpImporter` proposed in #6837 and [the matching discussion](https://github.com/orgs/assimp/discussions/6836): native import of SketchUp `.skp` files (both the legacy pre-2021 MFC `CArchive` container and the modern 2021+ VFF/ZIP container), by delegating all parsing to [OpenSKP](https://github.com/iamahsanmehmood/openskp) (MIT licensed), a from-scratch, dependency-light C++17 `.skp` reader — no Trimble SDK required, cross-platform including Linux and WASM.

## Design

OpenSKP's `build_instanced_scene()` already resolves a `.skp` file into an `InstancedScene`: flattened materials, deduplicated mesh resources (one per distinct definition + material/representation combination the file actually uses), and a scene graph of placed nodes that reference those resources by index. That shape maps onto `aiScene`/`aiNode`/`aiMesh` close to 1:1 — including SketchUp's own component-instancing, which becomes multiple `aiNode`s referencing the same `aiMesh` indices, exactly assimp's own established mesh-sharing convention. This importer's own job is purely that translation:

- **Materials/textures**: embedded (PNG/JPEG bytes as-is, `aiTexture` with `mHeight = 0`), PBR base color + metallic/roughness factors, falls back to `AI_MATKEY_COLOR_DIFFUSE` for non-PBR consumers (same dual-set pattern the glTF2 importer uses).
- **Meshes**: one `aiMesh` per `LocalPrimitive` (OpenSKP already splits by material, matching assimp's own per-mesh-one-material convention).
- **Node tree**: OpenSKP's `InstancedNode::matrix` is a 16-element column-major matrix (metres, Y-up, glTF's own convention); `aiMatrix4x4`'s `a1..d4` fields are row-major (translation at `a4/b4/c4`, confirmed against `aiMatrix4x4::Translation()`) — the two are transposes of each other read in the same flat element order, so the importer transposes on the way in. Verified empirically, not just derived on paper: a 20-inch instance translation round-trips to exactly `0.508` in the imported `aiNode`'s own transformation matrix.

## Integration

Vendors OpenSKP as a git submodule (`contrib/openskp`) and wires it in via `ADD_SUBDIRECTORY`, the same pattern `contrib/draco` already uses for an external CMake-built dependency (`OPENSKP_BUILD_TESTS`/`OPENSKP_BUILD_EXAMPLES` forced off so it builds as a lean library, not its own test suite).

Opt-in via `ASSIMP_BUILD_SKP_IMPORTER` (off by default), following the same precedent M3D/USD/VRML already set in this repo for exactly this situation: third-party code with its own style conventions, not written to assimp's CI rules. Concretely here, that's MSVC's C4275/C4251 (a dll-interface class exporting a `std::` base/member — OpenSKP's `SkpParseError : std::runtime_error`, and `InstancedScene`'s `std::vector`/`std::string` members) — scoped to just `SkpImporter.cpp` via `/wd4275 /wd4251 /WX-`, not disabled project-wide.

## Verification

- Built the real `assimp.dll`/`.lib` with the importer enabled and ran two OpenSKP-writer-generated `.skp` fixtures (`test/models/SKP/cube.skp` — a plain 6-face box; `test/models/SKP/textured_instances.skp` — one component placed twice with an embedded PNG) through the actual `Assimp::Importer::ReadFile()` API with `aiProcess_ValidateDataStructure`, confirming correct mesh/material/instancing/texture output and assimp's own structural validation passing.
- Added `test/unit/utSkpImportExport.cpp` (3 cases: basic import, geometry/material shape, instancing via shared mesh indices) — kept **commented out** in `test/CMakeLists.txt`, matching `utM3DImportExport.cpp`'s own established convention for an opt-in-off-by-default importer. Manually uncommented it locally to confirm all 3 pass, then reverted.
- Ran assimp's full existing unit suite (704 tests) with the importer enabled — zero regressions.
- Checked for `CanRead()` magic-byte collisions against every other importer's own signature check — none found.

**Known gaps, stated honestly:**
- Only tested against self-generated fixtures so far, not a real SketchUp-authored file.
- Only built/verified on Windows/MSVC locally — haven't run a Linux/macOS/clang/gcc build myself, so I'd genuinely appreciate CI (and reviewers) catching anything platform-specific I can't see from here.
- No malformed/corrupt-file fuzzing beyond what code review gives confidence in.

Happy to adjust the integration approach (submodule vs. something else, the opt-in flag naming, anything about the C4275/C4251 handling) if there's a preferred pattern I'm missing — this is my first PR to this codebase.

## Summary by CodeRabbit

- **New Features**
  - Added optional native import support for SketchUp (.skp) files.
  - Imports SketchUp geometry, materials, textures, transformations, and scene hierarchies.
  - Supports legacy and modern SketchUp file formats when enabled.

- **Documentation**
  - Documented SKP support and its build configuration requirement.

- **Tests**
  - Added coverage for SketchUp geometry, materials, textured components, and instancing.


turol:
We do not use submodules. Vendor the dependency. Enable the tests when the importer is enabled. Change the CI to enable SKP so it's actually tested.

iamahsanmehmood:
Thanks for the review — all three addressed in 84f846f:

- **Vendored, not submoduled.** Checked and confirmed: this repo has no `.gitmodules` at all before this PR, so submoduling was a real mistake on my part rather than following an existing pattern. `contrib/openskp` is now a plain committed copy of OpenSKP's `packages/cpp` (flattened, no monorepo nesting), with `ASSIMP_VENDORING.md` documenting what commit it's from and how to refresh it.
- **Tests enabled with the importer.** `utSkpImportExport.cpp` is now conditionally added to `test/CMakeLists.txt`'s `IMPORTERS` list via `if(ASSIMP_BUILD_SKP_IMPORTER)` — same pattern `utUSDImport.cpp` already uses for USD — instead of permanently commented out.
- **CI now builds it.** Added `-DASSIMP_BUILD_SKP_IMPORTER=ON` to the `ubuntu-latest-g++` matrix entry in `ccpp.yml`, so the importer and its tests actually run on every push/PR.

Also fixed the real bug CodeRabbit flagged: `aiTexture::pcData` was allocated as `unsigned char[]` and cast to `aiTexel*`, which is genuine UB against `~aiTexture()`'s `delete[]` (mismatched array-new/delete type, not just a size issue). Now allocated as `aiTexel[mWidth]`, matching `AssbinLoader`'s own pattern for compressed embedded textures exactly, including its overflow check.

Verified with a full clean rebuild (caught a second, smaller vendoring gap in the process — missed `contrib/openskp/cmake/`, needed by `OpenSkpConfig.cmake.in`) and a full run of the existing unit suite: 704/704 passing, including the 3 SKP tests now actually compiled in.

turol:
You didn't actually change the CI configuration. Enable the new importer in all CI configurations.

iamahsanmehmood:
You're right, that was a real miss — I only added the flag to one matrix entry in \`ccpp.yml\` and didn't touch \`sanitizer.yml\` at all. Fixed in e061d5e:

- \`ccpp.yml\`: removed the \`matrix.name\` condition, so it now applies to every entry in the build job (ubuntu-g++, macos-clang++, windows-cl.exe, windows-clang.exe, ubuntu-clang++).
- \`sanitizer.yml\`: added it to both the ASan and UBSan jobs — genuinely useful there specifically, since that's exactly what would have caught the \`aiTexel[]\` allocation bug from the first round if it recurred.

Didn't touch \`cifuzz.yml\` — that's OSS-Fuzz's own external build pipeline, not this repo's CMakeLists.txt directly, and adding real fuzz coverage for a new importer needs a dedicated harness target, which is a separate, bigger piece of work than what you asked for here. Let me know if you'd want that as a follow-up.

I can't fully verify the Linux/macOS-specific sanitizer runs myself (Windows machine, no local ASan/UBSan toolchain) — that's exactly the gap I flagged up front in the PR description, so I'm relying on CI here rather than claiming something I haven't actually checked.

iamahsanmehmood:
Looked into the SonarCloud Quality Gate failure. Both findings trace entirely to vendored code, not to `SkpImporter.h`/`.cpp` (my actual new code):

- **Duplication (4.1%, required ≤3%)** and the **D Security Rating**: every flagged issue is in `contrib/openskp/include/openskp/_fragments_fb/index_generated.h` — machine-generated by `flatc` from a public FlatBuffers schema, not hand-written. Things like "replace `enum` with `enum class`", "use `std::to_underlying`", "`void*` is not meaningful" are real code-smell patterns in generated FlatBuffers output generally, not something specific to this schema or something I can hand-edit without breaking regeneration and diverging from upstream. Two more (`create.hpp`, minor/major) are in OpenSKP's own hand-written but vendored code, same category.
- Checked whether existing `contrib/` vendoring (e.g. `contrib/draco`) is somehow excluded from this — it isn't; those files are tracked and scanned by SonarCloud the same way, they just aren't flagged as "new code" anymore since they were vendored a while back. So this looks like the same class of thing any large vendored addition would trigger the first time, not something specific to how I did this one.

I don't have visibility into (or access to change) this repo's SonarCloud project configuration — I can't tell whether `contrib/` is meant to be scanned like this project-wide or if there's meant to be an exclusion I'm just not seeing from here. If there's a preferred way to handle this (a `sonar-project.properties` exclusion for `contrib/`, or just waiving the gate for vendoring PRs), let me know and I'll make whatever change is needed on my end.

iamahsanmehmood:
The failing checks were a real mistake on my end, not a CI/environment issue - explaining exactly what happened:

1. When pushing the previous round's fixes, a `git add` with one stale pathspec (a file that had already been removed) silently failed to stage the other files listed alongside it. I misread `git status`'s output as confirming everything was staged, and pushed two commits (`84f846f`, `e061d5e`) that didn't actually contain the fixes I described - including the `CMakeLists.txt` path fix, which is exactly what made every job fail immediately at CMake configure (`ADD_SUBDIRECTORY given source "contrib/openskp/packages/cpp" which is not an existing directory`).
2. Fixed in `2d152f4` by carefully re-staging and verifying each file against `git diff --cached --stat` before committing.
3. To avoid trusting my own working tree again (which is what masked the problem the first time - my local builds were reading uncommitted changes, not what was actually pushed), I verified this fix from a completely separate, freshly-cloned copy of the branch rather than my working directory. That clone also surfaced a second, real bug: `OpenSkp`'s `add_library()` has no explicit `STATIC`/`SHARED`, so it was inheriting assimp's own `BUILD_SHARED_LIBS` and building as its own DLL on Windows - which nothing copied next to `unit.exe`, so the test binary failed to start (`STATUS_DLL_NOT_FOUND`). Fixed in `4bc3c0c` by forcing it static, since it's an internal/vendored dependency of the importer, not something meant to ship as its own shared library.
4. From that same clean clone (not the working tree): full configure + build of the `unit` target succeeds, and all 704 existing unit tests pass, including the 3 SKP-specific ones.

Pushed as `4bc3c0c`. The three workflow runs on this commit are sitting at `action_required` since I'm not able to approve my own first-time-contributor runs - could someone approve them so CI can confirm on its own infrastructure as well?

iamahsanmehmood:
Went through every open CodeRabbit finding on this round (not just the ones that block a build) and fixed the real ones properly rather than piecemeal:

**The vendoring finding was right, but the fix I initially had in mind (vendor miniz/tinygltf/flatbuffers too) was the wrong one.** SkpImporter only ever calls OpenSKP's *read* path (`parser.hpp`/`instanced_scene.hpp`/`errors.hpp`). tinygltf and FlatBuffers are only used by OpenSKP's GLB and Fragments *exporters* - code this importer never calls. So instead of vendoring two more third-party libraries to satisfy an API surface nothing here uses, I trimmed the vendored copy down to the read path: dropped the writer, editor, all format exporters (OBJ/STL/PLY/DXF/IFC/GLB/Fragments), and WASM bindings, and their now-unneeded `FetchContent` blocks. `contrib/openskp/ASSIMP_VENDORING.md` and `openskp.hpp`'s own comment document exactly what's kept and why. miniz stays - `core.cpp` needs it to read the zip-based container modern `.skp` files use.

That also made several other findings moot for the right reason (the flagged files - `ifc_export.cpp`, `ply_export.cpp`, `obj_export.cpp`, `glb.cpp`, `edit.cpp` - aren't vendored anymore) rather than ignored.

**Real bugs fixed in what's still vendored:**
- `scene.cpp`: texture dedup compared only a size+16-byte prefix, so two different same-size PNGs sharing their signature/IHDR bytes (common) would wrongly dedupe. Now compares full bytes before reusing an index.
- `scene.cpp`: sibling instances sharing a name produced the same `child_path`, so the later one's backfill entry silently overwrote the earlier one's - both siblings' meshes ended up with the last sibling's name/properties/attributes. Disambiguated with a `#N` suffix on repeat.
- `instanced_scene.cpp`: `to_gltf_matrix()` indexed `m[0]..m[8]` without a size check (unlike the guarded `m[9]..m[11]` reads right below it) - a malformed/truncated file with an empty matrix would read out of bounds. Now returns identity for anything under 9 elements.
- `SkpImporter.cpp`: replaced the `strncpy` SonarCloud flagged with an explicit bounded `memcpy` (the prior code was actually safe given the pre-zeroed buffer, but not worth arguing with the flag).
- `CMakeLists.txt`: `OpenSkp` now gets `POSITION_INDEPENDENT_CODE` explicitly, since the Hunter compiler-setup path skips `CMAKE_POSITION_INDEPENDENT_CODE` and a non-PIC static lib could fail to link into a shared assimp there.
- `ccpp.yml`: the SKP-args step wrote to `$GITHUB_OUTPUT` under Windows runners' default `pwsh` shell, where bare `$GITHUB_OUTPUT` isn't the runner's output file - the redirect was silently no-oping, so `-DASSIMP_BUILD_SKP_IMPORTER=ON` never actually reached CMake on Windows CI even though the step reported success. Pinned to `shell: bash`.

Pushed as `64e9a75`. Verified the same way as the last round: full configure + build of the `unit` target, and all 704 unit tests (including the 3 SKP-specific ones) passing, from a completely fresh clone of this commit.

iamahsanmehmood:
CodeRabbit's latest pass caught one more: \`instanced_scene.cpp\` has its own independent copy of the texture-dedup logic (\`build_instanced_scene_raw()\` doesn't share code with \`scene.cpp\`'s \`build_scene_raw()\`), so the size+16-byte-prefix bug fixed there last round was still live here too. Same fix applied. Checked this file for the other two bugs from last round as well: it doesn't use the \`path_updates\` backfill mechanism at all (each node's metadata is set inline, no deferred path-keyed lookup), so the sibling-name-collision fix doesn't apply here, and the \`to_gltf_matrix()\` bounds check was already part of this file's own previous fix.

Pushed as \`c50f99b\`, same verification as every round: full build + all 704 unit tests passing from a fresh clone of the commit.

iamahsanmehmood:
Re: the SonarCloud complexity/style findings in `contrib/openskp/` (Quality Gate passes, so not blocking) - opened [iamahsanmehmood/openskp#379](https://github.com/iamahsanmehmood/openskp/issues/379) upstream to track them for later, rather than leaving them unacknowledged. Not attempting that refactor in this PR itself - it'd mean touching already-shipped, production-used core parsing logic to satisfy a generic complexity heuristic, which is its own separate piece of work with its own risk profile.

